PR #29, opened 2017-06-16
vmBitwise and tangle improvements

nishantjr:


ehildenb:
Can you put the `k/*.k` files back under version control? We want people who don't have `pandoc-tangle` to be able to use the semantics.

nishantjr:
Ah cool, was wondering why it was in. I've dropped that commit.

PR #34, opened 2017-06-18
replace tangle with Build and a Makefile

nishantjr:
Let me know if you'll have nitpicks.

@msaxena2 Let me know if you want to keep something from 986217c

```
$ ./Build help

    usage:
       ./Build                  Build the definitions and kompile them
       ./Build defn             Only detangle the definitions
       ./Build test <test.json> krun the test file and check that it succeeds
       ./Build test <expr>      Use find to locate a test file in tests/tests-develop/VMTests/
```


msaxena2:
@nishantjr This looks great! I think we don't need anything from the old stuff.

PR #50, opened 2017-06-26
fast interpreter for evm

dwightguth:
I haven't hooked this into anything yet, and the parser probably won't work for the academic K, but I created a simple script that should do fast execution of the evm semantics. Tested on some of the json in the tests-develop directory and it takes about 100 ms to get stuck. It gets stuck pretty quickly (after <100 steps) so it's not much of a benchmark but it means there's essentially now negligible overhead and nearly all the user time is being spent executing steps.

Dunno how you want to hook it into your testing infrastructure; it looks like you're using ktest which means it won't really play well because it never actually calls krun. It also doesn't do any pretty printing, just spits out the configuration in kast format. But it should give you very very fast execution with pretty much no overhead. Note that you need to have a kserver running for it to work.

### DIFF
--- a/KRYPTO.ml
+++ b/KRYPTO.ml
@@ -2,5 +2,9 @@ open Constants.K
 
 let h = Cryptokit.Hash.keccak 256
 
-let hook_keccak256 c lbl sort config ff = match c with
-  [String input] -> [String (Cryptokit.hash_string h input)]
+let hook_keccak256 c lbl sort config ff = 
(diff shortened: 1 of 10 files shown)


dwightguth:
not ready yet: still depends on runtimeverification/rv-match#466

msaxena2:
@dwightguth We're not using ktest at the moment - We have 11 parallel builds on codeship that directly call krun, so it'll be easier to hook into.

PR #51, opened 2017-06-27
Fix #unparseByteStack

msaxena2:
@ehildenb @nishantjr. Only Keccak seems to use #unparseByteStack. In case you're planning to use it, this quickfix will help.


msaxena2:
Closing this. Handled in #52.

PR #52, opened 2017-06-27
Sha3 tests

msaxena2:
@ehildenb @nishantjr I think we can merge this. 4 tests don't pass because they use very large values - the tests themselves don't have "post" conditions, so the values might be invalid. Also, this contains commit from #51 . 

Also, this should only be merged after kframework/k#2312


msaxena2:
Build has to be triggered again after merging the PR in K. Tests should pass then.

PR #53, opened 2017-06-27
fast rvk interpreter

dwightguth:
tested on rvk and passes all issues except the one signextend test that tries to shift right on an enormous right operand.


dwightguth:
can we merge this?

PR #60, opened 2017-07-05
Opcode create

ehildenb:


ehildenb:
@dwightguth see here for fixes to CREATE.

PR #75, opened 2017-08-09
recover old build system that runs 12x faster on Jenkins

dwightguth:
The old build system, which ran the tests in parallel, is substantially faster than the new one that was created since I left, so I am reviving the old build system on rv-jenkins.


dwightguth:
Jenkins: test this please

ehildenb:
Branch develop now contains these fixes as well.

PR #104, opened 2017-11-10
ABI Verification + Updated Proofs

msaxena2:
WIP. Do not merge yet.

### DIFF
--- a/.build/uiuck/verification.k
+++ b/.build/uiuck/verification.k
@@ -206,8 +206,9 @@ module VERIFICATION
 // `<pc> 0 </pc>`.
 
 
-    syntax TypedArg ::= "#uint160"  "(" Int ")"
-                      | "#address"  "(" Int ")"
+    syntax TypedArg ::= "#uint160"      "(" Int ")"
+                      | "#address"      "(" Int ")"
+                      | "#uint256"      "(" Int ")"
 
     syntax TypedArgs ::= List{TypedArg, ","}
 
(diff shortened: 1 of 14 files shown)


.build/uiuck/verification.k
```diff
@@ -21,9 +22,122 @@ module VERIFICATION
       requires N =/=K 0
       [lemma]
 
-    rule (X -Int A) -Int B    => X -Int (A +Int B)    [smt-lemma]
-    rule ((X -Int A) >=Int B) => (X >=Int (A +Int B)) [smt-lemma]
-    rule (N +Int X <Int M)    => X <Int (M -Int N)    [smt-lemma]
+    rule (X -Int A) -Int B    =>  X -Int   (A +Int B)    [smt-lemma]
```
andreistefanescu: I am surprised these `smt-lemma` has any effect: the engine won't use it, just pass it to z3, which should be able to deal with linear arithmetic (decidable and efficient theory).

bmmoore: These were intended only for the engine and not Z3, so I should change this. In particular, they are only meant to apply when `A` and `B` are concrete, if you can suggest how to express that.

 I thought `smt-lemma` rules were used by the engine and by Z3 - are you sure the engine doesn't use them? I'm pretty sure the accumulated constraints during a proof got smaller after adding some of the rules along these lines, though I'm not sure exactly which. Before I added any extra lemmas we got expressions like `(((((Gas -Int 1) -Int 1) -Int 1) -Int 1) -Int 1)`, stuck on a symbolic variable.

andreistefanescu: I remember that `lemma` means engine (oriented left to right), `smt-lemma` means z3, and both tags combined means both, but I might be wrong. However, the engine will not work very well with lemmas that use boolean connectors, because boolean connectors are special. Anyway, there is some cleanup needed here.

PR #105, opened 2017-12-05
chop: %Int => modInt

daejunpark:
Do not merge. Just to see what happens.

### DIFF
--- a/data.md
+++ b/data.md
@@ -55,8 +55,8 @@ Some important numbers that are referred to often during execution:
 ```{.k .uiuck .rvk}
     syntax Int ::= chop ( Int ) [function]
  // --------------------------------------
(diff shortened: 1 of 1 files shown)


pirapira:
Oh, sorry, I rebased a branch that's not mine.

dwightguth:
Do we want to merge this?

daejunpark:
@dwightguth Not sure. Clearly, `%Int` is wrong when `I` is negative, although @ehildenb believes that no negative values can appear during the execution. Still this change makes sense and better to be merged just in case. But the thing is that there are many other places where such a questionable `%Int` is still used instead of `modInt`, and I'm not sure how much merging this single change will make the definition robust.

daejunpark:
close this and open #115

PR #107, opened 2017-12-19
ABI-Proofs Rebased

msaxena2:
ABI complaint hkg token proofs.

### DIFF
--- a/.build/uiuck/data.k
+++ b/.build/uiuck/data.k
@@ -2,8 +2,8 @@
 // =========
 
 // EVM uses bounded 256 bit integer words, and sometimes also bytes (8 bit words).
-// Here we provide the arithmetic of these words, as well as some data-structures
-// over them. Both are implemented using K's `Int`.
+// Here we provide the arithmetic of these words, as well as some data-structures over them.
+// Both are implemented using K's `Int`.
 
 
 requires "krypto.k"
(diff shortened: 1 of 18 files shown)


msaxena2:
Closing #107 - this PR has the related code. @daejunpark @ehildenb @bmmoore I'd like to get this in sometime today. The proofs are the same as the ones we used for the submissions. The proofs miss a few corner cases, and I'm making a note of them. Regardless, I'd like to get the abi-mechanism related code (which is independent of the proofs themselves) in before we proceed any further. The next step, of course would be to iron out the corner cases, and synchronize it with @grosu erc-20 repository.

msaxena2:
Ugh. The proofs pass but the tests fail. Looking into it now.

daejunpark:
I don't want to merge this as the current form. I'm not fully convinced with the verification lemmas. It seems to be too ad-hoc, which is ok for the experimental purpose but not for the master.

msaxena2:
@daejunpark Yeah I agree, we can clean the lemmas up a bit. Probably an explanation of what each rule with the "smt-lemma" attribute does would also be helpful. Is your current work focused on just cleanup, or does it involve adding lemmas?

daejunpark:
I'm trying a different approach, introducing an abstraction to avoid such low-level reasoning lemmas. Take a look at the viper branch.

daejunpark:
close this since we have a replacement

PR #98, opened 2017-10-14
Better test harness

ehildenb:
improvements to the `./Build` script and repository layout.

PR #108, opened 2017-12-20
use lua script to tangle markdown files

dwightguth:
This replaces part of the functionality contained in #98, in order to make it easier to build the EVM semantics without having to have access to pandoc-tangle.


ehildenb:
@dwightguth what was troublesome about the tests?

dwightguth:
@ehildenb they take too long to run and stall the jenkins box.

PR #112, opened 2018-01-08
Documentation

ehildenb:
beginning work on auto-documentation generation and cleanup other parts of documentation


ehildenb:
Turns out there are issues with the Makefile/Build script that need to be fixed for the tests passing to be meaningful (currently they are "passing", but the proofs are not actually being run).

Many of those features exist in branch `better-test-harness`, so in the interest of getting this dangling branch cleaned up, I'm subsuming this into https://github.com/kframework/evm-semantics/pull/98.

PR #119, opened 2018-01-19
Verification

daejunpark:


proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+```
+
+instead of specifying directly the underlying byte-array:
+
+```
+.Map[ RET_ADDR      := 0  ]
+    [ RET_ADDR + 1  := 0  ]
+    ...
+    [ RET_ADDR + 31 := 1 ]
+```
```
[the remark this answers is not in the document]
grosu: Is it really better to write it this way instead of 
```
RET_ADDR |-> 0
RET_ADDR + 1 |-> 0
...
```

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+-----------
+
+```{.k .erc20-spec .hkg-spec}
+[totalSupply]
+```
+
+`<k>` cell specifies that the execution eventually reaches RETURN instruction meaning that the program will successfully terminate. The RETURN instruction says that a 32-byte return value will be stored in the memory at the location RET_ADDR.
+
+```{.k .erc20-spec .hkg-spec}
+k: #execute => (RETURN RET_ADDR:Int 32 ~> _)
```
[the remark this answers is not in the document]
grosu: Explain also briefly what `#execute` is, and maybe refer to the EVM semantics for more info

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+
+callData: #abiCallData("totalSupply", .TypedArgs)
+```
+
+`<localMem>` cell specifies that the local memory is empty in the beginning, but in the end, it will contain some contents including the return value `TOTAL`, the total supply of the tokens.
+
+The other entries are represented by the wildcard symbol `_`, meaning that they are not relevant to the functional correctness of the program.
+
```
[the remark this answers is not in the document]
grosu: Wouldn't it be better, if possible, to verify the functions at their call boundaries rather than at their execution boundaries?

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+```{.k .erc20-spec .hkg-spec}
+localMem:
+      .Map
```
grosu: Is the local memory empty for sure when a function is executed?  Is this because you use the execution and not the caller's context when specifying the property, so the caller's local memory has already been stacked?  Is that stacking code then ever verified?

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+    =>
+    (
+      .Map[ RET_ADDR := #asByteStackInWidth(TOTAL, 32) ]
+      _:Map
+    )
+```
+
+`<gas>` cell specifies the maximum gas amount. Here we give a loose upper-bound for the demonstration purpose. In practice, one should set a reasonable amount of the gas limit to see if the program does not consume too much gas (i.e., no gas leakage).
```
grosu: `the demonstration purpose` => `demonstration purposes`.

grosu: So we are not verifying the exact gas consumption?

grosu: Say what `{gascap}` is and where it is defined.

proofs/erc20.md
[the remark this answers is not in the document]
grosu: What is the meaning of `_` in the RHS of a rule?  Explain briefly.

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+storage:
+    #hashedLocation({lang}, {pos(totalsupply)}, .IntList) |-> TOTAL
+    _:Map
+```
+
+The side-condition specifies the range of symbolic values based on their types.
+
+```{.k .erc20-spec .hkg-spec}
+requires:
+    andBool 0 <=Int TOTAL     andBool TOTAL     <Int (2 ^Int 256)
```
grosu: Why the large whitespaces above?

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+
+```{.k .erc20-spec .hkg-spec}
+refund: _
+```
+
+`<storage>` cell specifies that the value of `totaySupply` is `TOTAL` and other entries are not relevant (and could be arbitrary values). Specifying the irrelevant entries implicitly expresses the non-interference property. That is, the total supply value will be returned regardless of what the other entires of the storage are. This representation of the irrelevant part is used throughout the entire specification, ensuring one of the principal security properties.
+
+```{.k .erc20-spec .hkg-spec}
+storage:
+    #hashedLocation({lang}, {pos(totalsupply)}, .IntList) |-> TOTAL
```
grosu: Where is `{lang}` coming from?  Explain briefly.

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+
+```{.k .erc20-spec .hkg-spec}
+log: _
+```
+
+`<refund>` cell specifies that no gas is refunded. Note that it does not mean it consumes all the provided gas. The gas refund is different from returning the remaining gas after the execution. It is another notion to capture some specific gas refund events that happen, for example, when an unused storage entry is re-claimed (i.e., garbage-collected). The following specification ensures that no such event happens during the execution of the current function.
+
+```{.k .erc20-spec .hkg-spec}
+refund: _
+```
```
grosu: It would be nice to not have to specify the cells that do not change at all!  Narrowing should be able to do that for us.

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+[totalSupply]
+```
+
+`<k>` cell specifies that the execution eventually reaches RETURN instruction meaning that the program will successfully terminate. The RETURN instruction says that a 32-byte return value will be stored in the memory at the location RET_ADDR.
+
+```{.k .erc20-spec .hkg-spec}
+k: #execute => (RETURN RET_ADDR:Int 32 ~> _)
+
+callData: #abiCallData("totalSupply", .TypedArgs)
+```
```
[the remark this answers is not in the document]
daejunpark: That is due to the parsing issue. `.TypedArgs` specifies an empty list, meaning that there no argument for totalSupply.

proofs/erc20.md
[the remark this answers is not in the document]
daejunpark: Local variables are stored in the local memory.

> But what other junk is there, and why?

Not sure what you're asking, but find the RETURN instruction in the <k> cell which specifies that the RET_ADDR will be returned.

> And are we 100% sure that the value at RET_ADDR is what is returned by the function?



Definitely, it is better with the call boundaries, but I'll do it in the next version, as we discussed last time. There are some issues with K prover to achieve that.

> Wouldn't it be better, if possible, to verify the functions at their call boundaries rather than at their execution boundaries?

proofs/erc20.md
daejunpark: Yes, and will be done next version as I said in the above.

proofs/erc20.md
daejunpark: Specifying the *exact* gas consumption is fragile. Note that the different compiler version will affect it depending on their optimizations. Also, note that the exact gas consumption is indeed a *set* of the gas consumptions, since the gas price for each instruction varies by the current state of the storage.

> So we are not verifying the exact gas consumption?

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+    ...
+    [ RET_ADDR + 31 := 1 ]
+```
+
+### Storage
+
+The storage accommodates permanent data such as the `balances` map.
+
+A map is laid out in the storage where the map entries are scattered over the entire storage space for which the hash of each key is used to determine the location.
+
```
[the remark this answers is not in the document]
grosu: Although obvious, I would mention it in parentheses in passing, something like this:

... for which the (256-bit) hash ...

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+```
+hash(key2 ++ hash(key1 ++ pos(map)))
+```
+
+where `++` is the byte-array concatenation.
+
+The hashed location abstraction allows to uniformly specify the locations in a form parameterized by the underlying compilers. For example, the location of `map[key1][key2]` can be specified as follows, where `{lang}` is a place-holder to be replaced by the name of the compiler.
+
+```
+#hashedLocation({lang}, pos(map), key1 key2)
```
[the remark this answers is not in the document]
grosu: Then please say it so, because I was convinced it was a typo.

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+```
+
+This abstraction makes the specification independent of the underlying compilers, enabling it to be reused for differently compiled programs.
+
```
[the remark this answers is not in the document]
grosu: A link to the right place in `verification.md`, "for details on has collisions"?

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+
+### Gas Consumption
+
+Regarding the gas consumption of the program, the maximum gas amount is specified to ensure that the program does not consume more gas than the limit.
```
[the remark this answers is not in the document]
grosu: Ok, but I would still put a note here saying so.  I was confused.  You are also relying on the fact that the gas consumption does not depend on the input ... which means that probably you cannot prove the sum example yet?

proofs/erc20.md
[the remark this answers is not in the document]
grosu: I know :)  What I meant was add a short explanation somewhere, maybe in parentheses.

proofs/erc20.md
grosu: nice, but again, a short sentence about it now?  Later on, when we do things in the caller's context, we will go back and modify the text.  We will want to revise the entire text anyway.

proofs/erc20.md
grosu: Interesting ... we should say that somewhere, because people will be interested in gas consumption.
Also, once the dust settles, we should have a proof where we also calculate the exact gas consumption, symbolically.  Are we going to have the same complexity in IELE, where the gas price of each instruction varies by the current state of the storage?

grosu:
Alright, commented on @daejunpark 's comments to my comments; now moving on to continuing to add comments to the PR

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+storage:
+    #hashedLocation({lang}, {pos(balances)}, CALLER_ID) |-> (BAL_FROM => BAL_FROM -Int VALUE)
+```
+
+```{.k .erc20-spec}
+    #hashedLocation({lang}, {pos(balances)}, TO_ID)     |-> (BAL_TO   => BAL_TO   +Int VALUE)
+```
+
+##### HKG-specific:
+On the other hand, HKG token allows the additional overflow, “wrapping around” the balance if it is more than `2^256`, which is very unlikely but still a questionable behavior. See the difference where `+Word`, the addition modulo `2^256`, is used instead of `+Int`, the mathematical, arbitrary precision integer addition.
```
grosu: *the* HKG token

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+
+The side-condition ensures that no arithmetic overflow happens.
+
+```{.k .erc20-spec}
+    andBool VALUE <=Int BAL_FROM
+    andBool BAL_TO +Int VALUE <Int (2 ^Int 256) // viper overflow check: (VALUE ==Int 0 xorBool BAL_TO +Word VALUE >Int BAL_TO)
+```
+
+##### HKG-specific:
+As mentioned earlier, HKG token does not check the overflow thus we have no side-condition for that. Instead, HKG token does not allow the zero value to be requested to transfer, which is questionable though, and we have the side-condition for that.
```
grosu: *the* HKG

grosu: HKG is not proper ERC20 from what I can see.  I think we should say that upfront and then say that "We can also verify token which are not ERC20-compliant, but we have to change the specification accordingly.  We typically do that together with the customer, to make sure that that' indeed what they want.  It may very well be the case that the customer will actually want to stay ERC20-compliant.  But here we pretend that the HKG customer prefers their non-compliant ERC20 variant."

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+```
+
+The side-condition ensures that no arithmetic overflow (precisely underflow in this case) happens.
+
+```{.k .erc20-spec}
+    andBool VALUE <=Int BAL_FROM
+```
+
+##### HKG-specific:
+As in the sub-case 1, HKG token requires having the additional side-condition of VALUE’s being greater than 0.
```
grosu: I find it very confusing to present both the correct ERC20 variant and the HKG token variant at the same time, in parallel.  I would present each of them separately, and when presented the HKG token I would say that there are some differences and make those differences very clearly marked.  Each token is likely to be be somewhat different in one way or another.  So what are we going to do, to include all of them in the same file, and then lose track of the slight differences and where is one and where is the other?

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+    requires 0 <=Int ACCT_ID    andBool ACCT_ID    <Int (2 ^Int 160)
+     andBool 0 <=Int CALLER_ID  andBool CALLER_ID  <Int (2 ^Int 160)
+     andBool 0 <=Int ORIGIN_ID  andBool ORIGIN_ID  <Int (2 ^Int 160)
+     andBool 0 <=Int CALL_DEPTH andBool CALL_DEPTH <Int 1024
+     {requires}
+
+{comments}
+```
+
+Below are an abstract semantics of the four instruction, LT, GT, EQ, and ISZERO, that improves the performance of the underlying theorem prover’s symbolic reasoning by avoiding unnecessary case analyses.
```
grosu: shouldn't these go somewhere else, where we have lemmas and axioms?

verification.md
```diff
@@ -11,24 +12,296 @@ module VERIFICATION
...
 
-This `smt-lemma` helps Z3 reason about stack over/under flow.
+Abstractions
+------------
+
+We design abstractions that capture the EVM low-level specific details, allowing to specify specifications and reason about properties in a higher level similar to that of the surface languages (e.g., Solidity or Viper) in which smart contracts are written.
+
+### Memory Abstraction
+
+We present an abstraction for the EVM memory to allow the word-level reasoning. The word is considered as the smallest unit of values in the surface language level (thus in the contract developers’ mind as well), but the EVM memory is byte-addressable. Our abstraction helps to fill the gap and make the reasoning easier.
```
grosu: allow (the => .) word-level ...

grosu: MAybe have Philip or Everett revise the English when over.

verification.md
```diff
@@ -11,24 +12,296 @@ module VERIFICATION
...
+
+Specifically, we introduce uninterpreted function abstractions and refinements for the word-level reasoning.
+
+The term `nthbyteof(v, i, n)` represents the i-th byte of the two's complement representation of v in n bytes (0 being MSB), with discarding high-order bytes when v is not fit in n bytes.
+
+```{.k .uiuck}
+  syntax Int ::= nthbyteof(Int, Int, Int) [function, smtlib(smt_nthbyteof)]
```
grosu: explain the smtlib attribute.  also the function attribute

verification.md
```diff
@@ -11,24 +12,296 @@ module VERIFICATION
...
+```
+
+Precisely it is defined as follows.
+
+```{.k .uiuck}
+  rule nthbyteof(V, I, N) => nthbyteof(V /Int 256, I, N -Int 1) when N  >Int (I +Int 1) [concrete]
+  rule nthbyteof(V, I, N) =>           V %Int 256               when N ==Int (I +Int 1) [concrete]
+```
```
grosu: Explain the concrete attribute.  Actually, when did you guys add it?  Wouldn't it be better to have a `#concrete` symbol and add a `requires` side condition saying that the (meta-)variables V, I, N are concrete?

verification.md
```diff
@@ -11,24 +12,296 @@ module VERIFICATION
...
+              : nthbyteof(V, 25, 32)
+              : nthbyteof(V, 26, 32)
+              : nthbyteof(V, 27, 32)
+              : nthbyteof(V, 28, 32)
+              : nthbyteof(V, 29, 32)
+              : nthbyteof(V, 30, 32)
+              : nthbyteof(V, 31, 32)
+              : .WordStack ) => V
+    requires 0 <=Int V andBool V <Int (2 ^Int 256)
+
```
grosu: Explain the following intuitively.

verification.md
```diff
@@ -11,24 +12,296 @@ module VERIFICATION
...
+```
+
+Another type of byte-array manipulating operation is used to extract the function signature from the call data. The function signature is located in the first four bytes of the call data, but there is no atomic EVM instruction that can load only the four bytes, thus some kind of byte-twiddling operations are necessary.
+
+The extraction mechanism varies by language compilers. For example, in Viper, the first 32 bytes of the call data are loaded into the memory at the starting location 28 (i.e., in the memory range of 28 to 59), and the memory range of 0 to 31, which consists of 28 zero bytes and the four signature bytes, is loaded into the stack. In Solidity, however, the first 32 bytes of the call data are loaded into the stack, and the loaded word (i.e., a 256-bit integer) is divided by `2^(28*8)` (i.e., left-shifted by 28 bytes), followed by masked by 0xffffffff (i.e., 4 bytes of bit 1’s).
+
+The following lemmas essentially capture the signature extraction mechanisms. It reduces the reasoning efforts of the underlying theorem prover, factoring out the essence of the byte-twiddling operations.
+
+```{.k .uiuck}
+  // for viper
```
grosu: should we put the rule for Viper vs. for Solidity in different  modules/files?

verification.md
```diff
@@ -11,24 +12,296 @@ module VERIFICATION
...
+```
+
+### Abstraction for Hash
+
+We do not model the hash function as an injective function simply because it is not true due to the pigeonhole principle.
+
+Instead, we abstract it as an uninterpreted function that captures the possibility of the hash collision, even if the probability is very small.
+
+```
+  syntax Int ::= keccak ( WordStack ) [function, smtlib(smt_keccak)]
```
grosu: explain the smtlib attribute

verification.md
```diff
@@ -11,24 +12,296 @@ module VERIFICATION
...
+```
+
+However, one can avoid reasoning about the potential collision by assuming all of the hashed values appearing in each execution trace are collision-free. This can be achieved by instantiating the injectivity property only for the terms appearing in the symbolic execution, in a way analogous to the universal quantifier instantiation.
```
grosu: Not clear if you indeed did the above or not.  Where is it done?  How do you actually keep track of all the terms that appear in symbolic execution?  What is "universal quantifier instantiation"?

verification.md
```diff
@@ -11,24 +12,296 @@ module VERIFICATION
...
+  rule #hashedLocation("Solidity", BASE, OFFSET OFFSETS) => #hashedLocation("Solidity", keccakIntList(OFFSET BASE),       OFFSETS)
+
+  syntax Int ::= keccakIntList(IntList) [function]
+  rule keccakIntList(VS) => keccak(intList2ByteStack(VS))
+
+  syntax WordStack ::= intList2ByteStack(IntList) [function]
+  rule intList2ByteStack(.IntList) => .WordStack
+  rule intList2ByteStack(V VS) => #asByteStackInWidth(V, 32) ++ intList2ByteStack(VS)
+    requires 0 <=Int V andBool V <Int (2 ^Int 256)
+```
```
grosu: Don't take absolutely what I said before, that we should do things separately for Viper and Solidity.  Just think about it.  It may be better to keep them together after all ...

verification.md
```diff
@@ -11,24 +12,296 @@ module VERIFICATION
...
+
+These rules are specific to reasoning about EVM programs.
 
 ```{.k .uiuck}
-    rule #sizeWordStack ( _ , _ ) >=Int 0 => true [smt-lemma]
-    rule #sizeWordStack ( WS , N:Int )
-      => #sizeWordStack ( WS , 0 ) +Int N
-      requires N =/=K 0
-      [lemma]
+  syntax Bool ::= #isConcrete(K) [function, hook(KREFLECTION.isConcrete)]
```
grosu: hm, so what's the relationship between this and the the `[concrete]` attribute?

verification.md
```diff
@@ -11,24 +12,296 @@ module VERIFICATION
...
+
+  rule (I1 +Int I2) +Int I3 => I1 +Int (I2 +Int I3) when #isConcrete(I2) andBool #isConcrete(I3)
```
grosu: when => requires

proofs/erc20.md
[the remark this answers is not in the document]
daejunpark: I'm not relying on that. Indeed, the gas consumption *does* depend on the input. But it is not necessarily a function of the input. In the ERC20 token contracts, all possible amounts of the gas consumption are concrete. The sum example's gas consumption is constant.

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+
+k: #execute => (RETURN RET_ADDR:Int 32 ~> _)
+
+callData: #abiCallData("balanceOf", #address(OWNER))
+
+localMem:
+      .Map
+    =>
+    (
+      .Map[ RET_ADDR := #asByteStackInWidth(BAL, 32) ]
```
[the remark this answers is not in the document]
daejunpark: Not `[macro]` but `[function]`. But, yes, they are syntactic sugars.

Note that they are desugared into 32 map entries: `RET_ADDR |-> _ RET_ADDR + 1 |-> _` ... `RET_ADDR + 31 |-> _`. We do not want to write it directly because of two reasons: it is too verbose, and there is a performance bug in parsing them.

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+               | #parseHexWord(Keccak256(#generateSignature("Approval", #address(CALLER_ID), #address(SPENDER), #uint256(VALUE))))
+               : CALLER_ID
+               : SPENDER
+               : .WordStack
+               | #asByteStackInWidth(VALUE, 32)
+               })
+    )
+```
+
+`<refund>` cell specifies that a refund may be issued. This function will refund a gas if the allowance value is 0 while the previous value is greater than 0, re-claiming the corresponding entry of the storage.
```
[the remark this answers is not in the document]
daejunpark: The refund is part of EVM semantics. I think it would be great if IELE have a nice gas model *without* this, which I think could be yet another advantage over EVM.

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+
+transfer
+--------
+
+`transfer` function transfers the amount of VALUE from the caller to the account TO_ID.
+
+```{.k .erc20-spec .hkg-spec}
+[transfer]
+
+callData: #abiCallData("transfer", #address(TO_ID), #uint256(VALUE))
```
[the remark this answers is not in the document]
daejunpark: They are necessary for the function signature hash. Agree with the more information need.

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+               | #parseHexWord(Keccak256(#generateSignature("Transfer", #address(CALLER_ID), #address(TO_ID), #uint256(VALUE))))
+               : CALLER_ID
+               : TO_ID
+               : .WordStack
+               | #asByteStackInWidth(VALUE, 32)
+               })
+    )
+
+refund: _ => _
+
```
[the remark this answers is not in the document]
daejunpark: As I said in the above (maybe it's my fault you didn't get it), the refund happens when you update an existing value to 0 in the storage (e.g., in this case, the caller's balance becomes 0). So, the functions that touch the storage, i.e., `approve`, `transfer`, and `transferFrom`, may result in a refund.

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
+    andBool 0 <=Int VALUE     andBool VALUE     <Int (2 ^Int 256)
+    andBool 0 <=Int BAL_FROM  andBool BAL_FROM  <Int (2 ^Int 256)
+    andBool 0 <=Int BAL_TO    andBool BAL_TO    <Int (2 ^Int 256)
+
+comments:
+    // NOTE: negative VALUE is not possible since it is of `num256` type
+```
+
+There are two sub-cases depending on whether the caller is equal to the recipient.
+
```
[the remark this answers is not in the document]
daejunpark: I'll write about how this DSL works. Briefly, a section, the one enclosed by the squire brackets, e.g., `[transfer]` or `[transfer-success]`, inherits its parent. It overwrites the inherited fields, but accumulates when the current field starts with `+`, e.g., `+requires`.

proofs/erc20.md
```diff
@@ -0,0 +1,1030 @@
...
++requires:
+    andBool CALLER_ID =/=Int TO_ID
+```
+
+The side-condition ensures that no arithmetic overflow happens.
+
+```{.k .erc20-spec}
+    andBool VALUE <=Int BAL_FROM
+    andBool BAL_TO +Int VALUE <Int (2 ^Int 256) // viper overflow check: (VALUE ==Int 0 xorBool BAL_TO +Word VALUE >Int BAL_TO)
```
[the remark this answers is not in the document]
daejunpark: The two conditions are equivalent. The condition for Viper is just a comment, highlighting that it seems weird but we can still reason about it well. Will explain there.

proofs/erc20.md
daejunpark: I felt similarly once I wrote it. I did it in this way because I thought we can avoid any duplication of the spec, which helps to maintain it. But at this point, I think it'd be better to split the spec into two files and extend the inheritance figure, minimizing duplication in the second file.

proofs/erc20.md
daejunpark: I also want to do that, but I can't due to the limitation of the current k prover. But, we have an outstanding PR related to this, so hopefully we can improve that soon.

verification.md
daejunpark: I thought that as well, but it seems more succinct than having the lengthy condition when we require all of the variables to be concrete.

verification.md
daejunpark: I thought the similar thing. We can discuss more about it with others.

I wish I can put this rules in the spec file, but cannot due to the limitation of k prover's parsing mechanism.

verification.md
daejunpark: By the quantifier instantiation, I mean something similar to the Z3 trigger for selectively instantiating the universally quantified axioms.

I implicitly did it for at least for ERC20. In ERC20, hash is only used for the hashed location in the storage, and they appear in the spec as the keys of `Map`, which implicitly means they are different. 

Of course, for a program that generate and manipulate the hashed values, we need to implement some new feature for that.

verification.md
daejunpark: `[concrete]` attribute is a syntactic sugar for `#isConcrete`.

proofs/erc20.md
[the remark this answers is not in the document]
grosu: Cool, clear now.  Please add a comment in the text saying so.  This is useful info.

proofs/erc20.md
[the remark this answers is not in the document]
grosu: OK, sounds good.  So the `[function]` rules are now applied statically?  Didn't know that.  What if they are recursive?  I think we may want to get back to `[macro]`s, as the `[function]` attribute started to be overused.  OK for now, just add a comment somewhere about this, so people who do ERC20 proofs are aware of these things.

proofs/erc20.md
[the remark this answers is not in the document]
grosu: @daejunpark can you please open this discussion in the #iele channel on Slack?  I agree.  One question here is what is the right way to reclaim space.  In EVM, you reclaim when you write 0 in it, which results I guess in a refund, but is that such a big deal to be worth a refund?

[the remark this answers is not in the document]
grosu: I would also add a sentence where the refund _ => _ appears for the first time, saying that this is something EVM-specific and there is no way to go around it.

proofs/erc20.md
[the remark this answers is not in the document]
grosu: yes, i got it in the end.  It would be nice to include the refund, too, in the spec, because it would show everybody that we are very serious about the last detail.

proofs/erc20.md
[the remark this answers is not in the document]
grosu: Interesting.  How do you know which is its parent?  I hope it is not just the prefix of its name :)  it is OK to add and "extends" keyword or something

proofs/erc20.md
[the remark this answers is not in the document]
grosu: good, please explain.  if something confuses me, it will likely confuse others, too.

proofs/erc20.md
grosu: Can you separate them but still avoid duplication?  Introduce some notation in your DSL that you are extending another module, and then continue to use your overriding.

proofs/erc20.md
grosu: cool.  for now, add this as a comment; then we change it when we fix it

verification.md
grosu: ok, acceptable.  but add a comment stating the relationship between these two notations when you use each of them for the first time.

verification.md
grosu: report this issue to @dwightguth .  He may have a quick fix for us.

verification.md
grosu: Hm, this is tricky.  Please explain it as well as you can for now, including all the above.  We will then have to find a clean and general solution for it.  What is wrong with assuming them to be injective functions?

grosu:
Alright, done with revising @daejunpark 's comments.

proofs/erc20.md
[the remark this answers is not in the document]
daejunpark: Um, sorry, I made you more confusing. I think I should say it as DSLs instead of syntactic sugars. The `[function]` rules are not applied statically (if you mean static being applied before starting the execution). So, no issue with the recursions. Indeed, they are defined recursively.

proofs/erc20.md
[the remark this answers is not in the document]
daejunpark: Sorry it's the prefix separated by `-` to identify the parent. We can improve this DSLs though. The extends keyword sounds good.

verification.md
daejunpark: I'm not positive about it. It's quite complicated and messy. But will report it.

verification.md
daejunpark: Adding the universal injectivity axiom to z3 query is simply not sound.

Instantiating the injectivity axiom for each term created during the symbolic execution is already kinda done, since K considers non-functional symbols as constructor symbols. But when encoding z3 query, it doesn't add the universal injectivity axiom for them, which is still sound but not complete.

So, in some sense, the current symbolic execution engine is kind of doing what I explained in the above (except instantiating the injectivity axiom for each constructor term). But I'm not sure if it is still correct in terms of the new kore semantics and thus generating proof objects.

verification.md
daejunpark: @grosu Not entire sure if indeed it is good to do. If we want to extend the semantics, (even if it is just about defining DSLs that are nothing to do with the semantics itself), we still want to test it to make sure it doesn't screw up. In that sense, it is better to have separate modules for each language/DSL, but then, we have to kompile multiple version of extended semantics and test them every time, which may be too much overhead.

dwightguth:
Can this be closed now that we have the other PR? @ehildenb

ehildenb:
@dwightguth yes

PR #124, opened 2018-02-05
Added eip20 Token

iustines:
I have added the eip20 Token the transferFrom-succes still is giving OutOfMemoryError. Can you have a look?


msaxena2:
ok to test

PR #125, opened 2018-02-07
Cross platform build

msaxena2:
@ehildenb @nishantjr Merge?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -33,6 +33,7 @@ ocaml-deps: .build/local/lib/pkgconfig/libsecp256k1.pc
 		|| opam repository set-url k "$(K_SUBMODULE)/k-distribution/target/release/k/lib/opam"
 	opam update
 	opam switch 4.03.0+k
+	eval $$(opam config env) \
 	export PKG_CONFIG_PATH=$(PKG_CONFIG_LOCAL) ; \
 	opam install --yes mlgmp zarith uuidm cryptokit secp256k1 bn128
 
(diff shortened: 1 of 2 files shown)


msaxena2:
Also, @darko-marinov

dwightguth:
@msaxena2 have you tested this on a fresh installation of ArchLinux and High Sierra?

msaxena2:
High Sierra - Yes, I'm running 10.3.3, which I believe is the latest. Arch - @nishantjr will have to confirm.

dwightguth:
Can you please make a similar pull request [here](https://github.com/runtimeverification/rv-match/blob/master/k/k-distribution/INSTALL.md) and [here](https://github.com/runtimeverification/rv-match/blob/master/k/README.md)?

msaxena2:
@dwightguth Actually, it seems like running an OSX VM on OSX isn't too hard. I'll try the above on a fresh install.

dwightguth:
That would be helpful, yes.

Makefile
```diff
@@ -80,7 +81,7 @@ defn: $(defn_files)
 
 .build/ocaml/driver-kompiled/interpreter: $(ocaml_files) KRYPTO.ml deps
 	@echo "== kompile: $@"
-	eval $(shell opam config env) \
+	eval $$(opam config env) \
```
daejunpark: @msaxena2 what's it about?

Makefile
```diff
@@ -33,6 +33,7 @@ ocaml-deps: .build/local/lib/pkgconfig/libsecp256k1.pc
 		|| opam repository set-url k "$(K_SUBMODULE)/k-distribution/target/release/k/lib/opam"
 	opam update
 	opam switch 4.03.0+k
+	eval $$(opam config env) \
```
daejunpark: @msaxena2 why is it ok for ubuntu without this?

Makefile
dwightguth: I approve this. It should not have worked without this fix.

Makefile
dwightguth: Probably a different version of opam. This is fine though and I've seen issues in the past when running opam install without first running ``eval `opam config env`, so I'm inclined to say we should accept this change.

msaxena2:
@dwightguth I checked on a machine that didn't have rvk on it, and made a small change to the instructions. At this point, I don't think it's worth spend time to check the instructions on a fresh install. Let's get these changes in.

PR #126, opened 2018-02-07
update build instructions for mac os

dwightguth:
Improves install instructions as well as performs various minor fixes to the build system.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@
 all: build split-tests
 
 clean:
-	rm -rf .build/java .build/ocaml .build/logs tests/proofs
+	rm -rf .build/java .build/ocaml .build/logs tests/proofs .build/k/make.timestamp .build/local
 
 build: .build/ocaml/driver-kompiled/interpreter .build/java/driver-kompiled/timestamp
 
@@ -24,7 +24,7 @@ k-deps: $(K_SUBMODULE)/make.timestamp
 $(K_SUBMODULE)/make.timestamp:
 	git submodule update --init -- $(K_SUBMODULE)
 	cd $(K_SUBMODULE) \
-		&& mvn package -q -DskipTests
+		&& mvn package -q -DskipTests -U
 	touch $(K_SUBMODULE)/make.timestamp
 
 ocaml-deps: .build/local/lib/pkgconfig/libsecp256k1.pc
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 # Common to all versions of K
 # ===========================
 
-.PHONY: all clean deps ocaml-deps build defn sphinx split-tests \
+.PHONY: all clean deps k-deps ocaml-deps build defn sphinx split-tests \
 		test test-all vm-test vm-test-all bchain-test bchain-test-all proof-test proof-test-all
 
 all: build split-tests
@@ -18,7 +18,8 @@ K_SUBMODULE=$(CURDIR)/.build/k
 BUILD_LOCAL=$(CURDIR)/.build/local
 PKG_CONFIG_LOCAL=$(BUILD_LOCAL)/lib/pkgconfig
 
-deps: $(K_SUBMODULE)/make.timestamp ocaml-deps
+deps: k-deps ocaml-deps
+k-deps: $(K_SUBMODULE)/make.timestamp
 
 $(K_SUBMODULE)/make.timestamp:
 	git submodule update --init -- $(K_SUBMODULE)
--- a/Makefile
+++ b/Makefile
@@ -7,8 +7,7 @@
 all: build split-tests
 
 clean:
-	rm -r .build
-	find tests/proofs/ -name '*.k' -delete
+	rm -rf .build/java .build/ocaml .build/logs tests/proofs
 
 build: .build/ocaml/driver-kompiled/interpreter .build/java/driver-kompiled/timestamp
 
--- a/Makefile
+++ b/Makefile
@@ -71,14 +71,14 @@ defn: $(defn_files)
 
 # Java Backend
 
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -69,26 +69,21 @@ On ArchLinux:
 sudo pacman -S  base-devel rsync opam pandoc jre8-openjdk mpfr maven
 ```
```
msaxena2: I'm not sure you need to install XCode. You need just need the command line tools package, which most users would already have.

README.md
```diff
@@ -69,26 +69,21 @@ On ArchLinux:
...
 
-On OSX, using [Homebrew](https://brew.sh/):
+On OSX, using [Homebrew](https://brew.sh/), after installing XCode:
```
msaxena2: I'm not sure you need to install XCode. You need just need the command line tools package, which most users would already have.

README.md
```diff
@@ -69,26 +69,21 @@ On ArchLinux:
...
 brew tap caskroom/cask caskroom/version
 brew cask install java8
-brew install bison flex libtool autoconf gmp mpfr pandoc maven opam
-```
-
-Note that Homebrew install flex as a keg-only package. To install flex's library, use:
-
-```sh
-brew link --force flex
+brew install automake libtool gmp mpfr pkg-config pandoc maven opam python z3
```
msaxena2: OSX comes with python 2.7 pre-installed. Not sure you'd need to install it via homebrew.

README.md
dwightguth: I will update this to mention the command line tools explicitly.

README.md
dwightguth: I don't really know what to say except that python wasn't in the path when I built evm on a freshly upgraded high Sierra today, and this fixed it. Are you sure that's a correct statement?

dwightguth:
@msaxena2 please re-review.

Makefile
```diff
@@ -7,8 +7,8 @@
 all: build split-tests
 
 clean:
-	rm -r .build
-	find tests/proofs/ -name '*.k' -delete
+	rm -rf .build/java .build/ocaml .build/logs
+	-find tests/proofs/ -name '*.k' -delete
```
msaxena2: Is this change to `find` => `-find` intentional/necessary?

dwightguth: yes, find fails if tests/proofs doesn't exist, which it doesn't if you've just checked out the repository, so without this, the make clean target fails.

msaxena2:
@dwightguth Changes look good, and we should get these in asap. But looks like the PR has gone beyond just updating the Mac related instructions. I think either @nishantjr or @ehildenb should review the changes to the makefile and ubuntu sections, since I didn't write them. Also, I see you're still working on the PR, but before we merge, we should clean up the commits. I can take care of that you'd like, and merge.

Makefile
nishantjr: Would `[ ! -d tests/proofs ] || find tests/proofs/ -name '*.k' -delete` be more explicit and easier to understand?

ehildenb:
@dwightguth you can say "use `K_BIN` if set otherwise use `DEFAULT_VALUE`" by using `${K_BIN:-DEFAULT_VALUE}` in bash

kast-json.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python2
+#!/usr/bin/env python
```
nishantjr: Is this intended? On some systems where both `python2` and `python3` this can be a problem.

Makefile
ehildenb: I like @nishantjr's suggestion here.

dwightguth:
Jenkins: test this please

dwightguth:
Improve `make k-deps` a bit

ehildenb:
ah I see, my bad

dwightguth:
I will look at the test failures tomorrow morning

MrChico:
I can't build this version and it seems like the problem is that the module `Constants` imported in KRYPTO.ml cannot be found. Running `make all` yields;
```
[Error] Compiler: Had 1 parsing errors.
[Error] Internal: gcc returned nonzero exit code. See output for details.
File ".build/ocaml/driver-kompiled/constants.ml", line 1:
Error: I/O error: .build/ocaml/driver-kompiled/constants.ml: No such file or directory
File "KRYPTO.ml", line 1, characters 5-14:
Error: Unbound module Constants
File "_none_", line 1:
Error: Cannot find file KRYPTO.cmx
```

ehildenb:
@MrChico I think you need to run `make deps ; make all` now. The reason being that `make deps` always triggers a fetch from external sources, which is slow, so we wanted to separate it out into an explicit step.

dwightguth:
@MrChico There is an issue in the makefile causing it to continue past the first error it reaches. The error is actually "gcc returned nonzero exit code". This is happening because we no longer require or support the homebrew keg-only version of flex being linked. Can you try `brew unlink flex` and let me know if it fixes the issue on this branch? That will be the supported behavior going forward.

msaxena2:
Also, just to add onto to @dwightguth , you may want to check if you have flex (referred to as libl, not libfl on macs) in your linker's path. After unlinking flex via brew, try running "ld -ll" - that'd make sure you still have flex in your path.

dwightguth:
Jenkins: test this please

ehildenb:
@dwightguth commit messages

dwightguth:
This is now ready for final review.

nishantjr:
@dwightguth I force pushed a change that adds `z3` to ArchLinux install instructions.
Looks good to me.

MrChico:
Builds like a charm. Thanks!

PR #127, opened 2018-02-09
compute touched accounts of each transaction 

dwightguth:
computes the EIP 161 touched accounts of a transaction as it executes. This does not affect VM execution of a single transaction, but will eventually be necessary if we intend to verify entire blocks in KEVM.

### DIFF
--- a/evm.md
+++ b/evm.md
@@ -41,13 +41,14 @@ In the comments next to each cell, we've marked which component of the YellowPap
                       // Mutable during a single transaction
                       // -----------------------------------
 
-                      <output>        .WordStack </output>              // H_RETURN
-                      <memoryUsed>    0          </memoryUsed>          // \mu_i
-                      <callDepth>     0          </callDepth>
-                      <callStack>     .List      </callStack>
-                      <interimStates> .List      </interimStates>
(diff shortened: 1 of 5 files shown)


ehildenb:
@dwightguth should it ever be the case that the `touchedAccounts` state is rolled back? eg. on a `CALL` that throws an exception?

dwightguth:
The spec is a little unclear about this; I'm not entirely sure how it's supposed to go. I followed the same behavior that the Mantis client has; if this is the incorrect behavior, we will eventually find out because tests in the ethereum test suite will start to fail, at which point in time I can correct this. But this is a good start for the initial revision of this functionality.

evm.md
```diff
@@ -113,6 +113,7 @@ In the comments next to each cell, we've marked which component of the YellowPap
                       // ---------------
 
                       <activeAccounts> .Map </activeAccounts>
+                      <touchedAccounts> .Set </touchedAccounts>
                       <accounts>
```
ehildenb: I think perhaps this is better grouped with other TX data (perhaps in the `substateLog`), as data that is only pertinent to an individual transaction.

PR #130, opened 2018-02-13
Fixes #129

### LINKED ISSUE #129: Self modifying code?
Issues.md states:

> Currently in EVM, the *CODECOPY opcodes allow regarding program pieces as data, meaning that a translation back must always be maintained, because in theory a program can modify itself while executing.

The machine state (including memory) is defined in "9.4.1. Machine State" and the execution environment (including program code) is "9.3. Execution Environment."

CODECOPY is defined in the OPCODES table and brings code from the latter to the former.

How are you seeing that a program can modify itself?

fulldecent:


dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test this please

fulldecent:
Test passed.

Link broken, but test passed. Maybe the link is private and you can see it but I can't.

dwightguth:
Are you sure the link is broken? it works for me even when I'm not logged in. If it doesn't work for you, then something is wrong with our network setup or something.

PR #133, opened 2018-02-14
Changes to Core Semantics 

msaxena2:
Changes to two main evm semantics related files.

### DIFF
--- a/analysis.md
+++ b/analysis.md
@@ -23,7 +23,7 @@ This tool should be extended to take advantage of the symbolic execution engine 
 -   The mode `GASANALYZE` performs gas analysis of the program instead of executing normally.
 
 ```{.k .java .ocaml}
-    syntax Mode ::= "GASANALYZE"
+    syntax Mode ::= "GASANALYZE" [klabel(GASANALYZE)]
 ```
 
 We'll need to make summaries of the state which collect information about how much gas has been used.
--- a/data.md
+++ b/data.md
@@ -57,8 +57,8 @@ These can be used for pattern-matching on the LHS of rules as well (`macro` attr
 ```k
     syntax Int ::= chop ( Int ) [function, smtlib(chop)]
  // ----------------------------------------------------
-    rule chop ( I:Int ) => I %Int pow256 requires I  <Int 0  orBool I >=Int pow256    [concrete]
-    rule chop ( I:Int ) => I             requires I >=Int 0 andBool I  <Int pow256 // [concrete]
+    rule chop ( I:Int ) => I modInt pow256 requires I  <Int 0  orBool I >=Int pow256    [concrete]
+    rule chop ( I:Int ) => I               requires I >=Int 0 andBool I  <Int pow256 // [concrete]
 ```
 
 ### Boolean Conversions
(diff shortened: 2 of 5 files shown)


ehildenb:
do not merge yet, more changes to come

ehildenb:
@daejunpark  @msaxena2 I've pushed up my first round of review. Please see the commit messages marked with `!!!` for my issues with each commit. If you do have a fix, please directly fix the commit itself (not adding more commits afterward).

ehildenb:
@daejunpark if we make `pow256` a `macro`, will that solve the problem with fe67311?

ehildenb:
@dwightguth care to elaborate? would be good for us all to know about such issues with K.

dwightguth:
In the ocaml backend, macros that rewrite to functions perform syntactic replacement and the function applies each time the rule using the macro is applied, whereas pure functions with zero arguments are memoized and only ever invoked once.

ehildenb:
I see. So we could add another constant `pow256macro`, which can be used in the LHS of rules for matching purposes, while we keep using `pow256` in RHS of rules?

ehildenb:
@dwightguth another proposed solution, if we used the actual expanded out number (instead of `2 ^Int 256`) and made `pow256` a `macro`, we wouldn't take the performance hit and we would get matching capabilities on LHS of rules, right?

msaxena2:
@daejunpark I remember there was a specific reason to not use pow256, could you quickly elaborate? Ideally a macro should work, but not sure if the javabackend's proof compilation steps handle macros well, especially when they're in side conditions.

ehildenb:
@msaxena2 Daejun said above what the issue is. The number shows up on the LHS of the rule, so it won't actually match (the term will contain `BIG_NUMBER`, but it will be trying to match `pow256`). In the current setup, if we make `pow256` a `macro`, it will be trying to match `2 ^Int 256`, which also won't work.

msaxena2:
@daejunpark @ehildenb , Sorry. didn't fully understand the conversation above. Got the point now. So, your suggestion, to use a macro that desugars to the complete number should work fine, I guess.

msaxena2:
Regarding e35da4c - Is this going to be a performance hit? I can't recall exactly, but we did switch to tail recursion at some other place in the semantics in the semantics. @ehildenb note that if we decide to generalize the DSL (like we discussed yesterday), it'd be good to consider how the ocaml backend handles this. Perhaps @dwightguth can comment here. I don't have a strong opinion - this look readable to me, but I'll make the change if you guys think otherwise

ehildenb:
Well, the cases where we switched to tail recursion before were when we were building huge `WordStack`, with thousands of elements. A function signature will only have a handful of elements ever, so I don't think the performance will be affected.

ehildenb:
@daejunpark @dwightguth I've pushed up (tentative) changes which make `pow256` (and company) macros which go straight to the big expanded out number. I've also switched all uses of `%Int` to `modInt`. They are in separate commits so that they are easily dropped/squashed.

@dwightguth I would like to know what you think about the performance of this approach, and whether we will get correct matching behaviour on LHS of rules.

daejunpark:
@ehildenb 1b7b8b14ffb93b0e6cb1d31040ac59de525f6d8a
what i suggested is to have only the following single rule in the evm.md: (note that no side condition but having smt-lemma)
```
 rule chop ( I:Int ) => I modInt pow256 [concrete, smt-lemma]
```
and put the following in the verification.md:
```
 rule chop ( I:Int ) => I requires I >=Int 0 andBool I <Int pow256
```

Did you see the point here?

daejunpark:
regarding the macro of pow256, it would be great if it works.

dwightguth:
the performance hit was associated with running the `^Int` function so by replacing the powers of 2 with concrete integers it should be fine to use a macro, in my book.

PR #136, opened 2018-02-22
ERC20 Token Specifications

daejunpark:
ERC20 Token Specifications
- spec template
- spec parameters: both function-specific and program-specific ones
- script to instantiate spec template by parameters


ehildenb:
also subsumed

PR #139, opened 2018-02-23
Verification erc20 split

ehildenb:
@daejunpark version of Verification-ERC20 branch with fixes to README we discussed, and rebased on top of https://github.com/kframework/evm-semantics/pull/138


daejunpark:
made a quick fix of the typos in the readme. you can squash it if you want.

not sure what's the point of having the dummy history of the hkg.md, but that's ok, i don't want to delay to merge this any more. so, please merge if the test passes.

ehildenb:
@daejunpark , @msaxena2 felt that the individual commit messages on the HKG token he had were helpful to future contract developers. Because he had the entire history already intact (so it didn't require any work to add it), I conceded and put it in.

daejunpark:
Jenkins: test this please

daejunpark:
Jenkins: test this please

daejunpark:
@ehildenb I just wanted to let you know that I had to remove `tangle-deps` from the jenkins build commands for this.

daejunpark:
It turned out that the trusted rules are still needed. It sounds really scary. I guess I opened a can of worms, and it will take time to fix. Can we merge this for now so that @dwightguth can move forward? Once we fix the k prover, we can simply revert the last commit.

grosu:
@daejunpark what is *really scary*?  That sounds scary.
Scary that certain smart contract proofs do not go through and thus the results themselves may be wrong?  Or scary in the sense that you discovered more problems in the prover?

daejunpark:
@grosu The latter at this time. I found the root cause, but haven't found a good fix. Will work with @andreistefanescu. BTW, the found issue does not harm the soundness.

ehildenb:
@daejunpark before merging we need a better explanation of what the root cause of this is, preferably in the file next to the rules themselves as a `**TODO**: Explanation of root cause`. Otherwise, we'll simply forget about the fact that there are issues in K, and it will go unfixed forever.

For example, in the semantics where we've had differences between ocaml/java, we've been marking it explicitly, and that has helped @dwightguth to go through and figure out what needs to be merged between the two of them.

Please add an explanation in the file next to the lemmas (squash into the last commit), and remove the `FIXME` from the commit message.

daejunpark:
@ehildenb Done as you suggested. Could you take a look?

ehildenb:
@daejunpark why is there one more file to be generated for the `hobby` token than the others? See the updated `Makefile` for what I mean.

proofs/erc20/tmpl.k
```diff
@@ -97,4 +97,12 @@ module {MODULE}-SPEC
      andBool 0 <=Int CALL_DEPTH andBool CALL_DEPTH <Int 1024
      {REQUIRES}
 
+  // **TODO**: These semmingly unnecessary claims are to be an workaround for the K prover issue
+  // that unresolved functions are not fully evaluated in the normal semantics rule application steps.
+  // These should be dropped once the issue is fixed.
```
msaxena2: @daejunpark I was looking at the issue on Friday, and I don't really understand how putting these rules back in help. At least in the proof for viper's balanceOf case, the problem seemed to be that the `chop` rule from verification.md 
```
rule chop(I) => I requires 0 <=Int I andBool I <Int pow256
```
simplified expression in the term, but not in the lhs of the claim itself. Your K PR also seems to suggest a fix to that issue. So why is this required?

msaxena2: Also, just to be sure - do you want the chop to simplify the expression as much as possible? Would adding a concrete attribute to the rule above not work? At least for some the erc 20 proofs, this seemed to work, but it broke the sum-to-n proof.

daejunpark:
@ehildenb the approve function `hobby` has the following additional check:
```
if (_value != 0 && allowed[msg.sender][_spender] != 0) { return false; }
```
See https://github.com/ethereum/mist/issues/3301

proofs/erc20/tmpl.k
daejunpark: As I said, the normal semantics rule application failed to evaluate the unresolved terms that can be evaluated by applying the non-semantics rules, the spec rules in this case.

`[concrete]` attribute for the rule you mentioned should have no effect since the semantic rules cover that. Do NOT add that.

PR #141, opened 2018-02-27
Fix the path of a ./kvm prove example

pirapira:
Apparently, the file moved but the `README` description did not follow it.


dwightguth:
Jenkins: test this please

PR #148, opened 2018-03-05
Split Makefile for proofs (submodule)

daejunpark:


ehildenb:
@daejunpark not sure I like the way this is. I think a better solution would be to model it off of the `tests/ethereum-tests` submodule. There, the tests are still run from our `Makefile` (so that we can see in the KEVM repo what the tests are), but they are stored in the submodule. You could even think that others might adopt the same notation for specifying contract properties, and then mentioning `KEVM_DIR` in the Makefile doesn't make sense for them. So I think the submodule makefile should only have the code for splitting the proofs, not for running them, and the KEVM submodule has the code for running them as tests.

PR #154, opened 2018-03-11
Balance underflow fix

ehildenb:
This branch demonstrates a test which should not behave as it does.


ehildenb:
Turned into issue #166

PR #156, opened 2018-03-12
tests/ethereum-tests: update tests submodule

ehildenb:
updating to a newer testset


ehildenb:
Failing tests have been logged in the repo in file `failing-tests`. Looks like some issue with `shl`.

ehildenb:
Also, it's not clear that the added "slow" tests are also failing. They may have just run out of memory.

ehildenb:
Recorded as an issue #165 instead (moved to lower priority).

PR #183, opened 2018-04-10
Remove vmtest dependencies

ehildenb:
This removes the dependencies on VMTEST mode from the file `evm.md` (except for in the top-level execution loop).

Note that we've manually disabled problematic VMTests for now, but https://github.com/ethereum/tests/pull/444 should remove all problematic tests for us eventually.


ehildenb:
I'm currently investigating why the more specific exclusion supplied by @pirapira does not work.

ehildenb:
Turns out it doesn't have issues with the more specific exclusion, should work like this.

ehildenb:
Jenkins: test this please

ehildenb:
Jenkins: test this please

Issue #174, opened 2018-03-28
Update proofs to use `BYZANTIUM` instead of `DEFAULT`.


ehildenb:
The PR https://github.com/kframework/evm-semantics/pull/167 revealed that the Vyper proofs, which do not have `INVALID` in them, are still ending in `EVM_INVALID_INSTRUCTION` exceptions.

This is a combination of two things, (1) the proofs are run in `DEFAULT` mode instead of `BYZANTIUM`. (2) the current disassembler maps anything that is undefined to `INVALID` when it should instead map them to `UNDEFINED`.

These issues need to be fixed before https://github.com/kframework/evm-semantics/pull/167 can be merged.

ehildenb:
Not relevant for this repository anymore as tests are stored in https://github.com/runtimeverification/verified-smart-contracts.

PR #188, opened 2018-04-11
kevm, Makefile, tests/interactive/*.out: default to BYZANTIUM schedule

ehildenb:
-   VMTESTS run on DEFAULT schedule
-   output of interactive tests is updated

Addresses #174


ehildenb:
Need to also make sure the proofs are updated correctly.

PR #195, opened 2018-04-18
evm.md: add concrete attribute to #accountEmpty rule

yzhang90:


ehildenb:
@yzhang90 did you try on your machine switching the boolean condition to a requires clause instead? I would prefer that fix, though it also seems ugly/hacky.

yzhang90:
@ehildenb Why do you prefer switching the boolean condition to a requires clause?

yzhang90:
But `concrete` attribute has already appeared in the EVM semantics and they have to learn it anyway. My concern is that the abuse use of `concrete` attribute makes symbolic execution deviated from the concrete execution (looks like two set of semantics).

ehildenb:
That's another concern I have. Can you answer my question? Have you tried putting it in the requires clause? Does it work? I'm curious anyway, because I would like to know more about what makes K work and not.

I assume it's not equivalent, because by putting `concrete` you will probably add lemmas to make it reason about it or something?

yzhang90:
I tried and it worked. 

I'll add additional lemmas in `verification.k` to reason about `#accountEmpty`.

yzhang90:
Similarily, I need to change the rule
 ```
rule Cgascap(SCHED, GCAP, GAVAIL, GEXTRA)
      => #if GAVAIL <Int GEXTRA orBool Gstaticcalldepth << SCHED >> #then GCAP #else minInt(#allBut64th(GAVAIL -Int GEXTRA), GCAP) #fi
```

PR #205, opened 2018-04-27
add #range* predicates + update k and proofs submodules

daejunpark:


edsl.md
```diff
@@ -270,5 +270,19 @@ Specifically, `#hashedLocation` is defined as follows, capturing the storage lay
...
+```
+
+### Utilities
+
+```k
+    syntax Bool ::= "#range" "(" Int "<"  Int "<"  Int ")" [function]
+                  | "#range" "(" Int "<"  Int "<=" Int ")" [function]
+                  | "#range" "(" Int "<=" Int "<"  Int ")" [function]
+                  | "#range" "(" Int "<=" Int "<=" Int ")" [function]
+ // -----------------------------------------------------
```
ehildenb: Formatting, extend the `----` to the end of `[function]`.

daejunpark:
@ehildenb Ah you're right. I was worried about the performance, but shouldn't be that problem. Will revise it.

ehildenb:
Jenkins: test mantis

daejunpark:
@ehildenb please re-review and merge if it looks good to you

PR #206, opened 2018-04-30
Git ci setup

ehildenb:
Simple PR which enables me to easily/quickly get performance/failing numbers across a branch.

No real changes otherwise.


ehildenb:
@dwightguth I would like to merge this branch next, it's fairly trivial but will make it easier for me to do lightweight performance profiling across future PRs. Please review soon.

ehildenb:
I would like to keep a list of the smoke and performance tests in the repo, but didn't have a sensible format for doing so until now. The script that makes use of it is at https://github.com/cynic-net/git-ci/tree/record-build-artifacts.

dwightguth:
Yeah that's not what I'm asking about, it's the sort-logs command that seems misplaced.

ehildenb:
Well, the only better place I could think is in a different script, but I think it has to stay in-repo. The point of `git-ci` is that it's completely agnostic about the types of files which it tracks as test results, and so it can't know that it's supposed to sort this file before storing it or not.

Basically `git-ci` builds a secondary history of the files that are traditionally untracked but you tell it to track them as test output. Then you can just use standard git commands to display information about them. So in this case, if the files aren't sorted before storage, then `git-ci` becomes nearly useless.

ehildenb:
Jenkins: test mantis

dwightguth:
I'd still really like CI stuff in a separate script from the script that users use to execute programs.

ehildenb:
Well, that would require migrating all of the `./kevm` commands like `test`, `test-profile`, `get-failing`, and `interpret` as well. While it may be a bit cleaner, I don't have the time/energy for it now (`./kevm` does a decent amount of environment setup and such) and would like to make progress.

Perhaps open an issue about it instead?

PR #209, opened 2018-05-07
tests/proofs: update proofs submodule

daejunpark:
for test


daejunpark:
@ehildenb buildbot failed due to 1200s timeout, but one spec takes ~4000s.

ehildenb:
I will up the proof timeout time.

PR #213, opened 2018-05-15
BadJumpdest fix

MrChico:
I know this feature is in the works on the opcode-basic-blocks branch, but it seems like a fairly straight forward feature to merge already. It reduces a lot of overhead in the proof search. 

I guess it's up to your PR policy whether to merge this or to wait for the opcode-branch.

### DIFF
--- a/evm.md
+++ b/evm.md
@@ -350,7 +350,6 @@ The `#next` operator executes a single step by:
     rule <k> #exceptional? [ OP ]
           => #invalid?     [ OP ]
           ~> #stackNeeded? [ OP ]
-          ~> #badJumpDest? [ OP ]
           ~> #static?      [ OP ]
          ...
          </k>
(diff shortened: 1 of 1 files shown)


evm.md
```diff
@@ -1143,12 +1112,26 @@ The `JUMP*` family of operations affect the current program counter.
...
+         <program> PGM </program>
+      requires notBool (DEST in_keys(PGM))
 
     syntax BinStackOp ::= "JUMPI"
  // -----------------------------
-    rule <k> JUMPI DEST I => . ... </k> <pc> _      => DEST          </pc> requires I =/=Int 0
-    rule <k> JUMPI DEST I => . ... </k> <pc> PCOUNT => PCOUNT +Int 1 </pc> requires I  ==Int 0
+    rule <k> JUMPI DEST I => . ... </k>
+         <pc> PCOUNT => PCOUNT +Int 1 </pc>
+	     requires I ==Int 0
```
ehildenb: Why not pattern match for the `I ==Int 0` instead?

ehildenb:
It looks like you did intentionally change the pattern-matching to `==Int`. I'm doing a short performance analysis, but if you can provide an example proof where this impacts performance, I'm interested in adding it the performance test-suite. Also, it makes me wonder if other occurrences of pattern-matching on 0 directly are causing issues.

ehildenb:
add to whitelist

ehildenb:
Jenkins: add to whitelist

ehildenb:
Hmmmm, ok, so it seems that having the `==Int` is necessary for the prover to not continue executing the dead branch in the `sum-to-n` program (at least, that's what I've gathered about this). I guess this boils down to K not being able to prove that matching on `0` and the side-condition `=/=Int 0` are mutually exclusive and cover the entire range. I'll ask around tomorrow and see if we can make K smarter about this.

ehildenb:
Closing this in favor of one without a merge commit #214

PR #217, opened 2018-05-26
evm: no need to push/pop callstack on every opcode

ehildenb:
Seemed to pass conformance tests for me, should speed up execution a bit. Did not notice much speedup for the conformance tests locally, but perhaps the proofs will benefit.


ehildenb:
Jenkins: test this please

ehildenb:
Looks like tests pass @dwightguth. I had to disable the thing that made Jenkins fail if coverage report was not found for the Normal Jenkins job.

PR #218, opened 2018-05-31
casper iszero test

tomtomjhj:


ehildenb:
@tomtomjhj please look at `git log`, or even better, add `git hist` from https://github.com/ehildenb/dotfiles/blob/76b160b4037153736d6cc01c4af7aaf7ee9c1ad1/git/gitconfig#L48 as an alias and look at `git hist`.

You will see the commit message conventions we've had.

Also, we will *not* merge PRs which are not based on master, so you will need to rebase this PR ot be on top of `master`.

PR #221, opened 2018-06-05
.build/k: update k submodule

### LINKED ISSUE #220: Build issue due to ocaml and gcc mismatch
I'm running `gcc` 8.10 so I bump into an issue when the ocaml part of running `make deps` starts. The issue was fixed in this commit in ocaml: https://github.com/ocaml/ocaml/commit/ebcc2f8594138a5ba6b020083fc42d57a914433f
  
Is there a way to install with a newer ocaml version which includes this commit?

Thanks!

dwightguth:
Fixes #220


dwightguth:
@ehildenb this is ready for review.

PR #223, opened 2018-06-11
Clean up JelloPaper with its own intro/index page and clearer sidebar navigation

pdaian:
Currently, jellopaper.org only shows the README by default, and requires the users to click some counterintuitive navigation to explore the *actual* semantics.

This fix moves the semantics into the sidebar as a navigable first-class citizen, and adds a front splash page for the Jello Paper whose contents are shown below:

new jellopaper splash

### DIFF
--- a/.build/sphinx-docs/index.rst
+++ b/.build/sphinx-docs/index.rst
@@ -1,0 +1,47 @@
+JelloPaper: Human Readable Semantics of EVM in K
+================================================
+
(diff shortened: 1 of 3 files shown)


pdaian:
@ehildenb particularly wondering if you'd like to change the order of those documents in the index ToCtree; if so, feel free to commit directly, I have no preference on those.

ehildenb:
Jenkins: test mantis

PR #227, opened 2018-06-13
edsl.md: support for uint256

MrChico:


daejunpark:
@MrChico looks good to me. thanks!

BTW, it's support for `int256` not `uint256`, right?

MrChico:
Yeah, should be int256, that's a typo

PR #230, opened 2018-06-18
Makefile: separate repository dependencies from system dependencies

ehildenb:
Now you can call `make repo-deps` to avoid doing any of the user-wide ocaml stuff, and `make system-deps` if you want to do the system-wide ocaml stuff.

I've switched over the Jenkins CI to use `repo-deps` instead of `deps`, so that we avoid the race condition we've been having with the mantis vs normal test-suites with locking the user-wide ocaml directories.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,7 +18,8 @@ LUA_PATH:=$(PANDOC_TANGLE_SUBMODULE)/?.lua;;
 export TANGLER
 export LUA_PATH
 
-.PHONY: all clean deps k-deps tangle-deps ocaml-deps plugin-deps build build-ocaml build-java build-node defn split-tests \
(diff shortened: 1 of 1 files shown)


ehildenb:
@MrChico you also expressed interest in this separation. Now instead of `make deps`, I recommend using `make repo-deps`. You still need to do `make deps` at least once on the machine to setup OCaml stuff if you want concrete execution.

ehildenb:
Jenkins: test mantis

ehildenb:
Jenkins: test mantis

ehildenb:
Jenkins: test mantis

ehildenb:
Jenkins: test mantis

ehildenb:
Jenkins: test mantis

ehildenb:
@dwightguth looks like the tests will pass on this one, ready for review.

PR #233, opened 2018-06-24
kevm: allow extra options passed on command line for prover

ehildenb:
Passes options through to `./kevm prove`.


denis-bogdanas:
Looks good. We don't use this script btw. We call kprove directly as recommended by Daejun.

PR #239, opened 2018-07-12
edsl.md: add #array and #uint8 ABI encoding

daejunpark:


ehildenb:
@daejunpark I believe this is redundant with https://github.com/kframework/evm-semantics/pull/238/files

daejunpark:
@ehildenb that's my bad communication.

PR #241, opened 2018-07-13
Fix #bytes encoding and add #bytes(WordStack) encoding

tomtomjhj:


daejunpark:
@tomtomjhj could you explain a bit more about the encoding bug? it's opposite to my intuition. you mean the correct encoding pads zeros in the right? so, `#bytes( 10, 255 )` means that ` 11...11000....000` instead of `000...00011...11`?

tomtomjhj:
@daejunpark [yes](https://solidity.readthedocs.io/en/develop/abi-spec.html#use-of-dynamic-types)

daejunpark:
@yzhang90 please review this, as i remember that you implemented #bytes

edsl.md
```diff
@@ -153,9 +159,21 @@ where `F1 : F2 : F3 : F4` is the (two's complement) byte-array representation of
...
+ // --------------------------------------------------------------------
+    rule  #padRightToWidth(N, WS) => #padRightToWidthAux(N -Int #sizeWordStack(WS), WS, .WordStack)
+
+    syntax WordStack ::= #padRightToWidthAux ( Int , WordStack , WordStack ) [function]
+ // -----------------------------------------------------------------------------------
+    rule #padRightToWidthAux(0, WS, ZEROS) => WS ++ ZEROS
+
+    rule #padRightToWidthAux(N, WS, ZEROS) => #padRightToWidthAux(N -Int 1, WS, 0 : ZEROS)
+      requires N >Int 0
+
```
ehildenb: I think that `#padRightToWidth` should probably be put in `data.md` next to `#padToWidth` definition, @tomtomjhj.

yzhang90:
I took a look at the ABI spec again and I think jerry’s impl is correct. Btw, maybe add [concrete] to the padRightToWidth rule?

tomtomjhj:
@yzhang90 the contents inside `#byte` might be symbolic. [example](https://github.com/runtimeverification/verified-smart-contracts/blob/gnosis/gnosis/gnosis-spec.ini#L174)

daejunpark:
@yzhang90 why `[concrete]`? the call data could be symbolic, right?

PR #242, opened 2018-07-20
Add git to dependencies

asymmetric:
`git` is [used](https://github.com/kframework/evm-semantics/blob/cc9c659510383215e5333d7f60088a2fdf2d120d/Makefile#L56) in the `Makefile`.


ehildenb:
Jenkins: add to whitelist

PR #249, opened 2018-08-17
Rule tags, smt translations and hashing conveniences

MrChico:
This includes the uint48 commit so can either follow or subsume https://github.com/kframework/evm-semantics/pull/247

### DIFF
--- a/edsl.md
+++ b/edsl.md
@@ -302,14 +302,25 @@ Specifically, `#hashedLocation` is defined as follows, capturing the storage lay
     rule #hashedLocation("Vyper",    BASE, OFFSET OFFSETS) => #hashedLocation("Vyper",    keccakIntList(BASE) +Word OFFSET, OFFSETS)
     rule #hashedLocation("Solidity", BASE, OFFSET OFFSETS) => #hashedLocation("Solidity", keccakIntList(OFFSET BASE),       OFFSETS)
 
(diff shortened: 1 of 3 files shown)


edsl.md
```diff
@@ -302,14 +302,25 @@ Specifically, `#hashedLocation` is defined as follows, capturing the storage lay
...
       requires 0 <=Int V andBool V <Int pow256
+      
+    syntax IntList ::= wordStack2IntList( WordStack )      [function]
+                     | wordStack2IntList( WordStack, Int ) [function]
+ // -----------------------------------------------------------------
+    rule wordStack2IntList( WS ) => wordStack2IntList(WS, #sizeWordStack(WS) /Int 32)
+        requires #sizeWordStack(WS) %Int 32 ==Int 0
+    rule wordStack2IntList( WS , N ) => #asWord(WS [0 .. 32]) wordStack2IntList(#drop( 32, WS), N -Int 1)
+        requires N >Int 1
+    rule wordStack2IntList( WS , 1 ) => #asWord(WS [0 .. 32]) .IntList
```
daejunpark: This function may be fragile. First, `wordStack2IntList( .WordStack )` is not defined. Also, it doesn't capture the case where `WS` is not aligned with 32 bytes. Of course, we can use this function carefully in the well-defined cases, but who knows.

PR #229, opened 2018-06-14
Builtin bitrange

ehildenb:
Testing out https://github.com/kframework/k5/pull/99/

PR #253, opened 2018-09-05
data: switch over to using bitRangeInt builtin

ehildenb:
@dwightguth Github isn't smart enough to realize that the earlier commits in #229 were merged already, because the PR was opened before those commits were in `master`. So I've opened a new PR with the same content, to appease Github and show the correct diff here.


dwightguth:
looks like mantis timed out because of concurrent build load. I would try to restart the mantis job after Jenkins is less busy. My guess is it will pass then.

ehildenb:
Jenkins: test mantis

PR #254, opened 2018-09-06
Added Kore as a submodule

traiansf:


traiansf:
One issue that I see is that for both `stack` and `z3`, the Kore backend needs newer versions than those provided by package managers for e.g., Ubuntu Bionic.  I've documented that, but I still don't like it.

ehildenb:
@traiansf have you run any of the tests with this? Do you have a rough estimate about how many pass?

Also, if you post here how to invoke the Haskell backend (the command to use), I'll add a `./kevm haskell ...` command for running a program with the Haskell backend.

traiansf:
@ehildenb so far I haven't run any tests.  I was thinking next to modify `kevm` but if it's easier for you to do it, please take a look at  https://github.com/kframework/kore/blob/master/src/main/k/working/imp-monosorted/Makefile

We basically call `kast --sort K --kore` on the program to generate the input pattern, then call `stack exec -- kore-exec imp.kore --module IMP --is-program --pattern <pattern_file>`  to execute it.

However, I would not expect much (if anything) to run to completion yet (we don't even support builtin collections yet).  I thought of making this integration to better debug and see what works and what not.

ehildenb:
Jenkins: test this please

traiansf:
@ehildenb I'm now working on a PR to a Haskell backend for K ( https://github.com/kframework/k/pull/147 ) which should make invocation of the haskell-backend easier (through krun).

PR #260, opened 2018-09-28
remove nondeterministic map/set lookups

dwightguth:
This eases the ability to execute evm with the upcoming llvm backend by replacing the non-deterministic choice of map and set elements with a conversion to a list of unspecified order followed by a sequence of deterministic lookups.


evm.md
```diff
@@ -758,6 +763,8 @@ After executing a transaction, it's necessary to have the effect of the substate
            )
            ...
          </accounts>
+
+    rule #deleteAccounts(.List) => .
```
ehildenb: Wrap in a `<k>` cell, as this is not a `function`.

dwightguth:
As a matter of fact, no, these are the only Map/Set choice operations in the entire semantics. The reason I am not using kompile to rewrite these rules is because I have just transformed the rules into a strictly less non-deterministic behavior than the actual behavior we want kompile to infer automatically for map/set choices. Essentially, given the K term `SetItem(foo(BAR)) S:Set` on the left hand side, the intended K semantics is that we try each element in the set looking for one that matches `foo(BAR)` and if we find one, proceed with the rest of the match with the understanding that if the match fails at any point (eg because of a term `requires baz(BAR)` which evaluates to false, we must backtrack and try the remaining elements of the set. The meaning of the code I wrote in this pull request is quite different: namely, we first pick some unspecified total ordering of the elements in the set / keys in the map, and then proceed by matching the elements one at a time in that exact order. This is contrary to the behavior we want in the general case: in general, there is no guarantee that the elements in a set that is being iterated over will or will not match when the iteration proceeds in any given specific order. Thus the only safe way to proceed in general is to find some specific element each time which causes the match to succeed. However, there is a particular property to these rules which I am currently taking advantage of here. Namely, no matter what element is chosen first, some rule will successfully match it, and after all elements are matched, the final configuration is the same regardless of the order chosen. This allows us to be certain that it is always both successful nd deterministic no matter what ordering we pick; which is why we can take advantage of Set2List when in the general case we cannot.

Tl;dr: these are the only choices in the semantics and this transformation is not suitable for the general case of choices, however, it's perfectly safe in these specific cases and thus I wrote this PR so that we don't have to implement the general case in the llvm and Haskell backends in order to get EVM working.

dwightguth:
Btw don't merge this yet; I realized just now that Set2List was never implemented in the Java backend, which suggests we need better test coverage. I'm going to add the code needed to execute a single blockchain test with the java backend tomorrow.

dwightguth:
This should be ready for review now @ehildenb

ehildenb:
I see, these are the only rules which do not contain another part of the LHS which tells us specifically which map/set item to match.

Sorry I came across as rude, thank you for the patch!

PR #268, opened 2018-10-18
Fix hard coded path in llvm backend build

dwightguth:
Now that we have merged the blockchain-k-plugin changes, we update the submodule and point the path to the copy of the file inside the repo.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ export LIBRARY_PATH
 export PKG_CONFIG_PATH
 
 K_SUBMODULE:=$(BUILD_DIR)/k
-PLUGIN_SUBMODULE:=plugin
+PLUGIN_SUBMODULE:=$(abspath plugin)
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -178,7 +178,7 @@ endif
 	eval $$(opam config env) \
 		&& ${K_BIN}/kompile --debug --main-module ETHEREUM-SIMULATION \
 						    --syntax-module ETHEREUM-SIMULATION .build/ocaml/driver.k --directory .build/llvm \
-						    --backend llvm -ccopt ~/blockchain-k-plugin/plugin-c/crypto.cpp \
+						    --backend llvm -ccopt `pwd`/plugin/plugin-c/crypto.cpp \
```
ehildenb: Can we use variable `PLUGIN_SUBMODULE` instead of `pwd/plugin/plugin-c` for consistency?

PR #274, opened 2018-11-09
Update haskell backend

vladciobanu:
This outdates the `haskell-backend` branch (should be safe to remove) and makes `haskell` (through `kore`) work with the current semantics.

I updated both the `k` and `kore` submodules to their respective latest master and updated the `Makefile` and `kevm` to correctly support the `haskell` backend.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,9 +76,9 @@ $(PLUGIN_SUBMODULE)/make.timestamp:
 $(KORE_SUBMODULE)/make.timestamp:
 	@echo "== submodule: $@"
 	git submodule update --init -- $(KORE_SUBMODULE)
-	cd $(KORE_SUBMODULE_SRC) \
-		&& stack build kore:exe:kore-exec
-	touch $(KORE_SUBMODULE)/make.timestamp
+	cd $(KORE_SUBMODULE) \
+		&& stack install --local-bin-path $(abspath $(KORE_SUBMODULE))/bin kore:exe:kore-exec
+	touch ${KORE_SUBMODULE}/make.timestamp
 
 ocaml-deps: .build/local/lib/pkgconfig/libsecp256k1.pc
 	opam init --quiet --no-setup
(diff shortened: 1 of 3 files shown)


ehildenb:
Jenknis: add to whitelist

ehildenb:
Jenkins: add to whitelist

ehildenb:
@vladciobanu please look at `git log` (or even better, `git hist` https://github.com/ehildenb/dotfiles/blob/08a5bafaac1f060da62913d57b8ad08eb01828a8/git/gitconfig#L48) and amend your commits to follow our naming scheme (that is, `list of files changed: description`).

vladciobanu:
@ehildenb  Huh, I checked the Contributing section from the README and didn't see anything about it. Did I miss it? Should be updated otherwise.

Should I make a longer message, possibly multi-line for the commit, or would you rather I make this into multiple commits?

ehildenb:
@vladciobanu I think 2 commits would be good. (1) updating submodules, labeled something like `.build/{k,kore}: update submodules`, and (2) adding the build tooling, something like `kevm, Makefile: support for building/running Haskell backend`

ehildenb:
Also @vladciobanu don't forget to fix the documentation in the `./kevm help` string that is printed. See the end of the `kevm` script.

vladciobanu:
@ehildenb Does this look okay?

vladciobanu:
@ehildenb Thanks! It does not run on *any* EVM program, but so far I was able to run

```
$ ./kevm run-haskell tests/ethereum-tests/VMTests/vmArithmeticTest/add0.json
$ ./kevm run-haskell tests/ethereum-tests/VMTests/vmTests/suicide.json
```

I did not check all, and I know some fail (for example, `vmArithmeticTest/mul[N].json` do not work, and I think I also tried a few bitwise tests and they also failed).

ehildenb:
Haha, it seems to fail pretty hard on the EVM programs I've tried, but it at least parses/begins execution and prints out a final incorrect configuration of `add0.json`, which is all I was looking for (so that others can begin poking at it as well).

PR #281, opened 2018-11-28
New map model

yzhang90:


daejunpark:
could you drop the evm.smt2 file? i moved it to the verified-smart-contract repo.

yzhang90:
done. Do you have idea why jenkins failed?

PR #278, opened 2018-11-27
[WIP] Test with the haskell backend

15joeybloom:

PR #282, opened 2018-11-30
Test with haskell backend

15joeybloom:
Moved here from #278

### DIFF
--- a/kevm
+++ b/kevm
@@ -110,9 +110,14 @@ run_test() {
             local ocaml_output_file="$(mktemp "$test_file.ocaml.out.XXXXXX")"
             local haskell_output_file="$(mktemp "$test_file.haskell.out.XXXXXX")"
             trap "rm -rf $ocaml_output_file $haskell_output_file" INT TERM EXIT
-            run_krun ocaml "$test_file" "$@" > "$ocaml_output_file"
-            run_krun haskell "$test_file" "$@" > "$haskell_output_file"
-            pretty_diff "$ocaml_output_file" "$haskell_output_file"
+
+            # Ignore errors from krun
+            run_krun ocaml "$test_file" "$@" > "$ocaml_output_file" || true
+            run_krun haskell "$test_file" "$@" > "$haskell_output_file" || true
+
+            # Ignore whitespace diffs until we get the generated counter issue
+            # fixed
+            pretty_diff -w "$ocaml_output_file" "$haskell_output_file"
             ;;
         ocaml)
             case "$test_file" in
@@ -133,6 +138,7 @@ run_test() {
             ;;
         *)
             echo "Cannot test with $backend backend!"
+            exit 1
             ;;
     esac
 }
--- a/kevm
+++ b/kevm
@@ -55,17 +55,11 @@ run_krun() {
     run_env "$run_file"
     export K_OPTS=-Xss500m
     case "$backend" in
-        haskell)
-            args=(--haskell-backend-command $build_dir/kore/bin/kore-exec)
-            ;;
-        ocaml)
-            args=(--interpret)
-            ;;
-        *)
-            args=
-            ;;
+        haskell)    args=(--haskell-backend-command $build_dir/kore/bin/kore-exec)  ;;
+        ocaml)      args=(--interpret)                                              ;;
+        *)          args=()                                                         ;;
     esac
-    krun --directory "$build_dir/$backend/" -cSCHEDULE="$cSCHEDULE" -pSCHEDULE='printf %s' -cMODE="$cMODE" -pMODE='printf %s' "$run_file" "${args[@]}" "$@"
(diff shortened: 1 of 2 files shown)


kevm
```diff
@@ -54,11 +54,18 @@ run_krun() {
...
     export K_OPTS=-Xss500m
-    if [[ "$backend" == "haskell" ]]; then
-        krun --directory "$build_dir/$backend/" -cSCHEDULE="$cSCHEDULE" -pSCHEDULE='printf %s' -cMODE="$cMODE" -pMODE='printf %s' "$run_file" "$@" --haskell-backend-command "$build_dir/kore/bin/kore-exec"
-    else
-        krun --directory "$build_dir/$backend/" -cSCHEDULE="$cSCHEDULE" -pSCHEDULE='printf %s' -cMODE="$cMODE" -pMODE='printf %s' "$run_file" "$@"
-    fi
+    case "$backend" in
+        haskell)
+            args=(--haskell-backend-command $build_dir/kore/bin/kore-exec)
+            ;;
```
ehildenb: Let's shorten these to one line each:

```
haskel) args=(...) ;;
ocaml)  args=(...) ;;
```

Line up the `a` in `args` and the `;;` for each line.

For the default case can we do `args=()` instead of `args=`? It looks weird.

kevm
```diff
@@ -95,30 +102,54 @@ run_interpreter() {
...
+        *)
+            # Only care about --backend parameter if it's not a proof test or
+            # interactive test
+            case "$backend" in
+                haskell)
+                    local output_file="$(mktemp "$test_file.out.XXXXXX")"
+                    trap "rm -rf $output_file" INT TERM EXIT
+
+                    # Ignore whitespace diffs until we get the generated counter issue
+                    # fixed
```
ehildenb: compress comment to 1 line.

kevm
```diff
@@ -157,33 +188,48 @@ run_get_failing() {
...
+                shift 2
+                ;;
+        *)
+            params+=("$1")
+            shift
+            ;;
+    esac
+done
+
+set -- "${params[@]}"
```
ehildenb: This seems more complicated than we need. How about we *only* check if `--backend` is supplied as the *first* argument, and otherwise assume it's not supplied at all? Then can be done with an `if [[ $# -gt 0 ]] && [[ $1 == '--backend' ]]` check.

kevm
```diff
@@ -157,33 +188,48 @@ run_get_failing() {
...
+    sort-logs)    run_sort_logs                         ;;
+    get-failing)  run_get_failing               "$@"    ;;
 
     *) echo "
     normal usage
     ============
 
-        $0 [run|run-java|run-haskell|debug|search] <pgm>  <K args>*
-        $0 prove                                   <spec> <K args>*
+        $0 run            [--backend <backend>] <pgm>   <K args>*
```
ehildenb: Here we even explicitly say that it must come as the first argument after the `run_command`.

kevm
```diff
@@ -95,30 +102,54 @@ run_interpreter() {
...
+                    # fixed
+                    run_krun haskell "$test_file" "$@" > "$output_file" \
+                        || pretty_diff -w "tests/templates/output-success.json" "output_file"
+                    ;;
+                ocaml)
+                    run_interpreter "$test_file"
+                    ;;
+                *)
+                    echo "Cannot test with $backend backend!"
+                    exit 1
```
ehildenb: Use `die "Cannot test with $backend backend!"` instead of these two lines.

kevm
15joeybloom: I tried `args=()` and it doesn't work because then `${args[@]}` is an unbound variable. But what I can do instead is `${args[@]-}` so that if `args[@]` is unbound then we use the empty string.

kevm
```diff
@@ -96,36 +96,36 @@ run_interpreter() {
...
+    expected_file="tests/templates/output-success.json"
+    output_file="$(mktemp "$test_file.out.XXXXXX")"
+    trap "rm -rf $output_file" INT TERM EXIT
+
     case "$test_file" in
         *proofs/*)
             run_proof "$test_file" --debug "$@"
             ;;
         *interactive/*)
-            local expected_file="$1" ; shift
```
15joeybloom: Don't we still need this?

ehildenb: No, actually we don't. I just hardcode it above now.

ehildenb: I should update the `Makefile` though, 1 sec.

It's only used by `test-interactive`, not by any of the other ones. And even in that case it is hardcoded to `tests/templates/output-success.json`

ehildenb: Ah crap, we do need it :/ I'll push a fix.

ehildenb: @15joeybloom does it look correct now?

15joeybloom: lgtm!

PR #283, opened 2018-12-12
Separate out concrete execution performance test-set for Haskell backend

ehildenb:
This PR does:

-   Fixes the way that success is checked for on the Haskell backend.
-   Run the Makefile with `TEST_BACKEND` specified so that each test-set can be run on each backend.
-   Adds a target `test-vm-normal` which allows running the VMTests in `NORMAL` mode so the haskell backend can be compared performance-wise to the other backends.
-   Adds a target `test-vm-haskell-perf` which runs a specific set of slow VMTests on the Java and Haskell backend for measuring the slowdown. These tests are >4x the median slowdown over the testset.


ehildenb:
@ttuegel or @traiansf if you could try out:

```
make clean
make deps haskell-deps
make build-haskell build-java -j2
make test-vm-haskell-perf
```

and let me know if the haskell profiling options have been implemented properly (commit 783db8a) that would be helpful.

PR #284, opened 2018-12-12
Makefile: added haskell-defn command to build Haskell definition

traiansf:
Extracted a separate `haskell-defn` command to be able to use it from `verified-smart-contracts`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,8 @@ KORE_SUBMODULE:=$(BUILD_DIR)/kore
 KORE_SUBMODULE_SRC:=$(KORE_SUBMODULE)/src/main/haskell/kore
 
 .PHONY: all clean deps repo-deps system-deps k-deps tangle-deps ocaml-deps plugin-deps kore-deps \
-		build build-ocaml build-java build-node build-kore defn split-tests \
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -152,7 +154,7 @@ defn: $(defn_files)
 	$(K_BIN)/kompile --debug --main-module ETHEREUM-SIMULATION --backend java \
 					--syntax-module ETHEREUM-SIMULATION $< --directory .build/java -I .build/java
 
-.build/haskell/driver-kompiled/definition.kore: $(haskell_files)
+.build/haskell/driver-kompiled/definition.kore: haskell-defn
```
ehildenb: This should be `$(haskell_files)`, not `haskell-defn`. Having a phony target as a prerequisite means the phony target will be rebuilt too often.

Makefile
```diff
@@ -125,6 +125,8 @@ defn_files:=$(ocaml_files) $(java_files) $(node_files)
 
 defn: $(defn_files)
```
ehildenb: No extra space. Also add the corresponding targets `java-defn`, `ocaml-defn`, `node-defn` please.

traiansf:
@ehildenb is this what you meant?

PR #288, opened 2018-12-17
Use llvm backend for kevm node

dwightguth:
Ready for review. This isn't actually yet switching things over, because after speaking with Everett we've decided to wait until the GC is working better, but it is enough that the changes necessary can live on a very minimal branch that should be easy to maintain.


dwightguth:
Jenkins: test this please

dwightguth:
Jenkins: test mantis

PR #290, opened 2018-12-18
Update to latest Constantinople tests

dwightguth:
Do not review yet.

### DIFF
--- a/evm.md
+++ b/evm.md
@@ -1292,11 +1292,24 @@ For now, I assume that they instantiate an empty account and use the empty data.
          <account>
            <acctID> ACCT </acctID>
            <code> CODE </code>
+           <nonce> NONCE </nonce>
+           <balance> BAL </balance>
            ...
          </account>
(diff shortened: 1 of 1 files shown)


dwightguth:
Jenkins: test mantis

evm.md
```diff
@@ -1292,11 +1292,23 @@ For now, I assume that they instantiate an empty account and use the empty data.
...
 
 ```k
+     rule <k> EXTCODEHASH ACCT => 0 ~> #push ... </k>
+         <account>
+           <code> .WordStack </code>
+           <nonce> 0 </nonce>
+           <balance> 0 </balance>
+           ...
+         </account>
```
ehildenb: Should we be using `#accountNonexistent` instead? The question has two components:

-   Should we take into account the schedule constant `Gemptyisnonexistent` for this check?
-   Do we want to fold this rule and the following rule into one rule?

PR #295, opened 2019-01-18
.build/k: update k submodule

dwightguth:
This should fix builds of evm to use the new s3 endpoint, so it should be merged as soon as it passes.


ehildenb:
Jenkins: test this please

PR #299, opened 2019-01-26
Dockerfile, Jenkinsfile: jenkins pipeline based testing

ehildenb:


ehildenb:
@dwightguth I tried to move as much general installation stuff into the `Dockerfile`, and then only have repo setup stuff into the `Jenkinsfile` to maximize caching.

When installing from a `git` repository for an installation dependency I use the Jenkins `git` step module to have access to Jenkins credentials (eg `mantis-cardano` for the Mantis tests). I'm making sure to specify a tagged commit in all cases, so that if the tag changes, the `Dockerfile` caching is invalidated.

It runs the `Mantis`, `Concrete`, and `Proofs` tests in parallel, which I'm not really super amazed with because they all basically run in the same docker instance at the same time instead of all being run in separate docker instances AFAICT. It would be nice if executing them in parallel counted as using extra `Executor`s in Jenkins, to reflect the increased usage of resources.

Dockerfile
```diff
@@ -0,0 +1,56 @@
...
+RUN    echo "deb https://dl.bintray.com/sbt/debian /" >> /etc/apt/sources.list.d/sbt.list                    \
+    && apt-key adv --keyserver hkp://keyserver.ubuntu.com:80 --recv 2EE0EA64E40A89B84B2DF73499E82A75642AC823 \
+    && apt update                                                                                            \
+    && apt install --yes sbt
+
+RUN curl -sSL https://get.haskellstack.org/ | sh
+
+RUN cpan install App::FatPacker Getopt::Declare String::Escape String::ShellQuote UUID::Tiny
+
+RUN    git clone 'https://github.com/z3prover/z3' --branch=z3-4.8.4 \
```
dwightguth: Why can't we install the z3 package?

Jenkinsfile
```diff
@@ -0,0 +1,77 @@
...
+        script {
+          currentBuild.displayName = "PR ${env.CHANGE_ID}: ${env.CHANGE_TITLE}"
+        }
+      }
+    }
+    stage('Build') {
+      steps {
+        ansiColor('xterm') {
+          sh '''
+            git submodule update --init
```
dwightguth: I'd rather this be handled by the Jenkins job config so we consistently run all our vcs outside the container.

ehildenb: I would as well, but unfortunately that leads to a recursive submodule structure because of the VSC repository. So the SCM checkout does a normal `--init`, I make sure to do again here, and then manually do a `--recursive` inside .build/k.

Dockerfile
ehildenb: We need a specific version of Z3.

Makefile
```diff
@@ -82,11 +82,6 @@ $(KORE_SUBMODULE)/make.timestamp:
 	touch ${KORE_SUBMODULE}/make.timestamp
 
 ocaml-deps: .build/local/lib/pkgconfig/libsecp256k1.pc
-	opam init --quiet --no-setup
```
dwightguth: I think this should still call k-configure-opam, right?

ehildenb:
libffi is needed for installing the ocaml dependencies. I suspect it's pulled in by other packages on other installs instead of being listed explicitly. Without it, k-configure-opam-dev fails.

Makefile
ehildenb: No, I'm assuming the user has gone through the K setup. I'll add instructions to the README.

I figure the Makefile shouldn't be setting up things which happen outside the repo, and this one has caused strife for people.

ehildenb: @dwightguth I've added instructions to the README about setting up K's OCaml dependencies.

PR #307, opened 2019-02-06
Imap syntax trick

daejunpark:
on top of PR #306


daejunpark:
@ehildenb i'll rebase the K changes on the master once PR #308 is passed.

yzhang90:
We can only run the minimal tests once IMAP is introduced.

daejunpark:
@yzhang90 even with this small change? not sure..

PR #310, opened 2019-02-07
Dockerfile: install haskell backend dependencies ahead of time

ehildenb:


dwightguth:
It would be better to add this to the end of the dockerfile so that we can take advantage of previous caching, but I guess it's not that big of a deal. I will approve, but think about it in the future.

ehildenb:
Looks like this saved about 10min on the `Build` step.

PR #312, opened 2019-02-08
Update to newest haskell backend, add haskell backend integration tests to CI

ehildenb:
BLOCKED on https://github.com/kframework/k/pull/289, then the submodule should be updated before merge.

Ready for review.


ehildenb:
@dwightguth it is a bug in the frontend. Removing the generatedCounter cell worked for when I tested IMP, but apparently isn't working here yet.

dwightguth:
I approved. This seems fine for now. There is definitely room for improvement so you don't duplicate so much code though.

PR #315, opened 2019-02-11
Use previousGas cell as callGas cell, calculate all gas/refunds ahead of time

ehildenb:
These are simplifications/refactorings of the semantics which achieve a couple things:

-   Try to make the #call and #create machinery look more similar to each other.
-   Calculate all gas needed for CALL when calculating the execution gas (so that the semantics don't deal with gas directly, and we get cleaner separation of the "gas" vs "execution" semantics).
-   Similarly, calculate refunds at the same time as calculating execution gas.


daejunpark:
I added @yzhang90 as he is in a better position to review the details.

PR #317, opened 2019-02-13
Makefile: more informative test output

ehildenb:


kevm
```diff
@@ -217,73 +189,39 @@ run_get_failing() {
...
-    Examples:
-
-        $ $0 run   tests/ethereum-tests/VMTests/vmArithmeticTest/add0.json
-        $ $0 debug tests/interactive/gas-analysis/sumTo10.evm
-        $ $0 prove tests/proofs/specs/examples/sum-to-n-spec.k
-
-    ci usage
-    ========
+    # Running
+    run-@(ocaml|java|llvm|haskell|haskell-perf) )          run_krun         "$@" ;;
```
dwightguth: I don't like that you can no longer invoke a kevm program with the ocaml backend without invoking krun.

ehildenb: Fixed, I added back on option `interpret-@(ocaml|llvm)`.

ehildenb:
@dwightguth should be ready for re-review.

PR #318, opened 2019-02-16
evm.md: correct number of precompiled contracts to 8

livnev:


livnev:
@ehildenb plz

ehildenb:
Hey, sorry, don't review until the red X turns to a green check, but I think you probably fell victim to some weirdness in testing given how small this change is.

But I think this is subsumed by #333 now.

PR #319, opened 2019-02-22
added coverage mode

ana-pantilie:


evm.md
```diff
@@ -32,7 +32,7 @@ In the comments next to each cell, we've marked which component of the YellowPap
       <exit-code exit=""> 1 </exit-code>
       <mode> $MODE:Mode </mode>
       <schedule> $SCHEDULE:Schedule </schedule>
-      <analysis> .Map </analysis>
+      <analysis> "coverage" |-> .Set </analysis>
```
ehildenb: Let's leave the default as the empty map `.Map`.

evm.md
```diff
@@ -166,8 +166,9 @@ Our semantics is modal, with the initial mode being set on the command line via
 -   `VMTESTS` skips `CALL*` and `CREATE` operations.
 
 ```k
-    syntax Mode ::= "NORMAL"  [klabel(NORMAL), symbol]
-                  | "VMTESTS" [klabel(VMTESTS), symbol]
+    syntax Mode ::= "NORMAL"   [klabel(NORMAL), symbol]
+                  | "VMTESTS"  [klabel(VMTESTS), symbol]
+                  | "COVERAGE" [klabel(COVERAGE), symbol]
```
ehildenb: Can we declare this in `analysis.md` instead of `evm.md`? `evm.md` is meant to be the semantics of EVM only (as much as possible).

evm.md
```diff
@@ -339,6 +340,16 @@ The `#next` operator executes a single step by:
...
+
+      rule <mode> COVERAGE </mode>
+           <k> #next 
+            => #setMode NORMAL
+            ~> #next
+            ~> #setMode COVERAGE
+            ...
+           </k> 
+           <pc> PCOUNT </pc>
+           <analysis> ... "coverage" |-> (PCS:Set (.Set => SetItem(PCOUNT))) ... </analysis>
```
ehildenb: We should be able to put this rule in `analysis.md` instead of here as well.

driver.md
```diff
@@ -353,7 +354,12 @@ State Manipulation
     syntax EthereumCommand ::= "clear"
  // ----------------------------------
     rule <k> clear => clearTX ~> clearBLOCK ~> clearNETWORK ... </k>
-         <analysis> _ => .Map </analysis>
+         <analysis> _ => "coverage" |-> .Set </analysis>
+         <mode> EXECMODE </mode>
+      requires EXECMODE in (SetItem(NORMAL) SetItem(VMTESTS))
+
+    rule <k> clear => clearTX ~> clearBLOCK ~> clearNETWORK ... </k>
+         <mode> COVERAGE </mode>
```
daejunpark: um, why do we need this?

evm.md
daejunpark: @ana-pantilie i think you can add the following rule, and keep the default configuration of `<analysis> .Map </analysis>` in the above.

```
rule
<mode> COVERAGE </mode>
<k> #next </k>
<analysis> M => M [ "coverage" <- .Set ] </analysis>
requires "coverage" not in_keys(M)
```

evm.md
daejunpark: @ana-pantilie Just to make sure, you can append the new syntax declaration, by having the following:
```
syntax Mode ::= "COVERAGE" ...
```

driver.md
ana-pantilie: In `normal` or in `vmtest` mode the `analysis` cell has its contents removed at the `clear` computation so I needed to remove this behaviour when the mode is `coverage`.

PR #321, opened 2019-02-28
evm.md, driver.md: add PETERSBURG flag

MrChico:
Addresses https://github.com/kframework/evm-semantics/issues/320. Did I miss anything?

### DIFF
--- a/driver.md
+++ b/driver.md
@@ -532,6 +532,7 @@ The `"network"` key allows setting the fee schedule inside the test.
     rule #asScheduleString("Homestead")      => HOMESTEAD
     rule #asScheduleString("Byzantium")      => BYZANTIUM
     rule #asScheduleString("Constantinople") => CONSTANTINOPLE
+    rule #asScheduleString("Petersburg")     => PETERSBURG
 ```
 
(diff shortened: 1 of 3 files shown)


driver.md
```diff
@@ -532,6 +532,7 @@ The `"network"` key allows setting the fee schedule inside the test.
     rule #asScheduleString("Homestead")      => HOMESTEAD
     rule #asScheduleString("Byzantium")      => BYZANTIUM
     rule #asScheduleString("Constantinople") => CONSTANTINOPLE
+    rule #asScheduleString("Petersburg")     => PETERSBURG
```
dwightguth: This should actually be `"ConstantinopleFix"` on the lhs to match the current behavior of testeth.

MrChico:
Updated the schedulestring and the test repo.

dwightguth:
I believe that this failed due to the fact that the llvm backend cannot run some of the new tests that were added by this PR. The llvm backend still does not have adequate GC so we need to exclude a few tests that would otherwise crash with insufficient memory. I am running locally to determine which tests need to be excluded. I will push a fix to this PR once I have done this.

PR #322, opened 2019-03-04
Klab compatibility

MrChico:
Upgrades necessary to sync klab with this repo upstream

### DIFF
--- a/evm.md
+++ b/evm.md
@@ -283,8 +283,8 @@ OpCode Execution
 ```k
     syntax KItem ::= "#execute"
  // ---------------------------
-    rule <k> (. => #next) ~> #execute ... </k>
-    rule <k> #halt ~> (#execute => .) ... </k>
+    rule <k> (. => #next) ~> #execute ... </k> [tag(step)]
+    rule <k> #halt ~> (#execute => .) ... </k> [tag(step)]
 
     syntax InternalOp ::= "#execTo" Set
  // -----------------------------------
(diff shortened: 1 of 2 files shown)


evm.md
```diff
@@ -283,8 +283,8 @@ OpCode Execution
 ```k
     syntax KItem ::= "#execute"
  // ---------------------------
-    rule <k> (. => #next) ~> #execute ... </k>
-    rule <k> #halt ~> (#execute => .) ... </k>
+    rule <k> (. => #next) ~> #execute ... </k> [tag(step)]
+    rule <k> #halt ~> (#execute => .) ... </k> [tag(step)]
```
ehildenb: This should probably be a rule name/label instead of a rule tag, ie:

```
rule [step] : <k> (. => #next) ~> #execution ... </k>
rule [halt] : <k> #halt ~> (#execute => .) ... </k>
```

Note that rule labels must be unique. The reason I suggest a rule label instead of a `tag` is because using `tag` may mess up the strategy used for execution.

ehildenb: Note that rule labels/names are passed through as attributes too. So if you're just searching for the `step` and `halt` string as a key in the attributes to detect this, it should still work.

dwightguth: @ehildenb tags will not affect this definition at all because it doesn't import any strategy module.

ehildenb:
Subsumed by #323 I suppose.

PR #323, opened 2019-03-05
Klab mainline

mhhf:
replaces #322


MrChico:
Can we have the K bump and the optional haskell here as welm

ehildenb:
Optional Haskell merged by another PR, was testing if it worked without K bump, will put back in.

Seems we have a flaky OOM on the llvm backend going on here.

mhhf:
failed unfortunately :(

PR #325, opened 2019-03-05
Disable Haskell backend by default, allow enabling both haskell/llvm, documentation

ehildenb:
@bl4ck5un can you check the updated instructions and let me know if it makes sense/works for you?

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -14,15 +14,23 @@ pipeline {
         }
       }
     }
+    stage('Dependencies') {
+      steps {
+        ansiColor('xterm') {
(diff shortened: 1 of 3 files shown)


bl4ck5un:
That seems to work. I removed `stack` from my system and `make deps` doesn't fail.

dwightguth:
I believe you need to update the jenkinsfile to call all-deps.

ehildenb:
@dwightguth I made it build dependencies in a separate step (using `all-deps`), and also have it explicitly build the all versions of the semantics (including LLVM and Haskell).

PR #327, opened 2019-03-11
.build/k: update to latest master

ehildenb:


dwightguth:
Bus error isn't an out of memory error, that's a memory corruption error. Let me investigate and try and see what is causing the regression.

ehildenb:
We need to re-review this in light of the proof breakage on newest K master due to https://github.com/kframework/k/pull/290

Makefile
```diff
@@ -309,19 +309,29 @@ tests/interactive/gas-analysis/%.evm.test: tests/interactive/gas-analysis/%.evm
...
 
-$(proof_dir)/%.test: $(proof_dir)/% split-proof-tests
+$(proof_specs_dir)/%.test: $(proof_specs_dir)/% split-proof-tests
 	$(TEST) test --backend $(TEST_SYMBOLIC_BACKEND) $<
 
-split-proof-tests: tests/proofs/make.timestamp
-	$(MAKE) -C tests/proofs $@
+split-proof-tests: $(proof_repo)/git-submodule
+	$(MAKE) --directory tests/proofs/resources      $@
+	$(MAKE) --directory tests/proofs/bihu           $@
```
daejunpark: you would need to drop bihu

ehildenb: It passes, so we decided offline to keep it until it doesn't pass (so that we're notified of the change).

MrChico:
Would love a bump of this that includes at least https://github.com/kframework/k/commit/a1fc7e45f5fb85c26e4b6c243d880168c774eee9

PR #331, opened 2019-03-28
[WIP] Modification for coverage-mode

anvacaru:


driver.md
```diff
@@ -79,6 +79,7 @@ To do so, we'll extend sort `JSON` with some EVM specific syntax, and provide a
  // ----------------------------------
     rule <mode> NORMAL     </mode> <k> start => #execute    ... </k>
     rule <mode> VMTESTS    </mode> <k> start => #execute    ... </k>
+    rule <mode> COVERAGE   </mode> <k> start => #execute    ... </k>
```
yzhang90: We always start with `loadTx` and we can remove this rule for now.

evm.md
```diff
@@ -1362,6 +1363,21 @@ The various `CALL*` (and other inter-contract control flow) operations will be d
          <output>     _ => .WordStack </output>
          <wordStack>  _ => .WordStack </wordStack>
          <localMem>   _ => .Map       </localMem>
+         <programBytes> CODE </programBytes>
+         <program> PROGRAM </program>
+         <analysis> ANALYSIS => ANALYSIS["currentProgramHash" <- keccak(CODE)][keccak(CODE) <- .Set]["currentProgram" <- PROGRAM] </analysis>
```
yzhang90: are we still using the information in the "currentProgram" ?

yzhang90: VMTEST and NORMAL mode doesn't need to compute the program hash.

driver.md
anvacaru: Removed.

evm.md
anvacaru: "currentProgram" is used to get all the opcodes of the contract which are then displayed in the report.

yzhang90: We know that one test may contain "multiple" programs (e.g one program with constructor and one program without constructor). so there is no point in storing the "current program" in the cell. We need to discuss how to store the hash -> program map (in the k config? or can we parse the bytecode offline?)

ehildenb:
Closing for now.

PR #335, opened 2019-04-11
Configuration changes

ehildenb:
I have been wanting to rename `<previousGas> => <callGas>` for a while, because it's only used when making a call to record the previous gas used.

We need the wrapper `<kevm>` cell to make configuration composition possible with KEVM.


kore-json.py
```diff
@@ -35,11 +35,11 @@ def print_kast(data, sort="SortJSON"):
     for key in data:
       sys.stdout.write(')')
     sys.stdout.write(')')
-  elif isinstance(data, str) or isinstance(data, unicode):
+  elif isinstance(data, str) or isinstance(data, str):
```
discosultan: Duplicate checks

kore-json.py
```diff
@@ -35,11 +35,11 @@ def print_kast(data, sort="SortJSON"):
...
     sys.stdout.write("inj{SortString{}, " + sort + "{}}(\dv{SortString{}}("),
     sys.stdout.write(json.dumps(data))
     sys.stdout.write('))')
-  elif isinstance(data, long) or isinstance(data, int):
+  elif isinstance(data, int) or isinstance(data, int):
```
discosultan: Duplicate checks

kore-json.py
```diff
@@ -25,7 +25,7 @@ def print_kast(data, sort="SortJSON"):
     sys.stdout.write(')')
   elif isinstance(data, OrderedDict):
     sys.stdout.write("Lbl'LBraUndsRBraUnds'EVM-DATA'UndsUnds'JSONList{}(")
-    for key, value in data.items():
+    for key, value in list(data.items()):
```
discosultan: Why is it necessary to collect dictionary items to a list before iteration?

kore-json.py
ehildenb: Changed back to `unicode`.

kore-json.py
ehildenb: Changed back to `long`.

kore-json.py
ehildenb: Not sure, I changed it back.

ehildenb:
@discosultan I ran `2to3` on the script to convert to Python3, which made the duplicate checks and the additional `list` around the `data.items()`. I've reverted the changes you mention, but if you want another look I'm happy to incorporate feedback.

PR #337, opened 2019-04-24
data: compare with K instead of Int

MrChico:
It seems that when terms are sorted with `Int` rather than `K`, the K framework will not simplify expressions before they are sent to z3. This means that expressions like:
`W0 /Word abs(W1)` where we know `W1 ==Int 0` to be true will not simplified in the presence of a lemma `rule abs(B) ==Int 0 => B ==Int 0` if the condition is `requires W1 ==Int 0`, but it will if the condition is `requires W1 ==K 0` with lemma `rule abs(B) ==K 0 => B ==K 0`.

I'm not sure whether you consider this "intended behaviour" of K or not. If this is intended, I would suggest to change side conditions in kevm as suggested by this PR, if it is not intended, I can write some regression tests for this. Ping @daejunpark @dwightguth

### DIFF
--- a/data.md
+++ b/data.md
@@ -250,10 +250,10 @@ The corresponding `<op>Word` operations automatically perform the correct modulu
     rule W0 -Word W1 => chop( W0 -Int W1 ) requires W0 >=Int W1
     rule W0 -Word W1 => chop( (W0 +Int pow256) -Int W1 ) requires W0 <Int W1
     rule W0 *Word W1 => chop( W0 *Int W1 )
-    rule W0 /Word W1 => 0                    requires W1  ==Int 0
-    rule W0 /Word W1 => chop( W0 /Int W1 )   requires W1 =/=Int 0
-    rule W0 %Word W1 => 0                    requires W1  ==Int 0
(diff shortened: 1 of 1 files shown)


MrChico:
The issue I was having due to this is no longer an issue so I will close this PR. Someone may still want to check out the sorting curiosity though. For that I made this regression test https://github.com/MrChico/k/tree/int-sort/k-distribution/tests/regression-new/int-sort

daejunpark:
I put here @denis-bogdanas's response to the above regression test example.


----

I studied your example. There's no problem with kprove. It's a bad idea to write lemmas for `=/=K` because rules in domains.k will rewrite the term to `notBool _ ==K _`, and then your lemmas will no longer apply.

If you use these lemmas instead, both specs will pass:
    ```rule abs(B) ==K 0 => B ==K 0
    rule #unsigned(X) ==K 0 => X ==K 0```

Run the specs with options `--deterministic-functions --log-rules --log-func-eval`, on a recent enough K, to see what happens. Last option will print evaluation result of each functional term.

----

From https://matrix.to/#/!MfVXcmpKuyudZHcqil:matrix.org/$15567358071293037jsChL:matrix.org?via=matrix.org&via=feneas.org&via=kde.org

PR #338, opened 2019-04-26
bump k to latest master

MrChico:


ehildenb:
Subsumed by #314 . If that one passes, this one will be closed.

PR #343, opened 2019-05-03
Updates which make KEVM build on Arch Linux with LLVM backend

ehildenb:
Not ready for merge until the following is merged and this PR is updated to point at the new `master`.

https://github.com/runtimeverification/blockchain-k-plugin/pull/32

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ RUN    ln --symbolic --no-dereference --force /usr/share/zoneinfo/$TZ /etc/local
 RUN    apt update                                                                  \
     && apt upgrade --yes                                                           \
     && apt install --yes                                                           \
-        autoconf bison clang-6.0 cmake curl flex gcc libboost-test-dev             \
+        autoconf bison clang-6.0 clang-6.0++ cmake curl flex gcc libboost-test-dev \
         libcrypto++-dev libffi-dev libjemalloc-dev libmpfr-dev libprocps-dev       \
         libsecp256k1-dev libssl-dev libtool libyaml-cpp-dev lld-6.0 llvm-6.0-tools \
         make maven opam openjdk-8-jdk pandoc pkg-config python3 python-pygments    \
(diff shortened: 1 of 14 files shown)


Dockerfile
```diff
@@ -7,7 +7,7 @@ RUN    ln --symbolic --no-dereference --force /usr/share/zoneinfo/$TZ /etc/local
 RUN    apt update                                                                  \
     && apt upgrade --yes                                                           \
     && apt install --yes                                                           \
-        autoconf bison clang-6.0 cmake curl flex gcc libboost-test-dev             \
+        autoconf bison clang-6.0 clang-6.0++ cmake curl flex gcc libboost-test-dev \
```
dwightguth: I am reasonably certain that not only does a clang-6.0++ package not exist, clang-6.0 installs clang++.

evm-node.md
```diff
@@ -215,9 +215,9 @@ Because the same account may be loaded more than once, implementations of this i
 
 ```{.k .node}
     syntax KItem ::= vmResult ( return: String , gas: Int , refund: Int , status: Int , selfdestruct: List , logs: List , AccountsCell , touched: List )
-    syntax KItem ::= extractConfig ( GeneratedTopCell ) [function, symbol]
- // ----------------------------------------------------------------------
-    rule extractConfig ( <generatedTop>
+    syntax KItem ::= extractConfig ( KevmCell ) [function, symbol]
+ // --------------------------------------------------------------
+    rule extractConfig ( <kevm>
```
dwightguth: I'm fairly suspicious of this change, unless we intend to use the ocaml node in the future instead of the llvm node, which seems unlikely.

Dockerfile
ehildenb: Removed.

evm-node.md
ehildenb: @dwightguth when I switch back to GeneratedTop for this whole thing, it fails to build. Probably because the Node is built with the OCaml backend?

PR #346, opened 2019-05-10
Update k submodule, add kast command to kevm.

ehildenb:


ehildenb:
Must wait for https://github.com/kframework/k/pull/449

dwightguth:
I'll wait until the tests pass to review this, but I spotted something suspicious: you removed the shift command from run_krun, but didn't change anything else in that function.

ehildenb:
`run_file` is now set globally in the same way that `backend` and `run_command` are at the end of the script, so no need to set them in the procedures directly.

ehildenb:
This PR has had the build system refactors pulled from it.

PR #348, opened 2019-05-14
Refactors to kevm runner and Makefile

ehildenb:
This PR does a lot of things in refactoring the kevm running and the Makefile:

-   Adds `./kevm kast`, and calls that from `./kevm interpret`. `./kevm kast` detects if it's a JSON file, and if so calls the appropriate "fast kast-er" python script. Otherwise it goes through vanilla `kast` program. Tests are added for `./kevm kast` as well.
-   Adds more generic testing harnesses in `Makefile`, which handle all cases by setting Make variables for the MODE and SCHEDULE and overriding them when certain test-target patterns match.
-   Removes the `./kevm test` command, in favor of directly calling `./kevm interpret` and `./kevm prove`.
-   The `Jenkinsfile` now runs interactive tests on _all_ backends after running the proof tests, as well as running the `kast` tests on the OCaml backend.
-   The default `node` stack size is upped significantly for KLab executions.

The only part of `./kevm` that remains untested in `./kevm klab-(run|prove)`. I haven't thought of a good way to test it yet, but this PR still represents a significant improvement in the testing coverage of `./kevm`. I've tested `./kevm klab-run` locally and it seems to work correctly.


ehildenb:
We could pipe into diff, but I like having the output file around in case the test fails so it's easy to inspect without re-running. I admit that piping into diff is nice because then you can replace CHECK with cp and get a test updater, but that requires you to diff output -> expected instead of expected -> output, and I think the latter is more intuitive.

PR #349, opened 2019-05-14
Update macos install instructions

iamchrissmith:
I was trying to get this setup on MacOS v10.14.4 and had to do a few different/additional things to what was described in the README.  I updated the README to reflect the steps I took to get it working in the hopes they might help others.


dwightguth:
I made one slight tweak so that the fact that you need to run `make libsecp256k1` on Mac OS is more clear in the instructions. We might also consider having users use the latest jdk instead of specifically openjdk 8 now, as I don't think there's any longer any reason to use jdk 8 specifically. I haven't tested it with the main java formula for homebrew yet though so I didn't make that change myself.

ehildenb:
@iamchrissmith you'll have to sign the CLA, then I'll get this merged. It's the "Details" button next to the red X by `clahub`.

PR #356, opened 2019-05-21
Upstream the gen-spec.py script

ehildenb:
This PR simplifies slight and upstreams the gen-spec.py script used in the VSC repository.

Now three specs in `tests/specs` are generated from `gen-spec.py`, including `ds-token-erc20`, `hkg-erc20`, and `hobby-erc20`.


ehildenb:
Unlike the VSC `gen-spec.py` script, this one generates a _single_ `-spec.k` file for each method, and puts all the proof claims for that method into that one `-spec.k` file.

For example, the generated file `./tests/specs/ds-token-erc20/transfer-spec.k` contains 7 proof claims, generated with `python3 tests/gen-specs/gen-specs.py tests/specs/ds-token-erc20/spec.ini` (which is called by `make split-tests`.

ehildenb:
In addition, all of the proofs are made to use the same set of trusted lemma files in `tests/specs/trusted`, which are _not_ verified on CI but can be used by any of the proofs.

ehildenb:
This now uses the `gen-specs.py` script upstreamed into the K repository.

PR #358, opened 2019-05-22
Simplify branching structure of KEVM, less intermediate steps

ehildenb:
These refactors lead up to loading entire basic blocks at a time, which is not quite working yet.

It may be easier to review this commit-by-commit.

### DIFF
--- a/evm.md
+++ b/evm.md
@@ -1783,6 +1783,8 @@ In the YellowPaper, each opcode is defined to consume zero gas unless specified 
     rule #memory ( COP:CallOp     _ _ _ ARGSTART ARGWIDTH RETSTART RETWIDTH , MU ) => #memoryUsageUpdate(#memoryUsageUpdate(MU, ARGSTART, ARGWIDTH), RETSTART, RETWIDTH)
     rule #memory ( CSOP:CallSixOp _ _   ARGSTART ARGWIDTH RETSTART RETWIDTH , MU ) => #memoryUsageUpdate(#memoryUsageUpdate(MU, ARGSTART, ARGWIDTH), RETSTART, RETWIDTH)
 
+    rule #memory ( _ , MU ) => MU [owise]
+
     syntax Bool ::= #usesMemory ( OpCode ) [function]
  // -------------------------------------------------
     rule #usesMemory(OP) => isLogOp(OP)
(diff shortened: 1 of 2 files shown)


evm.md
```diff
@@ -289,48 +284,49 @@ When the `#next` operator cannot lookup the next opcode, it assumes that the end
...
 
-    rule <k> #stackNeeded? [ OP ] => #end EVMC_STACK_OVERFLOW ... </k>
+    rule <k> #next [ OP ] => #end EVMC_STACK_OVERFLOW ... </k>
          <wordStack> WS </wordStack>
       requires #stackOverflow(WS, OP)
 
-    rule <k> #stackNeeded? [ OP ] => . ... </k>
+    rule <k> #next [ OP ] => #end EVMC_STATIC_MODE_VIOLATION ... </k>
          <wordStack> WS </wordStack>
-      requires notBool ( #stackUnderflow(WS, OP) orBool #stackOverflow(WS, OP) )
```
dwightguth: I'm fairly sure you still need this, otherwise both rules could apply if the user is doing a STATICCALL and they try to change state but have a stack exception.

evm.md
```diff
@@ -526,19 +511,11 @@ The arguments to `PUSH` must be skipped over (as they are inline), and the opcod
  // ------------------------------------------
     rule <k> #pc [ OP ] => . ... </k>
          <pc> PCOUNT => PCOUNT +Int #widthOp(OP) </pc>
-      requires notBool isJumpOp(OP)
-
-    rule <k> #pc [ OP ] => . ... </k>
```
dwightguth: I think you have an off by one error here. You're going to skip executing the jumpdest instruction. Maybe that's fine, since JUMPDEST doesn't do anything, but it seems suspicious to me.

ehildenb: `#endBasicBlock` consumes everything up until the next `#execute`, so `#pc` isn't executed when `JUMP` affects the `<pc>` cell.

evm.md
ehildenb: Fixed.

evm.md
ehildenb: We definitely need to execute the `JUMPDEST` instruction when we arrive at it, because its gas needs to be deducted. If we were skipping it, no way we wolud pass tests.

PR #373, opened 2019-06-11
Kill old builds on updates to PRs

ehildenb:


Jenkinsfile
```diff
@@ -1,131 +1,157 @@
...
     }
-    stage('Test Execution') {
-      failFast true
-      parallel {
-        stage('Conformance (OCaml)') {
+    stage('Run CI') {
+      agent { label 'docker' }
+      stages {
+        stage('Init title') {
+          when { changeRequest() }
```
dwightguth: let's put this up in the agent none segment. we don't need a whole extra stage for this separate from the build killer stage.

ehildenb:
Avoided with more automated global build killer.

PR #377, opened 2019-06-20
Simpler `#halt` testing semantics, explicit state-splitting on `JUMPI` instruction

ehildenb:


tests/templates/output-success-llvm.json
```diff
@@ -0,0 +1,162 @@
+<kevm>
```
dwightguth: why is this file here? it shouldn't actually be using krun when running the llvm backend, right?

ehildenb: The `./kevm interpret ` command does `k-bin-to-text` on the output if it sees a non-zero exit code, and then the `Makefile` does `diff` on that and the backend-specific output file.

ehildenb: Note that both `k-bin-to-text` and `diff` are skipped on zero exit code.

ehildenb: We haven't needed the output file until now, because I haven't introduced bugs which we needed to display the diff for. Only on this PR (which introduced a bug) did the file ommission show up.

ehildenb: This was resolved in #380 .

PR #412, opened 2019-07-25
kevm: add search command

ehildenb:


kevm
```diff
@@ -140,6 +148,7 @@ if [[ "$run_command" == 'help' ]]; then
            $0 interpret : Run JSON EVM programs without K Frontend (external parser)
            $0 kast      : Parse an EVM program and output it in a supported format
            $0 prove     : Run an EVM K proof
+           $0 prove     : Run an EVM K proof
```
ttuegel: Maybe this instead?
```suggestion
           $0 search     : Search for a K pattern in an EVM program
```

ehildenb: Done, good catch.

PR #418, opened 2019-07-30
Install pinned version of Stack from Haskell backend

ttuegel:
See also: kframework/k#658


Dockerfile
```diff
@@ -44,10 +42,12 @@ RUN    cd /home/user \
     && ./.tmp-opam/bin/k-configure-opam-dev
 
 ENV LC_ALL=C.UTF-8
+ADD deps/k/haskell-backend/src/main/native/haskell-backend/scripts/install-stack.sh /home/user/.install-stack/
```
ehildenb:
```suggestion
ADD --chown=user:user deps/k/haskell-backend/src/main/native/haskell-backend/scripts/install-stack.sh /home/user/.install-stack/
```

ttuegel:
I have confirmed by inspecting the log that Stack now uses the cache of packages built in the Dockerfile stage.

PR #421, opened 2019-07-31
Jenkinsfile: comment out arch for now

dwightguth:


dwightguth:
merging to save time because it only affects release builds.

PR #426, opened 2019-08-06
Disassemble on the fly

dwightguth:
Essentially what this PR does is:

* deletes the `<program>` cell and renames `<programBytes>` to `<program>`
* deletes the `#dasmOpCodes` and `#asMapOpCodes`  functions
* modifies `#dasmOpCode` to disassemble every opcode.
* explicitly computes jump destination validity as a set of program counter values
* modifies the semantics of PUSH and JUMP to make everything work
* modifies the semantics of `#next` to disassemble on the fly
* deletes the OpCodes sort
* updates the necessary input and output files.


dwightguth:
Replaced by #448

PR #430, opened 2019-08-09
Multistage docker build

charala1:


.gitignore
```diff
@@ -5,3 +5,10 @@
 /package/pkg
 /package/src
 /media/*.pdf
+
+# Tests
```
charala1: Let me know if the following three lines are necessary, because I suspect they were cherry-picked accidentally.

charala1: If so, I'll remove them in the next PR.

ehildenb:
Too old to be useful.

PR #438, opened 2019-08-14
Start eth_sign implementation

gtrepta:
This will need to be finished once accounts have private keys.


Makefile
```diff
@@ -292,7 +292,7 @@ $(DEFN_DIR)/ocaml/$(MAIN_DEFN_FILE)-kompiled/interpreter: $(DEFN_DIR)/ocaml/$(MA
 	        && ocamlfind $(OCAMLC) -c -g -w -11-26 -package gmp -package zarith -package uuidm -package ethereum-semantics-plugin-ocaml -safe-string realdef.ml -match-context-rows 2 \
 	        && ocamlfind $(OCAMLC) $(LIBFLAG) -o realdef.$(DLLEXT) realdef.$(EXT) \
 	        && ocamlfind $(OCAMLC) -g -o interpreter constants.$(EXT) prelude.$(EXT) plugin.$(EXT) parser.$(EXT) lexer.$(EXT) hooks.$(EXT) run.$(EXT) interpreter.ml \
-	                               -package gmp -package dynlink -package zarith -package str -package uuidm -package unix -package ethereum-semantics-plugin-ocaml -linkpkg -linkall -thread -safe-string
+	                               -package gmp -package hex -package dynlink -package zarith -package str -package uuidm -package unix -package ethereum-semantics-plugin-ocaml -linkpkg -linkall -thread -safe-string
```
ehildenb: -   What is the additional `hex` package?
-   Do we need some counterpart for the LLVM backend build?
-   Do we need to install this package in `make ocaml-deps`?

gtrepta: - It has functions for converting strings to hex representations
- No, llvm has its own subroutine for it
- It's already in ocaml-deps.

web3.md
```diff
@@ -186,5 +189,21 @@ module WEB3
...
+    syntax KItem ::= #signMessage ( String , String )
+ // -------------------------------------------------
+    rule <k> #signMessage(KEY, MHASH) => #sendResponse( "result" : "0x" +String ECDSASign( MHASH, KEY ) ) ... </k>
+
+    syntax String ::= #hashMessage ( String ) [function]
+                    | #privateKey  ( Int )    [function]
+ // ----------------------------------------------------
+    rule #hashMessage( S ) => #unparseByteStack(#parseHexBytes(Keccak256("\x19Ethereum Signed Message:\n" +String Int2String(lengthString(S)) +String S)))
+    rule [[ #privateKey( ADDR ) => #unparseByteStack(#parseByteStack(KEY)) ]]
+         <accountKeys>... ADDR |-> KEY ...</accountKeys>
```
ehildenb: I would prefer to inline this directly instead of using a contexual function, so we have a better idea of where it's getting stuck (and so we can send back an error message if the key is not present).

```
    syntax KItem ::= "#eth_sign"
 // ----------------------------
    rule <k> #eth_sign => #signMessage(PRIV_KEY, #hashMessage(...)) ... </k>
         <accountKeys> ... #parseHexWord(ADDR) |-> KEY ... </accountKeys>
    rule <k> #eth_sign => #sendRespones("SOME ERROR MESSAGE") ... </k>

         <accountKeys> ACCTKEYS </accountKeys>
       requires notBool #parseHexWord(ADDR) in_keys(ACCTKEYS)
```

PR #441, opened 2019-08-15
Batch processing of web3 requests

dwightguth:


web3.md
```diff
@@ -71,15 +74,56 @@ module WEB3
...
+    rule <k> #loadFromBatch ~> _ => getRequest() </k>
+         <batch> [ .JSONList ] </batch>
+         <web3response> .List </web3response>
+
+    syntax JSON ::= List2JSON(List)           [function]
+                  | List2JSON(List, JSONList) [function, klabel(List2JSONAux)]
+ // --------------------------------------------------------------------------
+    rule List2JSON(L) => List2JSON(L, .JSONList)
+
+    rule List2JSON(L ListItem(J), JS) => List2JSON(L, J, JS)
```
ehildenb: This is confusing on first read, can we do `List2JSON(L, (J, JS))` on the RHS instead?

tests/web3/batch_partial_notification.out.json
```diff
@@ -0,0 +1 @@
+[{"jsonrpc":"2.0","id":1,"result":"28346"}]
```
ehildenb: I'm confused why this doesn't have two responses, is it intended?

ehildenb: Looks likee there are two requests in the file above.

dwightguth: This is the correct behavior. One of the requests is called a "notification" because it does not have an id field and thus it does not have a corresponding response.

PR #444, opened 2019-08-16
Istanbul

MrChico:
Fixes: https://github.com/kframework/evm-semantics/issues/442

### DIFF
--- a/evm.md
+++ b/evm.md
@@ -1984,9 +1984,9 @@ The intrinsic gas calculation mirrors the style of the YellowPaper (appendix H).
          <schedule> SCHED </schedule>
          <callData> DATA </callData>
 
-    rule <k> #gasExec(_, ECADD)     => 500   ... </k>
-    rule <k> #gasExec(_, ECMUL)     => 40000 ... </k>
-    rule <k> #gasExec(_, ECPAIRING) => 100000 +Int (#sizeByteArray(DATA) /Int 192) *Int 80000 ... </k> <callData> DATA </callData>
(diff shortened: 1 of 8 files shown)


evm.md
```diff
@@ -2349,6 +2362,27 @@ A `ScheduleConst` is a constant determined by the fee schedule.
...
+
+    rule SCHEDCONST < ISTANBUL > => SCHEDCONST < PETERSBURG >
+      requires notBool (SCHEDCONST ==K Gecadd orBool SCHEDCONST ==K Gecpairconst
+                orBool  SCHEDCONST ==K Gecmul orBool SCHEDCONST ==K Gecpaircoeff)
+
+    rule Ghaschainid << ISTANBUL >> => true
+
+    rule SCHEDFLAG   << ISTANBUL >> => SCHEDFLAG << PETERSBURG >>
+      requires notBool (SCHEDFLAG ==K Ghaschainid)
+//TODO
```
ehildenb: What's this TODO?

dwightguth:
I would very much like to see the history of this pr refactored into a sequence of commits, one per EIP, that each individually pass tests, and then merged without squashing. My experience with the Ethereum hardfork process is that they change their minds about things pretty frequently until fairly late in the process, so if we're going to be implementing Istanbul this early on, we need to do defensive coding to make super simple to remove any of these EIPs if they get dropped. This can wait until after the code has been written to do the history refactoring though.

ehildenb:
Subsumed by other PRs.

PR #448, opened 2019-08-20
Disassemble on the fly

dwightguth:


evm.md
```diff
@@ -1065,13 +1037,15 @@ The `JUMP*` family of operations affect the current program counter.
...
          <pc> _ => DEST </pc>
-         <program> ... DEST |-> OP ... </program>
+         <program> PGM </program>
+         <jumpDests> DESTS </jumpDests>
+      requires DEST <Int #sizeByteArray(PGM)
 
     rule <k> JUMP DEST => #end EVMC_BAD_JUMP_DESTINATION ... </k>
          <program> PGM </program>
-      requires notBool (DEST in_keys(PGM))
+      requires DEST >=Int #sizeByteArray(PGM)
```
ehildenb: Can we just have one rule which says `DEST in DESTS` and the other which says `notBool DEST in DESTS`? No need to check the size of the program or anything assuming that we've always computed `#validJumpDestinations` ahead of time.

evm.md
```diff
@@ -1372,12 +1345,38 @@ The various `CALL*` (and other inter-contract control flow) operations will be d
...
-         <wordStack>  _ => .WordStack </wordStack>
-         <localMem>   _ => .Map       </localMem>
+    rule <k> #initVM      => . ...      </k>
+         <pc>           _ => 0          </pc>
+         <memoryUsed>   _ => 0          </memoryUsed>
+         <output>       _ => .ByteArray </output>
+         <wordStack>    _ => .WordStack </wordStack>
+         <localMem>     _ => .Map       </localMem>
+         <jumpDests>    _ => #computeValidJumpDests(PGM) </jumpDests>
+         <program>      PGM             </program>
```
ehildenb: It seems to me that we should actually do `#computeValidJumpDests` above where the `PGM` is actually loaded into the `<program>` cell, to ensure that the invariant that `<jumpDests>` cell contents always corresponds to `<program>`.

`initVM` should only be doing stuff like clearing state, not loading program (according to how it was written before).

evm.md
```diff
@@ -1501,8 +1500,7 @@ For each `CALL*` operation, we make a corresponding call to `#call` and a state-
          <id> ACCT => ACCTTO </id>
          <gas> _ => GCALL </gas>
          <callGas> GCALL => 0 </callGas>
-         <program> _ => #asMapOpCodes(#dasmOpCodes(INITCODE, SCHED)) </program>
-         <programBytes> _ => INITCODE </programBytes>
+         <program> _ => INITCODE </program>
```
ehildenb: Let's `#computeValidJumpDests` here as well.

ehildenb: Unless you want to move program loading into `initVM` by passing it as an argument, which would reduce rule duplication but also make sure that the `<jumpDests>` and `<program>` cells always agree.

evm.md
```diff
@@ -1372,12 +1347,37 @@ The various `CALL*` (and other inter-contract control flow) operations will be d
...
-         <output>     _ => .ByteArray </output>
-         <wordStack>  _ => .WordStack </wordStack>
-         <localMem>   _ => .Map       </localMem>
+    rule <k> #initVM      => . ...      </k>
+         <pc>           _ => 0          </pc>
+         <memoryUsed>   _ => 0          </memoryUsed>
+         <output>       _ => .ByteArray </output>
+         <wordStack>    _ => .WordStack </wordStack>
+         <localMem>     _ => .Map       </localMem>
+         <program>      PGM             </program>
```
ehildenb: No need to mention `<program>` cell anymore.

PR #454, opened 2019-08-22
Automatically build brew bottle

dwightguth:


dwightguth:
My mistake, this is not actually ready yet.

Jenkinsfile
```diff
@@ -158,13 +158,16 @@ pipeline {
           }
         }
       }
-    }
+    }*/
```
ehildenb: remove comment tags

Jenkinsfile
```diff
@@ -158,13 +158,16 @@ pipeline {
...
     stage('Release') {
       agent { label 'docker' }
       options { skipDefaultCheckout() }
       environment {
         GITHUB_TOKEN    = credentials('rv-jenkins')
         KEVM_RELEASE_ID = '1.0.0'
+        PACKAGE = 'kevm'
+        VERSION = '1.0.0'
+        ROOT_URL = 'https://github.com/kframework/evm-semantics/releases/download/nightly'
```
ehildenb: Is there a reason for `VERSION` separate from `KEVM_RELEASE_ID`?

ehildenb: It doesn't seem to be used directly anywhere here.

Jenkinsfile
```diff
@@ -235,11 +238,11 @@ pipeline {
           }
         }
         stage('Deploy Release') {
-          when {
+/*          when {
             not { changeRequest() }
             branch 'master'
             beforeAgent true
-          }
+          }*/
```
ehildenb: remove comment tags

tests/node-firefly
```diff
@@ -27,11 +27,10 @@ cd openzeppelin-solidity
...
 npm install
-npx truffle compile
 
 # Wait for kevm-vm to start up
 while ! netcat -z localhost "$kevm_vm_port"; do
     sleep 1
 done
 
-npx truffle test
+node_modules/.bin/truffle test
```
ehildenb: Why these changes?

Jenkinsfile
dwightguth: This is used by the brew scripts that we are invoking that are part of the kframework package. They are used to control the CI so that it will test the correct commit on Mac OS without deploying it first.

tests/node-firefly
dwightguth: So that we always use the version of truffle specified by the package rather than the latest version which may not be backwards compatible.

PR #459, opened 2019-08-26
Isolate blockchain state for client

anvacaru:
- Wrap up current block state in `<kevm>` with a `<block>`
- Added `blockList` cell to store `<block>` and `network` substates as `syntax BlockchainItem ::= "{" NetworkCellFragment "|" BlockCell "}"`.
- Added rules for `#pushBlockchainState` and `#getBlockchainState(Int)`.


evm.md
```diff
@@ -182,7 +185,78 @@ Our semantics is modal, with the initial mode being set on the command line via
...
 ------------
+### The BlockchainStack
+
+The `blockList` cell stores a list of previous block and network states.
+-   `#pushBlockchainState` saves a copy of the block state and network state on the `blockList`.
+-   `#getBlockchainState(Int)` restores a blockchain state for a given block number.
+
+```k
+    syntax BlockchainItem ::= "{" NetworkCellFragment "|" BlockCell "}"
+ // -------------------------------------------------------------------
```
ehildenb: Looks like you're using a mix of `CellFragment` and `Cell`. I would personally just use `BlockCellFragment`, as I know from other projects that it works and it provides a simpler user-experience in my opinion.

evm.md
```diff
@@ -182,7 +185,78 @@ Our semantics is modal, with the initial mode being set on the command line via
...
+                  <gasLimit>         HL       </gasLimit>         // I_Hl
+                  <gasUsed>          HG       </gasUsed>          // I_Hg
+                  <timestamp>        HS       </timestamp>        // I_Hs
+                  <extraData>        HX       </extraData>        // I_Hx
+                  <mixHash>          HM       </mixHash>          // I_Hm
+                  <blockNonce>       HN       </blockNonce>       // I_Hn
+
+                  <ommerBlockHeaders> OBH </ommerBlockHeaders>
+                  <blockhash>         BH  </blockhash>
+                </block>
```
ehildenb: We definitely sohuld not have to mention all this state. We should be able to say `<block> BLOCK </block>`

ehildenb: Or, if you use `BlockCellFragment` instead, just `{ NETWORK | BLOCK }`

evm.md
```diff
@@ -182,7 +185,78 @@ Our semantics is modal, with the initial mode being set on the command line via
...
+           <gasLimit>         _ => HL       </gasLimit>         // I_Hl
+           <gasUsed>          _ => HG       </gasUsed>          // I_Hg
+           <timestamp>        _ => HS       </timestamp>        // I_Hs
+           <extraData>        _ => HX       </extraData>        // I_Hx
+           <mixHash>          _ => HM       </mixHash>          // I_Hm
+           <blockNonce>       _ => HN       </blockNonce>       // I_Hn
+
+           <ommerBlockHeaders> _ => OBH </ommerBlockHeaders>
+           <blockhash>         _ => BH  </blockhash>
+         </block>
```
ehildenb: And this would be `<block> _ => BLOCK </block>`

evm.md
anvacaru: I used `BlockCell` to be able to identify a block in `<blockList>` using the `BLOCKNUM` in `#getBlockchainState(Int)`

ehildenb: You can still do that by matching on `<block>-fragment ... </block>-fragment`

ehildenb: Actually, I'm confused about how `#getBlockchainState` is working. (1) is it correct to index based on `BLOCKNUM`, and (2) should we be dropping blocks as we go, or just going through and finding the one we need. @dwightguth seemed to think we should be keeping all the blocks. Would it make more sense to use configuration:

```
<blocks>
  <block multiplicity="*" type="Map">
    <number> ... </number>            // Index on block number if needed
    ...
    <txList> ... </txList>                   // Store list of relevant transaction identifiers here, but leave transactions in `<network>` state?
  </block>
</blocks>
```

PR #463, opened 2019-08-27
Fix evm snapshots 

anvacaru:
Modified `evm_snapshot` and `evm_revert` to store:
- current `blockList` cell
- current `block` cell
- current `network` cell

Added a new rule for the case in which an `evm_revert` request was received and the snapshot list is empty.


web3.md
```diff
@@ -362,30 +362,46 @@ WEB3 JSON RPC
...
-         <params> ( [ DATA:Int, .JSONList ] => [ .JSONList ] ) </params>
+         <params> ( [ DATA:Int, .JSONList ] ) </params>
          <snapshots> ( SNAPSHOTS => range(SNAPSHOTS, 0, DATA ) ) </snapshots>
+      requires size(SNAPSHOTS) >Int (DATA +Int 1)
+
+    rule <k> #evm_revert => #sendResponse( "error": {"code": -32000, "message": "Incorrect number of arguments. Method 'evm_revert' requires exactly 1 arguments. Request specified 0 arguments: [null]."} )  ... </k>
+         <params> [ .JSONList ] </params>
+
+     rule <k> #evm_revert => #sendResponse ( "result" : false ) ... </k>
+          <snapshots> .List </snapshots>
```
ehildenb: We need to handle the case where we're asked to revert to a snapshot outside the length of the list (so that KEVM doesn't get stuck). In general you need to handle all cases here, because KEVM getting stuck would make the client unusable and confuse people a lot.

anvacaru: Made an `owise` rule to handle all cases which are not already defined

PR #470, opened 2019-08-30
Add RPC method `firefly_addAccount` 

anvacaru:
1. Implemented `firefly_addAccount` which takes in:
```json
{
    "address": "0x0EFF7a5a881311f04505Dd0beD278D0632E88E9F",
     "balance": "0x100",
     "nonce": "0x3",
     "code": "0x6080604052348...",
     "storage": {
                        "0": "0x64"
                }
}
```

and inserts the account into the `<accounts>` cell if the address hasn't been used yet. It can also take a private account key instead of an address:
```json
{
     "key": "0x2b8ba85ff7f2c7fa62cddb20ae40ef2553b9fdbff086f60d1647ecaab15af867",
     "balance": "0x100",
     "nonce": "0x3",
     "storage": {
                    "0": "0x64"
                }
}
```
Only the `address` and `key` fields are mandatory.

2. Added test files for `eth_getBalance`, `eth_getStorageAt`, `eth_getCode`, and `eth_getTransactionCount` rpc methods.

3. Added `[owise]` rules for `eth_getBalance`, `eth_getStorageAt`, `eth_getCode`, and `eth_getTransactionCount` rpc methods to catch requests with a different number of arguments.


gtrepta:
My only comment is that this doesn't create a mapping to a private key in the `<accountKeys>` cell. If that doesn't matter for what this method is intended for then everything else looks good to me.

web3.md
```diff
@@ -587,5 +589,32 @@ WEB3 JSON RPC
...
+    syntax KItem ::= "#firefly_addAccount"
+ // --------------------------------------
+    rule <k> #firefly_addAccount ... </k>
+         <params> [ (ADDRESS => #parseHexWord(ADDRESS)),
+                    (BALANCE => #parseHexWord(BALANCE)),
+                    (NONCE => #parseHexWord(NONCE)),
+                    (CODE => #parseByteStack(CODE)),
+                    ({ STORAGE } => #parseMap({ STORAGE })),
+                    .JSONList ]
+         </params>
```
ehildenb: Format like so:

```
         <params>
             [ (ADDRESS     => #parseHexWord(ADDRESS))
             , (BALANCE     => #parseHexWord(BALANCE))
             , (NONCE       => #parseHexWord(NONCE))
             , (CODE        => #parseByteStack(CODE))
             , ({ STORAGE } => #parseMap({ STORAGE }))
             , .JSONList
             ]
         </params>
```

web3.md
```diff
@@ -587,5 +589,32 @@ WEB3 JSON RPC
...
+
+    rule <k> #firefly_addAccount => (#newAccount ADDRESS) ~> #loadAccountData ~> #sendResponse( "result": true ) ... </k>
```
ehildenb: No need for parans around `#newAccound ADDRESS` I don't think.

ehildenb: Also, we should handle the case that the account already exists, sending back an error.

web3.md
```diff
@@ -587,5 +589,32 @@ WEB3 JSON RPC
...
+         <params> [ ADDRESS, BALANCE, NONCE, CODE, STORAGE, .JSONList ] </params>
+
+
+    syntax KItem ::= "#loadAccountData"
+ // ----------------------------------
```
ehildenb: Formatting (add another `-`)

web3.md
```diff
@@ -587,5 +589,28 @@ WEB3 JSON RPC
...
+      requires isString( #getJSON("address", J) )
+
+    rule <k> #firefly_addAccount =>  #sendResponse( "error": {"code": -32025, "message":"Method 'firefly_addAccount' has invalid arguments"} ) ... </k> [owise]
+
+    syntax KItem ::= "#loadAccountData" JSON
+ // ----------------------------------------
+    rule <k> #loadAccountData J => . ... </k>
+         <account>
+           <acctID>  ACCTID  </acctID>
+           <balance>     ( _ => #if isString( #getJSON("balance",  J) ) #then #parseHexWord(#getString("balance", J)) #else 100                    #fi )     </balance>
```
ehildenb: Let's use the same default values that the configuration declaration provides.

web3.md
```diff
@@ -587,5 +589,28 @@ WEB3 JSON RPC
...
+           <code>        ( _ => #if isString( #getJSON("code",     J) ) #then #parseByteStack(#getString("code",  J)) #else .ByteArray:AccountCode #fi )        </code>
+           <nonce>       ( _ => #if isString( #getJSON("nonce",    J) ) #then #parseHexWord(#getString("nonce",   J)) #else 0                      #fi )       </nonce>
+           <storage>     ( _ => #if isString( #getJSON("storage",  J) ) #then #parseMap(#getString("storage",     J)) #else .Map                   #fi )     </storage>
+           <origStorage> ( _ => #if isString( #getJSON("storage",  J) ) #then #parseMap(#getString("storage",     J)) #else .Map                   #fi ) </origStorage>
```
ehildenb: Storage will be a `Map` not a `String`. So you just want to call `#getJSON` directly instead of `#getString`, and no need for the `isString`.

tests/web3/firefly_addAccount.in.json
```diff
@@ -0,0 +1,13 @@
...
+  "method": "firefly_addAccount",
+  "params": [
+    {
+      "address": "0xf2b39e6c322c897bbe7a60f301c7ba90ad1c7bc8814e53b467be423e38ac2eb6",
+      "balance": "0x100",
+      "nonce": "0x3",
+      "storage": { "0" : "0x64" }
+    }
+  ]
+}]
```
ehildenb: Let's also test `firefly_addAccount` multiple times, then do a few `getStorageAt`, `getBalanceAt`.

For now, just always use `pending`.

web3.md
```diff
@@ -637,5 +651,61 @@ WEB3 JSON RPC
...
+         <account>
+           <acctID>  ACCTID  </acctID>
+           <balance>     ( _ => #if            isString( #getJSON("balance",  J) ) #then #parseHexWord(#getString("balance", J)) #else 100                    #fi )     </balance>
+           <code>        ( _ => #if            isString( #getJSON("code",     J) ) #then #parseByteStack(#getString("code",  J)) #else .ByteArray:AccountCode #fi )        </code>
+           <nonce>       ( _ => #if            isString( #getJSON("nonce",    J) ) #then #parseHexWord(#getString("nonce",   J)) #else 0                      #fi )       </nonce>
+           <storage>     ( _ => #if notBool isJSONUndef( #getJSON("storage",  J) ) #then #parseMap(#getJSON("storage",       J)) #else .Map                   #fi )     </storage>
+           <origStorage> ( _ => #if notBool isJSONUndef( #getJSON("storage",  J) ) #then #parseMap(#getJSON("storage",       J)) #else .Map                   #fi ) </origStorage>
+         </account>
+      requires isString( #getJSON("key", J) )
+       andBool #addrFromPrivateKey(#getString("key", J)) ==Int ACCTID
```
ehildenb: This will turn into quite an inefficient map lookup, since we're computing the key in the side-condition. So instead of looking up exactly the account to load data into, we need to cycle through all the accounts and check "is this the right account".

Better to directly look up on the `ACCTID`, instead of constraining it in a side-condition.

PR #474, opened 2019-09-03
Update k submodule and remove references to rust

dwightguth:


dwightguth:
It would be better if we just moved the check into topCell and made it lazy instead of such an invasive change to the file.

PR #475, opened 2019-09-04
Quieter less redundant build

ehildenb:
-   Organize the Makefile a bit.
-   Don't do `mkdir -p ...` every time we `tangle`.
-   Let `make` tell us what it's doing (no need for `@echo` lines).


ehildenb:
Hmmm, some weirdness going on with rebuilds I'm investigating.

ehildenb:
weird rebuild issues are gone now

PR #481, opened 2019-09-11
Update test submodule/testing harness

ehildenb:
-   Updates to newest test submodule, which has moved the BlockchainTests/GeneralStateTests to a new subdirectory (but they are largely unchanged).
-   Explicitely lists the slow tests in a file, instead of using wildcards to expand them. This allows admitting more tests (since our wildcards were too generous). Timeout for being a "slow" test is 20s, passing or not.
-   Changes `test-slow-*` targets to `test-rest-*` targets, to run all tests which are either slow or failing quickly.

To collect slow/failing tests, the procedure used was (eg. for llvm backend):

-    Remove `tests/slow.llvm` and `tests/failing.llvm`
-    Run `make test-conformance TEST='profile .build/logs/test.profile timeout 20s ./kevm' -j16 -k TEST_CONCRETE_BACKEND=llvm`
-    Lines beginning with `124` in `.build/logs/test.profile` go into `tests/slow.llvm`.
-    Lines beginning with `1` in `.build/logs/test.profile` go into `tests/failing.llvm`.

Where the `profile` command is:

```
#!/usr/bin/env bash

set -euo pipefail

test_log="$1" ; shift
mkdir -p $(dirname "$test_log")
exit_status='0'
`which time` --quiet --format '%x %es %Us %Ss %MKB %C' --output "$test_log" --append \
    "$@" &>/dev/null \
    || exit_status="$?"
exit "$exit_status"
```


ehildenb:
Ok, before the update there are 11146 tests listed when doing make --dry-run test-conformance, and 11143 after.
It looks like several test files were shuffle around in their directories too.
Including some of the "slow" tests which we eliminated, which have been moved, so they are no longer eliminated.
I'm going to do a profiling run to find out which tests are absurdly slow in the new setup, and see if they correspond to similar tests from before.
I'm also going to count the total number of lines of test files before and after the update, as a measure of whether the contents changed very much.
After the update, there were a total of 5573288 lines worth of test files. Before the update there were 5723661. So I think the few test files that were deleted is all that changed, and otherwise we're safe to use the tests in LegacyTests instead, once we find out which ones are the poor performing ones.

PR #494, opened 2019-09-18
Create transaction receipt

anvacaru:
1. I've added a subconfiguration which stores the transaction receipt and has the following fields:
`txHash` - hash of the transaction which generated the current receipt
`txCumulativeGas` - the cumulative gas used in the block containing the transaction receipt as of immediately after the transaction has happened
`logSet` - the set of logs created through execution of the transaction
`bloomFilter` - the Bloom filter composed from information in those logs
`txStatus` - the status code of the transaction

2. Changed tangler from `node` execution to `concrete` execution.

3. Added a test which deploys an `ERC20` contract with 100 tokens assigned to an address.


web3.md
```diff
@@ -741,6 +750,55 @@ eth_sendRawTransaction
...
+               <logSet> LOGS </logSet>
+               <bloomFilter> LOGSBLOOM </bloomFilter>
+               <txStatus> 1 </txStatus>
+             </txReceipt>
+           )
+           ...
+         </txReceipts>
+         <statusCode> EVMC_SUCCESS </statusCode>
+         <logsBloom> LOGSBLOOM </logsBloom>
+         <gas> GLIMIT </gas>
```
ehildenb: We should probably be using `<gasUsed>` here instead of `<gas>`.

It's being updated each transaction anyway (double-check that `#executeTx` actually does this).

web3.md
```diff
@@ -741,6 +750,35 @@ eth_sendRawTransaction
...
+    rule <k> #makeTxReceipt TXID => . ... </k>
+         <txReceipts>
+           ( .Bag
+          => <txReceipt>
+               <txHash> "0x" +String #hashSignedTx (TXID) </txHash>
+               <txCumulativeGas> CGAS </txCumulativeGas>
+               <logSet> LOGS </logSet>
+               <bloomFilter> #bloomFilter(LOGS) </bloomFilter>
+               <txStatus> bool2Word(STATUSCODE ==K EVMC_SUCCESS) </txStatus>
+             </txReceipt>
```
ehildenb: It looks like here https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionreceipt that we'll also need the gas used by the transaction itself, _not_ just the cumulative gas used.

You can calculate it by doing `<txGasLimit> -Int <gas>`, I believe. Store it in a new cell `<txGas>`.

anvacaru: The tx receipt described in JSON RPC also contains some block info and has a few extra fields like `txGas`. I've only implemented the ones which are documented in the yellowpaper.
https://ethereum.github.io/yellowpaper/paper.pdf 
```
4.3.1:
...
The transaction receipt,R, is a tuple of four items com-
prising:  the cumulative gas used in the block containing
the transaction receipt as of immediately after the trans-
action has happened,Ru, the set of logs created through
execution of the transaction,Rland the Bloom filter com-
posed from information in those logs,Rband the status
code of the transaction,
Rz:(20)R≡(Ru,Rb,Rl,Rz)
```

PR #498, opened 2019-09-23
Implement `eth_estimateGas` in web3

anvacaru:
Fixes: runtimeverification/firefly#186


web3.md
```diff
@@ -1035,6 +1044,69 @@ loadCallSettings
...
+         <params> [ ({ _ } #as J), TAG, .JSONList ] </params>
+      requires notBool isString( #getJSON("from", J) )
+
+    syntax KItem ::= "#eth_estimateGas_intermediate"
+ // ------------------------------------------------
+    rule <k> #eth_estimateGas_intermediate => #prepareTx TXID ~> #eth_estimateGas_finalize ~> GUSED ... </k>
+         <txPending> ListItem( TXID ) ... </txPending>
+         <gasUsed> GUSED </gasUsed>
+
+    syntax KItem ::= "#eth_estimateGas_finalize"
```
ehildenb: Make the original gas used an argument to `#eth_estimateGas_finalize`.

eg.

```
syntax KItem ::= "#eth_estimateGas_finalize" Int
```

web3.md
```diff
@@ -1018,5 +1020,47 @@ loadCallSettings
...
+    syntax KItem ::= "#eth_estimateGas_finalize" Int
+ // ------------------------------------------------
+    rule <k> #eth_estimateGas_finalize INITGUSED:Int => #sendResponse ("result": #unparseQuantity( GUSED -Int INITGUSED )) ... </k>
+         <gasUsed> GUSED </gasUsed>
+
+    syntax Mode ::= "ACCUMULATE_GAS"
+ // --------------------------------
+    rule <k> G:Int ~> #deductGas => . ... </k>
+         <gas>  GUSED => GUSED +Int G </gas>
+         <mode> ACCUMULATE_GAS        </mode>
```
gtrepta: Make this rule priority 25, get rid of second rule.

ehildenb:
We need to investigate what to do with `<gas>` here. `eth_estimateGas` assumes that you either (1) supply a gas limit and still deduct gas, or (2) don't supply one, and it uses the block gas limit.

We should do the same instead of accumulating gas.

We should also check `ganache-cli`s behavior when not enough gas is supplied. Specifically, send an `eth_estimateGas`, and only supply `2` gas (or something small). Does `ganache-cli` report `OUT_OF_GAS` or just execute it anyway? If it reports `OUT_OF_GAS`, then we should deduct gas. If it executes anyway, we should accumulate instead of deducting.

web3.md
```diff
@@ -1018,5 +1020,47 @@ loadCallSettings
...
+          ~> load "transaction" : { !ID : J }
+          ~> signTX !ID #parseHexWord(#getString("from", J))
+          ~> #setMode ACCUMULATE_GAS
+          ~> #prepareTx !ID
+          ~> #eth_estimateGas_finalize GUSED
+         ...
+         </k>
+         <params> [ ({ _ } #as J), TAG, .JSONList ] </params>
+         <gasUsed> GUSED </gasUsed>
+      requires isString(#getJSON("from", J) )
```
gtrepta: Load the `<gasLimit>` cell into the `<gas>` cell, and then if `J` has the `gasLimit` field, it will overwrite it.

PR #503, opened 2019-09-24
loadCallState

gtrepta:
Fixes: https://github.com/runtimeverification/firefly/issues/189


state-loader.md
```diff
@@ -164,25 +164,71 @@ Here we load the environmental information.
...
+
+    rule <k> loadCallState { "gas" : GLIMIT:Int, REST => REST }      ... </k> <gas>       _ => GLIMIT </gas>
+    rule <k> loadCallState { "gasPrice" : GPRICE:Int, REST => REST } ... </k> <gasPrice>  _ => GPRICE </gasPrice>
+    rule <k> loadCallState { "value" : VALUE:Int, REST => REST }     ... </k> <callValue> _ => VALUE  </callValue>
+    rule <k> loadCallState { "data" : DATA:ByteArray, REST => REST } ... </k> <callData>  _ => DATA   </callData>
+    rule <k> loadCallState { "nonce" : _, REST => REST } ... </k>
+
+    rule <k> loadCallState { .JSONList } => . ... </k>
+
+    rule <k> loadCallState TXID:Int
```
anvacaru: This rule is specific for `web3.md`, perhaps we should keep it there?

driver.md
```diff
@@ -266,8 +266,26 @@ Note that `TEST` is sorted here so that key `"network"` comes before key `"pre"`
     rule <k> run TESTID : { KEY : (VAL:JSON) , NEXT , REST } => run TESTID : { NEXT , KEY : VAL , REST } ... </k>
       requires KEY in #execKeys
 
-    rule <k> run TESTID : { "exec" : (EXEC:JSON) } => load "exec" : EXEC ~> start ~> flush ... </k>
+    rule <k> run TESTID : { "exec" : (EXEC:JSON) } => loadCallState EXEC ~> start ~> flush ... </k>
     rule <k> run TESTID : { "lastblockhash" : (HASH:String) } => #startBlock ~> startTx    ... </k>
+
+    rule <k> load "exec" : J => loadCallState J ... </k>
```
ehildenb: Seems that this rule and the rule above for `run TESTID : { "exec" : EXEC } => loadCallState EXEC ~> start ~> flush ... </k>` are redundant. Is this one needed?

gtrepta: I don't know if the rule above is needed. But the rule I added is needed for some `.evm` test scripts.

ehildenb: Can we translate them to `loadCallState` instead?

gtrepta: It didn't work right away.

PR #505, opened 2019-09-25
Web3:  Transaction is successfully executed even if `txGasLimit` is exceeded

anvacaru:
Fixes: https://github.com/runtimeverification/firefly/issues/191

### DIFF
--- a/web3.md
+++ b/web3.md
@@ -541,6 +541,7 @@ eth_sendTransaction
     rule <k> #eth_sendTransaction_load J => mkTX !ID:Int ~> #loadNonce #parseHexWord( #getString("from",J) ) !ID ~> load "transaction" : { !ID : J } ~> signTX !ID #parseHexWord( #getString("from",J) ) ~> #prepareTx !ID ~> #eth_sendTransaction_final !ID ... </k>
 
     rule <k> #eth_sendTransaction_final TXID => #sendResponse( "result": "0x" +String #hashSignedTx( TXID ) ) ... </k>
+        <statusCode> EVMC_SUCCESS </statusCode>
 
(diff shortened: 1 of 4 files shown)


web3.md
```diff
@@ -824,11 +831,48 @@ loadCallSettings
...
+           <msgID>      TXID     </msgID>
+           <txGasLimit> GLIMIT   </txGasLimit>
+           <data>       DATA     </data>
+           ...
+         </message>
+      requires ( GLIMIT -Int G0(SCHED, DATA, false) ) <Int 0
+
+    rule <k> #validateTx TXID => #executeTx TXID ~> #makeTxReceipt TXID ... </k>
+         <schedule> SCHED </schedule>
+         <callGas> _ => GLIMIT -Int G0(SCHED, CODE, true) </callGas>
```
ehildenb: Can we make this one rule instead of two by putting `ACCOUNT_TO ==K .Account` as the third argument to `G0`

ehildenb:
Maybe we should check if the intrinsic gas of the transaction is available before even trying to execute it.

web3.md
```diff
@@ -823,12 +830,35 @@ loadCallSettings
...
+    rule <k> #validateTx TXID => . ... </k>
+         <statusCode> ( _ => EVMC_OUT_OF_GAS) </statusCode>
+         <schedule> SCHED </schedule>
+         <message>
+           <msgID>      TXID     </msgID>
+           <txGasLimit> GLIMIT   </txGasLimit>
+           <data>       DATA     </data>
+           ...
+         </message>
+      requires ( GLIMIT -Int G0(SCHED, DATA, false) ) <Int 0
```
ehildenb: Should this actually say `G0(SCHED, DATA, ACCTTO ==K .Account)`?

anvacaru: Yes it should, for both rules.

PR #506, opened 2019-09-25
Replace #sizeWordStack(_, _) helper with direct recursion

ttuegel:
The change is only applied to the symbolic backends; the tail-recursive helper
is retained for concrete backends. The change allows using the `memo` attribute
with the Haskell backend, reducing runtime of the sum-to-n spec by over 50%. We
are also able to remove lemmas from several specs.

### DIFF
--- a/data.md
+++ b/data.md
@@ -451,14 +451,23 @@ A cons-list is used for the EVM wordstack.
 -   `#sizeWordStack` calculates the size of a `WordStack`.
 -   `_in_` determines if a `Int` occurs in a `WordStack`.
 
-```k
+```{.k .symbolic}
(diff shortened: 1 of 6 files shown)


dwightguth:
This is okay, I suppose, but I'd be happier with a solution that just made computing the size be constant time instead.

ttuegel:
> I'd be happier with a solution that just made computing the size be constant time instead.

After discussion with @ehildenb this morning, we don't think there is a way to do that in the semantics.

ttuegel:
I would have to coordinate changes to verified-smart-contracts to make this change, and I don't want to do that. However, we can probably achieve the same effect by applying `memo` elsewhere.

PR #509, opened 2019-09-27
Web3: Implement `NOGAS` mode

anvacaru:
Fixes: https://github.com/runtimeverification/firefly/issues/177

### DIFF
--- a/web3.md
+++ b/web3.md
@@ -1051,15 +1051,6 @@ NOGAS Mode
 ```k
     syntax Mode ::= "NOGAS"
  // -----------------------
-    rule <k> #gas [ OP , AOP ]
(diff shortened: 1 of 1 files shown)


web3.md
```diff
@@ -1040,6 +1041,31 @@ Transaction Receipts
...
+    syntax Mode ::= "NOGAS"
+ // -----------------------
+    rule <k> #gas [ OP , AOP ]
+          => #if #usesMemory(OP) #then #memory [ AOP ] #else .K #fi
+          ~> #gas [ AOP ]
+         ...
+        </k>
+         <mode> MODE </mode>
+     requires MODE =/=K NOGAS
+     [priority(25)]
```
ehildenb: Pretty sure we don't need this rule, if the rule below doesn't fire the one from the original semantics will.

PR #514, opened 2019-10-03
Sum to n haskell proof

ehildenb:


yzhang90:
The k version and the haskell backend version is quite old...

PR #518, opened 2019-10-15
Web3: Implement eth_accounts

anvacaru:
Fixes: https://github.com/runtimeverification/firefly/issues/210.


web3.md
```diff
@@ -334,6 +336,17 @@ WEB3 JSON RPC
...
+    syntax KItem ::= "#eth_accounts"
+ // --------------------------------
+    rule <k> #eth_accounts => #sendResponse( "result" : [List2JSONList(Set2List(AS))] ) ...</k>
+         <activeAccounts> AS </activeAccounts>
+    
+    syntax JSONList ::= List2JSONList ( List ) [function]
+ // -----------------------------------------------------
+    rule List2JSONList (.List)                           => .JSONList
+    rule List2JSONList (ListItem(I:Int) L)               => I, List2JSONList(L)
+    rule List2JSONList (ListItem({I:Int | _:OpCode }) L) => I, List2JSONList(L)
```
ehildenb: Probably the accounts need to be converted to hex strings (not just the integers) (20 bytes wide).

ehildenb: And let's name this `AccountList2JSONList`, not `List2JSONList`, since there are other times we want to convert a `List` to a `JSONList`.

PR #519, opened 2019-10-15
Merkle tree

gtrepta:


ehildenb:
Before merge let's switch to just using `build-llvm` instead of `build-merkle`, since the extra build target is extra maintenance.

PR #520, opened 2019-10-16
make ByteArray a synonym and remove casts

dwightguth:


denis-bogdanas:
It works with kprove now. I just had to remove casts. Please merge.

PR #523, opened 2019-10-17
Suhabe benchmarks

denis-bogdanas:


kevm
```diff
@@ -57,7 +57,18 @@ run_kast() {
...
+    run_dir=$(dirname $run_file)
+    smt_file=$(ls ${run_dir}/*.smt2 2>/dev/null | xargs || true)
+    if [ -z "$smt_file" ]; then
+      kprove --directory "$backend_dir" "$run_file" "$@"
+    elif [ -f "$smt_file" ]; then
+      # echo "SMT file:" $smt_file
+      kprove --directory "$backend_dir" --smt-prelude "$smt_file" "$run_file" "$@"
+    else
+      fatal "More than one SMT file found: '$smt_file'"
+    fi
```
ehildenb: Let's give the smt2 file a specific name so we don't have to do this weird search for it. For now, let's always assume it's called `evm.smt2`.

Then the code can become:

```
run_dir="$(dirname "$run_file")"
smt_file="$run_dir/evm.smt2"
args=()
[[ -z "$smt_file" ]] || args=(--smt-prelude "$smt_file")
kprove --directory "$backend_dir" "${args[@]-}" "$run_file" "$@"

denis-bogdanas:
@ehildenb I restored the original smt2 discovery. In VSC we have different names for smt2 file for different projects. So if you want kevm script to be usable in VSC, we better keep this general discovery mechanism.

ehildenb:
@denis-bogdanas this is still too fragile in my opinion. I either want the smt file (i) passed in as a paramater or (2) set as an environment variable with some sane default.

So you can say `KEVM_SMT_PRELUDE="${KEVM_SMT_PRELUDE:-evm.smt2}"` somewhere before the line, and then do the check if the file exists. Then VSC can override the smt prelude for each individual file if it likes.

PR #527, opened 2019-10-20
Web3: Implement eth_getTransactionReceipt

anvacaru:
Fixes: runtimeverification/firefly#218


ehildenb:
For `blockhash` field, let's put a dummy value there for now.

Transaction index probably refers to the order of transactions in the block itself (not a global list of all transactions).

PR #533, opened 2019-11-01
stateRoot Implementation

gtrepta:
fixes: https://github.com/runtimeverification/firefly/issues/196


ehildenb:
Let's add tests of the following cases:

-   Only code present (has code, but no balance and no nonce)
-   Only nonce present (has non-zero nonce, but no balance and no code)
-   All three present (has code, balance, and nonce).

PR #544, opened 2019-11-12
CREATE opcode proof

yzhang90:


tests/specs/opcodes/create-summary-spec.k
```diff
@@ -0,0 +1,175 @@
...
+      <acctID> #newAddr(ACCT_ID, NONCE) </acctID>
+      <balance> VALUE </balance>
+      <nonce> 1 </nonce>
+      <code> #generatedCode </code>
+      ...
+    </account>
+  )
+    ...
+  </accounts>
+</network>
```
ehildenb: Remove this duplicate, since it could become out of sync with the proof itself. Leave the comment for explanation.

PR #545, opened 2019-11-13
Fix account creation case in eth_sendTransaction

gtrepta:
fixes: https://github.com/runtimeverification/firefly/issues/241


web3.md
```diff
@@ -852,7 +852,7 @@ Transaction Receipts
          </message>
          <account>
            <acctID> ACCTFROM </acctID>
-           <balance> BAL => BAL -Int (GLIMIT *Int GPRICE) </balance>
+           <balance> BAL => BAL -Int 2 *Int (GLIMIT *Int GPRICE) </balance>
```
gtrepta: For some reason, the `from` account ends with too much gas being refunded (as if it is being refunded twice), and so far I haven't been able to find out where this is happening.

ehildenb: @gtrepta it looks like `finalizeTx` is already refunding gas to the account (at the same time that it awards the miner gas.

So you probably don't need `#catchHaltTx _` to do a `#refund GAVAIL`

web3.md
```diff
@@ -891,9 +886,21 @@ Transaction Receipts
...
- // -------------------------------
-    rule <k> #halt ~> #catchHaltTx => . ... </k>
+    syntax KItem ::= "#catchHaltTx" Account
+ // ---------------------------------------
+    rule <statusCode> _:ExceptionalStatusCode </statusCode>
+         <k> #halt ~> #catchHaltTx _ => #popCallStack ~> #popWorldState ... </k>
+
+    rule <statusCode> EVMC_REVERT </statusCode>
+         <k> #halt ~> #catchHaltTx _ => #popCallStack ~> #popWorldState ~> #refund GAVAIL ... </k>
+         <gas> GAVAIL </gas>
```
ehildenb: Here you probably don't need to do the `#refund GAVAIL`

ehildenb:
You'll have to look carefully at the refunds that `finalizeTx` and `mkCodeDeposit` make. it looks like `finalizeTx` already awards a gas refund, and so does `mkCodeDeposit` when you have `EVMC_SUCCESS` by calling `finalizeCodeDeposit`.

web3.md
gtrepta: I don't think that's it. `#refund` adds to the `<gas>` cell. It's used after a `#popCallStack` to preserve the remaining gas.

PR #547, opened 2019-11-15
Newline outputs on tests/web3 for easy diffing

ehildenb:
Fixes: runtimeverification/firefly#256

Updates the tests in `tests/web3` to use `jq` to format the output.

This makes diffing large tests easier.


Makefile
```diff
@@ -415,6 +416,11 @@ tests/%.run-web3: tests/%.in.json
 	$(CHECK) tests/$*.expected.json tests/$*.out.json
 	rm -rf tests/$*.out.json
 
+tests/%.update-web3: tests/%.in.json
+	tests/web3/runtest.sh $< tests/$*.out.json $(KEVM_WEB3_ARGS)
+	$(UPDATE) tests/$*.out.json tests/$*.expected.json
+	rm -rf tests/$*.out.json
+
```
gtrepta: I'm guessing you used this to update the *.expected.json files. Does this helper functionality need to remain in the PR? And wouldn't it have been a better idea to pipe the expected files through `jq` to update them instead of overwriting them with kevm's output?

ehildenb: I'd like to keep the helper in, can be useful when needing to update from time to time.

And my approach is usually to just update the expected files then use various `git diff` incantations to see the difference. Either way works.

Several other repos have `update-*` targets too, just need to introduce them here.

PR #564, opened 2019-11-21
Erc20detailed partial test fixes

gtrepta:
fixes: https://github.com/runtimeverification/firefly/issues/281


tests/web3/erc20/erc20detailed-partial.in.json
```diff
@@ -182,6 +194,12 @@
     false
   ]
 }
+{
+  "jsonrpc": "2.0",
+  "id": 0,
+  "method": "evm_increaseTime",
+  "params": [ 10 ]
+}
```
ehildenb: Maybe we should be increasing the time when we do the `firefly_genesisBlock`?

gtrepta: Why should we do that? We would still need `evm_increaseTime` for subsequent blocks.

ehildenb: Does `evm_mine` increase the time in Ganache? If so, we need to do the same.

Ethereum blocks come roughly 10s apart, so my guess is that Ganache is adding 10s to the time each time a block is mined.

ehildenb:
Please also update the other two tests `tests/web3/*-partial.*`.

PR #589, opened 2019-12-02
Replace [concrete] attributes by rule tags

daejunpark:


daejunpark:
Following from PR #575, (especially, https://github.com/kframework/evm-semantics/pull/575#discussion_r350962237), this PR fixes the semantics rule duplication problem in the specs. Essentially, there are three sets of concrete rules; one used by Bihu specs, another used by ERC20 specs, and another by the other specs.

Now no `[concrete]` attribute appears in the semantics, and no semantic rule duplication is made in the specs.

ehildenb:
Does the Haskell backend understand this `--concrete-rules` option?

ttuegel:
@ehildenb No. Is this implemented in the backend only?

daejunpark:
For now, the option is implemented in the java backend only. I think it can be implemented at the frontend level though. Should we wait until that happens?

daejunpark:
@ttuegel What do you think how to proceed? Is it acceptable to implement the option in the haskell backend? If so, should we wait to merge this until that happens? Or, do you have another opinion?

daejunpark:
Seems that Jenkins build failure has nothing to do with this PR. Re-ran it but failed again at a different step.

daejunpark:
@ehildenb this is ready to review.

PR #590, opened 2019-12-02
Update test submodule to post-Istanbul tests

ehildenb:
This just updates the test submodule to include the newest changes. @MrChico recommended tracking the develop branch instead of the release tags.

This also updates the failing/slow test lists to incorporate the changes to the testing submodule.

A timeout of 120s and memory limit of 10GB on my server was used to determine if a test is "slow".

Also use `python` instead of `python3` in a place.


kast-json.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
dwightguth: python3 is not necesarily going to exist in your path when you have python 3 installed ,so this is not a good solution. It's better to have the script work with both python 2 or python 3, which is what it did before.

ehildenb:
Regarding the `"postStateHash"` key, I've added an issue here: https://github.com/ethereum/tests/issues/668

Otherwise, now it looks like we're just moving some tests from `failing` to `slow`, and removing some tests from `slow`.

ehildenb:
Alright, I'm not going to touch this one anymore. I don't really care about the OCaml backend lists.

For the LLVM backend lists, the additional `failing` testst that used to be `slow` are due to the increased timeout I used for making a test slow. The new timeout is 120s, and <10G RAM. If it uses >120s or >10G RAM, it's `slow` regardless of whether it's passing or not.

The additional `slow` tests come from several sources. Some of them are because the additional `postStateHash` allows the test to get further. Most of them are because of the additional restriction I added that it must use <10GB RAM to not be `slow`.

kore-json.py
```diff
@@ -48,13 +48,14 @@ def print_kast(data, sort="SortJSON"):
...
 print_kast(data)
-print("))),Lbl'UndsPipe'-'-GT-Unds'{}(inj{SortKConfigVar{}, SortKItem{}}(\dv{SortKConfigVar{}}(\"$SCHEDULE\")),inj{SortSchedule{}, SortKItem{}}(", end=' ')
+sys.stdout.write("))),Lbl'UndsPipe'-'-GT-Unds'{}(inj{SortKConfigVar{}, SortKItem{}}(\dv{SortKConfigVar{}}(\"$SCHEDULE\")),inj{SortSchedule{}, SortKItem{}}( ")
 print_klabel(sys.argv[2])
-print("()))),Lbl'UndsPipe'-'-GT-Unds'{}(inj{SortKConfigVar{}, SortKItem{}}(\dv{SortKConfigVar{}}(\"$MODE\")),inj{SortMode{}, SortKItem{}}(", end=' ')
+sys.stdout.write("()))),Lbl'UndsPipe'-'-GT-Unds'{}(inj{SortKConfigVar{}, SortKItem{}}(\dv{SortKConfigVar{}}(\"$MODE\")),inj{SortMode{}, SortKItem{}}( ")
 print_klabel(sys.argv[3])
-print("()))))")
-print()
+sys.stdout.write("()))))")
```
dwightguth: this should have a \n too.

ehildenb:
```suggestion
sys.stdout.write("()))))\n")
```

ehildenb: Done.

PR #592, opened 2019-12-02
Allow user to override opam command

ttuegel:
The user may override the opam command by setting the OPAM environment
variable. The default is equivalent to `OPAM=opam`.

opam is not required (or even available) in some environment, such as under Nix.


ehildenb:
What problem does this solve? You don't need to run `make ocaml-deps` if you don't want ocaml, and `make build` doesn't build the OCaml backend by default.

ttuegel:
@ehildenb It lets me run `make build-ocaml` and `make test-conformance` with the OCaml backend.

PR #605, opened 2019-12-06
evm.md: Implement EIP 1108 for Istanbul

ehildenb:
Fixes: runtimeverification/firefly#342

Factored out from @MrChico 's PR #444.

Implements https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1108.md.


ehildenb:
No, not without the other EIPs. This really means that the test-set doesn't have good separation of concerns.

PR #632, opened 2019-12-16
Data.md refactoring

gtrepta:
fixes: https://github.com/runtimeverification/firefly/issues/395

This does the following refactorings:

- Moves Word operations, WordStack, ByteArray, and Account related semantics into an `EVM-TYPES` module
- Moves hashing/address related computations, string format conversions, rlp encoding/decoding, and merkle patricia tries into a `SERIALIZATION` module
- Moves the `JSON` and `JSON-RPC` modules into a `json.md` file
- Some rules that didn't have an obvious place were moved on a case-by-case basis, check the commits


evm.md
```diff
@@ -1263,7 +1279,18 @@ These rules reach into the network state and load/store from account storage:
...
+
+-   `#lookup` looks up a key in a map and returns 0 if the key doesn't exist, otherwise returning its value.
 
+```k
+    syntax Int ::= #lookup ( Map , Int ) [function]
+ // -----------------------------------------------
+    rule [#lookup.some]: #lookup( (KEY |-> VAL) M, KEY ) => VAL
+    rule [#lookup.none]: #lookup(               M, KEY ) => 0 requires notBool KEY in_keys(M)
+```
+
```
ehildenb: This should probably stay in either `data.md` or `evm-types.md`.

PR #640, opened 2019-12-19
Decontextualize more

gtrepta:
fixes: https://github.com/runtimeverification/firefly/issues/397

This finishes removing all of the contextual functions in web3.md


web3.md
```diff
@@ -1797,17 +1773,28 @@ Mining
     rule <k> #clearGas => . ... </k>
          <gas> _ => 0 </gas>
 
-    rule <k> #getParentHash => . ... </k>
+    rule <k> #getParentHash => #setParentHash #getBlockByNumber( "latest", BLOCKLIST, {<network> NETWORK </network> | <block> BLOCK </block>} ) ... </k>
```
ehildenb: Get rid of `#getParentHash` (which isn't used anywhere else and is unintuitively named), and just go directly to `#setParentHash` on the rule for `#mineBlock` which uses it.

PR #643, opened 2019-12-20
symtest

daejunpark:
Add two tests: openzeppelin erc20 and bectoken.  The second test is expected to fail due to the bectoken bug.  The failed test output is not deterministic due to nondeterministic variable renaming, so the second test is put in the failing list.


daejunpark:
The second commit fixes the mistake made in https://github.com/kframework/evm-semantics/pull/632/commits/223f21ce46625298962413953e388f5eccfefd58

daejunpark:
I don't delete the branch for the future reference.

PR #648, opened 2020-01-06
HelloWorld Truffle test fix

gtrepta:
fixes: https://github.com/runtimeverification/firefly/issues/434


tests/truffle/HelloWorld/contracts/Migrations.sol
```diff
@@ -0,0 +1,23 @@
+pragma solidity >=0.4.21 <0.7.0;
```
ehildenb: In contracts/HelloWorld.sol and test/HelloWorld.sol, we require Solidity = 0.5.8. Let's be consistent, either pick 0.5.8 or set these bounds on those other files.

PR #655, opened 2020-01-09
 Factor out actual enum for `BlockchainIdentifier`

anvacaru:
Fixes: https://github.com/runtimeverification/firefly/issues/462


web3.md
```diff
@@ -74,11 +74,11 @@ The `blockList` cell stores a list of previous blocks and network states.
 
     syntax BlockchainItem ::= #getBlockByNumber ( BlockIdentifier , List , BlockchainItem ) [function]
  // --------------------------------------------------------------------------------------------------
-    rule #getBlockByNumber( _:Int      , .List                 , _     ) => .BlockchainItem
-    rule #getBlockByNumber( "earliest" , _ ListItem( BLOCK )   , _     ) => BLOCK
-    rule #getBlockByNumber( "latest"   ,   ListItem( BLOCK ) _ , _     ) => BLOCK
-    rule #getBlockByNumber( "pending"  , _                     , BLOCK ) => BLOCK
-    rule #getBlockByNumber( _:String   , .List                 , BLOCK ) => BLOCK
+    rule #getBlockByNumber( BLOCKID          , .List                 , BLOCK ) => .BlockchainItem requires BLOCKID =/=K LATEST andBool BLOCKID =/=K PENDING
+    rule #getBlockByNumber( LATEST           , .List                 , BLOCK ) => BLOCK
```
ehildenb: AFAICT, we don't need the side-condition on the first rule here, and the second rule becomes redundant.

anvacaru: Without the side-condition and the second rule, line 1739 would become something like
` #setParentHash .BlockchainItem` which would cause the configuration to crash.

ehildenb: Ah I see, I read it wrong, looks good then.

PR #666, opened 2020-01-14
Have eth_accounts only return accounts owned by client

gtrepta:
fixes: https://github.com/runtimeverification/firefly/issues/456


ehildenb:
Let's add a test of `firefly_addAccount` with private key, then `firefly_addAccount` without private key, and make sure that `eth_accounts` only returns the one added with private key.

PR #670, opened 2020-01-16
Make EDSL and lemmas.k compatible with bytes version of semantics

ehildenb:
Fixes: runtimeverification/firefly#496

Currently EDSL uses the sort `WordStack` directly in many places where it needs to be able to also accept a `Bytes` in the `Bytes` version of the semantics.

This converts it to use the `ByteArray` abstraction sort so that it's compatible with the `Bytes` version of the semantics.

This requires several changes:

-   Implement `#take` and `#drop` for the `Bytes` type in `evm-types.md`.
-   Make the rule simplifying `#sizeByteArray` a `concrete` rule globally, so all lemmas can be written in terms of it.
-   Switch all appropriate occurances of `WordStack` to `ByteArray` in `edsl.md`, `tests/specs/lemmas.k`, and `tests/specs/functional/lemmas-spec.k`.
-   Add implementation of `#padRightToWidth` for `Bytes` type.


ehildenb:
BLOCKED on #673

edsl.md
```diff
@@ -159,58 +159,64 @@ where `F1 : F2 : F3 : F4` is the (two's complement) byte-array representation of
...
+    rule #encBytes(N, BS) => #enc(#uint256(N)) ++ BS ++ #buf(#ceil32(N) -Int N, 0)
 
     //Byte array buffer. Lemmas defined in evm-data-symbolic.k
     // SIZE, DATA // left zero padding
     syntax ByteArray ::= #buf ( Int , Int ) [function, smtlib(buf)]
  // ---------------------------------------------------------------
+    rule #buf(SIZE, DATA) => #padToWidth(SIZE, #asByteStack(DATA))
+      requires #range(0 <= DATA < (2 ^Int (SIZE *Int 8))) [concrete]
+
+    syntax ByteArray ::= #bufSeg ( ByteArray , Int , Int ) [function, smtlib(bufSeg)] // BUFFER, START, WIDTH
```
yzhang90: @daejunpark I'm not sure if this should be put here or in verification.k. I put it here mainly because the haskell backend will fail to unparse terms if the definition is in verification.k

edsl.md
```diff
@@ -340,38 +346,25 @@ Specifically, `#hashedLocation` is defined as follows, capturing the storage lay
...
-    syntax ByteArray ::= intList2ByteStack( IntList ) [function]
+    syntax ByteArray ::= intList2ByteArray( IntList ) [function]
  // ------------------------------------------------------------
-    rule intList2ByteStack(.IntList) => .WordStack
-    rule intList2ByteStack(V VS)     => #padToWidth(32, #asByteStack(V)) ++ intList2ByteStack(VS)
+    rule intList2ByteArray(.IntList) => .ByteArray
+    rule intList2ByteArray(V VS)     => #buf(32, V) ++ intList2ByteArray(VS)
       requires 0 <=Int V andBool V <Int pow256
-
-    syntax IntList ::= byteStack2IntList ( WordStack )       [function]
```
yzhang90: why this part is removed in edsl.md? I don't have this part in edsl2.md because I haven't used it in the specs I have proved so far. I'm not sure if they will be used in other specs.

tests/specs/lemmas.k
```diff
@@ -147,7 +147,7 @@ module LEMMAS
     // for terms came from bytecode not via #hashedLocation
     rule keccak(WS) => keccakIntList(byteStack2IntList(WS))
```
yzhang90: byteStack2IntList is used here but it is removed from edsl.md.

@daejunpark I think we should keep `keccak(WS)` as the canonical term in the configuration instead of converting it to `keccakIntList`. is there any specific reason that we want to do it?

edsl.md
daejunpark: It is fine to put these here, but the unparsing issue should be fixed. We wouldn't want to put all the syntax declarations here.

edsl.md
daejunpark: Please put these back if possible.

edsl2.md
```diff
@@ -0,0 +1,291 @@
+eDSL High-Level Notations
```
daejunpark: Can we remove this file?

ehildenb: Done, was a mistake to add it back.

tests/specs/lemmas.k
daejunpark: It is better to keep this for now, unless we cannot put `keccakIntList` back in the edsl.md.

Your suggestion is not infeasible, but requires changing existing proofs. Let's keep this PR minimal.

edsl.md
ehildenb: I've removed it for now, since it's not needed to make this PR work.

tests/specs/lemmas.k
ehildenb: I've changed it back to how it was before.

edsl.md
ehildenb: I've put it back.

PR #676, opened 2020-01-17
Switch proofs to #buf abstraction

ehildenb:
Fixes: runtimeverification/firefly#495

These are the changes I had to make to get `#buf` working on current proofs (to avoid use of `nthbyteof`).

I'm not sure if they're the best changes, but they seem to work for many of the proofs. Still testing locally if they work for all the proofs.

I specifically aimed to remove the one rule which introduces the `nthbyteof` abstraction, then add lemmas about `#buf` to make the proof `tests/specs/erc20/hobby/allowance-spec.k` go through.


ehildenb:
@denis-bogdanas I'm working on getting these specs to the point where they even run on Haskell backend at all, which requires removing `WordStack` sort in many places, which requires removing `nthbyteof`.

I'm not really focused on updating VSC specs to haskell backend, but it sounds nice to do. I'm more worried about providing some simple examples which can be used for haskell backend (currently the goal is just to make the specs even run on haskell backend, not to pass).

I think it makes more sense to update KEVM specs before VSC specs, since there are fewer of them and they're smaller. If there is some breaking change to VSC that happens because of changes to KEVM, it should be added as a regression test to KEVM. That's how all our other repos work too. We work changes from dependencies to dependents, and regressions get passed from dependents to dependencies.

denis-bogdanas:
Let's discuss on Slack about what specs to keep in KEVM.

tests/specs/lemmas.k
```diff
@@ -74,17 +74,58 @@ module LEMMAS
...
+    rule #drop(N, BUF1 ++ BUF2)  => #drop(N, BUF1) ++ BUF2                   requires N <Int #sizeWordStack(BUF1)
 
+    rule #asWord(#buf(N, BUF)) => BUF
 
+    rule #asWord(W0 : W1 : W2 : W3 : #take(28, BUF)) /Int 26959946667150639794667015087019630673637144422540572481103610249216 => #asWord(W0 : W1 : W2 : W3 : .WordStack)
 
+    rule M [ N := BUF ] => M requires #sizeWordStack(BUF) <=Int 0
 
+    rule M [ N := BUF:WordStack ] [ L <- V ]    => M [ L <- V ] [ N := BUF ]                  requires L  <Int N
+    rule M [ N := BUF           ] [ L <- V ]    => M [ L <- V ] [ N := BUF ]                  requires L >=Int N +Int #sizeWordStack(BUF)
```
daejunpark: you can consider to merge this two rules into one by OR-ring the side condition.

tests/specs/lemmas.k
```diff
@@ -74,17 +74,58 @@ module LEMMAS
...
+    rule M [ N := BUF           ] [ N <- V ]    => M [ N <- V ] [ N +Int 1 := #drop(1, BUF) ]
```
daejunpark: Does this really happen?

ehildenb: The scenario is when we get a term like this:

```
M [ 32 := #buf(32, OWNER) ] [ 32 <- 0 ] [ 33 <- 0 ] ... [ 63 <- 0 ]
```

So this, with the rules:

```
rule #drop(N, #drop(M, BUF)) => #drop(N +Int M, BUF)
rule M [ N := BUF ] => M requires #sizeWordStack(BUF) <=Int 0
```

Simplify this term down to:

```
M [ 32 <- 0 ] [ 33 <- 0 ] ... [ 63 <- 0 ]
```

tests/specs/lemmas.k
ehildenb: I'll try it and push it up if it works.

tests/specs/lemmas.k
```diff
@@ -74,17 +74,58 @@ module LEMMAS
...
+    rule #buf(W, BUF) [ L .. K ] => .ByteArray requires K <=Int 0
+    rule #buf(W, BUF) [ 0 .. W ] => #buf(W, BUF)
 
+    rule #range(M, N, K) => .ByteArray requires notBool K >Int 0
 
+    rule #range(M [ N := #padToWidth(W, BUF) ], L, K) => #range(M, L, K)
+      requires L +Int K <=Int N
+        orBool N +Int W  <Int L
 
+    rule #range(M [ N := #padToWidth(W, BUF) ], N, W) => BUF
```
daejunpark: #padToWidth shouldn't appear at first place, as it should have been reduced to #buf by applying the following rule:
```
    rule #padToWidth(32, #asByteStack(V)) => #buf(32, V)
      requires 0 <=Int V andBool V <Int pow256 andBool #notKLabel(V, "#asWord")
```
I guess this rule is not applied because it fails to prove `V <Int pow256`. And, you may need to add the following:
```
    syntax Bool ::= isStorage(Map) [function]
    rule 0 <=Int select(M, _)             => true requires isStorage(M)
    rule         select(M, _) <Int pow256 => true requires isStorage(M)
```
and add `isStorage` predicate for the storage map.

tests/specs/lemmas.k
```diff
@@ -74,17 +74,58 @@ module LEMMAS
...
+    rule M [ N := BUF ] => M requires #sizeWordStack(BUF) <=Int 0
 
+    rule M [ N := BUF:WordStack ] [ L <- V ]    => M [ L <- V ] [ N := BUF ]                  requires L  <Int N
+    rule M [ N := BUF           ] [ L <- V ]    => M [ L <- V ] [ N := BUF ]                  requires L >=Int N +Int #sizeWordStack(BUF)
+    rule M [ N := BUF           ] [ N <- V ]    => M [ N <- V ] [ N +Int 1 := #drop(1, BUF) ]
+    rule M [ N := BUF           ] [ N := BUF' ] => M [ N := BUF' ]                            requires #sizeWordStack(BUF) ==Int #sizeWordStack(BUF')
 
+    rule #buf(W, BUF) [ L .. K ] => .ByteArray requires L  >Int W
+    rule #buf(W, BUF) [ L .. K ] => .ByteArray requires K <=Int 0
+    rule #buf(W, BUF) [ 0 .. W ] => #buf(W, BUF)
```
daejunpark: For java backend, this rule is applied only when the term matched by the two `W`s are structurally equal. For the haskell backend, it may be different, but I don't know.

tests/specs/lemmas.k
```diff
@@ -74,17 +74,58 @@ module LEMMAS
...
+       andBool L <Int N
+
+    rule #range(M [ N := BUF ], L, K) => BUF [ L -Int N .. minInt(K, #sizeWordStack(BUF) -Int (L -Int N)) ] ++ #range(M, N +Int #sizeWordStack(BUF), K -Int minInt(K, #sizeWordStack(BUF) -Int (L -Int N)))
+      requires K  >Int 0
+       andBool L >=Int N
+       andBool L  <Int N +Int #sizeWordStack(BUF)
+
+    rule #range(M [ N := BUF ], L, K) => #range(M, L, K)
+      requires K  >Int 0
+       andBool L >=Int N +Int #sizeWordStack(BUF)
```
daejunpark: ok this seems to make sense, but it is quite tricky especially because of using minInt, so i cannot trust my eyeball-checking at this point.  how about writing test specs for all possible cases, especially including the corner cases as well?

tests/specs/lemmas.k
daejunpark: hmm, that scenario could be problematic.

i think what happened is that initially we have `... [ 32 := #buf(32, 0) ]`, but it is somehow evaluated to `... [ 32 <- 0 ] [ 33 <- 0 ] ... [ 63 <- 0 ]`. 

it is not incorrect, but in the end, we have `M [ 32 <- 0 ] [ 33 <- 0 ] ... [ 63 <- 0 ]` instead of `M [ 32 := #buf(32, 0) ]`. 

ideally, we want to have `M [ _ := _ ]` form only to allow word-level read/write to local mem, unless we have MSTORE8 opcode in the bytecode.

ehildenb: Why is this problematic?

```
M [ 32 := #buf(32, OWNER) ] [ 32 <- 0 ] => M [ 32 <- 0 ] [ 33 := #drop(1, #buf(32, OWNER)) ]
```

is true.

tests/specs/lemmas.k
ehildenb: I'm fine with this.

tests/specs/lemmas.k
ehildenb: Combining the lemmas causes two specs to fail: `tests/specs/benchmarks/requires01-a0le0-spec.k` and `tests/specs/benchmarks/ecrecoverloop02-sig1-invalid-spec.k`.

tests/specs/lemmas.k
ehildenb: The thing is, pushing a `[ 32 <- 0 ]` often allows the update operation to simplify out fully, the the overall resulting term becomes much simpler. So it's easier to debug the proofs and make the final states line up.

daejunpark: I'm not saying that this lemma is problematic. I'm saying that you shouldn't met such term `_ [ _ <- _ ]` at the first place.  `_ [ _ := _ ]` is ok, but `_ [ _ <- _ ]` is not.  We'd like to keep `_ [ _ := _ ]` terms, not allowing them to be further evaluated to nested `_ [ _ <- _ ]` terms.  That's the whole point of having `#buf` abstraction, instead of `nthbyteof`.

daejunpark: OK. I found what happend. You need to add `EVM-TYPES.mapWriteBytes.recursive` in the concrete-rules.txt file, so that the following rule is not applied for the term `M [ 32 := #buf(32, OWNER) ] [ 32 <- 0 ] [ 33 <- 0 ] ... [ 63 <- 0 ]`:
```
    rule [mapWriteBytes.recursive]: WM[ N := W : WS     ] => (WM[N <- W])[N +Int 1 := WS]
```
And then, you need to add another set of lemmas as I said above.

tests/specs/erc20/concrete-rules.txt
```diff
@@ -1,27 +1,28 @@
...
+EVM-TYPES.powmod.nonzero
+EVM-TYPES.powmod.zero
+EVM-TYPES.#range
+EVM-TYPES.#signed.negative
+EVM-TYPES.#signed.positive
+EVM-TYPES.signextend.invalid
+EVM-TYPES.signextend.negative
+EVM-TYPES.signextend.positive
+EVM-TYPES.#unsigned.negative
+EVM-TYPES.#unsigned.positive
```
daejunpark: Is this just a reordering change?

daejunpark: OK. You reordered and also added `EVM-TYPES.#range`, which is good.

Indeed, note that we have the following three additional items in the concrete-rule.txt file, in the latest proofs:
```
EVM-TYPES.mapWriteBytes.recursive
EVM-TYPES.#lookup.some
EVM-TYPES.#lookup.none
```

I already suggested to add `EVM-TYPES.mapWriteBytes.recursive` in another comment.

The remaining two items for lookup may not be needed. So you can just ignore them for now.

tests/specs/lemmas.k
ehildenb: Ah great. I already have the lemmas which handles `M [ N := BUF1 ] [ N := BUF2 ]`, I'll make the correct rule concrete. Thanks!

tests/specs/erc20/concrete-rules.txt
ehildenb: I'll add the recursive `mapWriteBytes` lemma, and see if the other two are needed. My guess is not for now, but maybe when you start adding more regression tests for downstream proofs.

ehildenb: And yes, it was just a reordering.

tests/specs/lemmas.k
ehildenb: Confirmed that I can remove all the lemmas about `_ [ _ <- _ ]` by marking the appropriate rule concrete. The only remaining rule is:

```
rule M [ N := BUF1 ] [ N := BUF2 ] => M [ N := BUF2 ] requires #sizeWordStack(BUF1) ==Int #sizeWordStack(BUF2)
```

This is enough for now, so I don't think we need to generalize to more cases (where the wordstacks are different sizes for instance) yet.

tests/specs/lemmas.k
ehildenb: I've just removed the lemmas and it seems to work OK.

tests/specs/lemmas.k
ehildenb: I've added tests in `tests/specs/functional/lemmas-spec.k`. Have a look.

tests/specs/functional/lemmas-spec.k
```diff
@@ -0,0 +1,50 @@
...
+ // #range selection operation
+ // --------------------------
+
+    // Non-positive range
+    rule <k> runLemma ( #range(M, N,  0) ) => doneLemma ( .WordStack ) ... </k>
+    rule <k> runLemma ( #range(M, N, -3) ) => doneLemma ( .WordStack ) ... </k>
+
+    // Range outside of previous buffer write
+    rule <k> runLemma ( #range(M [ 32 := BUF:WordStack ], 20, 4) ) => doneLemma ( #range(M, 20, 4) ) ... </k>
+    rule <k> runLemma ( #range(M [ 32 := BUF:WordStack ], 48, 4) ) => doneLemma ( #range(M, 48, 4) ) ... </k> requires #sizeWordStack(BUF) <=Int 12
```
daejunpark: Please add the following corner cases (ranges of size 1, right before and right after the previous buffer):
```
rule <k> runLemma ( #range(M [ 32 := BUF:WordStack ], 31, 1) ) => doneLemma ( #range(M, 31, 1) ) ... </k>
rule <k> runLemma ( #range(M [ 32 := #buf(1, DATA) ], 33, 1) ) => doneLemma ( #range(M, 33, 1) ) ... </k>
```

tests/specs/functional/lemmas-spec.k
```diff
@@ -0,0 +1,50 @@
...
+    rule <k> runLemma ( #range(M [ 32 := BUF:WordStack ]            , 30 , 8) ) => doneLemma ( #range(M, 30, 2) ++ (BUF [ 0 .. 6 ])                             ) ... </k> requires #sizeWordStack(BUF) >=Int 6
+    rule <k> runLemma ( #range(M [ 32 := (0 : 1 : 2 : .WordStack) ] , 30 , 8) ) => doneLemma ( #range(M, 30, 2) ++ (0 : 1 : 2 : .WordStack) ++ #range(M, 35, 3) ) ... </k>
+
+    // Range starting at same point as buffer write
+    rule <k> runLemma ( #range(M [ 32 := BUF:WordStack ]            , 32 , 7) ) => doneLemma( BUF [ 0 .. 7 ]                               ) ... </k> requires #sizeWordStack(BUF) >=Int 7
+    rule <k> runLemma ( #range(M [ 32 := (0 : 1 : 2 : .WordStack) ] , 32 , 7) ) => doneLemma( (0 : 1 : 2 : .WordStack) ++ #range(M, 35, 4) ) ... </k>
+
+    // Range starting in the middle of the buffer write
+    rule <k> runLemma ( #range(M [ 32 := (0 : 1 : 2 : 3 : 4 : .WordStack) ] , 35 , 2) ) => doneLemma( (3 : 4 : .WordStack)                     ) ... </k>
+    rule <k> runLemma ( #range(M [ 32 := (0 : 1 : 2 : 3 : 4 : .WordStack) ] , 35 , 8) ) => doneLemma( (3 : 4 : .WordStack) ++ #range(M, 37, 6) ) ... </k>
```
daejunpark: Please add the following corner cases:
```
rule <k> runLemma ( #range(M [ 32 := (0 : 1 : 2 : .WordStack) ] , 33 , 1) ) => doneLemma( (1 : .WordStack)                     ) ... </k>
rule <k> runLemma ( #range(M [ 32 := (0 : 1 : 2 : .WordStack) ] , 33 , 2) ) => doneLemma( (1 : 2 : .WordStack)                     ) ... </k>
rule <k> runLemma ( #range(M [ 32 := (0 : 1 : 2 : .WordStack) ] , 33 , 3) ) => doneLemma( (1 : 2 :  .WordStack) ++ #range(M, 35, 1)                     ) ... </k>
```

tests/specs/lemmas.k
```diff
@@ -74,17 +74,47 @@ module LEMMAS
...
+    rule #drop(N, BUF)           => .WordStack                               requires #sizeWordStack(BUF) <=Int N
+    rule #drop(N, #drop(M, BUF)) => #drop(N +Int M, BUF)
+    rule #drop(N, BUF1 ++ BUF2)  => #drop(N -Int #sizeWordStack(BUF1), BUF2) requires N >=Int #sizeWordStack(BUF1)
+    rule #drop(N, BUF1 ++ BUF2)  => #drop(N, BUF1) ++ BUF2                   requires N <Int #sizeWordStack(BUF1)
 
+    rule #asWord(#buf(N, BUF)) => BUF
 
+    rule #asWord(W0 : W1 : W2 : W3 : #take(28, BUF)) /Int 26959946667150639794667015087019630673637144422540572481103610249216 => #asWord(W0 : W1 : W2 : W3 : .WordStack)
 
+    rule M [ N := BUF ] [ N := BUF' ] => M [ N := BUF' ] requires #sizeWordStack(BUF) ==Int #sizeWordStack(BUF')
```
daejunpark: You will need additional lemmas to simplify, e.g., the following term:
```
M [ N1 := BUF1 ] [ N2 := BUF2 ] [ N1 := BUF3 ]
```
into
```
M [ N1 := BUF3 ] [ N2 := BUF2 ]
```
where N1 and N2 are apart enough.  Without such lemmas, the term will grow too big to debug.

But as you said in another comment, we can add such lemmas later.

ehildenb: Yeah it's not needed for now, so I won't add it until it is needed.

tests/specs/functional/lemmas-spec.k
```diff
@@ -0,0 +1,50 @@
...
+    rule <k> runLemma ( #range(M [ 32 := BUF:WordStack ], 20, 4) ) => doneLemma ( #range(M, 20, 4) ) ... </k>
+    rule <k> runLemma ( #range(M [ 32 := BUF:WordStack ], 48, 4) ) => doneLemma ( #range(M, 48, 4) ) ... </k> requires #sizeWordStack(BUF) <=Int 12
+
+    // Range starting before buffer write
+    rule <k> runLemma ( #range(M [ 32 := BUF:WordStack ]            , 30 , 8) ) => doneLemma ( #range(M, 30, 2) ++ (BUF [ 0 .. 6 ])                             ) ... </k> requires #sizeWordStack(BUF) >=Int 6
+    rule <k> runLemma ( #range(M [ 32 := (0 : 1 : 2 : .WordStack) ] , 30 , 8) ) => doneLemma ( #range(M, 30, 2) ++ (0 : 1 : 2 : .WordStack) ++ #range(M, 35, 3) ) ... </k>
+
+    // Range starting at same point as buffer write
+    rule <k> runLemma ( #range(M [ 32 := BUF:WordStack ]            , 32 , 7) ) => doneLemma( BUF [ 0 .. 7 ]                               ) ... </k> requires #sizeWordStack(BUF) >=Int 7
+    rule <k> runLemma ( #range(M [ 32 := (0 : 1 : 2 : .WordStack) ] , 32 , 7) ) => doneLemma( (0 : 1 : 2 : .WordStack) ++ #range(M, 35, 4) ) ... </k>
```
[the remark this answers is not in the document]
ehildenb: Added!

tests/specs/functional/lemmas-spec.k
ehildenb: Added!

tests/specs/functional/lemmas-spec.k
ehildenb: Added!

tests/specs/functional/lemmas-spec.k
```diff
@@ -0,0 +1,50 @@
...
+    rule <k> runLemma ( #range(M, N,  0) ) => doneLemma ( .WordStack ) ... </k>
+    rule <k> runLemma ( #range(M, N, -3) ) => doneLemma ( .WordStack ) ... </k>
+
+    // Range outside of previous buffer write
+    rule <k> runLemma ( #range(M [ 32 := BUF:WordStack ], 20, 4) ) => doneLemma ( #range(M, 20, 4) ) ... </k>
+    rule <k> runLemma ( #range(M [ 32 := BUF:WordStack ], 48, 4) ) => doneLemma ( #range(M, 48, 4) ) ... </k> requires #sizeWordStack(BUF) <=Int 12
+
+    // Range starting before buffer write
+    rule <k> runLemma ( #range(M [ 32 := BUF:WordStack ]            , 30 , 8) ) => doneLemma ( #range(M, 30, 2) ++ (BUF [ 0 .. 6 ])                             ) ... </k> requires #sizeWordStack(BUF) >=Int 6
+    rule <k> runLemma ( #range(M [ 32 := (0 : 1 : 2 : .WordStack) ] , 30 , 8) ) => doneLemma ( #range(M, 30, 2) ++ (0 : 1 : 2 : .WordStack) ++ #range(M, 35, 3) ) ... </k>
```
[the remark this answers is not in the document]
ehildenb: Added!

PR #687, opened 2020-01-22
Execute transaction sent by eth_sendRawTransacion

anvacaru:
Fixes: https://github.com/runtimeverification/firefly/issues/502


ehildenb:
We should add a test which this change fixes, and confirm that it is broken before this fix.

PR #690, opened 2020-01-22
Remove "from" requirement from eth_call and eth_estimateGas

gtrepta:
fixes: https://github.com/runtimeverification/firefly/issues/503


web3.md
```diff
@@ -1361,17 +1370,26 @@ Transaction Execution
...
+          => #pushNetworkState
+          ~> #loadTx {keys_list(ACCTS)[0]}:>Int J
+          ~> #eth_estimateGas_finalize GUSED
+         ...
+         </k>
          <params> [ ({ _ } #as J), TAG, .JSONs ] </params>
-      requires notBool isString( #getJSON("from", J) )
+         <gasUsed>  GUSED  </gasUsed>
+         <accountKeys> ACCTS </accountKeys>
+      requires notBool isString(#getJSON("from", J) )
```
ehildenb: Could we also just have a rule which does something like this:

```
rule <k> #eth_estimateGas ... </k>
      <params> [ { _ , (.JSONs => "from": {keys_list(ACCTS)[0]}:>Int , .JSONs) } #as J ], TAG, .JSONs </params>
  requires notBool isString(#getJSON("from", J))
```

that way we only have one rule describing the semantics of `#eth_estimateGas` instead of duplicating the logic?

web3.md
```diff
@@ -1300,17 +1299,27 @@ Transaction Execution
...
+          ~> #setMode NOGAS
+          ~> #loadTx {keys_list(ACCTS)[0]}:>Int J
+          ~> #eth_call_finalize EXECMODE
+         ...
+         </k>
          <params> [ ({ _ } #as J), TAG, .JSONs ] </params>
-      requires notBool isString( #getJSON("from", J) )
+         <mode> EXECMODE </mode>
+         <accountKeys> ACCTS </accountKeys>
+      requires notBool isString( #getJSON("from" , J) )
```
ehildenb: Same here.

PR #694, opened 2020-01-23
Make lemma-specs valid on both Java/Haskell backends

ehildenb:


tests/specs/lemmas.k
```diff
@@ -88,118 +94,105 @@ module LEMMAS
...
-    rule keccakIntList(V1:Int V2:Int .IntList) => hash2(V1, V2)
+    rule keccakIntList(V:Int .IntList) => hash1(V)               [simplification]
+    rule keccakIntList(V1:Int V2:Int .IntList) => hash2(V1, V2)  [simplification]
 
     // for terms came from bytecode not via #hashedLocation
     rule keccak(WS) => keccakIntList(byteStack2IntList(WS))
-      requires ( notBool #isConcrete(WS) )
-       andBool ( #sizeByteArray(WS) ==Int 32 orBool #sizeByteArray(WS) ==Int 64 )
+      requires ( #sizeByteArray(WS) ==Int 32 orBool #sizeByteArray(WS) ==Int 64 )
+      [simplification]
```
daejunpark: this change will be problematic.  we want to concretely evaluate keccak(WS) when WS is concrete.  so, i think we will need to use a rule priority to ensure that this lemma is applied only when WS is symbolic.

tests/specs/lemmas.k
```diff
@@ -88,118 +94,105 @@ module LEMMAS
...
-    rule I1 +Int (B +Int I3) => B +Int (I1 +Int I3) when #isConcrete(I1) andBool notBool #isConcrete(B) andBool #isConcrete(I3)
-    rule I1 -Int (B +Int I3) => (I1 -Int I3) -Int B when #isConcrete(I1) andBool notBool #isConcrete(B) andBool #isConcrete(I3)
-    rule (I1 -Int B) +Int I3 => (I1 +Int I3) -Int B when #isConcrete(I1) andBool notBool #isConcrete(B) andBool #isConcrete(I3)
-
-    rule I1 +Int (I2 +Int C) => (I1 +Int I2) +Int C when #isConcrete(I1) andBool #isConcrete(I2) andBool notBool #isConcrete(C)
-    rule I1 +Int (I2 -Int C) => (I1 +Int I2) -Int C when #isConcrete(I1) andBool #isConcrete(I2) andBool notBool #isConcrete(C)
-    rule I1 -Int (I2 +Int C) => (I1 -Int I2) -Int C when #isConcrete(I1) andBool #isConcrete(I2) andBool notBool #isConcrete(C)
-    rule I1 -Int (I2 -Int C) => (I1 -Int I2) +Int C when #isConcrete(I1) andBool #isConcrete(I2) andBool notBool #isConcrete(C)
-
-    rule I1 &Int (I2 &Int C) => (I1 &Int I2) &Int C when #isConcrete(I1) andBool #isConcrete(I2) andBool notBool #isConcrete(C)
```
daejunpark: you just dropped this set of lemmas, right?  i think it is fine for now, and revisit later if you need.

tests/specs/lemmas.k
```diff
@@ -88,118 +94,105 @@ module LEMMAS
...
     rule N &Int MASK => N  requires MASK ==Int (2 ^Int (log2Int(MASK) +Int 1)) -Int 1 // MASK = 0xffff...f
                             andBool 0 <=Int N andBool N <=Int MASK
+      [simplification]
 
     // for gas calculation
-    rule A -Int (#if C #then B1 #else B2 #fi) => #if C #then (A -Int B1) #else (A -Int B2) #fi
+    rule A -Int (#if C #then B1 #else B2 #fi) => #if C #then (A -Int B1) #else (A -Int B2) #fi  [simplification]
 
-    rule (#if C #then B1 #else B2 #fi) -Int A => #if C #then (B1 -Int A) #else (B2 -Int A) #fi
-        when notBool #isConcrete(A) andBool #notKLabel(A, "#if_#then_#else_#fi_K-EQUAL")
```
daejunpark: github diff doesn't work smartly here.  so, it seems that this lemma is dropped, and all the changes below are simply adding the attribute, right?  it is fine to drop this lemma for now.

tests/specs/lemmas.k
virgil-serbanuta: Yes. `#isConcrete` will probably never work on the Haskell backend.

tests/specs/lemmas.k
virgil-serbanuta: Yes, I dropped it. Yes, the diff is not helpful, it's easier to follow the git commits.

I'm also adding 4 new lemmas below, and I'm deleting `#notKLabel`'s definition. The latter causes other tests to fail in the continuous build.

ehildenb:
@virgil-serbanuta I'm closing this because I want to review the changes here, but I was the one who opened the PR. Please re-open a PR against master with the new set of changes you need now that #721 is merged.

PR #714, opened 2020-02-03
Accept "amount" fields in "eth_call" methods

anvacaru:
Fixes: https://github.com/runtimeverification/firefly/issues/570


ehildenb:
Let's look through the ganache codebase to confirm that it interprets amount as value.

PR #722, opened 2020-02-05
Useless rules removal

ehildenb:
Fixes runtimeverification/firefly#551

This fixes two warnings the llvm backend produces about useless rules


dwightguth:
this looks okay with me but I'd rather someone else approve since I don't know this code very well. In particular, I'm not familiar with the BlockIdentifier sort.

PR #724, opened 2020-02-06
Hkg with the Haskell backend 

virgil-serbanuta:


ehildenb:
Bits and pieces merged in other PRs, one working HKG proof on haskell backend.

PR #727, opened 2020-02-06
Better rebuildability and allowing external builds of kevm-vm and kevm-client

ehildenb:


ehildenb:
Because runtimeverification/firefly#585 is passing, this is ready for review.

PR #728, opened 2020-02-06
Remove coverage collection from web3 client

ehildenb:


ehildenb:
Because runtimeverification/firefly#585 is passing, this is ready for review.

PR #731, opened 2020-02-07
Script to extract function signatures from compiled data

anvacaru:
Fixes: https://github.com/runtimeverification/firefly/issues/563


ehildenb:
This script should live over in runtimeverification/firefly.

ehildenb:
This will be moved over to runtimeverification/firefly.

PR #735, opened 2020-02-10
improve gas calculation for verification

### LINKED ISSUE #726: HKG with Haskell - rewrite lemma with function in LHS
The Haskell backend cannot unify functions, so it cannot find a value for the `MEM'` variable, so it crashes. The variable is wrapped in the `#symCmem` function.

https://github.com/kframework/evm-semantics/blob/master/tests/specs/erc20/abstract-semantics-segmented-gas.k#L26

yzhang90:
fix #726

### DIFF
--- a/evm.md
+++ b/evm.md
@@ -64,7 +64,7 @@ In the comments next to each cell, we've marked which component of the YellowPap
               <wordStack>   .WordStack </wordStack>           // \mu_s
               <localMem>    .Memory    </localMem>            // \mu_m
               <pc>          0          </pc>                  // \mu_pc
-              <gas>         0          </gas>                 // \mu_g
+              <gas>    #gas(0, 0, 0)   </gas>                 // \mu_g
               <memoryUsed>  0          </memoryUsed>          // \mu_i
               <callGas>     0          </callGas>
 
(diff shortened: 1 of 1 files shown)


evm.md
```diff
@@ -1843,17 +1843,24 @@ Ethereum Gas Calculation
...
+-   `#gas(AvailGas, ExecGas, MemoryGas)` represents the remaining gas, which is calculated by `AvailGas - ExecGas - MemoryGas`.
 -   `#gas` calculates how much gas this operation costs, and takes into account the memory consumed.
 -   `#deductGas` is used to check that there won't be a gas underflow (throwing `EVMC_OUT_OF_GAS` if so), and deducts the gas if not.
 -   `#deductMemory` checks that access to memory stay within sensible bounds (and deducts the correct amount of gas for it), throwing `EVMC_INVALID_MEMORY_ACCESS` if bad access happens.
+-   `#resolveGas` updates the remaining gas by substracting `ExecGas` and `MemoryGas` from `AvailGas`.
 
 ```k
+
+    syntax Gas      ::= #gas ( AvailGas , Int , Int )
+    syntax AvailGas ::= "InfGas" | Int  // Providing InfGas prevents programs from EVMC_OUT_OF_GAS
```
ehildenb: Documentation should be in the markdown, not inline.

evm.md
```diff
@@ -1863,15 +1870,27 @@ Overall Gas
...
+    rule <k> #resolveGas => . ... </k>
+         <gas> #gas(G:Int , EXECGAS, MEMGAS) => #gas(G -Int EXECGAS -Int MEMGAS, 0, 0) </gas>
+       requires G >=Int (EXECGAS +Int MEMGAS)
+
+    rule <k> #resolveGas => #end EVMC_OUT_OF_GAS ... </k>
+         <gas> #gas(G:Int , EXECGAS, MEMGAS) </gas>
+       requires notBool (G >=Int (EXECGAS +Int MEMGAS))
+
+    rule <k> #resolveGas => . ... </k>  // For Verification
+         <gas> #gas(InfGas, EXECGAS, MEMGAS) </gas>
```
ehildenb: No need to say that this is for Verification, it will also be for other things (eg. when we want to compute concrete gas usage).

PR #738, opened 2020-02-18
evm-types.md: simplify sizeWordStack

MrChico:
There's plenty of instances employing a pattern like this, where an `*aux` version of a production is used with an extra accumulating argument. These constructions seem to introduce unnecessary complexity and complicate symbolic execution to the point where [additional rules](https://github.com/kframework/evm-semantics/blob/master/tests/specs/lemmas.k#L13) need to be introduced, which would seem redundant if the rules were just written in a straight forward manner in the first place. 

Is there any reasoning to keeping rules on this form that I'm unaware about? If there is no good reason, there are several [function] productions that can be simplified

### DIFF
--- a/evm-types.md
+++ b/evm-types.md
@@ -434,11 +434,9 @@ A cons-list is used for the EVM wordstack.
 
 ```k
     syntax Int ::= #sizeWordStack ( WordStack )       [function, functional, smtlib(sizeWordStack)]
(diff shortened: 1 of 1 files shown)


asymmetric:
Seems like the build [timed-out](https://office.runtimeverification.com/jenkins/blue/organizations/jenkins/evm-semantics/detail/PR-738/1/pipeline/#log-660).

ehildenb:
We have the semantics presented like this for efficiency reasons. It's inconvenient for the prover, but the functions need to be tail recursive for the LLVM backend memory usage.

PR #739, opened 2020-02-18
Hkg lemmas for the Haskell backend

virgil-serbanuta:


virgil-serbanuta:
Sorry, I forgot that this should have been opened on the main repository and not a fork, I'll close it.

PR #743, opened 2020-02-26
prove hkg-totalSupply using haskell backend

yzhang90:


yzhang90:
@denis-bogdanas  the tests passed

PR #747, opened 2020-03-09
Json Specification conformant with ECMA-404

ehildenb:
Fixes: runtimeverification/firefly#630

This makes it so that the JSON specification we use is actually the ECMA-404 standard (or much closer, modulo the regexes used for parsing integers and strings).


ehildenb:
the JSON spec is being migrated to upstream K, we can work on this there.

PR #754, opened 2020-03-17
 Unit test for failing Arbitrum `eth_call`

anvacaru:
Fixes: https://github.com/runtimeverification/firefly/issues/664


gtrepta:
Reopening because the test passes now.

PR #765, opened 2020-04-01
New slow lists

ehildenb:
Slow/failing lists of VM tests are updated.

The new slow lists are determined by any program which takes >20s to run. I may adjust it more if that proves to take too long still.


tests/failing.haskell
```diff
@@ -1,2 +1,6 @@
-tests/ethereum-tests/VMTests/vmSystemOperations/suicide0.json
-tests/ethereum-tests/VMTests/vmSystemOperations/suicideSendEtherToMe.json
+tests/ethereum-tests/VMTests/vmLogTest/log0_emptyMem.json
+tests/ethereum-tests/VMTests/vmLogTest/log0_logMemsizeZero.json
+tests/ethereum-tests/VMTests/vmLogTest/log0_nonEmptyMem.json
+tests/ethereum-tests/VMTests/vmLogTest/log0_nonEmptyMem_logMemSize1.json
+tests/ethereum-tests/VMTests/vmLogTest/log0_nonEmptyMem_logMemSize1_logMemStart31.json
+tests/ethereum-tests/VMTests/vmLogTest/log_2logs.json
```
ttuegel, line 6: Are these tests new, or are these regressions?

ehildenb: They were removed from the `slow` list, because they finished quick enough to fail in some other way.

PR #771, opened 2020-04-09
Update <blockhashes> in web3

gtrepta:
fixes: https://github.com/runtimeverification/firefly/issues/744

Currently, the `<blockhashes>` cell is not being updated anywhere in the web3 semantics, which creates an issue with at least the BLOCKHASH instruction.

- Keeps the `<blockhashes>` cell in sync with the `<blockList>` cell
- Refactors upkeep of the `<number>` cell to prevent an off by one error with the BLOCKHASH instruction
- Moves transaction receipt generation into the block mining semantics (tx receipts aren't supposed to be available until they are mined). This was necessary for this PR as tx receipts were broken by the above changes.


ehildenb:
If this fixes the unit test at #754 , then add that in this PR too so I can see it's passing.

web3.md
```diff
@@ -69,6 +69,7 @@ The `blockList` cell stores a list of previous blocks and network states.
     rule <k> #pushBlockchainState => . ... </k>
          <blockList> (.List => ListItem({ <network> NETWORK </network> | <block> BLOCK </block> })) ... </blockList>
          <network> NETWORK </network>
+         <blockhashes> (.List => ListItem(#blockchainItemHash({ <network> NETWORK </network> | <block> BLOCK </block> }))) _ </blockhashes>
```
ehildenb, line 72:
```suggestion
         <blockhashes> (.List => ListItem(#blockchainItemHash({ <network> NETWORK </network> | <block> BLOCK </block> }))) ... </blockhashes>
```

ehildenb:
Does this end up fixing the LeapDAO tests?

ehildenb:
Basically, anywhere where the current `<block>` changes or the `<blockList>` changes, the `<blockHashes>` also needs to change. These three cells need to be kept in sync.

web3.md
```diff
@@ -69,14 +69,15 @@ The `blockList` cell stores a list of previous blocks and network states.
...
          <block>   BLOCK   </block>
 
     syntax BlockchainItem ::= #getBlockByNumber ( BlockIdentifier , List , BlockchainItem ) [function]
  // --------------------------------------------------------------------------------------------------
     rule #getBlockByNumber( _:Int            , .List                 , _     ) => .BlockchainItem
     rule #getBlockByNumber( LATEST           , .List                 , BLOCK ) => BLOCK
     rule #getBlockByNumber( LATEST           ,   ListItem( BLOCK ) _ , _     ) => BLOCK
-    rule #getBlockByNumber( PENDING          , _                     , BLOCK ) => BLOCK
+    rule #getBlockByNumber( PENDING => LATEST, _                     , _     )
     rule #getBlockByNumber( EARLIEST         , .List                 , BLOCK ) => BLOCK
```
ehildenb, line 81: I don't think we should be treating `PENDING` and `LATEST` as the same cases, it seems like these should be treated differently because otherwise why would there be two separate identifiers for it in other clients?

gtrepta: I'm just mimicking ganache here. Ganache literally does a check on the blocknumber like this:
```js
if (number === "latest" || number === "pending") {
    return this.getHeight(callback);
}
```
If we want to change this to make more sense and actually follow the web3 rpc spec, we'll also have to change some of our tests since they expect ganache's behavior.

ehildenb: Can you link the code where Ganache uses this?

gtrepta: https://github.com/trufflesuite/ganache-core/blob/25de0e2348c0e49dd6a30f69e3a6889a2cca8048/lib/blockchain_double.js#L228

PR #777, opened 2020-04-25
Make `#range` and `_[_.._]` total on Bytes builds

ehildenb:
Currently both `#range` and `_[_.._]` are marked as partial, because there were some cases that were undefined.

I've added a case to `_[_.._]` to make it total, so that we don't produce definedness conditions for this symbol.

### DIFF
--- a/evm-types.md
+++ b/evm-types.md
@@ -590,9 +590,12 @@ The local memory of execution is a byte-array (instead of a word-array).
 
     syntax ByteArray ::= ByteArray "[" Int ".." Int "]" [function, functional]
  // --------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


denis-bogdanas:
It all depends on how `ByteArray` is used throughout the semantics. Technically these operations are not defined when START or WIDTH is <0. But this adds performance penalty at runtime because of `#Ceil` rules. So I guess we have to make such compromises.

ehildenb:
@denis-bogdanas they are defined for `START <Int 0` and `WIDTH <Int 0` because of the `owise` case. So they should be total. If you can find an example where they aren't total, then this should definitely be reverted. We should not be stating false things to the prover.

PR #778, opened 2020-04-27
loadTransaction: Ignore fields that aren't part of the web3 spec

gtrepta:
fixes: https://github.com/runtimeverification/firefly/issues/801

This PR simply adds a rule that drops any fields it doesn't know what to do with when loading in a JSON with transaction parameters.


ehildenb:
I'm investigating if there has been a performance regression.

PR #783, opened 2020-04-30
Improve timestamp updates

gtrepta:
fixes: https://github.com/runtimeverification/firefly/issues/822

This PR introduces a `<lastMineTime>` cell, which helps to track the passage of time from the last mining of a block. This is necessary because `<timestamp>` can be manipulated in ways such as custom initialization, `evm_revert` and `evm_increaseTime`, which makes just updating the cell with `#time()` incorrect.


web3.md
```diff
@@ -22,6 +22,7 @@ module WEB3
         </blockchain>
         <defaultGasPrice> 20000000000 </defaultGasPrice>
         <defaultGasLimit> 90000       </defaultGasLimit>
+        <lastMineTime>    0           </lastMineTime>
```
ehildenb, line 25: Should this start at `#time()` instead?

Also, can we call it `lastTime`, instead of `lastMineTime`? It seems that we are using it for other things too, eg. on `#evm_revert` and `#evm_snapshot`, which are not mining.

web3.md
```diff
@@ -512,8 +513,9 @@ WEB3 JSON RPC
     syntax KItem ::= "#evm_revert"
  // ------------------------------
     rule <k> #evm_revert => #popNetworkState ~> #rpcResponseSuccess(true) ... </k>
-         <params>    [ DATA:Int, .JSONs ] </params>
-         <snapshots> SNAPSHOTS            </snapshots>
+         <params>       [ DATA:Int, .JSONs ] </params>
+         <snapshots>    SNAPSHOTS            </snapshots>
+         <lastMineTime> _ => #time()         </lastMineTime>
```
ehildenb, line 518: Why do we update the timestamp on revert?

ehildenb:
I'm somewhat skeptical of all the places we've added updates to the `<lastMineTime>` cell. Are we sure it's necessary in all of them, and not just when we're mining a block?

It may only actually be needed at the beginning of execution (by putting `#time()` in the initial configuration), and when we actually use the cell (the only place being where we say `<lastMineTime> LASTTIME => #time() </lastMineTime>`).

web3.md
```diff
@@ -1728,7 +1731,7 @@ Timestamp Calls
  // -----------------------------------
     rule <k> #firefly_setTime => #rpcResponseSuccess(true) ... </k>
          <params> [ TIME:String, .JSONs ] </params>
-         <timestamp> _ => #parseHexWord( TIME ) </timestamp>
+         <timeDiff> _ => #parseHexWord( TIME ) -Int #time() </timeDiff>
```
ehildenb, line 1734: Should we also be adjusting `<timestamp>` here?

gtrepta: Nope. `<timestamp>` is just updated by `<timeDiff>` for tx executions now.

ehildenb:
Let's add just an `evm_increaseTime` test which demonstrates the same failure we see on CI.

web3.md
gtrepta: And also when `evm_mine` is called. Basically whenever the blockchain is going to get updated.

PR #785, opened 2020-05-02
Use DockerHub images for CI

ehildenb:
~Not ready for merge.~

@dwightguth this is ready for review now.


dwightguth:
Hardcoding /usr/lib/kframework will not work on Mac OS, which installs it into /usr/local/lib/kframework. You probably want something like `which kompile`/../lib/kframework.

PR #786, opened 2020-05-04
Remove node build from semantics

ehildenb:
The Node build is fully subsumed by the Web3 build now, so no need to keep it around, which should allow for some simplifications in the blockchain-k-plugin as well.


evm.md
```diff
@@ -461,63 +460,6 @@ The `CallOp` opcodes all interperet their second argument as an address.
...
 
-### Helpers
-
--   `#addr` decides if the given argument should be interpreted as an address (given the opcode).
-
-```k
-    syntax InternalOp ::= "#load" "[" OpCode "]"
- // --------------------------------------------
-    rule <k> #load [ OP:OpCode ] => #loadAccount #addr(W0) ... </k>
-         <wordStack> (W0 => #addr(W0)) : WS </wordStack>
```
dwightguth, line 472: Removing these three rules entirely looks suspicious.

ehildenb:
Should be ready for re-review, I've tested locally, and the only thing I'm not convinced of is that the proofs will all pass.

evm.md
```diff
@@ -461,61 +460,27 @@ The `CallOp` opcodes all interperet their second argument as an address.
...
-    rule #addr?(SELFDESTRUCT) => true
-    rule #addr?(EXTCODEHASH)  => true
-    rule #addr?(OP)           => false requires (OP =/=K BALANCE) andBool (OP =/=K SELFDESTRUCT) andBool (OP =/=K EXTCODEHASH)
-
-    syntax Bool ::= "#code?" "(" OpCode ")" [function]
- // --------------------------------------------------
-    rule #code?(EXTCODESIZE)  => true
-    rule #code?(EXTCODECOPY)  => true
-    rule #code?(OP)           => false requires (OP =/=K EXTCODESIZE) andBool (OP =/=K EXTCODECOPY)
+    rule <k> #addr [ OP:OpCode ] => . ... </k> [owise]
```
dwightguth, line 483: pretty sure the java backend doesn't understand owise on non-function rules.

ehildenb: @dwightguth I converted to an explicit side condition.

dwightguth:
Looks fine. I wonder if we can make that side condition faster by converting it to structural matching, but we can worry about that later.

PR #793, opened 2020-05-11
update blockchain-k-plugin and build process

sskeirik:
blockchain-k-plugin now includes its dependencies as submodules
and build logic for them in its Makefile; we update the KEVM
Makefile to call the blockchain-k-plugin Makefile to build these
libraries.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,7 +30,7 @@ pipeline {
             stage('Tests') { steps { sh 'make split-tests -j3'   } }
           }
         }
-        stage('Build') { steps { sh 'make build -j4' } }
+        stage('Build') { steps { sh 'make CC=clang-8 CXX=clang++-8 libcryptopp -j4 && make CC=clang-8 CXX=clang++-8 build -j4' } }
         stage('Test Execution') {
           failFast true
           options { timeout(time: 20, unit: 'MINUTES') }
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -30,7 +30,7 @@ pipeline {
             stage('Tests') { steps { sh 'make split-tests -j3'   } }
           }
         }
-        stage('Build') { steps { sh 'make CC=clang-8 CXX=clang++-8 libcryptopp -j4 && make CC=clang-8 CXX=clang++-8 build -j4' } }
+        stage('Build') { steps { sh 'make build -j4' } }
         stage('Test Execution') {
           failFast true
           options { timeout(time: 20, unit: 'MINUTES') }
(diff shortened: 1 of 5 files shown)


Jenkinsfile
```diff
@@ -30,7 +30,7 @@ pipeline {
             stage('Tests') { steps { sh 'make split-tests -j3'   } }
           }
         }
-        stage('Build') { steps { sh 'make build -j4' } }
+        stage('Build') { steps { sh 'make CC=clang-8 CXX=clang++-8 libcryptopp -j4 && make CC=clang-8 CXX=clang++-8 build -j4' } }
```
ehildenb, line 33: If these are needed by default, they should be built _into_ the Makefile, not here. The top-level commands like `make build` should "just work" on most systems. And if we can't make it work on most systems then they should "just work" on Ubuntu 18.04.

Makefile
```diff
@@ -70,42 +70,27 @@ distclean:
...
 
-$(libff_out): $(DEPS_DIR)/libff/CMakeLists.txt
-	@mkdir -p $(DEPS_DIR)/libff/build
-	cd $(DEPS_DIR)/libff/build \
-	    && CC=$(LIBFF_CC) CXX=$(LIBFF_CXX) cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$(BUILD_LOCAL) $(LIBFF_CMAKE_FLAGS) \
-	    && make -s -j4 \
-	    && make install
+libcryptopp: $(libcryptopp_out)
+$(libcryptopp_out): $(PLUGIN_SUBMODULE)/make.timestamp
+	cd deps/plugin && $(MAKE) CC=$(CC) CXX=$(CXX) PREFIX=$(BUILD_LOCAL) -- libcryptopp
```
ehildenb, line 93: Why do we need to build libcryptopp from source? If it's because of a bug that only affects the hooks used in Tezos, then please continue to use the distribution installed version here. We are bundling Firefly as a debian package, and anything that can be installed via normal Ubuntu dependencies makes that process _way_ easier.

Jenkinsfile
sskeirik: I added another commit that removes these because in my testing, they are not needed on Ubuntu 18.04.

Makefile
sskeirik: I added another commit that removes this from the Makefile.

ehildenb:
@sskeirik this looks good to me! If you can:

-   Review and get merged this one: https://github.com/runtimeverification/blockchain-k-plugin/pull/97
-   Then rebase yours on top and get it merged: https://github.com/runtimeverification/blockchain-k-plugin/pull/96

Then this should be good to go!

PR #815, opened 2020-06-05
Fix typo in #gasExec for SSTORE

daejunpark:
Unless it is intentional.


ehildenb:
Good catch. I guess since most of our proofs are done over ISTANBUL this wasn't caught before.

PR #825, opened 2020-06-16
Updates to kevm script to make proving easier

ehildenb:


denis-bogdanas:
Please add this: on `--debug`, print kevm command and underlying K command. This will simplify adoption a lot as it will be simpler to understand what happens under the hood.

ehildenb:
I've made some updates. On `--debug` it will print out the `kprove` command it is using, but having it print out the KEVM command it's using is more trouble than it's worth, because you can just tell whatever script you're calling KEVM with to print that out. If it's a shell, then it's whatever you typed, if it's a Makefile, then make automatically prints out that command, and if it's another script, you can do a `set -x` before the call to KEVM.

denis-bogdanas:
sounds good

PR #826, opened 2020-06-18
Misc lemmas for symbolic testing

denis-bogdanas:


ehildenb:
@denis-bogdanas please add tests which would not be provable with haskell backend before this and now are to file `tests/specs/functional/lemmas-spec.k`.

denis-bogdanas:
@ehildenb I need a formula to represent `b"0000000000000000000000000000000000000000000000000000000000000001"` instead of `String2Bytes("00000000000000000000000000000001")` in the tests. If we put the `Bytes` expression above, we get parse error.

denis-bogdanas:
@ehildenb Nevermind, fixed.

tests/specs/lemmas.k
```diff
@@ -291,6 +291,11 @@ module LEMMAS-HASKELL [symbolic, kore]
...
     rule 0 <=Int lengthBytes ( _ )        => true [smt-lemma]
 
-    rule bool2Word( B:Bool ) ==Int 1 => B         [simplification]
-    rule bool2Word( B:Bool ) ==Int 0 => notBool B [simplification]
+    rule I ==Int bool2Word( B:Bool ) => bool2Word(B) ==Int I            [simplification, concrete(I)]
+    rule bool2Word( B:Bool ) ==Int I => B ==K word2Bool(I)              [simplification, concrete(I)]
+
+    rule BA:ByteArray ==K #buf( 32, DATA ) => #buf( 32, DATA ) ==K BA   [simplification, concrete(BA)]
+    rule #buf( 32, DATA ) ==K BA:ByteArray => DATA ==Int #asInteger(BA)
+      requires #sizeByteArray(BA) <=Int 32                              [simplification, concrete(BA)]
```
ehildenb, line 299: I think you may also need `0 <=Int DATA andBool DATA <Int pow256` here, because otherwise `#buf` is ill-defined.

ehildenb: And if `#buf` is ill-defined, we shouldn't get `false` here, we should get `#False` (`\bottom`).

denis-bogdanas: This is the job of `#Ceil(#buf)` which I disabled for now because it's prohibitive for performance. If we add it here we'll likely forget it and have lingering performance issues.

tests/specs/functional/lemmas-spec.k
```diff
@@ -3,7 +3,7 @@ requires "../lemmas.k"
 module VERIFICATION
     imports LEMMAS
 
-    syntax StepSort ::= Map | ByteArray
+    syntax StepSort ::= Map | ByteArray | Int | Bool
  // -----------------------------------
```
ehildenb, line 7: Formatting.

tests/specs/functional/lemmas-spec.k
```diff
@@ -57,4 +57,16 @@ module LEMMAS-SPEC
...
+    rule <k> runLemma ( bool2Word( B:Bool ) ==Int 1 ) => doneLemma( B ==K true ) ... </k>
+    rule <k> runLemma ( 1 ==Int bool2Word( B:Bool ) ) => doneLemma( B ==K true ) ... </k>
+
+    //#buf simplification
+    rule <k> runLemma ( #buf ( 32 , bool2Word ( B:Bool ) ) ==K #buf(32, 1) )
+             => doneLemma( B ==K true ) ... </k>
+    rule <k> runLemma ( #buf ( 32 , I:Int ) ==K #buf(32, 1) )
+             => doneLemma( I ==K 1 ) ... </k>
+    rule <k> runLemma ( #buf(32, 1) ==K #buf ( 32 , I:Int ))
+             => doneLemma( I ==K 1 ) ... </k>
```
ehildenb, line 70: These can all be one line, and lined up horizontally, formatting.

Makes scanning the file a lot easier.

denis-bogdanas:
All done now. Including a version for Java.

ehildenb:
It looks OK to me, but I want @daejunpark to take a look too.

tests/specs/lemmas.k
ehildenb: @denis-bogdanas I don't understand "disabled for now" here? I still see this `#Ceil` condition in `edsl.md`, and it looks like it's still enabled.

I think the only reason this is sound without the additional range checks on `DATA` is because of that `#Ceil` condition, right?

denis-bogdanas:
Too much bureaucracy IMO.

daejunpark:
Thanks @ehildenb to have me here. I think I will need to review any lemma changes, unless it is trivial.

tests/specs/lemmas.k
```diff
@@ -225,6 +226,14 @@ module LEMMAS-JAVA [symbolic, kast]
...
     rule chop(I) => I requires 0 <=Int I andBool I <Int pow256  [simplification]
 
+    //Simplification of bool2Word() ==Int CONCRETE, #buf() ==K CONCRETE
+    rule I                   ==K bool2Word( B:Bool ) => bool2Word(B) ==Int I    requires #isConcrete(I)  [simplification]
+    rule bool2Word( B:Bool ) ==K I                   => B ==K word2Bool(I)      requires #isConcrete(I)  [simplification]
+
+    rule BA:ByteArray     ==K #buf( 32, DATA ) => #buf( 32, DATA ) ==K BA       requires #isConcrete(BA) [simplification]
+    rule #buf( 32, DATA ) ==K BA:ByteArray     => DATA ==Int #asInteger(BA)
+      requires #sizeByteArray(BA) <=Int 32                                      andBool  #isConcrete(BA) [simplification]
+
```
daejunpark, line 236: I think this is the nontrivial change in this PR.

@denis-bogdanas Could you please explain why this is needed? I mean, why the previous lemmas that you are going to delete above are not enough?

tests/specs/lemmas.k
```diff
@@ -291,6 +297,12 @@ module LEMMAS-HASKELL [symbolic, kore]
...
-    rule bool2Word( B:Bool ) ==Int 1 => B         [simplification]
-    rule bool2Word( B:Bool ) ==Int 0 => notBool B [simplification]
+    //Simplification of bool2Word() ==Int CONCRETE, #buf() ==K CONCRETE
+    rule I                   ==Int bool2Word( B:Bool ) => bool2Word(B) ==Int I  [simplification, concrete(I)]
+    rule bool2Word( B:Bool ) ==Int I                   => B ==K word2Bool(I)    [simplification, concrete(I)]
+
+    rule BA:ByteArray     ==K #buf( 32, DATA ) => #buf( 32, DATA ) ==K BA       [simplification, concrete(BA)]
+    rule #buf( 32, DATA ) ==K BA:ByteArray     => DATA ==Int #asInteger(BA)
+      requires #sizeByteArray(BA) <=Int 32                                      [simplification, concrete(BA)]
+
```
daejunpark, line 307: I think this is the nontrivial change in this PR.

@denis-bogdanas Could you please explain why this is needed? I mean, why the previous lemmas that you are going to delete above are not enough?

tests/specs/lemmas.k
denis-bogdanas: Right. That's why we don't need extra checks here. It is disabled in symbolic testing branches.

tests/specs/lemmas.k
daejunpark: Also, could you explain the scenario when the equality `#buf() ==K CONCRETE` occurs?

denis-bogdanas: 1. Because tests in `lemmas-spec.k` are run on both Java and Haskell.
2. To increase compatibility between the 2 sets of lemmas.

`#buf` equality occurs in symbolic testing, in `assume()` custom function. It could have been written differently that wouldn't trigger this term but chances are we'll need this lemma in other places. See `lemmas-spec.k` for real-life tests.

daejunpark: I cannot see why lemmas-spec.k is relevant for bool2Word lemmas.

daejunpark: Also, do you think that it is better to have the lemmas specific to the symbolic testing in this common lemma file?

ehildenb: @daejunpark I think we should put them in the common lemmas file unless (i) we have a specific reason not to, or (ii) someone shows an example where it actually hurts performance.

We should not have a huge fragmented set of lemmas spread out everywhere because we're afraid it _might_ hurt performance. If you have concrete evidence that it _does_ hurt performance somewhere, we need to make a test for it.

denis-bogdanas: 1. I'm not sure what you mean. There are tests for `bool2Word` lemmas in lemmas-spec.k.
2. I think it's better to have the setup for Haskell and Java as close as possible. When there is no big effort doing that. This will help porting stuff from one backend to another, and who knows maybe comparing deductive power of the 2 backend. If maintenance is a concern then we should add comments about why each group of lemmas was needed. Or better, for each test, if it was inspired from real life stuck configurations, write where it was taken from.

3. Regarding performance. In Haskell regular rules and builtins take priority, so lemmas will be tried only if regular rules couldn't apply. In Java Builtins apply first then all the rest. In addition function evaluations are cached so performance drawback is minimal.

Generally lemmas hugely affect performance when:
- they match very general terms, especially terms in side conditions of many rules, like `A <=Int MaxUInt256`
- they have side conditions
- side conditions don't match. Triggering attempts to apply lemmas over and over again.

Terms above are not that common.

ehildenb: @daejunpark to clarify, we have a spec called `tests/specs/functional/lemmas-spec.k` which tests the functional simplifications that show up in the `lemmas.k` file. This test is run on both Java and Haskell backend, and is the place to put regressions in terms of simplification power that occur (to ensure that the same terms will remain simplifiable in the future). @denis-bogdanas has provided some tests of his changes there (at my request), meaning they need to be simplifiable with both Java nad Haskell backend.

I agree with the principle of keeping the lemmas for Java and Haskell in sync as much as possible.

ehildenb:
I will continue to ask @daejunpark to review changes to the lemmas, but I also ask @daejunpark to change what he focuses on in his reviews. Specifically, I'm _not_ interested in reviews where "maybe this will affect performance negatively", if you can instead focus on:

-   Actual errors in the simplification lemmas which cause unsoundness,
-   Actual concrete examples where performance _would_ be hurt, but not worries about how it _might_ affect performance, accompanied by benchmarks showing the problem (either from existing tests or new tests written for that purpose), and
-   Missing test cases that should be added to the `lemmas-spec.k` file along with the new lemmas.

We cannot expect to have perfect evolution of this lemmas file from the start, we can only learn from experience. So if you do find an example where performance is _definitely_ hurt, then by all means, make or suggest a regression test and add it to the test-suite so that we can make sure it doesn't happen again or that the lemmas are adjusted. But I don't want to have the same argument over and over again about how something _might_ affect performance of a project in the future negatively.

I want to shift the way we think about verification and lemmas away from guesswork to something we back up with (i) benchmarks and (ii) regression tests, much the same way we deal with concrete execution. They are no different from the perspective of maintaining high quality software.

tests/specs/lemmas.k
daejunpark: @denis-bogdanas Sorry I missed the newly added tests in the lemmas-spec.k. But my original question was the motivation of the change from:
```
    rule bool2Word( B:Bool ) ==Int 1 => B         [simplification]
    rule bool2Word( B:Bool ) ==Int 0 => notBool B [simplification]
```
to:
```
    rule bool2Word( B:Bool ) ==Int I => B ==K word2Bool(I)    [simplification, concrete(I)]
```
Did you find an actual case where `I` is other than 0 and 1, or is this just a refactoring change?  If the former, I want to know when it happens.

Yes, the new lemmas don't seem to affect the performance much, but I wanted to double check if you also thought about it.

daejunpark:
@ehildenb Thanks for the suggestion. I'll keep that in mind.

A question. Do we have an automatic way to detect performance regression in CI?

tests/specs/lemmas.k
denis-bogdanas: There are no such cases and they are not possible of course. The change was to (1) minimize number of lemmas in Haskell because they have symmetric equivalent so duplication gets multiplied, and (2) keep Java in sync with Haskell.

ehildenb:
Yes we do catch big performance regressions on CI, because we have timeouts on every test step that are set to roughly 1.5x how long it should run at quickest. So if there is a really bad one, it will be caught.

PR #827, opened 2020-06-18
Make benchmark proofs parse on Haskell backend dry run

ehildenb:
Fixes runtimeverification/firefly#510

This PR does several things to make begin working on the haskell backend proofs for benchmarks easier:

-   Removes all `#bufSeg`, `#bufElm`, `select`, `selectRange`, `store`, `storeRange` abstractions.
-   Switches a bunch of cases of `WordStack` to `ByteArray`.
-   Removes lots of uneeded lemmas.
-   Consolidates the various lemma files into `verification.k`.

This makes it so that we can call `./kevm prove --backend haskell tests/specs/benchmarks/*-spec.k VERIFICATION --dry-run` on each of the benchmark proofs (and enables it on CI), so that we know that not only (i) the proofs pass on the Java backend, but also (ii) the proofs are parsable on the Haskell backend build of the semantics.


tests/specs/erc20/evm-symbolic.k
```diff
@@ -4,79 +4,11 @@ module EVM-SYMBOLIC     [symbolic]
...
  // -----------------------------------------------------------------------------------
- /*
-  startGas - gas available at the beginning of execution
-  nonMemory - gas consumed on non-memory operations
-  memory - gas consumed on memory operation
-
-  Concrete semantics:
-
-  rule #gas ( StartGas , NonMemory , Memory ) => StartGas -Int NonMemory -Int Memory
- */
```
denis-bogdanas, line 46: this has to be preserved

denis-bogdanas:
Also please split this into 2 PRs. One for common lemmas and one for benchmarks. In case we want to review changes to lemmas in future, I want to see them in a separate commit.

ehildenb:
@denis-bogdanas I've removed all the changes to the global lemmas file and the ERC20 proofs, I'll get those in a separate PR.

PR #829, opened 2020-06-19
Consistent integer comparisons

ehildenb:
This switches `>Int` for `<Int` and `>=Int` for `<=Int` throughout the semantics so that we only need to write lemmas for one way of comparing things.


dwightguth:
I'm not really sure how I feel about this; it makes the semantics a lot less readable, introduces an implicit assumption about all future changes to the repository that will make things break if it's violated, but isn't being enforced anywhere, and it's not even as reductive as you might want; I can easily go further by transforming `I <=Int J` into `notBool J <Int I`

PR #830, opened 2020-06-20
Incremental updates to the state trie

gtrepta:
fixes: https://github.com/runtimeverification/firefly/issues/882

This PR adds a `<stateTrie>` cell to web3's configuration which keeps an up-to-date merkle tree of the blockchain state. It gets updated any time a block is mined and is also included in pushes and pops of the `<snapshots>` cell.

This is an alternative to the previous, slower method of recomputing the entire state trie on every block mine.


web3.md
```diff
@@ -1840,6 +1852,9 @@ Mining
                    | "#updateStateRoot"
                    | "#updateTransactionsRoot"
                    | "#updateReceiptsRoot"
+                   | "#initStateTrie"
+                   | "#updateStateTrie"
+                   | #updateStateTrie ( JSONs )
  // --------------------------------------
```
ehildenb, line 1858: formatting

web3.md
```diff
@@ -1874,6 +1882,24 @@ Mining
...
+    rule <k> #updateStateTrie => #updateStateTrie(OMMERS) ... </k>
+         <ommerBlockHeaders> [ OMMERS ] </ommerBlockHeaders>
+
+    rule <k> #updateStateTrie( [ _ , _ , OMMER , _ , _ , _ , _ , _ , _ , _ ] , REST ) => #updateStateTrie( REST ) ... </k>
+         <touchedAccounts> ... .Set => SetItem(OMMER) ... </touchedAccounts>
+
+    rule <k> #updateStateTrie( .JSONs ) => . ... </k>
+         <schedule> SCHED </schedule>
+         <stateTrie> TREE => #putAccountsInTrie( TREE, Set2List(SetItem(MINER) (ACCTS -Set #precompiledAccounts(SCHED))), <accounts> ACCTSCELL </accounts> ) </stateTrie>
+         <touchedAccounts> ACCTS => .Set </touchedAccounts>
```
ehildenb, line 1899: Why do we need to do `ACCTS -Set #precompiledAccounts(SCHED)` here? Where is `#precompiledAccounts(SCHED)` being added to the `<touchedAccounts>`?

gtrepta: If a contract ever calls one of the precompiled contracts it will be added to `<touchedAccounts>`. And we can't do an update on a precompiled contract in the trie because it doesn't have an associated account so it will remove it from the trie otherwise.

PR #832, opened 2020-06-22
Precompiled contracts in state root based on schedule

gtrepta:
fixes: https://github.com/runtimeverification/firefly/issues/899

This PR makes the web3 client reference the `<schedule>` cell when adding precompiled contracts to the state trie, making it more future-proof as well as backwards compatible with previous hardforks. The web3 tests needed to be updated to reflect this change and are now specified as istanbul tests. So any future EIPs that add more precompiled contracts will not require these tests to be modified again.


serialization.md
```diff
@@ -697,17 +697,13 @@ Tree Root Helper Functions
...
-         #parseByteStackRaw( Hex2Raw( #unparseData( 6, 20 ) ) ) |-> #emptyContractRLP
-         #parseByteStackRaw( Hex2Raw( #unparseData( 7, 20 ) ) ) |-> #emptyContractRLP
-         #parseByteStackRaw( Hex2Raw( #unparseData( 8, 20 ) ) ) |-> #emptyContractRLP
+    syntax Map ::= #precompiledAccountsMap   ( Set )       [function]
+                 | #precompiledAccountsMapAux( List, Map ) [function]
+ // -----------------------------------------------------------------
+    rule #precompiledAccountsMap( ACCTS ) => #precompiledAccountsMapAux( Set2List( ACCTS ), .Map )
+
+    rule #precompiledAccountsMapAux( .List, M ) => M
+    rule #precompiledAccountsMapAux( (ListItem( ACCT ) => .List) _, (.Map => #parseByteStackRaw( Hex2Raw( #unparseData( ACCT, 20 ) ) ) |-> #emptyContractRLP) _ )
```
ehildenb, line 706: We shouldn't be in the habit of doing map updates like this because it can crash when there are conflicts in the keys.

Instead do `MAP => MAP [ KEY <- VALUE ]`, which handles the case where the key doesn't exist or does exist already smoothly.

PR #835, opened 2020-06-23
Simplify range definition, add more tooling to KLab integration

ehildenb:
This PR:

-   Removes all the additional `_ [ _ <- undef ]` nodes that are inserted by having `range` remove map elements as it processes them.
-   Updates the klab integration to use the correct cell names, and newest setting used by DappHub guys.

These changes were inspired by working on one of the MKR proofs for MCD.


evm-types.md
```diff
@@ -509,10 +509,9 @@ Most of EVM data is held in local memory.
     syntax ByteArray ::= #range ( Memory , Int , Int )             [function, functional]
     syntax ByteArray ::= #range ( Memory , Int , Int , ByteArray ) [function, functional, klabel(#rangeAux)]
  // --------------------------------------------------------------------------------------------------------
-    rule [#range]:         #range(WM, START, WIDTH) => #range(WM, START +Int WIDTH -Int 1, WIDTH, .WordStack)
-    rule [#rangeAux.base]: #range(WM,           END, WIDTH, WS) => WS                                           requires notBool WIDTH >Int 0
-    rule [#rangeAux.none]: #range(WM,           END, WIDTH, WS) => #range(WM, END -Int 1, WIDTH -Int 1, 0 : WS) requires (WIDTH >Int 0) andBool notBool END in_keys(WM)
-    rule [#rangeAux.some]: #range(END |-> W WM, END, WIDTH, WS) => #range(WM, END -Int 1, WIDTH -Int 1, W : WS) requires (WIDTH >Int 0)
+    rule [#range]: #range(WM, START, WIDTH) => #range(WM, START +Int WIDTH -Int 1, WIDTH, .WordStack)
+    rule [#rangeAux.base]: #range(WM, END, WIDTH, WS) => WS requires notBool 0 <Int WIDTH
+    rule [#rangeAux.rec]:  #range(WM, END => END -Int 1, WIDTH => WIDTH -Int 1, WS => #lookup(WM, END) : WS) requires 0 <Int WIDTH
```
ehildenb, line 514: This change might not be good because `#range` is used for accessing the local memory, which has bytes, not words, and @denis-bogdanas is working on making `#lookup` assert that it returns words.

So it's not incorrect, it's just that we should have a tighter assertion about the fact that this is a `#byteLookup` and not a `#lookup`.

evm-types.md
```diff
@@ -660,13 +652,22 @@ Addresses
...
+ // ----------------------------------------------------------------------------------------
+    rule [#lookup.some]:       #lookup( (KEY |-> VAL:Int) M, KEY ) => VAL requires          0 <=Int VAL andBool VAL <Int pow256
+    rule [#lookup.outOfRange]: #lookup( (KEY |-> VAL:Int) M, KEY ) => 0   requires notBool (0 <=Int VAL andBool VAL <Int pow256)
+    rule [#lookup.none]:       #lookup(                   M, KEY ) => 0   requires notBool KEY in_keys(M)
+    rule [#lookup.notInt]:     #lookup( (KEY |-> VAL    ) M, KEY ) => 0   requires notBool isInt(VAL)
+
+    rule [#lookupMemory.some]:       #lookupMemory( (KEY |-> VAL:Int) M, KEY ) => VAL requires          0 <=Int VAL andBool VAL <Int pow256
+    rule [#lookupMemory.outOfRange]: #lookupMemory( (KEY |-> VAL:Int) M, KEY ) => 0   requires notBool (0 <=Int VAL andBool VAL <Int pow256)
+    rule [#lookupMemory.none]:       #lookupMemory(                   M, KEY ) => 0   requires notBool KEY in_keys(M)
+    rule [#lookupMemory.notInt]:     #lookupMemory( (KEY |-> VAL    ) M, KEY ) => 0   requires notBool isInt(VAL)
```
daejunpark, line 670: Two questions:
1. What is the difference between #lookup and #lookupMemory here?
2. Where does the range condition come from?  (It wasn't there before right?)

ehildenb: `#lookup` ensures that it will return something between `0` and `pow256`.

`#lookupMemory` ensures that it will return something between `0` and `256`.

daejunpark: Oh, then the condition for lookupMemory needs to be fixed.

ehildenb: Ah yeah, copy-paste error.

daejunpark:
The #range rules are not used by the haskell backend, so this shouldn't affect the performance.

PR #839, opened 2020-06-24
porting address00-spec to haskell backend

daejunpark:
Changes:
- The `matching` attribute is not supported yet by the haskell backend, and thus the #gas rule doesn't work. So, the spec is simply modified to have a concrete initial gas in the precondition.
- The `#isConcrete` function is not supported by the haskell backend. (It has a nicer alternative.) For now, all the lemmas that use `#isConcrete` are disabled in the haskell backend.

### DIFF
--- a/edsl.md
+++ b/edsl.md
@@ -219,7 +219,7 @@ where `F1 : F2 : F3 : F4` is the (two's complement) byte-array representation of
     rule #getValue(   #bool( DATA )) => DATA
       requires #range(0 <= DATA <= 1)
 
-    syntax Int ::= #ceil32 ( Int ) [function, smtlib(ceil32), smt-prelude]
+    syntax Int ::= #ceil32 ( Int ) [function, smt-hook(ceil32), smt-prelude]
  // -----------------------------------------
     rule [#ceil32]: #ceil32(N) => ((N +Int 31) /Int 32) *Int 32
 ```
--- a/edsl.md
+++ b/edsl.md
@@ -220,7 +220,7 @@ where `F1 : F2 : F3 : F4` is the (two's complement) byte-array representation of
       requires #range(0 <= DATA <= 1)
 
     syntax Int ::= #ceil32 ( Int ) [function, smt-hook(ceil32), smt-prelude]
- // -----------------------------------------
+ // ------------------------------------------------------------------------
     rule [#ceil32]: #ceil32(N) => ((N +Int 31) /Int 32) *Int 32
 ```
 
--- a/tests/failing-symbolic.haskell
+++ b/tests/failing-symbolic.haskell
@@ -1,4 +1,3 @@
-tests/specs/benchmarks/address00-spec.k
 tests/specs/benchmarks/bytes00-spec.k
 tests/specs/benchmarks/ecrecover00-siginvalid-spec.k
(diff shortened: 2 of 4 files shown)


Makefile
```diff
@@ -349,6 +349,7 @@ tests/%.parse: tests/%
 tests/specs/functional/%.prove:                 TEST_SYMBOLIC_BACKEND = haskell
 tests/specs/examples/%.prove:                   TEST_SYMBOLIC_BACKEND = haskell
 tests/specs/erc20/hkg/totalSupply-spec.k.prove: TEST_SYMBOLIC_BACKEND = haskell
+tests/specs/benchmarks/address00-spec.k.prove:  TEST_SYMBOLIC_BACKEND = haskell
```
daejunpark, line 352: @ehildenb I'd like to run these spec proof tests for both java and haskell backends. Could you please reorg  Makefile to do that? I think we want to do:
- run all proof tests with the java backend
- _dry-run_ all proof tests with the haskell backend (this can be dropped once the porting is done)
- run all _ported_ proof tests with the haskell backend

ehildenb: I would rather not re-org the makefile to do this, we used to do this but I don't really care to support the Java backend with ported proofs.

ehildenb: But if you absolutely want to, it's better to do it in the Jenkinsfile. We can have separate stages for running all the proofs with haskell vs. java (the ones that are supported).

daejunpark: OK. I'll think about it. Let's focus on merging this PR then, and later let me open another PR if it is really needed.

ehildenb: I'll have a PR ready in half an hour or so that enables running them on both.

edsl.md
```diff
@@ -219,7 +219,7 @@ where `F1 : F2 : F3 : F4` is the (two's complement) byte-array representation of
     rule #getValue(   #bool( DATA )) => DATA
       requires #range(0 <= DATA <= 1)
 
-    syntax Int ::= #ceil32 ( Int ) [function, smtlib(ceil32), smt-prelude]
+    syntax Int ::= #ceil32 ( Int ) [function, smt-hook(ceil32), smt-prelude]
  // -----------------------------------------
```
ehildenb, line 223:
```suggestion
    syntax Int ::= #ceil32 ( Int ) [function, smt-hook(ceil32), smt-prelude]
 // ------------------------------------------------------------------------
```

ehildenb:
This looks good to me, I'm happy-enough with the hardcoded gas, or you can calculate the exact gas expression and put it in the post condition.

Please resolve the merge conflicts with the Makefile though, and make sure to remove this from the corresponding new failing lists.

daejunpark:
> Would it be possible to replace `[matching(#gas)]` with a custom equality rule like:
> 
> ```
> rule #gas(A, B, C) ==Int #gas(A2, B2, C2) => true
>   requires A ==Int A2 andBool B ==Int B2 andBool C ==Int C2
> ```
> 
> ?
> 
> It is sound, but might affect performance.

@denis-bogdanas Thanks for the suggestion. We will consider this when we get there.

daejunpark:
> Please resolve the merge conflicts with the Makefile though, and make sure to remove this from the corresponding new failing lists.

@ehildenb Done. Thanks!

PR #841, opened 2020-06-26
timeFreeze for web3 tests

gtrepta:
fixes: https://github.com/runtimeverification/firefly/issues/894

Sister PR: https://github.com/runtimeverification/blockchain-k-plugin/pull/110

These PRs add a --timeFreeze parameter to the web3 client.

When launched with --timeFreeze, kevm-client will not use `#time()` when updating the `<timestamp>` cell, making it much easier to control what timestamps are generated for unit tests.


ehildenb:
@gtrepta I wonder about a slightly different way of doing this.

What if instead we had a new function:

```
syntax Int ::= #time ( Bool ) [function]
rule #time(false) => 0
rule #time(true)  => #time()
```

Then we just replace every occurance of `#time()` in the semantics with `#time(TIMEFREEZE)` (pass in the contents of the `<timeFreeze>` cell. Would that achieve the same thing? I think it would reduce the amount of rule duplication if it works.

PR #844, opened 2020-06-26
Pop Listitem in txPending & txOrder when #eth_sendTransaction fails

WenhaoSu:
Fixes: runtimeverification/firefly#910

Remove topmost Listitem in `<txPending>` and `<txOrder>` when `#eth_sendTransaction` fails.
We'll also want to find a way to reformat and clean it up since there's a lot of duplicate code for similar logic inside `web3.md`.

This is linked to the [issue 910](https://github.com/runtimeverification/firefly/issues/910) of Firefly.


ehildenb:
@WenhaoSu , `eth_estimateGas`, `eth_sendRawTransaction`, and `eth_call` all also deal with executing transactions, and so may not be popping the `<txPending>` and `<txOrder>` cells correctly. Can you please check them as well?

Also, I would make a new construct called `#popTx` which does this:

```
syntax KItem ::= "#popTx"

rule <k> #popTx => . ... </k>
     <txPending> (ListItem(_) => .) ... </txPending>
     <txOrder> (ListItem(_) => .) ... </txOrder>
```

WenhaoSu:
@ehildenb Sure, I'll get them checked. And yep, get a KItem `#popTx` makes sense!

WenhaoSu:
Below is the observation result:

- :heavy_check_mark: `#eth_sendTransaction`: Passes all submodule test cases, fixed the hanging situation in runtimeverification/firefly#910.
- :heavy_check_mark: `#eth_sendRawTransaction`: It requires popping Txs, and it is already implemented in the original version. We can just use the new Kitem `#popTxLists` to replace the original code.
- :heavy_multiplication_x: `#eth_estimateGas`: By adding `#popTxLists` to the failure case of `#eth_estimateGas_finalize` will cause test failure for `nexus-mutual` submodule. It could be that the failure cases of `#eth_estimateGas_finalize` already handled the logic of `#popTxLists`.
- :heavy_multiplication_x: `#eth_call`: By adding `#popTxLists` to the failure cases of `#eth_call_finalize`, it caused hanging for `ethwall_contract` test. It's better to keep the origin logic.

web3.md
```diff
@@ -624,13 +624,13 @@ eth_sendTransaction
          <output> RD </output>
          <endPC> PCOUNT </endPC>
 
-    rule <k> _:String ~> #eth_sendTransaction_final => #rpcResponseError(-32000, "base fee exceeds gas limit") ... </k>
+    rule <k> _:String ~> #eth_sendTransaction_final => #popTxLists ~> #rpcResponseError(-32000, "base fee exceeds gas limit") ... </k>
```
gtrepta, line 627: There's a rule above this:

```
    rule <k> (TXID:Int => "0x" +String #hashSignedTx(TN, TP, TG, TT, TV, TD, TW, TR, TS)) ~> #eth_sendTransaction_final ... </k>
         <message>
           <msgID> TXID </msgID>
           <txNonce>    TN </txNonce>
           <txGasPrice> TP </txGasPrice>
           <txGasLimit> TG </txGasLimit>
           <to>         TT </to>
           <value>      TV </value>
           <sigV>       TW </sigV>
           <sigR>       TR </sigR>
           <sigS>       TS </sigS>
           <data>       TD </data>
         </message>
```

I think this is a good place to pop TXID from. You can make another rule that matches the LHS of the k cell like this one that pops TXID from the tx lists, and this rule can match when TXID is not at the top of those lists.

PR #846, opened 2020-06-29
evm-types: #rangeUInt expressed from pow256, pow160. faster symbolic

denis-bogdanas:


ehildenb:
Awesome, thanks. This is a change I've wanted made for a while.

daejunpark:
@denis-bogdanas @ehildenb 

`#range*` were written using `[min, max]` instead of `[min, max+1)` because for `#range*Fixed`, `max+1` doesn't make much sense.  So, for all `#range*` to be uniform, the `[min, max]` form was used.

Also, the intention was (although never had a chance to do) to change all `X < pow256` terms to `X <= maxUInt256`, and `0 <= X andBool X < pow256` to `#rangeUInt(256, X)` in lemma files.

Also, in the mean time, we used to have the normalization:
```
    rule X <=Int maxUInt256 => X <Int pow256
    rule X <=Int maxUInt160 => X <Int pow160
    rule X <=Int 255        => X <Int 256
```

ehildenb:
I still prefer the interval `[0,pow256)` to `[0,maxUInt256]`. But if you want to open a PR making all the semantics more uniform in one direction or the other, feel free.

Another thing we should do is make all the comparison directions use `_<Int_` or `_<=Int_`, and never `_>Int_` or `_>=Int_`, because it speeds up removing side-conditions via just matching/rewriting instead of asking SMT.

Anyway, @denis-bogdanas reported that this PR resulted in roughly a 7-12% speedup for him, so I think it's good.

daejunpark:
I largely agree with you.  Just for the record:

> Another thing we should do is make all the comparison directions use _<Int_ or _<=Int_, and never _>Int_ or _>=Int_, because it speeds up removing side-conditions via just matching/rewriting instead of asking SMT.

For this, we used to have:
```
    rule X >Int Y => Y <Int X
    rule X >=Int Y => Y <=Int X
```

@denis-bogdanas Is it the case we don't have the normalization rules any longer?  If so, why?  Do they hurt the performance?

denis-bogdanas:
Inequality direction normalization lemmas are still present in VSC and in KEVM/LEMMAS-JAVA module. But not in KEVM/LEMMAS-HASKELL Adding lemmas above is unlikely to increase performance because in all usage places the arguments are expected to be concrete. Only when we deal with variable size byte arrays those lemmas should have a lot of impact.

@virgil-serbanuta You did some early work on cleaning up lemmas for Haskell+ERC20. Do you remember why it was necessary to remove lemmas above? Or you just tried to have as few lemmas as possible to reduce chances of problems.

virgil-serbanuta:
I don't remember anything about lemmas like these. If I removed them (did I?), most likely it was because there were too many lemmas that did not work with the Haskell backend, and I removed everything that didn't seem mandatory. Do you know in which commit/PR they were removed?

denis-bogdanas:
Judging by version history, @yzhang90 created module `LEMMAS-HASKELL`. But it doesn't matter now. We'll add lemmas from Java backend back to Haskell when they at least should not decrease performance.

PR #851, opened 2020-07-01
benchmark transformation for Haskell

ssaha6:


ehildenb:
@ssaha6 I see these are pretty generic transformations. Can you apply these transformations to _all_ the benchmark proofs in this PR? Or do some of them not work if you do that?

Also, if any of them are now passing on Haskell backend, please make sure to remove them from the `tests/failing-symbolic.haskell` list.

daejunpark:
@ssaha6 Please upload only specs that are passed by the haskell backend (as well as the java backend).  These are failed ones.  Please upload the version that works.  Also, please make the commit message and PR title more specific.

ehildenb:
@daejunpark it's OK to upload partial progress like this, in fact I think it's good.

If you can change the specs to all be closer to passing on Haskell backend (maybe with fewer warnings on dry-run), but still passing on the Java backend, then when we need to review the actual semantic changes, the overall diff will be smaller and easier to see.

So I would argue that these standard transformations should be applied to all the specs, and we get that merged first.

daejunpark:
@ehildenb Your suggestion of applying the systematic changes to all the benchmark specs sounds good. @ssaha6 will make another PR for that.  For this PR, we already have the version of specs that are passed by the haskell backend.  He will push them here.

tests/specs/benchmarks/overflow00-nooverflow-spec.k
```diff
@@ -122,8 +123,8 @@ _
 andBool #rangeUInt(256, CONTRACT_BAL)
 andBool #rangeUInt(256, CALLEE_BAL) andBool #range(0 <= CD < 1024)
 andBool #rangeAddress(MSG_SENDER) andBool #range(0 <= A0 < 2 ^Int 256 -Int 1)
-    ensures ?N ==Int A0 +Int 1
-
+    //ensures ?N ==Int A0 +Int 1
+    ensures true
```
ehildenb: We don't need the comment or ensures clause.

Please don't leave the old version commented out in the proofs, just update them. It will just confuse future people who need to work on these proofs.

tests/specs/benchmarks/overflow00-overflow-spec.k
```diff
@@ -11,7 +11,8 @@ module OVERFLOW00-OVERFLOW-SPEC
     <schedule> ISTANBUL </schedule>
     <ethereum>
       <evm>
-        <output> _ => #encodeArgs(#uint256(?N)) </output>
+        //<output> _ => #encodeArgs(#uint256(?N)) </output>
+        <output> _ => #encodeArgs(#uint256(0)) </output>
```
ehildenb, line 15: Same, remove the comment.

tests/specs/benchmarks/overflow00-overflow-spec.k
```diff
@@ -122,8 +123,8 @@ _
 andBool #rangeUInt(256, CONTRACT_BAL)
 andBool #rangeUInt(256, CALLEE_BAL) andBool #range(0 <= CD < 1024)
 andBool #rangeAddress(MSG_SENDER) andBool A0 ==Int 2 ^Int 256 -Int 1
-    ensures ?N ==Int 0
-
+    //ensures ?N ==Int 0
+    ensures true
```
ehildenb: Remove the comment.

tests/specs/benchmarks/overflow00-nooverflow-spec.k
ssaha6: I removed all the comments.

tests/specs/benchmarks/overflow00-nooverflow-spec.k
```diff
@@ -110,23 +110,21 @@ _
...
+     andBool CONTRACT_ID  >Int 0 andBool (notBool CONTRACT_ID  in #precompiledAccounts(BYZANTIUM))
+     andBool CALLEE_ID  >Int 0 andBool (notBool CALLEE_ID  in #precompiledAccounts(BYZANTIUM))
 
 andBool #rangeUInt(256, CONTRACT_BAL)
-andBool #rangeUInt(256, CALLEE_BAL)andBool #range(0 <= CD < 1024)
-andBool #rangeAddress(MSG_SENDER)andBool #range(0 <= A0 < 2 ^Int 256 -Int 1)
-andBool N ==Int A0 +Int 1
+andBool #rangeUInt(256, CALLEE_BAL) andBool #range(0 <= CD < 1024)
+andBool #rangeAddress(MSG_SENDER) andBool #range(0 <= A0 < 2 ^Int 256 -Int 1)
     ensures true
```
ehildenb, line 125: @daejunpark is there any reason to keep `ensures true`? Or should we remove that too?

daejunpark: I think this was auto-generated. But having this makes it easier to move some constraints from `requires` to `ensures` and back, because we don't need to remove `andBool` and add back.  (Indeed, that happened for the haskell backend porting, because the haskell backend strictly distinguishes `requires` from `ensures`, while the java backend does not.) So, I think it is not that bad to keep this dummy. But I'm OK with removing this, if you want.

daejunpark:
@ehildenb thanks for reviewing and helping him to fix.  Now it looks good to me.  As I said, he will make another PR that applies the systematic changes to all the spec in the benchmarks directory.  If you'd like to have that PR to merge first, let us know.  Otherwise, please merge this if you're happy with.

ehildenb:
Looks great to me too! And yes please, let's get a PR with all the systematic changes in so that it's easier to review the remaining changes.

ssaha6:
Thank you @ehildenb for your help. I will get another PR done with all the systematic changes.

PR #860, opened 2020-07-08
add lemmas for Map #lookup and update

daejunpark:
I have no idea why these lemmas have not been added earlier.  @denis-bogdanas @ehildenb any idea or history?

### DIFF
--- a/tests/specs/benchmarks/storagevar00-spec.k
+++ b/tests/specs/benchmarks/storagevar00-spec.k
@@ -69,8 +69,8 @@ module STORAGEVAR00-SPEC
             <acctID> CONTRACT_ID </acctID>
             <balance> CONTRACT_BAL </balance>
             <code> #parseByteStack("0x608060405260043610603f576000357c0100000000000000000000000000000000000000000000000000000000900463ffffffff16806361461954146044575b600080fd5b348015604f57600080fd5b506056606c565b6040518082815260200191505060405180910390f35b6000805490509056fea165627a7a7230582042a42ab720fca3a1db0ba7ac19c662cf1a6fb8fdb252593eb42de55a510e09470029") </code>
-            <storage>     #hashedLocation("Solidity", 0, .IntList) |-> N  _:Map </storage>
-            <origStorage> #hashedLocation("Solidity", 0, .IntList) |-> N  _:Map </origStorage>
+            <storage>     S [ #hashedLocation("Solidity", 0, .IntList) <- N ] </storage>
+            <origStorage> S [ #hashedLocation("Solidity", 0, .IntList) <- N ] </origStorage>
             <nonce> _ </nonce>
           </account>
 
--- a/tests/specs/benchmarks/verification.k
+++ b/tests/specs/benchmarks/verification.k
@@ -26,6 +26,9 @@ module VERIFICATION
     imports VERIFICATION-COMMON
     imports VERIFICATION-JAVA
 
+    rule ( M:Map [ K1 <- _  ] ) [ K2 <- V2 ] =>   M [ K1 <- V2 ]                requires K1 ==Int K2
+    rule ( M:Map [ K1 <- V1 ] ) [ K2 <- V2 ] => ( M [ K2 <- V2 ] ) [ K1 <- V1 ] requires K1  >Int K2
+
     rule Rsstore(BYZANTIUM, NEW, CURR, ORIG) => 0 requires NEW =/=Int 0
 
   // ########################
--- a/tests/specs/benchmarks/verification.k
+++ b/tests/specs/benchmarks/verification.k
@@ -26,8 +26,8 @@ module VERIFICATION
     imports VERIFICATION-COMMON
     imports VERIFICATION-JAVA
 
-    rule ( M:Map [ K1 <- _  ] ) [ K2 <- V2 ] =>   M [ K1 <- V2 ]                requires K1 ==Int K2
-    rule ( M:Map [ K1 <- V1 ] ) [ K2 <- V2 ] => ( M [ K2 <- V2 ] ) [ K1 <- V1 ] requires K1  >Int K2
+    rule M:Map [ K1 <- _  ] [ K2 <- V2 ] => M [ K1 <- V2 ]              requires K1 ==Int K2
+    rule M:Map [ K1 <- V1 ] [ K2 <- V2 ] => M [ K2 <- V2 ] [ K1 <- V1 ] requires K2  <Int K1
 
     rule Rsstore(BYZANTIUM, NEW, CURR, ORIG) => 0 requires NEW =/=Int 0
 
--- a/tests/specs/functional/lemmas-spec.k
+++ b/tests/specs/functional/lemmas-spec.k
@@ -23,6 +23,11 @@ module LEMMAS-SPEC
     // rule <k> runLemma ( M [ L := WS:ByteArray ] [ N := .ByteArray ]                             ) => doneLemma ( M [ L := WS ]                        ) ... </k> requires N <=Int L
     // rule <k> runLemma ( M [ 32 := (0 : 1 : 2 : .ByteArray) ] [ 32 := (3 : 4 : 5 : .ByteArray) ] ) => doneLemma ( M [ 32 := (3 : 4 : 5 : .ByteArray) ] ) ... </k>
 
+    // rule <k> runLemma ( M:Map [ I1 <- 1 ] [ I2 <- 2 ] ) => doneLemma ( M [ I1 <- 1 ] [ I2 <- 2 ] ) ... </k> requires I1  <Int I2
+    // rule <k> runLemma ( M:Map [ I1 <- 1 ] [ I2 <- 2 ] ) => doneLemma ( M [ I2 <- 2 ] [ I1 <- 1 ] ) ... </k> requires I1  >Int I2
+    // rule <k> runLemma ( M:Map [ I1 <- 1 ] [ I2 <- 2 ] ) => doneLemma ( M [ I1 <- 2 ]             ) ... </k> requires I1 ==Int I2
+    // rule <k> runLemma ( M:Map [ I1 <- 1 ] [ I2 <- 2 ] ) => doneLemma ( M [ I2 <- 2 ]             ) ... </k> requires I1 ==Int I2
+
  // #lookup simplifications
  // -----------------------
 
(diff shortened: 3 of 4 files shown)


tests/specs/lemmas.k
```diff
@@ -241,6 +241,12 @@ module LEMMAS-HASKELL [symbolic, kore]
     imports EVM
     imports EDSL
 
+    rule #lookup ( M:Map [ K1 <- V1 ] , K2 ) => V1                 requires K1 ==K  K2
+    rule #lookup ( M:Map [ K1 <- _  ] , K2 ) => #lookup ( M , K2 ) requires K1 =/=K K2
+
+    rule ( M:Map [ K1 <- _  ] ) [ K2 <- V2 ] =>   M [ K1 <- V2 ]                requires K1 ==K  K2
+    rule ( M:Map [ K1 <- V1 ] ) [ K2 <- V2 ] => ( M [ K2 <- V2 ] ) [ K1 <- V1 ] requires K1 =/=K K2
```
daejunpark, line 248: This (the last rule) could be problematic for the debugging purpose, as it could swap the order of updates (but in a sound way).  To preserve the order, we need to check if `K1 in keys(M)`, but the current haskell backend seems to have some issues regarding the reasoning about `in_keys`.

daejunpark:
I'll add tests soon.

tests/specs/lemmas.k
denis-bogdanas: 1. For first 2 rules I have this version in erc20-verifier, which is probably faster but it doesn't work, because there are bugs in Haskell:
```
  rule #lookup( M [KEY1 <- VAL:Int], KEY2 )
       => #if KEY1 ==K KEY2 #then VAL #else #lookup( M, KEY2 ) #fi                    [simplification]
```
So if you need this change now, then (a) add `V1:Int` (b) add [simplification] everywhere
. If not, then maybe it's better to postpone until Haskell gets fixed.

2. For 4th rule, looks to me like it'll cause infinite loop. Waht exactly are you verifying.

ehildenb: The last rule should be removed. I use these rules for ordering updates and pushing single-value updates inside of range updates on the MCD proofs:

```
    // Memory assignment sorting/normalization
    rule MAP [ N := .ByteArray ] => MAP
    rule MAP [ N := W : WS     ] => MAP [ N <- W ] [ N +Int 1 := WS ]

    rule MAP:Memory [ N <- V           ] [ N' <- V' ] => MAP [ N' <- V' ] [ N <- V ] requires N'                          <Int N
    rule MAP:Memory [ N := V:ByteArray ] [ N' <- V' ] => MAP [ N' <- V' ] [ N := V ] requires N'                          <Int N orBool N +Int #sizeByteArray(V) <=Int N'
    rule MAP:Memory [ N := V:ByteArray ] [ N' := V' ] => MAP [ N' := V' ] [ N := V ] requires N' +Int #sizeByteArray(V') <=Int N
    rule MAP:Memory [ N <- V           ] [ N' := V' ] => MAP [ N' := V' ]            requires N' <=Int N andBool N <Int N' +Int #sizeByteArray(V')
```

ehildenb: These rules I ahve found tend to minimize the size of the map terms that show up.

tests/specs/benchmarks/storagevar00-spec.k
```diff
@@ -69,8 +69,8 @@ module STORAGEVAR00-SPEC
             <acctID> CONTRACT_ID </acctID>
             <balance> CONTRACT_BAL </balance>
             <code> #parseByteStack("0x608060405260043610603f576000357c0100000000000000000000000000000000000000000000000000000000900463ffffffff16806361461954146044575b600080fd5b348015604f57600080fd5b506056606c565b6040518082815260200191505060405180910390f35b6000805490509056fea165627a7a7230582042a42ab720fca3a1db0ba7ac19c662cf1a6fb8fdb252593eb42de55a510e09470029") </code>
-            <storage>     #hashedLocation("Solidity", 0, .IntList) |-> N  _:Map </storage>
-            <origStorage> #hashedLocation("Solidity", 0, .IntList) |-> N  _:Map </origStorage>
+            <storage>     S [ #hashedLocation("Solidity", 0, .IntList) <- N ] </storage>
+            <origStorage> S [ #hashedLocation("Solidity", 0, .IntList) <- N ] </origStorage>
```
ehildenb, line 73: I don't think we should use this structure. I think instead it's better for composability to use:

```
<storage> S => ?S </storage>
<origStorage> OS => ?OS </origStorage>
requires #lookup(S, #hashedLocation("Solidity", 0, .IntList)) ==Int N
 andBool #lookup(OS, #hashedLocation("Solidity", 0, .IntList)) ==Int N
 ensures #lookup(S, #hashedLocation("Solidity", 0, .IntList)) ==Int N
 andBool #lookup(OS, #hashedLocation("Solidity", 0, .IntList)) ==Int N
```

tests/specs/lemmas.k
daejunpark: OK. The last rule is problematic. Thanks @denis-bogdanas. 

@denis-bogdanas Regarding your simpler lookup rule, could you elaborate more on it?  Why does it work?  Also, what's the current issue of the haskell backend?

daejunpark: @ehildenb Your rules are for the integer-key maps, right?  My intention was to have rules for arbitrary maps.  Originally I would like to add the following, but wasn't sure if it works for the haskell backend:
```
rule ( M:Map [ K1 <- V1 ] ) [ K2 <- V2 ] => ( M [ K2 <- V2 ] ) [ K1 <- V1 ] requires K1 =/=K K2 andBool K1 in_keys(M)
```

tests/specs/benchmarks/storagevar00-spec.k
daejunpark: Have you tried this before?  At least, this doesn't work for the java backend in many cases.  I'm not sure if the haskell backend would be better for this either.

daejunpark:
We discussed further in-person.  Here is the summary of what we agreed: 

1. We will try to use the builtin K lemmas (`module MAP` in [domains.k](https://github.com/kframework/k/blob/master/k-distribution/include/kframework/builtin/domains.k)) as much as possible, and introduce integer-map-specific lemmas if needed.

   BTW, we need specific lemmas for `#lookup`, since the following is not sound:
   ```
   rule #lookup(M,K) => M[K]
   ```

2. For (1), we need to rearrange the lemma structure, and make [the map-symbolic-tests](https://github.com/kframework/k/tree/master/k-distribution/tests/regression-new/map-symbolic-tests) to work with the haskell backend as well.

3. For `#lookup` we will go for the initially proposed lemmas, with fixing the sort of variables, like:
   ```
       rule #lookup ( M:Map [ K1 <- V1:Int ] , K2 ) => V1                 requires K1 ==Int  K2
       rule #lookup ( M:Map [ K1 <- _  ] , K2 ) => #lookup ( M , K2 ) requires K1 =/=Int K2
   ```
   But @denis-bogdanas told us that currently the haskell backend has some issues reasoning about equality conditions, so the above lemmas may not be applied as expected.  The `#if-#then-#else` version suggested by @denis-bogdanas would be helpful for debugging in that case.

4. For the map updates, we need to normalize the nested update terms.  For the integer maps, this can be easily done by sorting over their integer keys, as @ehildenb suggested.  For the general builtin maps, my proposal is to use the `in_keys` side condition, but we have a performance concern with that.  We will need to profile its performance overhead first.  If the general approach doesn't work, then we will add the integer map normalization lemma here.

5. The way to specify the map update in the spec that @ehildenb proposed is better if it works.  We will try that for the benchmarks proofs as well.

daejunpark:
@ehildenb I fixed the #lookup lemmas, which are needed anyway.  I also moved the update lemmas to the benchmarks/verification.k to move forward, which will be deleted once the builtin lemmas are made to work.  I added tests for both #lookup and update lemmas, but commented out for the update lemmas for now.  I'd like to merge this to make a incremental progress, if you're ok with it.

tests/specs/functional/lemmas-spec.k
```diff
@@ -23,6 +23,11 @@ module LEMMAS-SPEC
     // rule <k> runLemma ( M [ L := WS:ByteArray ] [ N := .ByteArray ]                             ) => doneLemma ( M [ L := WS ]                        ) ... </k> requires N <=Int L
     // rule <k> runLemma ( M [ 32 := (0 : 1 : 2 : .ByteArray) ] [ 32 := (3 : 4 : 5 : .ByteArray) ] ) => doneLemma ( M [ 32 := (3 : 4 : 5 : .ByteArray) ] ) ... </k>
 
+    // rule <k> runLemma ( M:Map [ I1 <- 1 ] [ I2 <- 2 ] ) => doneLemma ( M [ I1 <- 1 ] [ I2 <- 2 ] ) ... </k> requires I1  <Int I2
+    // rule <k> runLemma ( M:Map [ I1 <- 1 ] [ I2 <- 2 ] ) => doneLemma ( M [ I2 <- 2 ] [ I1 <- 1 ] ) ... </k> requires I1  >Int I2
+    // rule <k> runLemma ( M:Map [ I1 <- 1 ] [ I2 <- 2 ] ) => doneLemma ( M [ I1 <- 2 ]             ) ... </k> requires I1 ==Int I2
+    // rule <k> runLemma ( M:Map [ I1 <- 1 ] [ I2 <- 2 ] ) => doneLemma ( M [ I2 <- 2 ]             ) ... </k> requires I1 ==Int I2
```
daejunpark, line 29: BTW, the haskell backend failed to prove the last one, with the update lemmas in the benchmarks/verification.k :(

tests/specs/functional/lemmas-spec.k
```diff
@@ -36,6 +41,11 @@ module LEMMAS-SPEC
...
+    rule <k> runLemma ( #lookup( M:Map [ KEY <- 33 ] [ KEY' <- 728 ] [ KEY'' <- (pow256 +Int 5) ] [ KEY''' <- "hello" ] , KEY'   ) ) => doneLemma ( 728 ) ... </k> requires KEY =/=Int KEY' andBool KEY =/=Int KEY'' andBool KEY =/=Int KEY''' andBool KEY' =/=Int KEY'' andBool KEY' =/=Int KEY''' andBool KEY'' =/=Int KEY'''
+    // rule <k> runLemma ( #lookup( M:Map [ KEY <- 33 ] [ KEY' <- 728 ] [ KEY'' <- (pow256 +Int 5) ] [ KEY''' <- "hello" ] , KEY''  ) ) => doneLemma ( 5   ) ... </k> requires KEY =/=Int KEY' andBool KEY =/=Int KEY'' andBool KEY =/=Int KEY''' andBool KEY' =/=Int KEY'' andBool KEY' =/=Int KEY''' andBool KEY'' =/=Int KEY'''
+    // rule <k> runLemma ( #lookup( M:Map [ KEY <- 33 ] [ KEY' <- 728 ] [ KEY'' <- (pow256 +Int 5) ] [ KEY''' <- "hello" ] , KEY''' ) ) => doneLemma ( 0   ) ... </k> requires KEY =/=Int KEY' andBool KEY =/=Int KEY'' andBool KEY =/=Int KEY''' andBool KEY' =/=Int KEY'' andBool KEY' =/=Int KEY''' andBool KEY'' =/=Int KEY'''
```
daejunpark, line 47: Note that we need the explicit side conditions here.

tests/specs/benchmarks/verification.k
```diff
@@ -26,6 +26,9 @@ module VERIFICATION
     imports VERIFICATION-COMMON
     imports VERIFICATION-JAVA
 
+    rule ( M:Map [ K1 <- _  ] ) [ K2 <- V2 ] =>   M [ K1 <- V2 ]                requires K1 ==Int K2
+    rule ( M:Map [ K1 <- V1 ] ) [ K2 <- V2 ] => ( M [ K2 <- V2 ] ) [ K1 <- V1 ] requires K1  >Int K2
```
ehildenb, line 30:
```suggestion
    rule ( M:Map [ K1 <- V1 ] ) [ K2 <- V2 ] => ( M [ K2 <- V2 ] ) [ K1 <- V1 ] requires K2  <Int K1
```

Let's have all new added lemmas use `<Int` and `<=Int` instead (since it reads lowest to highest that way).
I also don't think you need the paranthesis.

tests/specs/functional/lemmas-spec.k
ehildenb: What about the first 3? If it can prove those first 3, let's add them.

tests/specs/lemmas.k
```diff
@@ -241,6 +241,9 @@ module LEMMAS-HASKELL [symbolic, kore]
     imports EVM
     imports EDSL
 
+    rule #lookup ( M:Map [ K1 <- V1 ] , K2 ) => #lookup ( K1 |-> V1 , K1 ) requires K1 ==Int  K2 [simplification]
+    rule #lookup ( M:Map [ K1 <- _  ] , K2 ) => #lookup ( M         , K2 ) requires K1 =/=Int K2 [simplification]
+
```
daejunpark, line 246: Note that the first rule relies on the original #lookup rules in the semantics, to handle the different cases, in the same way, no matter how they will be changed in the future.

ehildenb, line 246: Can these be added to just `LEMMAS` instead to be used on both Haskell and Java? Or is it only needed on Haskell?

tests/specs/functional/lemmas-spec.k
daejunpark: No, the update lemmas exist only in the benchmarks/verification.k which is not imported here.  I don't think we want to import such non-standard lemmas here, right?

tests/specs/lemmas.k
daejunpark: let me try

tests/specs/benchmarks/verification.k
daejunpark: done

tests/specs/lemmas.k
daejunpark: done. see if tests pass

tests/specs/functional/lemmas-spec.k
ehildenb: Ah Ok.

tests/specs/benchmarks/verification.k
```diff
@@ -26,6 +26,9 @@ module VERIFICATION
     imports VERIFICATION-COMMON
     imports VERIFICATION-JAVA
 
+    rule M:Map [ K1 <- _  ] [ K2 <- V2 ] => M [ K1 <- V2 ]              requires K1 ==Int K2
+    rule M:Map [ K1 <- V1 ] [ K2 <- V2 ] => M [ K2 <- V2 ] [ K1 <- V1 ] requires K2  <Int K1
+
```
daejunpark, line 31: @ehildenb @denis-bogdanas I realized that this doesn't work.  Since the symbolic map is used only for the storage for the haskell backend version of KEVM, and the storage keys are the hashed location, whose ordering is unclear.  So, I think we can just keep the nested map update terms without simplifying, because usually the storage is not much updated.  Or, I can bring the simplification lemmas used for the java backend here and profile the performance overhead.  Anyway, this is the custom lemmas for the benchmarks proof, so the effect should be minimal.

denis-bogdanas: It might be worth benchmarking performance with those lemmas. If benchmarks become faster, chances are that ERC20 will too.

ehildenb: Well we can also just leave them here in the benchmarks proofs for now.

ehildenb: But these lemmas _will_ help with local memory, if you change it from `M:Map` to `M:Memory`. Then `_[_<-_]` is interpreted at Map-update on Java backend, and Bytes-update on Haskell backend.

daejunpark: Oh, that's right. In that case, it would be better to replace this with yours:
https://github.com/kframework/evm-semantics/pull/860#discussion_r451641065

We will do that when we get there (probably other benchmarks proofs would need that.)

ehildenb: I will need those lemmas for the MCD proofs, once I'm ready to merge those.

PR #852, opened 2020-07-01
evm-types: functional #lookup, #lookupMemory. Version with modInt

denis-bogdanas:

PR #861, opened 2020-07-08
fix invalid #lookup.* rule names in concrete-rules.txt

daejunpark:
Fix a bug introduced at https://github.com/kframework/evm-semantics/commit/223f21ce46625298962413953e388f5eccfefd58#diff-558a5f8e4337c04713ebea82c99dde22R19-R20

### DIFF
--- a/tests/specs/concrete-rules.txt
+++ b/tests/specs/concrete-rules.txt
@@ -5,8 +5,6 @@ EVM.Cgascap
 EVM.Cmem
 EVM.Csstore.new
 EVM.Csstore.old
-EVM-TYPES.#lookup.none
-EVM-TYPES.#lookup.some
 EVM.#memoryUsageUpdate.some
 EVM.Rsstore.new
 EVM.Rsstore.old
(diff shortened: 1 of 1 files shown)


daejunpark:
@denis-bogdanas @ehildenb I made this as a separate PR, to see if this doesn't break the existing proofs.

ehildenb:
@daejunpark I'm pretty sure the file `tests/specs/concrete-rules.txt` isn't used by _any_ of the proofs. Can you just remove it?

The `Makefile` currently does this:

```
tests/%.prove: tests/%
    $(TEST) prove $(TEST_OPTIONS) --backend $(TEST_SYMBOLIC_BACKEND) $< $(KPROVE_MODULE) --format-failures $(KPROVE_OPTS) \
        --concrete-rules $(shell cat $(dir $@)concrete-rules.txt | tr '\n' ',')
```

which only looks for `concrete-rules.txt` next to the current proof, not at this global file.

ehildenb:
Ah, I just saw that this file is symlinked from many proofs, which is why it's there.

@daejunpark since this doesn't seem to break the proofs, can you just _remove_ those lines from `concrete-rules.txt` instead? I would like that file to be as small as possible, and it seems we discovered a place where we can make it smaller.

denis-bogdanas:
Those 2 lines must be removed. They don't help with #852

daejunpark:
@denis-bogdanas Ah, I misunderstood that you mentioned it could help.  OK I just removed them.

@ehildenb I agree with you.

PR #869, opened 2020-07-14
Check upfront cost for transactions in web3 client

gtrepta:
fixes: https://github.com/runtimeverification/firefly/issues/960

This PR adds an `EVMC_BALANCE_UNDERFLOW` exception case to the `#validateTx` rule that triggers when the sender account's balance doesn't meet the upfront cost for a transaction.


web3.md
```diff
@@ -1255,28 +1256,44 @@ Transaction Execution
...
            <msgID>      TXID   </msgID>
+           <txGasPrice> GPRICE </txGasPrice>
            <txGasLimit> GLIMIT </txGasLimit>
            <data>       DATA   </data>
            <to>         ACCTTO </to>
            ...
          </message>
-      requires ( GLIMIT -Int G0(SCHED, DATA, (ACCTTO ==K .Account)) ) <Int 0
+      requires GLIMIT -Int G0(SCHED, DATA, (ACCTTO ==K .Account)) <Int 0
+        orBool BAL <Int GLIMIT *Int GPRICE
```
ehildenb, line 1276: I didn't catch this before, but we should likely do instead `<k> #validateTx TXID => #if BAL <Int GLIMIT *Int GPRICE #then #end EVMC_BALANCE_UNDERFLOW #else #end EVMC_OUT_OF_GAS #fi ... </k>`

`#end ...` is used to halt execution and signal an exception.

gtrepta: That will leave a `#halt` at the top of the cell which I'm pretty sure will cause a problem, as it will do `<k> #halt ~> (_:Int => .) ... </k>`, and we have `TXID:Int` in the k cell for `#eth_sendTransaction_final`.

PR #870, opened 2020-07-15
Some more functional marked total, add --bug-report option

ehildenb:
The function `#sizeOfDynamicType` gets some weird behavior on static types, to become total.

This makes a lot of the functions that `#encodeArgs` uses total, but unfortunately because of the nature of `getValue` we can't make `#enc` total, so we can't make `#encodeArgs` total.

### DIFF
--- a/edsl.md
+++ b/edsl.md
@@ -137,8 +137,8 @@ where `F1 : F2 : F3 : F4` is the (two's complement) byte-array representation of
     rule #isStaticType(   #string( _ )) => false
     rule #isStaticType(#array(_, _, _)) => false
 
-    syntax Int ::= #sizeOfDynamicType ( TypedArg ) [function, functional]
- // ---------------------------------------------------------------------
+    syntax Int ::= #sizeOfDynamicType ( TypedArg ) [function]
+ // ---------------------------------------------------------
     rule #sizeOfDynamicType(#bytes(BS)) => 32 +Int #ceil32(#sizeByteArray(BS))
 
     rule #sizeOfDynamicType(#array(T, N, _)) => 32 *Int (1 +Int N)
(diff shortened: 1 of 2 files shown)


daejunpark:
Do we need to make `#sizeOfDynamicType` functional?  As you said, `#enc` cannot be functional anyway, and `#sizeOfDynamicType` is used only inside `#inc`.

daejunpark:
How about the following functions?

- #signatureCallData
- #generateSignature
- #generateSignatureArgs
- #isStaticType

They should be functional, right?

Also, this PR is only for the "ABI Call Data" section?

daejunpark:
NOTE: Although it is better to make any total function [functional] for the readability and consistency, only `#ceil` function should benefit from the functional attribute, because all the other functions should be fully evaluated, that is, those function symbols should not appear in the configuration terms during the symbolic execution.

ehildenb:
@daejunpark I've removed the extra cases from `#sizeOfDynamicType*` cases, and removed `functional`. I have _not_ marked `#signatureCallData` as `functional`, because it uses `Keccak256` which is not marked functional, and I don't know if it can be marked functional (we would have to consult the backend hooked implementations).

I agree that it's only important to mark ones which get `#Ceil` conditions during proving as `functional`, but if we take the time to mark things `functional` when they are total, it makes it easier down the road to see if a functions dependencies are all total or not.

edsl.md
```diff
@@ -152,7 +152,7 @@ where `F1 : F2 : F3 : F4` is the (two's complement) byte-array representation of
     rule #sizeOfDynamicTypeAux(TARG, TARGS) => #sizeOfDynamicType(TARG) +Int #sizeOfDynamicTypeAux(TARGS)
       requires notBool #isStaticType(TARG)
 
-    rule #sizeOfDynamicTypeAux(.TypedArgs) => 0
+    rule #sizeOfDynamicTypeAux(.TypedArg) => 0
```
daejunpark, line 155: typo?

ehildenb:
```suggestion
    rule #sizeOfDynamicTypeAux(.TypedArgs) => 0
```

PR #871, opened 2020-07-15
Less failure prone eth_estimateGas

gtrepta:
fixes: https://github.com/runtimeverification/firefly/issues/919

How this PR works (so far):
- A `#binSearchGas( Int, Int )` KItem is used, where the two integers represent the low and high numbers in a binary search for the gas amount.
- The transaction is ran with `mid` ((LO + HI) / 2) as the gas limit.
- If the transaction succeeds, then HI is set to `mid`
- If the transaction fails, then LO is set to `mid`
- Once LO and HI have a difference smaller than or equal to one, the transaction is finally ran with HI as the gas limit
- If the transaction succeeds, then HI is returned, otherwise an error is reported.


ehildenb:
Last we discussed, we found that we are still calculating the state trie (mining the block) after `eth_call` and `eth_estimateGas` and computing `G0` before (even though it's the same each time). Both of these are redundant and expensive, so we should not do these more than once (or zero) times as appropriate.

tests/web3/eth_estimateGas2.expected.json
```diff
@@ -12,6 +12,6 @@
   {
     "jsonrpc": "2.0",
     "id": 3,
-    "result": "0x5659"
+    "result": "0xd8cb"
```
ehildenb, line 15: Can you post here what ganache gets on these examples so we have an idea of the difference?

gtrepta: I can't get ganache's output for `eth_estimateGas2.in.json` because it's using a very specific `firefly_addAccount` call. The output ganache gives for the estimateGas call in `eth_estimateGas3.in.json` is `0x5208`, as opposed to firefly's `0xd8cb`.

ehildenb: 0x5208 is way lower than 0xd8cb. Why the huge discrepancy?

web3.md
```diff
@@ -1248,16 +1248,33 @@ Transaction Execution
...
+
+    syntax KItem ::= "#setup_G0" Int
+ // --------------------------------
+    rule <k> #setup_G0 TXID => . ... </k>
+         <schedule> SCHED </schedule>
+         <callGas> _ => G0(SCHED, DATA, (ACCTTO ==K .Account) ) </callGas>
+         <message>
+           <msgID>      TXID   </msgID>
+           <data>       DATA   </data>
+           <to>         ACCTTO </to>
```
ehildenb, line 1269:
```suggestion
           <msgID> TXID   </msgID>
           <data>  DATA   </data>
           <to>    ACCTTO </to>
```

tests/web3/eth_estimateGas2.expected.json
gtrepta: Because the binary search is being cut short by the first allotment of gas that results in a successful execution of the transaction, which is the amount that gets returned. Allowing the binary search to finish completely wasn't being reliable and was sometimes returning gas amounts that didn't allow the transaction to complete, presumably because there is some sort of error in the algorithm I wrote.

PR #872, opened 2020-07-16
lemmas.k: lemmas for ERC20 verifier

denis-bogdanas:


tests/specs/lemmas.k
```diff
@@ -9,6 +9,9 @@ module LEMMAS
     rule #lookup ( M:Map [ K1 <- V1 ] , K2 ) => #lookup ( K1 |-> V1 , K1 ) requires K1 ==Int  K2 [simplification]
     rule #lookup ( M:Map [ K1 <- _  ] , K2 ) => #lookup ( M         , K2 ) requires K1 =/=Int K2 [simplification]
 
+    rule 0 <=Int #lookup( _M:Map , _ )             => true                                       [simplification]
+    rule         #lookup( _M:Map , _ ) <Int pow256 => true                                       [simplification, smt-lemma]
```
ehildenb, line 13: Why not `smt-lemma` on both?

tests/specs/lemmas.k
```diff
@@ -244,6 +247,12 @@ module LEMMAS-HASKELL [symbolic, kore]
     imports EVM
     imports EDSL
 
+    //Simplification of double negation in term constraint
+    rule (B ==Bool false) ==Bool false => B                                                              [simplification]
+    //Simplifies { false #Equals EXP ==Bool false }
+    //see https://github.com/kframework/kore/issues/1944#issuecomment-657733745
+    rule  B ==Bool false               => notBool B                                                      [simplification]
+
```
ehildenb, line 255: These rules will have a race condition. `(X ==Bool false) ==Bool false` may simplify down to either `B` (first rule), or to `notBool (X ==Bool false)`.

I think you should keep the second rule, and add a rule which says `notBool (X ==Bool false) => X`.

tests/specs/lemmas.k
```diff
@@ -269,6 +278,9 @@ module LEMMAS-HASKELL [symbolic, kore]
     rule maxUInt160 &Int N => N
       requires #rangeUInt(160, N) [simplification]
 
+    rule N modInt pow160 => N
+      requires #rangeUInt(160, N) [simplification]
+
```
ehildenb, line 283: Any reason for this to be in `LEMMAS-HASKELL` and not just `LEMMAS`?

tests/specs/lemmas.k
denis-bogdanas: Because I only need it for the 2nd. Performance differs I guess.

tests/specs/lemmas.k
denis-bogdanas: closer to its brothers above : D

denis-bogdanas:
@ehildenb What exactly is to test here? Rules except bool simplification are trivial. For Bool simplification I guess I'll add some tests.

tests/specs/lemmas.k
ehildenb: The rules shouldn't have a race on their own, even if the Haskell backend is doing some magic (which it probably shouldn't be doing, but I'm not sure what magic you refer to).

Please make the rules presented here confluent, so that they work independent of what's going on in the backend.

ehildenb:
Tests on the bool simplification would be good then. And please make the above rules I left unresolved confluent.

PR #873, opened 2020-07-17
evm-types: switched remaining UInt ranges to #range(0 <= X < powXX)

denis-bogdanas:


evm-types.md
```diff
@@ -95,16 +99,16 @@ These can be used for pattern-matching on the LHS of rules as well (`macro` attr
...
+    rule #rangeUInt    (  16 ,      X ) => #range ( minUInt16       <= X <  pow16           ) [macro]
+    rule #rangeUInt    (  48 ,      X ) => #range ( minUInt48       <= X <  pow48           ) [macro]
+    rule #rangeUInt    ( 128 ,      X ) => #range ( minUInt128      <= X <  pow128          ) [macro]
     rule #rangeUInt    ( 160 ,      X ) => #range ( minUInt160      <= X <  pow160          ) [macro]
     rule #rangeUInt    ( 256 ,      X ) => #range ( minUInt256      <= X <  pow256          ) [macro]
     rule #rangeSFixed  ( 128 , 10 , X ) => #range ( minSFixed128x10 <= X <= maxSFixed128x10 ) [macro]
     rule #rangeUFixed  ( 128 , 10 , X ) => #range ( minUFixed128x10 <= X <= maxUFixed128x10 ) [macro]
     rule #rangeAddress (            X ) => #range ( minUInt160      <= X <  pow160          ) [macro]
-    rule #rangeBytes   (   N ,      X ) => #range ( 0               <= X <= #nBytes(N)      ) [macro]
+    rule #rangeBytes   (   N ,      X ) => #range ( 0               <= X <  1 <<Byte N      ) [macro]
```
ehildenb, line 111: @denis-bogdanas why this change?

daejunpark: +1

daejunpark:
Looks good to me, other than the above change.

evm-types.md
denis-bogdanas: That's the only change that matters actually. `#Ceil(#buf)` uses `#rangeBytes`, while other range lemmas are all `X < pow256`. So this change is needed to avoid invoking z3.

daejunpark: @denis-bogdanas Hmm, I'm not sure why this makes the difference.

It's just two rule applications vs. single rule application:
```
    rule #nBits(N)  => (1 <<Int N) -Int 1 requires N >=Int 0
    rule #nBytes(N) => #nBits(N *Int 8)   requires N >=Int 0
    rule N <<Byte M => N <<Int (8 *Int M)
```

BTW, @ehildenb why `#nBytes` requires the positivity, but `<<Byte` does not?

denis-bogdanas: Problem is not in the number of rule applications, but in getting the `#rangeBytes` proved through other lemmas, The other lemmas match the new `#rangeBytes` better.

The rule that uses it is:
```
rule #Ceil(#buf(SIZE, DATA)) => {(0 <=Int SIZE) andBool #rangeBytes(SIZE, DATA) #Equals true}  [anywhere, simplification]
```

and it has to match rules like this:
```
rule         keccak( _ ) <Int pow256 => true                                                            [simplification, smt-lemma]
```

ehildenb: Well, `<<Bytes` does not require positivity because `_ <<Byte - N` is the same as `_ >>Byte N` (it gets this property from `<<Int`. At least, that's what I remember.

So maybe it actually matters that we use `#nBytes` to ensure positivity?

denis-bogdanas: All these always true side conditions affect performance. If `N` is negative then range will be vacuous and it will be as if the whole term evaluated to `Bottom`.

ehildenb: I won't merge something incorrect for performance, so when we're talking about something's correctness, please refrain from making arguments about why we should merge it anyway for performance.

In this case, I think this one actually doesn't affect correctness, because it will just make a vacuously `false` (not `#Bottom`) result. The only danger here is that the user supplies a negative number to `#rangeBytes` in the side-condition of a proof, which results in the whole proof being vacuously true.

denis-bogdanas: @daejunpark approved earlier. Can we add automerge?

ehildenb: Well @daejunpark 's review says "other than the above change", so if he's satisfied with this now, he can put `automerge`.

PR #877, opened 2020-07-22
Fix simplification rules for #Ceil(#buf)

ttuegel:
This pull request fixes two problems with the simplification rules for `#Ceil(#buf)`:

1. The rule in `lemmas.k` conflicts with the rule in `edsl.md`.
1. The rule in `edsl.md` cannot be applied in most cases.

### DIFF
--- a/edsl.md
+++ b/edsl.md
@@ -194,7 +194,7 @@ where `F1 : F2 : F3 : F4` is the (two's complement) byte-array representation of
       [anywhere, simplification]
 ```
 
-```
+```k
     syntax Int ::= #getValue ( TypedArg ) [function]
  // ------------------------------------------------
     rule #getValue(#uint160( DATA )) => DATA
--- a/edsl.md
+++ b/edsl.md
@@ -184,13 +184,17 @@ where `F1 : F2 : F3 : F4` is the (two's complement) byte-array representation of
     rule #buf(SIZE, DATA) => #padToWidth(SIZE, #asByteStack(DATA))
       requires #range(0 <= DATA < (2 ^Int (SIZE *Int 8)))
       [concrete]
+```
 
+``` {.k .symbolic .ceil}
     rule #Ceil(#buf(@SIZE, @DATA))
       =>      {(0 <=Int @SIZE) andBool #rangeBytes(@SIZE, @DATA) #Equals true}
          #And #Ceil(@SIZE)
          #And #Ceil(@DATA)
       [anywhere, simplification]
+```
 
+```
     syntax Int ::= #getValue ( TypedArg ) [function]
  // ------------------------------------------------
     rule #getValue(#uint160( DATA )) => DATA
--- a/edsl.md
+++ b/edsl.md
@@ -185,9 +185,10 @@ where `F1 : F2 : F3 : F4` is the (two's complement) byte-array representation of
       requires #range(0 <= DATA < (2 ^Int (SIZE *Int 8)))
       [concrete]
 
-    rule #Ceil(#buf(@SIZE, @DATA)) =>
-      {(0 <=Int @SIZE) andBool #rangeBytes(@SIZE, @DATA) #Equals true}
-      #And #Ceil(@SIZE) #And #Ceil(@DATA)
+    rule #Ceil(#buf(@SIZE, @DATA))
+      =>      {(0 <=Int @SIZE) andBool #rangeBytes(@SIZE, @DATA) #Equals true}
+         #And #Ceil(@SIZE)
+         #And #Ceil(@DATA)
       [anywhere, simplification]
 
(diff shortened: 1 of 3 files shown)


edsl.md
```diff
@@ -185,7 +185,10 @@ where `F1 : F2 : F3 : F4` is the (two's complement) byte-array representation of
       requires #range(0 <= DATA < (2 ^Int (SIZE *Int 8)))
       [concrete]
 
-    rule #Ceil(#buf(SIZE, DATA)) => {(0 <=Int SIZE) andBool #rangeBytes(SIZE, DATA) #Equals true}  [anywhere, simplification]
+    rule #Ceil(#buf(@SIZE, @DATA)) =>
+      {(0 <=Int @SIZE) andBool #rangeBytes(@SIZE, @DATA) #Equals true}
+      #And #Ceil(@SIZE) #And #Ceil(@DATA)
+      [anywhere, simplification]
```
ehildenb, line 191:
```suggestion
    rule #Ceil(#buf(@SIZE, @DATA))
      =>      {(0 <=Int @SIZE) andBool #rangeBytes(@SIZE, @DATA) #Equals true}
         #And #Ceil(@SIZE)
         #And #Ceil(@DATA)
      [anywhere, simplification]
```

denis-bogdanas: @ttuegel LGTM. Please add automerge after fixing formatting.

edsl.md
```diff
@@ -185,7 +185,11 @@ where `F1 : F2 : F3 : F4` is the (two's complement) byte-array representation of
       requires #range(0 <= DATA < (2 ^Int (SIZE *Int 8)))
       [concrete]
 
-    rule #Ceil(#buf(SIZE, DATA)) => {(0 <=Int SIZE) andBool #rangeBytes(SIZE, DATA) #Equals true}  [anywhere, simplification]
+    rule #Ceil(#buf(@SIZE, @DATA))
```
denis-bogdanas, line 188: By the way, how is `@SIZE` different from `SIZE`?

ehildenb:
Hmmm, I wonder why it passed once but not the second time because of teh `#And` symbol which shouldn't be fed to Java backend...

edsl.md
ttuegel: `SIZE` is an element (singular) variable; the rule with `SIZE` will only match a term where the first argument of `#buf` is known to be defined. `@SIZE` is a set variable, so the rule will match any arguments, and check the `#Ceil` later. It's usually preferred to write simplification rules with set variables.

ehildenb:
@ttuegel maybe we should either (i) put this rule behind the `tangle_haskell` tag, or (ii) move it to `LEMMAS-HASKELL`. I mildly prefer (i) because then it keeps the definedness condition close to the definition of `#buf`, but otherwise have little preference.

denis-bogdanas:
It's very strange why it fails. Java backend doesn't complain about `#Ceil` but complains about `#And` which is defined in the same module `ML-SYNTAX`.

ttuegel:
> put this rule behind the `tangle_haskell` tag

@ehildenb I tried to do that, but I don't understand how `tangle_haskell` works. Can you advise?

edsl.md
```diff
@@ -184,9 +184,17 @@ where `F1 : F2 : F3 : F4` is the (two's complement) byte-array representation of
...
-    rule #Ceil(#buf(SIZE, DATA)) => {(0 <=Int SIZE) andBool #rangeBytes(SIZE, DATA) #Equals true}  [anywhere, simplification]
+``` {.k .symbolic .ceil}
+    rule #Ceil(#buf(@SIZE, @DATA))
+      =>      {(0 <=Int @SIZE) andBool #rangeBytes(@SIZE, @DATA) #Equals true}
+         #And #Ceil(@SIZE)
+         #And #Ceil(@DATA)
+      [anywhere, simplification]
+```
 
+```
```
ehildenb, line 197:
```suggestion
```k
```

ehildenb:
@ttuegel it looks like there may have been a simple typo in your update, but I think your solution worked. I checked with `make build-haskell HASKELL_KOMPILE_OPTS=-E haskell_main_file=edsl.md haskell_main_module=EDSL | less` and looked for `#Ceil`.

Then I checked `make build-java JAVA_KOMPILE_OPTS=-E java_main_file=edsl.md java_main_module=EDSL | less` and `make build-llvm KOMPILE_OPTS=-E llvm_main_file=edsl.md llvm_main_module=EDSL | less` to make sure it wasn't there.

ehildenb:
Ahhhh, wait a second though. This still won't work because that file is not included in `kompile`, it's included in `kprove`. `kprove` doesn't understand the tangle tag selectors...

PR #882, opened 2020-07-30
Enable benchmark/storagevar proofs on Haskell backend

ehildenb:
This largely takes two changes to the proofs:

-   Inline constants and expressions for integers when/where possible.
-   Use the `requires` clause for specifying pre-conditions on the storage, so that we are not committing to any particular structure for the pre-state, just that certain values are present.
-   Use the sequence of writes approach for post-conditions on the storage, which makes showing that (i) the storage is not changed except for some values, and (ii) the sequence of writes happen in a specific order, easy to show.


ehildenb:
storagevar02-nooverflow cannot be enabled on the Haskell backend, because of this issue: https://github.com/kframework/kore/issues/2032

tests/specs/benchmarks/verification.k
```diff
@@ -298,6 +298,19 @@ rule #ecrecData(SIGS, I, DATA_HASH) => #encodeArgs(#bytes32(DATA_HASH), #uint8(#
...
+    imports VERIFICATION-COMMON
+
+    syntax Int ::= "#infGas" [function, functional]
+ // -----------------------------------------------
+    rule #gas(_, _, _) => #infGas
+
+    rule #infGas  <Int _ => false
+    rule #infGas >=Int _ => true
+
+    rule #infGas -Int _ => #infGas
```
daejunpark, line 311: `#infGas -Int #infGas` is not necessarily `#infGas`.

The inequality rules above are similarly problematic too.

To be safe, we need to ensure that `#infGas` does not appear in the other side.  That was done using K-REFLECTION in the java backend, but I have no good idea how to do in the haskell backend.

tests/specs/benchmarks/storagevar01-spec.k
```diff
@@ -3,15 +3,15 @@ requires "verification.k"
 module STORAGEVAR01-SPEC
   imports VERIFICATION
 
-  // fn-execute 
+  // fn-execute
   rule
     <k> (#execute => #halt) ~> _ </k>
     <exit-code> 1 </exit-code>
     <mode> NORMAL </mode>
     <schedule> BYZANTIUM </schedule>
```
daejunpark, line 11: Please update this to ISTANBUL, which was part of the initial goal.

tests/specs/benchmarks/verification.k
ehildenb: Concrete initial gas didn't end up working out because the haskell backend could not prove that `NNNNNNNNN <Int Csstore(_, _, _, _)` for example. But maybe we can add lemmas for that.

By the way, does `rule #infGas <Int G => false requires G =/=Int #infGas` not work for some reason?

daejunpark: Not including Csstore in the concrete-rules.txt, z3 should be able to prove it, if this issue (https://github.com/kframework/kore/issues/1962) was properly fixed.

daejunpark: what If G is foo(#infGas)?

tests/specs/benchmarks/storagevar01-spec.k
ehildenb: Done.

tests/specs/benchmarks/verification.k
ehildenb: Ok, instead I've only simplified it in cases where (i) it's needed and (ii) it's safe.

ehildenb:
@daejunpark please re-review. I've removed the notion of `#infGas`, instead preferring one where we say it's higher than any concrete value, and also higher than any `Csstore`. In time, we will likely also have to say it's higher than things like `Cgascap` for instance, but for these proofs it was enough. I figure this is OK because it's just in the benchmarks/verification.k file for now.

PR #884, opened 2020-08-03
Make benchmarks/bytes00 pass on haskell backend

ehildenb:
This makes the `bytes00` proof pass on the Haskell backend, but it takes roughly 5 hours on my machine, so should not be merged yet.

### DIFF
--- a/kevm
+++ b/kevm
@@ -52,16 +52,12 @@ run_kast() {
 }
 
 run_prove() {
-    local def_module run_dir additional_proof_args repl_cmd repl_script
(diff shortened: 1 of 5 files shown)


daejunpark:
Could you elaborate on " kevm: remove manual --repl in favor of frontend supported --debugger --debug-script"? Could you provide the equivalent command of `kevm --repl ...`?

ehildenb:
I pulled out non-semantic changes to this PR: https://github.com/kframework/evm-semantics/pull/887

PR #886, opened 2020-08-03
Upstream KEVM hook code

sskeirik:
This PR is mostly straightforward. However, it introduces two issues to be resolved:

### Pulling source files from submodules

However, there is a question about the best way to point to the `krypto.md` file that is upstream. There are, in my mind, two candidates:

1. Copy it from the plugin subdir into the standard location (this is what I have done currently).
2. Update the Makefile to point to the new source location.

Of course, eventually, `krypto.md` will disappear and be replaced by smaller source files --- however, the issue of how to import those still remains the same.

### Updating/documenting the build process

Also, as far as I can tell, this upstreaming means that: *all* KEVM builds will need to clone the plugin submodule to obtain the needed syntax for these operations, even if they aren't hooked.

Since submodules aren't cloned by default, we either need to:

1. instruct users to clone the submodule before building
2. update the Makefile to pull it for them automatically

I haven't done either of these yet, because I wasn't sure what was preferred.


ehildenb:
It's assumed that users know to clone submodules for building KEVM. We can add it to the README, but I've been intentionally removing that functionality from the Makefile, because the Makefile should work largely independently of the VCS used.

sskeirik:
@ehildenb So, I took a look into the README, and found that it already instructs the user to clone the plugin submodule before building anything. Thus, for issue 2 above, we are already covered.

If you have any thoughts about the best way to solve issue 1, I'd be happy to implement that.

Makefile
```diff
@@ -94,7 +95,9 @@ K_JAR := $(K_SUBMODULE)/k-distribution/target/release/k/lib/java/kernel-1.0-SNAP
 deps: repo-deps
 repo-deps: k-deps plugin-deps
 k-deps: $(K_JAR)
-plugin-deps: $(PLUGIN_SUBMODULE)/client-c/main.cpp
+plugin-deps: $(PLUGIN_SUBMODULE)/client-c/main.cpp $(wildcard $(PLUGIN_SUBMODULE)/plugin/*)
```
ehildenb, line 98: Why the added dependent (the `$(wildcard ...)` one)?

Makefile
```diff
@@ -22,6 +22,7 @@ K_LIB := $(K_RELEASE)/lib/kframework
 export K_RELEASE
 
 LIBRARY_PATH       := $(LOCAL_LIB)
+K_INCLUDE_PATH     := $(BUILD_LOCAL)/include/k
```
ehildenb, line 25: Is there a technical reason to have this at `include/k`? I think we should just have it at `include`, because then people can just do `requires "krypto.md"` from a global install, instead of having to do `requires "k/krypto.md"`, if they have `-I /usr/bin/include/kframework` set.

Though, our global install of K puts files at `/usr/bin/include/kframework`, so maybe we should install to `$(BUILD_LOCAL)/include/kframework/blockchain-k-plugin/krypto.md`, then do `K_INCLUDE_PATH=$(BUILD_LOCAL)/include/kframework`, and do `requires "blockchain-k-plugin/krypto.md"` in our files. This seems most consistent with (i) how a global K install will work, and (ii) some concept of separate "packages" you can install with different K includes in each.

@dwightguth thoughts? I would like to use this to setup some notion of global packages, with the hope of one day, for example, having the Firefly build able to just depend on a globally installed KEVM instead of a submodule.

dwightguth: Logically, if you want to install a k package globally, you would want to install it to `/usr/local/include/kframework`. It's fairly normal for includes to use directory names as namespaces, so presumably you would install and require under a subdirectory. Are you sure you don't have a typo in your comment, though? The path `/usr/bin/include` seems incredibly sketchy to me.

Makefile
sskeirik: The idea is that building the semantics depends on the K source files inside the blockchain plugin, i.e. everything under the `plugin/` subdir.

Makefile
sskeirik: I updated the `requires` in the source files to have the form: `requires "blockchain-k-plugin/krypto.md"` as mentioned above.

I also updated the makefile for the blockchain plugin to install inside the `blockchain-k-plugin` subdir at the installation prefix.

sskeirik:
This PR is superceded by #917.

PR #887, opened 2020-08-04
Updates to --repl and formatting on benchmarks

ehildenb:
This PR does 4 things:

-   Change the `--repl` option of KEVM to use the now supported `--debugger --debug-script` options of the K frontend.
-   Formats all the `benchmarks` specs.
-   Fixes warnings about anonymous variables in all the `benchmarks/` specs.
-   Formats the `lemmas` specifications to have `<k>` tags around the claims.

### DIFF
--- a/kevm
+++ b/kevm
@@ -52,16 +52,13 @@ run_kast() {
 }
 
 run_prove() {
-    local def_module run_dir additional_proof_args repl_cmd repl_script
+    local def_module run_dir additional_proof_args
 
     def_module="$1" ; shift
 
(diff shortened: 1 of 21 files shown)


daejunpark:
What is an alternative way to modify the value of `--haskell-backend-command` in the --repl option?

ehildenb:
@daejunpark there is no such way, it's now considered a bug in the K frontend to specify both.

ehildenb:
If you really need to, you can just do `./kevm prove --backend haskell test/specs/prov-spec.k VERIFICATION --haskell-backend--comand "kore-exec --repl --repl-script kast.kscript --some-other-option"`.

PR #889, opened 2020-08-06
Passing Benchmark Tests

ehildenb:
With these changes, now the following tests are passing:

-   `requires01-a0gt0`
-   `requires01-a0le0`
-   `staticloop00-a0lt10`
-   `staticarray00`: not included in this PR because the performance is too bad, passes in 0.43 hrs.
-   `bytes00`: not included in this PR because the performance is too bad, passes in 4.72 hrs.

### DIFF
--- a/tests/specs/benchmarks/verification.k
+++ b/tests/specs/benchmarks/verification.k
@@ -211,6 +211,10 @@ module VERIFICATION
 
     rule #ceil32(X) => X requires X modInt 32 ==Int 0
 
+    rule 0 <=Int #ceil32(X) -Int X => true requires 0 <=Int X [simplification]
+
+    rule 0 <Int 1 <<Int N => true requires 0 <=Int N [simplification]
+
     rule X <=Int #ceil32(X) => true
       requires X >=Int 0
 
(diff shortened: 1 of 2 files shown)


tests/specs/benchmarks/verification.k
```diff
@@ -211,6 +211,10 @@ module VERIFICATION
 
     rule #ceil32(X) => X requires X modInt 32 ==Int 0
 
+    rule 0 <=Int #ceil32(X) -Int X => true requires 0 <=Int X [simplification]
```
denis-bogdanas, line 214: equivalent to 2 lemmas below. Lemma below likely just requires [simplification]

tests/specs/benchmarks/verification.k
```diff
@@ -307,12 +311,37 @@ module VERIFICATION-HASKELL [symbolic, kore]
 
     rule #gas(G) -Int G' => #gas(G +Int G') [simplification]
 
+    rule #gas(_)  <Int I => false requires I <=Int pow256 [simplification]
```
denis-bogdanas, line 314: Put this and all related into a new section "Infinite gas assumption"

ehildenb: I'm not sure it makes sense to do this. We currently have infinite gas assumption for Java backend, it's just done using the modified `#deductGas` rules instead of using this technique, so I don't see why it's different.

These are also the only lemmas existing in the `VERIFICATION-HASKELL`, so I'm not sure a separate section makes sense. And this is only in the `benchmarks/verification.k`, so it's not affecting any other proofs.

tests/specs/benchmarks/verification.k
ehildenb: Which lemmas specifically? I don't see them.

tests/specs/benchmarks/verification.k
denis-bogdanas: It's not particularly important, but it doesn't hurt adding a comment explaining the reason for these lemmas. This sort of comments is one of the documentation forms lacking in RV. But I won't hold the review on this. Just a recommendation.

tests/specs/benchmarks/verification.k
```diff
@@ -211,6 +211,10 @@ module VERIFICATION
 
     rule #ceil32(X) => X requires X modInt 32 ==Int 0
 
+    rule 0 <=Int #ceil32(X) -Int X => true requires 0 <=Int X [simplification]
+
+    rule 0 <Int 1 <<Int N => true requires 0 <=Int N [simplification]
+
     rule X <=Int #ceil32(X) => true
```
denis-bogdanas, line 218: this one

tests/specs/benchmarks/verification.k
```diff
@@ -211,6 +211,10 @@ module VERIFICATION
...
       requires X >=Int 0
```
denis-bogdanas, line 219: @ehildenb this one

tests/specs/benchmarks/verification.k
ehildenb: Ah ok, I thought you meant a separate module. I've added a comment explaining how it works, and the benefits.

tests/specs/benchmarks/verification.k
ehildenb: I'm trying with the other lemma removed and this one marked `simplification`, but I don't think it will work, because it's specifically to simplify out the form `0 <=Int #ceil32(X) -Int X`, not for simplifying out `X <=Int #ceil32(X)`. But if it works I'll push that up.

tests/specs/benchmarks/verification.k
ehildenb: @denis-bogdanas I've done as you say here, and at least for the proofs here, they still pass with your suggested change, so I've pushed it up.

PR #892, opened 2020-08-11
MKR regression tests: vat-move-diff

ehildenb:
This adds a regression test supplied by MKR.


tests/specs/mcd/vat-move-diff-spec.k
```diff
@@ -0,0 +1,212 @@
...
+          <activeAccounts> SetItem(ACCT_ID)
+          SetItem(1)
+          SetItem(2)
+          SetItem(3)
+          SetItem(4)
+          SetItem(5)
+          SetItem(6)
+          SetItem(7)
+          SetItem(8)
+          SetItem(9) _ </activeAccounts>
```
denis-bogdanas, line 74: Not clear why precompiled accounts have to be filled in like this. They don't seem to be used by the test. Proof side condition correctly have `andBool #notPrecompileAddress(ACCT_ID)`. There are probably other things that could be optimized but this is probably not worth the trouble.

Also, what exactly is this test testing?

ehildenb: This spec is largely generated from the ACT specification here (with some tweaks to make it work in our testing harness): https://reports.makerfoundation.com/k-dss/33d35d2934cfc47ccb98/#Vat_move(address,address,uint256)

So I don't want to change this because I want it to be relevant for the MKR folks who want this proof to work. The short answer is that all their proofs are auto-generated using a template which takes into account all the precompiles in this way.

denis-bogdanas: Ok then not clear what's the point of review. You can just ask for a formal approval (or I guess you can merge without approval since you are admin)

ehildenb: Well I was hoping for you to look over `verification.k` actually, because it add lemmas above and beyond `lemmas.k`, which I would like an audit on. I deleted their `verification.k` file and started from scratch, since theirs had lots of old lemmas which actually weren't used anymore.

tests/specs/mcd/verification.k
```diff
@@ -0,0 +1,77 @@
...
+requires "storage.k"
+
+module HASKELL-HASH2 [symbolic, kore]
+    imports EDSL
+
+    syntax Int ::= hash2(Int, Int) [function]
+ // -----------------------------------------
+endmodule
+
+module VERIFICATION
```
ehildenb, line 12: @denis-bogdanas can you look over the lemmas added here for sanity?

tests/specs/mcd/verification.k
```diff
@@ -0,0 +1,77 @@
...
+
+    // arithmetic/data-structures
+    rule I modInt N => I requires 0 <=Int I andBool I <Int N
+    rule I in (SetItem(I') REST) => I in REST requires I =/=Int I'
+
+    // chop
+    rule chop(hash2(V1, V2) +Int 1) => hash2(V1, V2) +Int 1
+
+    // sizeWordStack/sizeByteArray
+    rule #sizeByteArray(WS ++ WS')                        => #sizeByteArray(WS) +Int #sizeByteArray(WS')
```
denis-bogdanas, line 30: duplicate with lemmas.k

tests/specs/mcd/verification.k
```diff
@@ -0,0 +1,77 @@
...
+    rule #sizeByteArray(#padToWidth(32, #asByteStack(I))) => 32 requires 0 <=Int I andBool I <Int pow256
+    rule #sizeByteArray(_WS [ _START .. WIDTH ])          => WIDTH
```
denis-bogdanas, line 32: requires #range(0 < WIDTH <= #sizeByteArray(WS) -Int START)

tests/specs/mcd/verification.k
```diff
@@ -0,0 +1,77 @@
...
+
+    rule 0 <=Int #sizeWordStack(_, _) => true [smt-lemma]
+    rule 0 <=Int #sizeByteArray(_)    => true [smt-lemma]
```
denis-bogdanas, line 35: already in lemmas.k

tests/specs/mcd/verification.k
```diff
@@ -0,0 +1,77 @@
...
+
+    // ranged lookups in ByteArray
+    rule (WS ++ WS') [ 0 .. WIDTH ] => WS ++ (WS' [ 0 .. WIDTH -Int #sizeByteArray(WS) ]) requires #sizeByteArray(WS) <=Int WIDTH
+
+    rule (WS ++ _WS') [ START .. WIDTH ] => WS  [ START                         .. WIDTH ] requires START +Int WIDTH <=Int #sizeByteArray(WS)
+    rule (WS ++  WS') [ START .. WIDTH ] => WS' [ START -Int #sizeByteArray(WS) .. WIDTH ] requires #sizeByteArray(WS) <=Int START
```
denis-bogdanas, line 41: in more rigurous form, those are present in `LEMMAS-HASKELL` : https://github.com/kframework/evm-semantics/blob/3fc250d96249feefadbd3c5785875dd2bae5a934/tests/specs/lemmas.k#L266-L271
Can be moved to `LEMMAS`.

tests/specs/mcd/verification.k
```diff
@@ -0,0 +1,77 @@
...
+    // #asWord
+    rule #asWord(WS) >>Int M                           => #asWord(WS [ 0 .. #sizeByteArray(WS) -Int (M /Int 8) ]) requires 0 <=Int M andBool M modInt 8 ==Int 0
+    rule #asWord(#padToWidth(32, #asByteStack(VALUE))) => VALUE                                                   requires 0 <=Int VALUE andBool VALUE <Int pow256
+
+    // #unsigned
+    rule #unsigned(#sizeWordStack(WS, N) +Int M) => #sizeWordStack(WS, N) +Int M requires 0 <=Int M andBool #sizeWordStack(WS, N) +Int M <Int maxSInt256
+
+    // #storage lookup
+    rule #lookup( MAP [ KEY <- _VALUE ], KEY') => #lookup(MAP, KEY') requires KEY =/=K KEY'
+    rule #lookup(_MAP [ KEY <-  VALUE ], KEY') => VALUE              requires KEY  ==K KEY'
```
denis-bogdanas, line 55: already in lemmas.k

tests/specs/mcd/verification.k
```diff
@@ -0,0 +1,77 @@
...
+    // #computeValidJumpDests
+    rule #computeValidJumpDests(#padToWidth(32, #asByteStack(ADDR)) => .ByteArray, I => I +Int 32, _RESULT) requires 0 <=Int ADDR andBool ADDR <Int pow256
+
+    // Memory assignment sorting/normalization
+    rule MAP [ _N := .ByteArray ] => MAP
+    rule MAP [  N := W : WS     ] => MAP [ N <- W ] [ N +Int 1 := WS ]
+
+    rule MAP:Memory [ N <- V           ] [ N' <- V' ] => MAP [ N' <- V' ] [ N <- V ] requires N'                          <Int N
+    rule MAP:Memory [ N := V:ByteArray ] [ N' <- V' ] => MAP [ N' <- V' ] [ N := V ] requires N'                          <Int N orBool N +Int #sizeByteArray(V) <=Int N'
+    rule MAP:Memory [ N := V:ByteArray ] [ N' := V' ] => MAP [ N' := V' ] [ N := V ] requires N' +Int #sizeByteArray(V') <=Int N
```
denis-bogdanas, line 75: already in lemmas.k

denis-bogdanas:
I checked for duplicates with lemmas.k. Versions in lemmas.k often have more rigorous side conditions, so there might be incompatibilities.

tests/specs/mcd/verification.k
ehildenb: Actually, AFAICT, all we need is `0 <=Int WIDTH andBool 0 <=Int _START`, and I'm not even sure we need `0 <=Int _START`. But if it passes with both conditions, I think we might as well add them.

tests/specs/mcd/verification.k
ehildenb: Removed.

tests/specs/mcd/verification.k
ehildenb: Removed.

tests/specs/mcd/verification.k
ehildenb: Removed.

tests/specs/mcd/verification.k
ehildenb: Removed.

tests/specs/mcd/verification.k
ehildenb: Removed.

tests/specs/mcd/verification.k
ehildenb: Removed.

ehildenb:
@denis-bogdanas I've addressed all your comments, please have another look.

ehildenb:
@denis-bogdanas I've remove all lemmas which are not needed for this proof (but are needed for a later proof I'm working on), to make it clear which proofs the lemmas are needed for.

Since I've only deleted lemmas since your last review, I'm going to add `automerge`, but feel free to review again if you like.

PR #896, opened 2020-08-13
MKR regression tests: cat-exhaustiveness

ehildenb:
This is a proof that the Cat contract always correctly reverts on invalid function encoding inputs.


tests/specs/mcd/verification.k
```diff
@@ -28,4 +28,10 @@ module VERIFICATION
 
     rule hash2(V1, _) ==K hash2(V2, _) => false requires V1 =/=Int V2
 
+    // ### cat-exhaustiveness
+
+    rule #sizeByteArray(_WS [ START .. WIDTH ]) => WIDTH requires 0 <=Int START andBool 0 <=Int WIDTH
```
denis-bogdanas, line 33: `andBool START +Int WIDTH <=Int #sizeByteArray(WS)` , with `_WS` renamed to `WS`. In case it matters.

ehildenb: I don't think it does matter does it? `_[_.._]` pads with zeros if `START +Int WIDTH >Int #sizeByteArray(_)`, so I think this lemma is correct either way, and the simpler side condition will perform better. Can you say why you think the extra side condition is needed?

ehildenb: In fact, with the additional side-condition, this proof then fails to simplify some terms, and so it cannot be proven.

denis-bogdanas: Forgot about padding. Then it's all good.

PR #898, opened 2020-08-17
lemmas.k: Removed lemmas already in in domains.md, moved one lemma to common LEMMAS module

denis-bogdanas:
No tests to add here. Lemma move helps simplifying a term but that paraphrases the lemma.

### DIFF
--- a/tests/specs/lemmas.k
+++ b/tests/specs/lemmas.k
@@ -76,6 +76,8 @@ module LEMMAS
     rule 0 <=Int bool2Word(_B)             => true   [simplification]
     rule         bool2Word(_B) <Int pow256 => true   [simplification]
 
+    rule N /Int 1 => N  [simplification]
+
 endmodule
 
 module LEMMAS-JAVA [symbolic, kast]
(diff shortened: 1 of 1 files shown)


tests/specs/lemmas.k
```diff
@@ -170,17 +172,11 @@ module LEMMAS-JAVA [symbolic, kast]
...
     rule 1 *Int N => N  [simplification]
     rule N *Int 1 => N  [simplification]
     rule 0 *Int _ => 0  [simplification]
     rule _ *Int 0 => 0  [simplification]
 
-    rule N /Int 1 => N  [simplification]
-
     rule 0 |Int N => N  [simplification]
     rule N |Int 0 => N  [simplification]
     rule N |Int N => N  [simplification]
```
ehildenb, line 182: I'm approving, but why do we not also move the other arith lemmas into the common one? You only need this one right now?

ehildenb:
Feel free to add `automerge` if you don't want to make further changes.

PR #901, opened 2020-08-20
MKR regression tests: dai-symbol-pass

ehildenb:
This PR:

-   Adds the `asByteStackInWidth` abstraction (which uses the `nythbyteof` abstraction), which is used in 3 MKR rough proofs. 
-   Clears some of the warnings that have been popping up regarding variable names in `lemmas.k`, and removes the `simplification` attribute from Java-only lemmas.
-   Separates the MCD `VERIFICATION` module from a `VERIFICATION-SYNTAX` module.
-   Sets the `-Xss` stack size to 512m globally for Java, which is 4x the K default of 32m. Before we were setting 500m for `krun`, but nothing for kprove. This avoids the flaky infinite recursion we see with `#sizeWordStack` sometimes.

### DIFF
--- a/evm.md
+++ b/evm.md
@@ -1330,8 +1330,8 @@ The various `CALL*` (and other inter-contract control flow) operations will be d
 
 ```{.k .nobytes}
     rule #computeValidJumpDests(.WordStack, _, RESULT) => List2Set(RESULT)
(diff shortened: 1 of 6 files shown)


denis-bogdanas (approved):
Removing `[simplification]` might create headaches in future if we move lemmas from JAVA to common module and forget adding it. Not a big deal however.

PR #904, opened 2020-08-24
MKR regression tests: switch to direct key differences in side-conditions

ehildenb:
In order to avoid the problems pointed out by @daejunpark in https://github.com/kframework/evm-semantics/pull/902, here we're instead directly encoding the storage key differences needed for each individual proof directly in the proof itself.

This can still be auto-generated by KLab, but should avoid some of the soundness issues that Daejun pointed out.

I also switch the remaining specs over to abstract-pre concrete-post storage model, and organize the `storage.k` file a bit.


daejunpark:
I like this direction.  It requires more work to write (or generate) the spec, but should be safer.

PR #907, opened 2020-08-30
Remove unsigned/signed, add rangeBool, tighter definition of `sgn` and `abs`.

ehildenb:
This PR does 3 things:

-   `#unsigned/#signed` are removed. `#signed` is never used anywhere, and `#unsigned` is only used in `#getValue`, so the uses are just inlined.
-   `#rangeBool` is added with a simpler predicate for being boolean than `#rangeUInt(1, _)`, and is used in `edsl.md`.
-   `abs` and `sgn` have tighter bounds and go to `0` on nonsense cases.

### DIFF
--- a/edsl.md
+++ b/edsl.md
@@ -231,7 +231,7 @@ where `F1 : F2 : F3 : F4` is the (two's complement) byte-array representation of
       requires #rangeUInt(256, DATA)
 
     rule #getValue(   #bool( DATA )) => DATA
-      requires #range(0 <= DATA <= 1)
+      requires #rangeBool(DATA)
 
     syntax Int ::= #ceil32 ( Int ) [function, functional, smt-hook(ceil32), smt-prelude]
  // ------------------------------------------------------------------------------------
--- a/edsl.md
+++ b/edsl.md
@@ -194,15 +194,6 @@ where `F1 : F2 : F3 : F4` is the (two's complement) byte-array representation of
       [anywhere, simplification]
 ```
 
-- #unsigned : sInt256 -> uInt256  (i.e., [minSInt256..maxSInt256] -> [minUInt256..maxUInt256])
-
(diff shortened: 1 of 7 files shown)


denis-bogdanas (approved):

daejunpark:
`#signed` is intended to be used in writing specs for functions that take signed integers.  For example, we want to say a signed integer argument `#signed(-1)`, which is more intuitive than writing `pow256 - 1`.  `#unsigned` is paired with `#signed`, so both are needed.  I remember we used them in writing and proving some Maker specs.  Please put them back.  (It can be just put in the edsl.md.)

Also, you seem to want to keep adding "infeasible" rules, especially in a way that they are not clearly distinguished from other normal feasible rules.  As I mentioned before, this practice, at best, hurts the readability as the executable standard.  (Moreover, although I do not have a concrete counterexample, repeating this practice would bite us eventually, allowing false specs to be passed by kprove.)  I recommend you to consider a way to make them distinguishable (at least putting some comments), so that we can easily correct them later when the haskell backend is ready.

ehildenb:
@daejunpark I'm putting them back in Maker verification file directly, and we can upstream them into lemmas file and edsl as they become more widely used. But they definitely don't belong in the semantics proper.

As far as `abs` and `sgn` goes, I think the definition before was definitely too loose. I think you will be hard-pressed to find an example where it going directly to zero allows a proof to pass when it shouldn't.

ehildenb:
I think it's arguably better to go to a garbage value on poorly defined cases (in this case out of range), then to continue execution as normal (which was what happened before with these definitions).

daejunpark:
> I'm putting them back in Maker verification file directly, and we can upstream them into lemmas file and edsl as they become more widely used. 

Ah, that makes sense.

PR #909, opened 2020-09-01
MKR regression tests: dstoken-burn-self-fail-rough, flipper-ttl-pass, vat-addui-fail-rough, vat-mului-pass

ehildenb:
These proofs demonstrate two new things:

-   More advanced of `#WordPack*` abstractions.
-   Use of signed arithmetic ~(with `#unsigned` abstraction)~. I have replaced use of `#unsigned(X)` with just `chop(X)`, because we have `#rangeSInt(256, X) impliesBool (#unsigned(X) ==Int chop(X))`, and we never use `#unsigned(X)` without `#rangeSInt(256, X)` as an additional constraint.
-   Make `/sWord` concrete so that we can write a lemma directly about it instead of about arguments `sgn` and `abs`.

### DIFF
--- a/tests/specs/mcd/verification.k
+++ b/tests/specs/mcd/verification.k
@@ -199,12 +199,12 @@ module VERIFICATION
     //        => (X *Int Y) /Int Y
     //        => X
     //
-    // Via Z3 ?.?.?:
-    // ; 2^6 and 2^5 instead of 2^256 and 2^255 for small model proof
+    // Via Z3 4.7.1:
+    // ; 2^4 and 2^3 instead of 2^256 and 2^255 for small model proof
     // (declare-const pow256 Int)
-    // (assert (= pow256 64))
+    // (assert (= pow256 16))
     // (declare-const pow255 Int)
-    // (assert (= pow255 32))
+    // (assert (= pow255 8))
     //
     // ; sgn
     // (declare-fun sgn (Int) Int)
--- a/tests/specs/mcd/verification.k
+++ b/tests/specs/mcd/verification.k
@@ -69,6 +69,11 @@ module VERIFICATION-SYNTAX
     rule nthbyteof(V, I, N) => nthbyteof(V /Int 256, I, N -Int 1) when N  >Int (I +Int 1) [concrete]
     rule nthbyteof(V, I, N) =>           V modInt 256             when N ==Int (I +Int 1) [concrete]
 
+    syntax Int ::= #unsigned ( Int ) [function]
+ // -------------------------------------------
+    rule #unsigned(DATA) => DATA             requires 0 <=Int DATA andBool DATA <=Int maxSInt256
+    rule #unsigned(DATA) => pow256 +Int DATA requires minSInt256 <=Int DATA andBool DATA <Int 0
+
 endmodule
 
 module VERIFICATION
--- a/tests/specs/mcd/verification.k
+++ b/tests/specs/mcd/verification.k
@@ -147,5 +147,48 @@ module VERIFICATION
     // ### vat-mului-fail-rough
 
     rule #unsigned(Y) ==K 0 => Y ==K 0 requires #rangeSInt(256, Y) [smt-lemma]
+    rule       abs(Y) ==K 0 => Y ==K 0 requires #rangeUInt(256, Y) [smt-lemma]
+
+    // Proof:
+    // case sgn(#unsigned(Y)) ==Int 1                   // #unsigned(Y) ==Int Y andBool 0 <=Int Y
+    //      chop(X *Int #unsigned(Y))
+    //   => chop(X *Int Y)                              // #unsigned(Y) ==Int Y
+    // case sgn(#unsigned(Y)) ==Int -1                  // #unsigned(Y) ==Int Y +Int pow256 andBool Y <Int 0
+    //      chop(X *Int #unsigned(Y))
+    //   => chop(X *Int (Y +Int pow256))                // #unsigned(Y) ==Int Y +Int pow256
+    //   => chop(X *Int Y +Int X *Int pow256)
+    //   => chop((X *Int Y) +Int chop(X *Int pow256))
+    //   => chop(X *Int Y)
+    rule chop(X *Int #unsigned(Y)) => chop(X *Int Y) requires #rangeSInt(256, Y)
+
+    // Proof:
+    // case sgn(#unsigned(Y)) ==Int 1                           // 0 <=Int Y andBool #unsigned(Y) ==Int Y
+    //      sgn(chop(X *Int Y)) *Int sgn(#unsigned(Y))
+    //   => sgn(X *Int Y) *Int sgn(Y)                           // 0 <=Int X andBool 0 <=Int Y andBool #rangeSInt(256, X *Int Y) andBool #unsigned(Y) ==Int Y
+    //   => 1 *Int 1                                            // 0 <=Int X *Int Y andBool #rangeSInt(256, X *Int Y)
+    //   => 1
+    // case sgn(#unsigned(Y)) ==Int -1                          // Y <Int 0 andBool #unsigned(Y) ==Int Y +Int pow256
+    //      sgn(chop(X *Int Y)) *Int sgn(#unsigned(Y))
+    //   => sgn(X *Int Y +Int pow256) *Int sgn(Y +Int pow256)   // 0 <=Int X andBool Y <Int 0 andBool #rangeSInt(256, X *Int Y) andBool #unsigned(Y) ==Int Y +Int pow256
+    //   => -1 *Int -1                                          // X *Int Y <Int 0 andBool #rangeSInt(256, X *Int Y) andBool Y <Int andBool #rangeSInt(256, Y)
+    //   => 1
+    rule sgn(chop(X *Int Y)) *Int sgn(#unsigned(Y)) => 1 requires #rangeUInt(256, X) andBool #rangeSInt(256, Y) andBool #rangeSInt(256, X *Int Y)
+
+    // Proof (assuming Y =/=Int 0):
+    // case sgn(#unsigned(Y)) ==Int 1                                                   // 0 <=Int Y andBool #unsigned(Y) ==Int Y
+    //      abs(chop(X *Int Y)) /Int abs(#unsigned(Y))
+    //      abs(X *Int Y) /Int abs(Y)                                                   // #rangeSInt(256, Y) andBool #unsigned(Y) ==Int Y
+    //      X *Int Y /Int Y                                                             // #rangeSInt(256, X *Int Y) andBool 0 <=Int X *Int Y andBool #rangeSInt(256, Y) andBool 0 <=Int Y
+    //      X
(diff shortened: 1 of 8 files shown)


denis-bogdanas (commented):
waiting the fix in PM.

denis-bogdanas (approved):

tests/specs/mcd/verification.k
```diff
@@ -23,13 +23,23 @@ module VERIFICATION-SYNTAX
...
+
+    rule maxUInt160 &Int  ADDR_UINT48_UINT48              => ADDR     requires ADDR_UINT48_UINT48 ==Int #WordPackAddrUInt48UInt48(ADDR, UINT48_1, UINT48_2) andBool #rangeAddress(ADDR) andBool #rangeUInt(48, UINT48_1) andBool #rangeUInt(48, UINT48_2) [smt-lemma]
+    rule maxUInt48  &Int (ADDR_UINT48_UINT48 /Int pow160) => UINT48_1 requires ADDR_UINT48_UINT48 ==Int #WordPackAddrUInt48UInt48(ADDR, UINT48_1, UINT48_2) andBool #rangeAddress(ADDR) andBool #rangeUInt(48, UINT48_1) andBool #rangeUInt(48, UINT48_2) [smt-lemma]
+    rule maxUInt48  &Int (ADDR_UINT48_UINT48 /Int pow208) => UINT48_2 requires ADDR_UINT48_UINT48 ==Int #WordPackAddrUInt48UInt48(ADDR, UINT48_1, UINT48_2) andBool #rangeAddress(ADDR) andBool #rangeUInt(48, UINT48_1) andBool #rangeUInt(48, UINT48_2) [smt-lemma]
+
+    rule maxUInt160 &Int  ADDR_UINT8              => ADDR  requires ADDR_UINT8 ==Int #WordPackAddrUInt8(ADDR, UINT8) andBool #rangeAddress(ADDR) andBool #rangeUInt(8, UINT8) [smt-lemma]
+    rule maxUInt8   &Int (ADDR_UINT8 /Int pow160) => UINT8 requires ADDR_UINT8 ==Int #WordPackAddrUInt8(ADDR, UINT8) andBool #rangeAddress(ADDR) andBool #rangeUInt(8, UINT8) [smt-lemma]
```
daejunpark, line 42: How could this work?  Is it what @denis-bogdanas fixed recently?

daejunpark: BTW, why did you write these lemmas in this way?  I mean, what's the problem of writing, e.g.,:
```
rule maxUInt48 &Int  #WordPackUInt48UInt48(UINT48_1, UINT48_2)  => UINT48_1 requires #rangeUInt(48, UINT48_1) andBool #rangeUInt(48, UINT48_2)
```
(no `[smt-lemma]`)?

daejunpark: Another question.  Could you explain why the commented lemmas don't work?  I guess that z3 is not capable of reasoning about the bitwise operations using the nonlinear expressions, but want to double-check.

denis-bogdanas: No, my fix is unrelated and is not merged. I'm surprised this works, but in case it does, seems correct.

tests/specs/mcd/verification.k
```diff
@@ -109,4 +119,53 @@ module VERIFICATION
...
+    //        => X
+    //      case Y <Int 0
+    //           abs(chop(X *Int Y)) /Int abs(chop(Y))
+    //        => abs(X *Int Y +Int pow256) /Int abs(Y +Int pow256)                           // #rangeSInt(256, X *Int Y) andBool chop(Y) ==Int Y +Int pow256
+    //        => (0 -Word (X *Int Y +Int pow256)) /Int (0 -Word (Y +Int pow256))             // #rangeSInt(256, X *Int Y) andBool X *Int Y <Int 0 andBool #rangeSInt(256, Y) andBool Y <Int 0
+    //        => (pow256 -Int (X *Int Y +Int pow256)) /Int (pow256 -Int (Y +Int pow256))     // 0 <Int (X *Int Y +Int pow256) andBool 0 <Int (Y +Int pow256)
+    //        => (X *Int Y) /Int Y
+    //        => X
+    rule chop(X *Int Y) /sWord chop(Y) => X requires #rangeUInt(256, X) andBool #rangeSInt(256, Y) andBool #rangeSInt(256, X *Int Y) andBool Y =/=Int 0
+
```
daejunpark, line 170: Have you tried to prove this using z3?

tests/specs/mcd/verification.k
```diff
@@ -109,4 +119,53 @@ module VERIFICATION
...
+    // Proof:
+    // case sgn(X) ==Int  1
+    //      X s<Word 0
+    //   => X <Word 0           // sgn(X) ==Int 1 andBool sgn(0) ==Int 1
+    //   => bool2Word(false)    // #rangeUInt(256, X)
+    // case sgn(X) ==Int -1
+    //      X s<Word 0
+    //   => bool2Word(true)     // sgn(X) ==Int -1 andBool sgn(0) ==Int 1
+    rule X s<Word 0 => bool2Word(sgn(X) ==Int -1) requires #rangeUInt(256, X)
```
daejunpark, line 132: A z3 proof: https://rise4fun.com/Z3/plzQl

(I wish we have a good way to link an external mechanized proof, if any, to each lemma.)

tests/specs/mcd/verification.k
daejunpark: A worst case scenario: It could be the case that this smt-lemma is encoded incorrectly, leading to a _false_ statement, which will make all z3 queries always _unsat_, making all specs to be blindly passed.  So, we might be better to double-check how this is exactly handled by the java backend.

tests/specs/mcd/verification.k
daejunpark: A partial proof with a small model (for 6-bit integers): https://rise4fun.com/Z3/Eslc8

Provable by z3 4.6.0 or 4.7.1, but not 4.8.x.
(The z3 in the web interface failed to prove.)

daejunpark: Although I didn't read the hand-written proof, since z3 proved the lemma with a smaller bitwidth, and the lemma doesn't seem dependent on the particular bitwidth, I think the lemma is fine.

daejunpark:
@ehildenb Please feel free to merge after checking the following:
- double-check if the smt-lemma is correctly encoded by the java backend
- review the two z3 proofs

daejunpark:
Also, next time, please consider writing z3 proofs for nontrivial lemmas.  Even a partial proof would be helpful for the long term.

tests/specs/mcd/verification.k
ehildenb: @daejunpark I cannot write it how you would like (as `maxUInt48 &Int #WordPackUInt48UInt48(UINT48_1, UINT48_2) => ...`), because the actual term looks like `maxUInt48 &Int #lookup(STORAGE, SOME_KEY)`, and only via the side-condition do we see that `#lookup(STORAGE, SOME_KEY)` is equal `SOME_VARIABLE`, and `SOME_VARIABLE` is equal to `#WordPackUInt48UInt48(UINT48_1, UINT48_2)`. So the unification/matching just doesn't seem to apply the rule as needed.

tests/specs/mcd/verification.k
ehildenb: Hmmm, I can't get this Z3 proof to work on any version of Z3 I've tried, even though it looks OK.

daejunpark: Have you tried 4.6.0 or 4.7.1?

tests/specs/mcd/verification.k
```diff
@@ -109,4 +119,137 @@ module VERIFICATION
...
+    //      case Y <Int 0
+    //           abs(chop(X *Int Y)) /Int abs(chop(Y))
+    //        => abs(X *Int Y +Int pow256) /Int abs(Y +Int pow256)                           // #rangeSInt(256, X *Int Y) andBool chop(Y) ==Int Y +Int pow256
+    //        => (0 -Word (X *Int Y +Int pow256)) /Int (0 -Word (Y +Int pow256))             // #rangeSInt(256, X *Int Y) andBool X *Int Y <Int 0 andBool #rangeSInt(256, Y) andBool Y <Int 0
+    //        => (pow256 -Int (X *Int Y +Int pow256)) /Int (pow256 -Int (Y +Int pow256))     // 0 <Int (X *Int Y +Int pow256) andBool 0 <Int (Y +Int pow256)
+    //        => (X *Int Y) /Int Y
+    //        => X
+    //
+    // Via Z3 ?.?.?:
+    // ; 2^64 and 2^32 instead of 2^256 for small model proof
```
daejunpark, line 203: They are 2^6 (= 64) and 2^5 (= 32). Quite small, but z3 takes too long beyond that.

tests/specs/mcd/verification.k
```diff
@@ -109,4 +119,137 @@ module VERIFICATION
...
+    //        => X
+    //      case Y <Int 0
+    //           abs(chop(X *Int Y)) /Int abs(chop(Y))
+    //        => abs(X *Int Y +Int pow256) /Int abs(Y +Int pow256)                           // #rangeSInt(256, X *Int Y) andBool chop(Y) ==Int Y +Int pow256
+    //        => (0 -Word (X *Int Y +Int pow256)) /Int (0 -Word (Y +Int pow256))             // #rangeSInt(256, X *Int Y) andBool X *Int Y <Int 0 andBool #rangeSInt(256, Y) andBool Y <Int 0
+    //        => (pow256 -Int (X *Int Y +Int pow256)) /Int (pow256 -Int (Y +Int pow256))     // 0 <Int (X *Int Y +Int pow256) andBool 0 <Int (Y +Int pow256)
+    //        => (X *Int Y) /Int Y
+    //        => X
+    //
+    // Via Z3 ?.?.?:
```
daejunpark, line 202: It worked with z3 4.6.0 and 4.7.1.  (More precisely, z3-4.6.0-x64-osx-10.11.6, and z3-4.7.1-x64-osx-10.11.6, downloaded from the z3 github release page.)

ehildenb: For those versions of Z3, I get `unknown` on this query every time.

ehildenb: Also 4.8.7

tests/specs/mcd/verification.k
ehildenb: I put the Z3 proof in the PR, thanks!

tests/specs/mcd/verification.k
daejunpark: Hmm, I tried the same versions in the ubuntu (rv-work machine), but it didn't work.  Wow, I didn't know that the same version works differently on different OSes.

OK, the ubuntu version works if you further reduce the bitwidth to 4, that is, 16 for pow256, and 8 for pow255.

tests/specs/mcd/verification.k
daejunpark: Oh I see. Then this makes sense.

tests/specs/mcd/verification.k
ehildenb: This worked for me too, thanks!

PR #910, opened 2020-09-03
Enable more specs with the Haskell backend

ttuegel:
The `#Ceil(#buf)` rule is removed because it is obsolete. In the past, it was necessary to define `#Ceil(#buf)` so that we could use the SMT solver to refute the `#Not(#Ceil(#buf))` remainder after applying rules. This is no longer necessary because the backend correctly infers `#Ceil(#buf)` from the `#buf` appearing in the configuration; then the remainder is refuted by the obvious contradiction.


edsl.md
```diff
@@ -186,14 +186,6 @@ where `F1 : F2 : F3 : F4` is the (two's complement) byte-array representation of
...
 ```
 
-``` {.k .symbolic .ceil}
-    rule #Ceil(#buf(@SIZE, @DATA))
-      =>      {(0 <=Int @SIZE) andBool #rangeBytes(@SIZE, @DATA) #Equals true}
-         #And #Ceil(@SIZE)
-         #And #Ceil(@DATA)
-      [anywhere, simplification]
-```
-
```
ehildenb, line 196: Why is this not needed anymore?

ttuegel: I added it to the description of the pull request.

ttuegel:
There are some issues with the SMT solver on Jenkins, so we probably need to wait for the latest version of K before all these tests will work.

ttuegel:
I thought this would use the latest K now, but I'm still getting weird responses from the SMT solver, so I need to investigate locally.

ehildenb (approved):

PR #915, opened 2020-09-08
Use common infinite gas abstraction for Haskell backend tests

ttuegel:
With these changes, the ERC20 specs do not run out of gas, but get stuck evaluating some `#hashedLocation` rules.


tests/specs/infinite-gas.k
```diff
@@ -0,0 +1,55 @@
...
+ // This allows (i) computing final gas used, and (ii) never stopping because of out-of-gas.
+ //
+
+    syntax Int ::= #gas ( Int ) [function, functional, no-evaluators]
+ // -----------------------------------------------------------------
+    rule #gas(_, _, _) => #gas(0)
+
+    rule #gas(G) -Int G' => #gas(G +Int G') [simplification]
+
+    rule #gas(_)  <Int I => false requires I <=Int pow256 [simplification]
```
ttuegel, line 24: @ehildenb Why not this?

```suggestion
    rule #gas(_)  <Int I => notBool (I <=Int pow256) [simplification]
```

tests/specs/infinite-gas.k
```diff
@@ -0,0 +1,55 @@
...
+    rule #gas(_) <=Int I => false requires I <=Int pow256 [simplification]
+    rule #gas(_) >=Int I => true  requires I <=Int pow256 [simplification]
+    rule I <=Int #gas(_) => true  requires I <=Int pow256 [simplification]
+
+    rule #gas(_)  <Int _I => false [concrete(_I), simplification]
```
ttuegel, line 29: Isn't this block of rules redundant to the rules above? And contradictory?

tests/specs/infinite-gas.k
ehildenb: @ttuegel we don't want this change because it's not always the case that we can simplify out `#gas(_) <Int I` using the fact that `I <=Int pow256`, it's just a heuristic. So basically, `I <=Int pow256 IMPLIES notBool #gas(_) <Int I`, but _not_ the case that `I <=Int pow256 EQUALS notBool (I <=Int pow256)`.

For example, if you have `#gas(_) <Int (2 *Int pow256)`, and you simplify it to `notBool (2 *Int pow256 <=Int pow256)`, you'll conclude `true`, when instead it should take the rule `rule #gas(_) <Int I => false [concrete(_I)]` to determine `true`.

The reason for this rule is that many values which are compared in EVM are bounded by `pow256`, so this is a good heuristic for simplifying out using the fact that `#gas(_)` is larger than any particular number.

tests/specs/infinite-gas.k
ehildenb: I'm not sure how it's redundant or contradictory. Here we're saying that `#gas(_)` is not smaller than any particular number that is nameable as a concrete number. It's just a heuristic for saying that `#gas(_)` is larger than any particular number, this heuristic is for any number that is concrete, which means it's some particular integer.

ehildenb (changes requested):
@ttuegel please add a few tests to `tests/specs/functional/infiinite-gas-spec.k` of these lemmas (in the same style as `tests/specs/functional/lemmas-no-smt-spec.k`, and you'll need to update the Makefile in the same way to turn off the SMT solver).

Just a few tests of `#gas(_) < ...`, `#gas(_) <= ...`, `... < #gas(_)`, and `... <= #gas(_)` would be good.

tests/specs/infinite-gas.k
ttuegel: > instead it should take the rule `rule #gas(_) <Int I => false [concrete(_I)]` to determine `true`.

Do you mean `false`?

ehildenb: Yes, sorry, edited.

ehildenb (approved):

ehildenb:
@ttuegel I added some tests, and made a tweak to the rules for `/Int`. If you agree with the changes I made, feel free to add `automerge`.

tests/specs/infinite-gas.k
```diff
@@ -0,0 +1,55 @@
...
+    syntax Int ::= #gas ( Int ) [function, functional, no-evaluators]
+ // -----------------------------------------------------------------
+    rule #gas(_, _, _) => #gas(0)
+
+    rule #gas(G) -Int G' => #gas(G +Int G') [simplification]
+
+    rule #gas(_)  <Int I => false requires I <=Int pow256 [simplification]
+    rule #gas(_) <=Int I => false requires I <=Int pow256 [simplification]
+    rule #gas(_) >=Int I => true  requires I <=Int pow256 [simplification]
+    rule I <=Int #gas(_) => true  requires I <=Int pow256 [simplification]
```
ttuegel, line 27: @ehildenb Can we mark these rules `symbolic` to ensure that the `concrete` rules below are always preferred?

ehildenb: Hmmmm, I'm not opposed, but I guess I'd be surprised it if led to a performance improvement. These are all heuristics, so whichever is the first to apply is good enough.

Up to you though, if you have reason to believe it will improve performance somehow then it makes sense.

daejunpark:
@ehildenb Could you please advise me which part I should focus on?  Is there any change made other than copying or moving the set of lemmas to different files?

ehildenb:
@daejunpark I just want another set of eyes on the lemmas now that they're being promoted to the top-level, and then any tests you would like to see please suggest. But if you're happy with the state of it, you can just approve.

The only difference is the added side-conditions that divisors are non-zero here.

daejunpark (approved):

PR #916, opened 2020-09-09
MKR regression tests: flipper-bids-pass-rough, flopper-file-pass-rough, flopper-tick-pass-rough

ehildenb:
This updates the WordPack abstractions to remove the use of `smt-lemma`. Instead, equalities that look like `#lookup(_, _) ==Int WordPack(...)` are unfolded into their individual components, and there are lemmas provided for refolding them up when those storage locations are updated.

-   cat-ilks-pass is replaced with flipper-bids-pass-rough, which shows a bigger Solidity struct we need to handle hash non-overflow for.
-   flopper-file-pass-rough shows repacking a WordPackUInt48UInt48.
-   flopper-tick-pass-rough shows repacking a WordPackAddrUInt48UInt48.


tests/specs/mcd/verification.k
```diff
@@ -23,23 +23,22 @@ module VERIFICATION-SYNTAX
...
-    rule maxUInt48  &Int (ADDR_UINT48_UINT48 /Int pow160) => UINT48_1 requires ADDR_UINT48_UINT48 ==Int #WordPackAddrUInt48UInt48(ADDR, UINT48_1, UINT48_2) andBool #rangeAddress(ADDR) andBool #rangeUInt(48, UINT48_1) andBool #rangeUInt(48, UINT48_2) [smt-lemma]
-    rule maxUInt48  &Int (ADDR_UINT48_UINT48 /Int pow208) => UINT48_2 requires ADDR_UINT48_UINT48 ==Int #WordPackAddrUInt48UInt48(ADDR, UINT48_1, UINT48_2) andBool #rangeAddress(ADDR) andBool #rangeUInt(48, UINT48_1) andBool #rangeUInt(48, UINT48_2) [smt-lemma]
+    rule UINT48_UINT48 ==K #WordPackUInt48UInt48(UINT48_1, UINT48_2) => UINT48_1 ==Int maxUInt48 &Int UINT48_UINT48 andBool UINT48_2 ==Int maxUInt48 &Int (UINT48_UINT48 /Int pow48)
+    rule (UINT48_2 *Int pow48) |Int (maskWordPackUInt48UInt48_1 &Int UINT48_UINT48) => #WordPackUInt48UInt48(maxUInt48 &Int UINT48_UINT48, UINT48_2                                 ) requires #rangeUInt(48, UINT48_2)
+    rule  UINT48_1             |Int (maskWordPackUInt48UInt48_2 &Int UINT48_UINT48) => #WordPackUInt48UInt48(UINT48_1                    , maxUInt48 &Int (UINT48_UINT48 /Int pow48)) requires #rangeUInt(48, UINT48_1)
```
daejunpark, line 39: No range condition for `UINT48_UINT48`?

daejunpark: Also, it is hard to get the intuition of the above two lemmas.  Could you write a comment that explains them intuitively?

tests/specs/mcd/verification.k
```diff
@@ -23,23 +23,22 @@ module VERIFICATION-SYNTAX
...
 
-    rule maxUInt160 &Int  ADDR_UINT8              => ADDR  requires ADDR_UINT8 ==Int #WordPackAddrUInt8(ADDR, UINT8) andBool #rangeAddress(ADDR) andBool #rangeUInt(8, UINT8) [smt-lemma]
-    rule maxUInt8   &Int (ADDR_UINT8 /Int pow160) => UINT8 requires ADDR_UINT8 ==Int #WordPackAddrUInt8(ADDR, UINT8) andBool #rangeAddress(ADDR) andBool #rangeUInt(8, UINT8) [smt-lemma]
+    rule ADDR_UINT8 ==K #WordPackAddrUInt8(ADDR, UINT8) => ADDR ==Int maxUInt160 &Int ADDR_UINT8 andBool UINT8 ==Int maxUInt8 &Int (ADDR_UINT8 /Int pow160)
```
daejunpark, line 41: So, essentially you specified how to unify two terms when one of them is a WordPack term.  My concern is that this could be applied even if not necessary, leading to blowing up the size of constraints and serious performance issue.  Also, you may want to have a different unification strategy for some cases, e.g., `#WordPackAddrUInt8(X, Y) ==K #WordPackAddrUInt8(Z, W)`.

You can just go with these for now, but keep in mind the potential issue.

daejunpark: BTW, no range condition for `ADDR_UINT8` needed?

tests/specs/mcd/verification.k
```diff
@@ -252,4 +251,9 @@ module VERIFICATION
     // (check-sat)
     rule chop(X *Int Y) /sWord chop(Y) => X requires #rangeUInt(256, X) andBool #rangeSInt(256, Y) andBool #rangeSInt(256, X *Int Y) andBool Y =/=Int 0
 
+    // ### dsvalue-peek-pass-rough
+
+    rule (B ==K false) ==K false => B
```
denis-bogdanas, line 256: On Haskell this is not needed: https://github.com/kframework/evm-semantics/blob/master/tests/specs/functional/lemmas-no-smt-spec.k#L37
Because of this: https://github.com/kframework/evm-semantics/blob/fc44b9ddd1ba29f433778303e922ea73ab748df2/tests/specs/lemmas.k#L268-L269

Lemmas don't have side conditions, so they don't carry performance degradation risks. Should be made common for Java & Haskell.

tests/specs/mcd/verification.k
```diff
@@ -252,4 +251,9 @@ module VERIFICATION
...
+    rule bool2Word((UINT8 ==K 0) ==K false) => UINT8 requires #rangeBool(UINT8)
```
denis-bogdanas, line 257: With lemmas from above comment, this can be replaced with lower-level lemmas:
```
rule notBool (X ==Int 0) => X ==Int 1 requires #rangeBool(X)
rule bool2Word(X ==Int 1)  => X         requires #rangeBool(X)

//Rcommended for symmetry, to avoid dealing with them next time
rule notBool (X ==Int 1) => X ==Int 0   requires #rangeBool(X)
```
And above rule should be made test instead.

denis-bogdanas (changes requested):

tests/specs/mcd/verification.k
ehildenb: I've added comments and re-organized the lemmas to hopefully be more intuitive.

tests/specs/mcd/verification.k
ehildenb: The re-organization of the lemmas and added comments should help make it clearer. Basically `_ ==K #WordPack(_, _)` itself is a large predicate which unfolds into several smaller predicates, which makes it easier to reason about the individual values packed into the storage slot.

ehildenb:
I've removed `bool2Word` changes into a separate PR (https://github.com/kframework/evm-semantics/pull/918) so that they can be reviewed independently. This one has the refactored `WordPack` lemmas, which basically amounts to rules which (i) change predicates about `#WordPack` into predicates about the individual components, and (ii) fold updates to individual components back up into `#WordPack` over the entire word.

denis-bogdanas (approved):

tests/specs/mcd/flipper-bids-pass-rough-spec.k
```diff
@@ -0,0 +1,233 @@
...
+     andBool #Flipper.bids[ABI_n].lot =/=Int #Flipper.bids[ABI_n].guy_tic_end
+     andBool #Flipper.bids[ABI_n].lot =/=Int #Flipper.bids[ABI_n].usr
+     andBool #Flipper.bids[ABI_n].lot =/=Int #Flipper.bids[ABI_n].gal
+     andBool #Flipper.bids[ABI_n].lot =/=Int #Flipper.bids[ABI_n].tab
+     andBool #Flipper.bids[ABI_n].guy_tic_end =/=Int #Flipper.bids[ABI_n].usr
+     andBool #Flipper.bids[ABI_n].guy_tic_end =/=Int #Flipper.bids[ABI_n].gal
+     andBool #Flipper.bids[ABI_n].guy_tic_end =/=Int #Flipper.bids[ABI_n].tab
+     andBool #Flipper.bids[ABI_n].usr =/=Int #Flipper.bids[ABI_n].gal
+     andBool #Flipper.bids[ABI_n].usr =/=Int #Flipper.bids[ABI_n].tab
+     andBool #Flipper.bids[ABI_n].gal =/=Int #Flipper.bids[ABI_n].tab
```
kmbarry1, line 231: A bit surprised this storage address pairwise disequality check is necessary...shouldn't these all be of the form `hash + offset` where `hash` stays the same but `offset` is different for every slot? Not saying there's anything wrong with it, just surprised `K` doesn't recognize that they aren't the same.

ehildenb: I started by adding lemmas about the disequalities of hashes on different inputs and at different offsets, but a discussion in #902 with @daejunpark made me realize it may be easy to add lemmas which _seem_ OK, but are not.

So I switched to generating exactly the disequalities needed in the proofs in #904 (and made KLab generate these disequalities directly on my branch `pr` of my fork). Because this (i) didn't seem to impact performance at all, (ii) was easy to make KLab do correctly, and (iii) expresses _exactly_ the disequalities needed for a given proof rather than the over-approximation given by the hash2 lemmas, it seemed like the better way to go.

ehildenb: Specifically if you want to see the type of `hash2` disequalities that were removed, check out: https://github.com/kframework/evm-semantics/pull/902/commits/7d81a7bcaca9eb6c107566ebab6b1c99feada14b

tests/specs/mcd/verification.k
```diff
@@ -19,27 +19,54 @@ module VERIFICATION-SYNTAX
...
+
+    rule    UINT48_UINT48 ==K #WordPackUInt48UInt48(UINT48_1, UINT48_2)
+         => UINT48_1 ==Int maxUInt48 &Int  UINT48_UINT48
+    andBool UINT48_2 ==Int maxUInt48 &Int (UINT48_UINT48 /Int pow48)
+    andBool #rangeUInt(96, UINT48_UINT48)
+
+    rule    ADDR_UINT8 ==K #WordPackAddrUInt8(ADDR, UINT8)
+         => ADDR  ==Int maxUInt160 &Int  ADDR_UINT8
+    andBool UINT8 ==Int maxUInt8   &Int (ADDR_UINT8 /Int pow160)
+    andBool #rangeUInt(168, ADDR_UINT8)
```
daejunpark, line 49: Mentioning that the arguments are packed in the reverse order would help others to understand these rules.

tests/specs/mcd/verification.k
```diff
@@ -19,27 +19,54 @@ module VERIFICATION-SYNTAX
...
+    //
+    // When updating variables which originally contained `WordPack*`, the Solidity compiler will mask the original value everywhere _except_ where you're updating.
+    // Then the update will be bitwise `|Int` with the masked value, and the new value written back.
+    // Here we provide the masks which are used by the Solidity compiler, and rules which fold these bitwise updates back up into semantic `#WordPack*` arguments for the purposes of matching the RHS of proofs.
+    //
+
+    syntax Int ::= "maskWordPackUInt48UInt48_1" | "maskWordPackUInt48UInt48_2"
+ // --------------------------------------------------------------------------
+    rule maskWordPackUInt48UInt48_1 => 115792089237316195423570985008687907853269984665561335876943319951794562400255 [macro] // pow256 -Int pow96 +Int pow48 -Int 1
+    rule maskWordPackUInt48UInt48_2 => 115792089237316195423570985008687907853269984665640564039457583726438152929280 [macro] // pow256 -Int pow48
```
daejunpark, line 60: Having these hex numbers in the comment would be helpful to understand.
```
// maskWordPackUInt48UInt48_1: 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF000000000000FFFFFFFFFFFF
// maskWordPackUInt48UInt48_2: 0xFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFFF000000000000
```

tests/specs/mcd/verification.k
```diff
@@ -19,27 +19,54 @@ module VERIFICATION-SYNTAX
...
 
-    rule maxUInt48 &Int  UINT48_UINT48             => UINT48_1 requires UINT48_UINT48 ==Int #WordPackUInt48UInt48(UINT48_1, UINT48_2) andBool #rangeUInt(48, UINT48_1) andBool #rangeUInt(48, UINT48_2) [smt-lemma]
-    rule maxUInt48 &Int (UINT48_UINT48 /Int pow48) => UINT48_2 requires UINT48_UINT48 ==Int #WordPackUInt48UInt48(UINT48_1, UINT48_2) andBool #rangeUInt(48, UINT48_1) andBool #rangeUInt(48, UINT48_2) [smt-lemma]
+    rule  UINT48_1             |Int (maskWordPackUInt48UInt48_2 &Int UINT48_UINT48) => #WordPackUInt48UInt48( UINT48_1 , (maskWordPackUInt48UInt48_2 &Int UINT48_UINT48) /Int pow48 ) requires #rangeUInt(256, UINT48_UINT48) andBool #rangeUInt(48, UINT48_1)
+    rule (UINT48_2 *Int pow48) |Int (maxUInt48                  &Int UINT48_UINT48) => #WordPackUInt48UInt48( maxUInt48 &Int UINT48_UINT48 , UINT48_2                               ) requires #rangeUInt(256, UINT48_UINT48) andBool #rangeUInt(48, UINT48_2)
```
daejunpark, line 63: Isn't it `#rangeUInt(96, UINT48_UINT48)`?

The second rule might be fine even if the higher bits are non-zero, but the first rule would _not_.  The higher bits are not cleared by ANDing with `maskWordPackUInt48UInt48_2`.

daejunpark: BTW, these rules are quite confusing mainly because of the reverse order of the arguments being packed.  Would be the spec hard to read if we reverse the order of the arguments in the WordPack* terms?

tests/specs/mcd/verification.k
```diff
@@ -19,27 +19,54 @@ module VERIFICATION-SYNTAX
...
 
-    rule maxUInt160 &Int  ADDR_UINT8              => ADDR  requires ADDR_UINT8 ==Int #WordPackAddrUInt8(ADDR, UINT8) andBool #rangeAddress(ADDR) andBool #rangeUInt(8, UINT8) [smt-lemma]
-    rule maxUInt8   &Int (ADDR_UINT8 /Int pow160) => UINT8 requires ADDR_UINT8 ==Int #WordPackAddrUInt8(ADDR, UINT8) andBool #rangeAddress(ADDR) andBool #rangeUInt(8, UINT8) [smt-lemma]
+    rule (UINT48_2 *Int pow208) |Int (maxUInt208 &Int ADDR_UINT48_UINT48) => #WordPackAddrUInt48UInt48(maxUInt160 &Int ADDR_UINT48_UINT48, maxUInt48 &Int (ADDR_UINT48_UINT48 /Int pow160), UINT48_2) requires #rangeUInt(256, ADDR_UINT48_UINT48) andBool #rangeUInt(48, UINT48_2) andBool maxUInt48 &Int (ADDR_UINT48_UINT48 /Int pow160) =/=Int 0
+    rule (UINT48_2 *Int pow208) |Int (maxUInt208 &Int ADDR_UINT48_UINT48) => #WordPackAddrUInt48UInt48(maxUInt160 &Int ADDR_UINT48_UINT48, 0                                              , UINT48_2) requires #rangeUInt(256, ADDR_UINT48_UINT48) andBool #rangeUInt(48, UINT48_2) andBool maxUInt48 &Int (ADDR_UINT48_UINT48 /Int pow160)  ==Int 0
```
daejunpark, line 69: Could you please explain why this separation is needed?

tests/specs/mcd/verification.k
```diff
@@ -19,27 +19,54 @@ module VERIFICATION-SYNTAX
...
+    //
 
     syntax Int ::= #WordPackUInt48UInt48     (       Int , Int ) [function, no-evaluators, smtlib(WordPackUInt48UInt48)]
                  | #WordPackAddrUInt48UInt48 ( Int , Int , Int ) [function, no-evaluators, smtlib(WordPackAddrUInt48UInt48)]
                  | #WordPackAddrUInt8        (       Int , Int ) [function, no-evaluators, smtlib(WordPackAddrUInt8)]
  // -----------------------------------------------------------------------------------------------------------------
-    // rule #WordPackUInt48UInt48     (            UINT48_1 , UINT48_2 ) => UINT48_2 *Int pow48 +Int UINT48_1                        requires #rangeUInt(48, UINT48_1) andBool #rangeUInt(48, UINT48_2)
-    // rule #WordPackAddrUInt48UInt48 (     ADDR , UINT48_1 , UINT48_2 ) => UINT48_2 *Int pow208 +Int UINT48_1 *Int pow160 +Int ADDR requires #rangeAddress(ADDR) andBool #rangeUInt(48, UINT48_1) andBool #rangeUInt(48, UINT48_2)
-    // rule #WordPackAddrUInt8        (     ADDR , UINT8               ) => UINT8 *Int pow160 +Int ADDR                              requires #rangeAddress(ADDR) andBool #rangeUInt(8, UINT_8)
```
daejunpark, line 32: I think it would be good to keep these commented rules for the documentation purpose, as they give the "definition" of the WordPack* abstraction, as well as their definedness conditions.

daejunpark (commented):
Can we have some tests for the WordPack lemmas, as you did before for other tricky lemmas?

tests/specs/mcd/verification.k
ehildenb: I now mention it at the beginning of the WordPack explanation comment.

tests/specs/mcd/verification.k
ehildenb: Added.

tests/specs/mcd/verification.k
ehildenb: Good catch, I've changed it to `#rangeUInt(96, UINT48_UINT48)`.

I would prefer not to change the argument order for the WordPack, because that's the order the KLab currently is generating them in. It's not huge change to make for it, but would rather focus on getting more proofs passing.

If you really want me to change the argument order (instead of just leaving the comment about it), let me know.

tests/specs/mcd/verification.k
ehildenb: I added a comment/TODO explaining why this is needed on the Java backend.

tests/specs/mcd/verification.k
ehildenb: Added them back.

ehildenb:
@daejunpark I've addressed all your comments, except about adding tests. The tests that are added with this PR will not pass without these additional lemmas, and there are maybe 10 other MKR proofs which also won't pass with these lemmas. If they are bumped up to `tests/specs/work-pack.k` (so they can be used elsewhere), then I'm happy to add unit-tests for them, but for now I think the MKR integration tests should suffice.

daejunpark (approved):

PR #917, opened 2020-09-09
Upstream kevm hooks v2

sskeirik:
This PR is identical to #886 but the branch lives in the evm-semantics repo directly instead of on my forked repo. See #886 for discussion on relevant issues for this PR.

Moving the PR here resolves an issue with testing modified Jenkinsfiles --- as a security precaution, Jenkins CI does not test Jenkinsfiles updates from forks.


sskeirik:
@ehildenb @dwightguth So, continuing the discussion from #886, this PR integrates the blockchain plugin currently into evm-semantics in the following manner:

1) In the `Makefile`, the K include path was updated as follows:

    ```
    K_INCLUDE_PATH     := $(BUILD_LOCAL)/include/kframework/
    ```

2) The blockchain plugin was updated with an `install` target:

    ```
    PLUGIN_NAMESPACE := blockchain-k-plugin
    install:
            @mkdir -p $(PREFIX)/$(PLUGIN_NAMESPACE)
            cp plugin/* $(PREFIX)/$(PLUGIN_NAMESPACE)
    ```

3) The `Makefile` target `plugin-deps` was extended with an action:
    ```
    cd deps/plugin && \
        make PREFIX=$(K_INCLUDE_PATH) install
    ```

4) The `requires krypto.md` clauses in kevm source files were updated to `requires blockchain-k-plugin/krypto.md`

5) The Jenkinsfile was updated to do `make plugin-deps` before `make build ...`

Questions:
- With regards to step (5), it is possible to have the Makefile `build-XXX` targets depend on the plugin directly, so an extra build step in the Jenkinsfile is not needed. Do you guys have preferences?
- With regards to step (4), we have a potential issue that local includes and not distinct from global includes, i.e. the `#include "file.h"` vs. `#include <file.h>` --- and that package directory names (such as blockchain-k-plugin) could overlap with subdirectories included in user-defined semantics. Obviously, there are basically no packages now, so this isn't a big deal, but it seems like an potential future issue to me and I'd rather we all be aware of it.

Jenkinsfile
```diff
@@ -22,7 +22,13 @@ pipeline {
...
     stage('Build and Test') {
       stages {
-        stage('Build') { steps { sh 'make build RELEASE=true -j6' } }
+        stage('Build') {
+          steps { sh '''
+                    make plugin-deps
+                    make build RELEASE=true -j6
+                  '''
+          }
+        }
```
ehildenb, line 31:
```suggestion
        stage('Deps')  { steps { sh 'make plugin-deps'            } }
        stage('Build') { steps { sh 'make build RELEASE=true -j6' } }
```

Makefile
```diff
@@ -22,6 +22,7 @@ K_LIB := $(K_RELEASE)/lib/kframework
 export K_RELEASE
 
 LIBRARY_PATH       := $(LOCAL_LIB)
+K_INCLUDE_PATH     := $(BUILD_LOCAL)/include/kframework/
```
ehildenb, line 25:
```suggestion
K_INCLUDE_PATH     += :$(BUILD_LOCAL)/include/kframework/
```

To allow users to set their own `K_INCLUDE_PATH`, and just extend that one.

Makefile
```diff
@@ -94,7 +95,9 @@ K_JAR := $(K_SUBMODULE)/k-distribution/target/release/k/lib/java/kernel-1.0-SNAP
 deps: repo-deps
 repo-deps: k-deps plugin-deps
 k-deps: $(K_JAR)
-plugin-deps: $(PLUGIN_SUBMODULE)/client-c/main.cpp
+plugin-deps: $(PLUGIN_SUBMODULE)/client-c/main.cpp $(wildcard $(PLUGIN_SUBMODULE)/plugin/*)
```
ehildenb, line 98: Why do you need to add the `$(wildcard ...)`? At best it is depending on too many things. At worst, those files you're trying to pick up won't exist but Makefile will silently ignore the fact that they don't exist. If there are specific files you need to be present, list them here.

Alternatively, you can just remove all dependencies for this top-level `plugin-deps`.

sskeirik: I'm in favor of either being precise or not worrying about it. Since being precise becomes more complicated over time, I have removed explicit dependencies from this target.

Makefile
sskeirik: @ehildenb @dwightguth Currently, we use the variable `K_INCLUDE_PATH` as an argument to `kompile` on like 146 which says:

```
KOMPILE_OPTS += --hook-namespaces "$(HOOK_NAMESPACES)" -I $(K_INCLUDE_PATH)
```

However, it seems that K does not support the colon-delimited path inclusion convention. At least, after this change was integrated, the build seems to be broken. Is there a way to support this correctly, or did I make a mistake?

ehildenb: Ok, if we cannot use the `:` convention as expected, please just call it something slightly different, like `K_LOCAL_INCLUDE_PATH`.

sskeirik: Done.

Makefile
```diff
@@ -94,7 +96,9 @@ K_JAR := $(K_SUBMODULE)/k-distribution/target/release/k/lib/java/kernel-1.0-SNAP
 deps: repo-deps
 repo-deps: k-deps plugin-deps
 k-deps: $(K_JAR)
-plugin-deps: $(PLUGIN_SUBMODULE)/client-c/main.cpp
+plugin-deps:
+	cd deps/plugin && \
+	    make INSTALL_PREFIX=$(LOCAL_K_INCLUDE_PATH) install
```
ehildenb, line 101:
```suggestion
	    make INSTALL_PREFIX=$(BUILD_LOCAL) install
```

Makefile
```diff
@@ -140,7 +143,7 @@ tangle_haskell  := k            & ( ( ! ( concrete | nobytes ) ) | symbolic | by
 
 HOOK_NAMESPACES = KRYPTO JSON
 
-KOMPILE_OPTS += --hook-namespaces "$(HOOK_NAMESPACES)"
+KOMPILE_OPTS += --hook-namespaces "$(HOOK_NAMESPACES)" -I "$(LOCAL_K_INCLUDE_PATH)"
```
ehildenb, line 146:
```suggestion
KOMPILE_OPTS += --hook-namespaces "$(HOOK_NAMESPACES)" -I $(LOCAL_K_INCLUDE_PATH)
```

ehildenb (approved):

PR #923, opened 2020-09-17
MKR regression tests: flipper-addu48u48-fail-rough, vat-subui-fail-rough, vat-subui-pass-rough

ehildenb:
-   flipper-addu48u48-fail-rough: arithmetic modulo pow48
-   vat-subui-fail-rough and vat-subui-pass-rough: signed difference arithmetic

### DIFF
--- a/tests/specs/mcd/verification.k
+++ b/tests/specs/mcd/verification.k
@@ -320,4 +320,9 @@ module VERIFICATION
     rule sgn(chop(Y)) ==K 1 => 0 <=Int Y requires #rangeSInt(256, Y)
     rule X <Int X -Word chop(Y) => X <Int X -Word Y requires 0 <=Int Y andBool Y <=Int maxSInt256
 
+    // ### vat-subui-pass-rough
+
+    rule 0 <Int chop(Y) => true requires #rangeSInt(256, Y)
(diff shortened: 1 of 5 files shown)


tests/specs/mcd/verification.k
```diff
@@ -309,4 +309,17 @@ module VERIFICATION
...
+
+    // ### vat-subui-fail-rough
+
+    rule 0 s<Word X => bool2Word(sgn(X) ==Int 1 andBool X =/=Int 0) requires #rangeUInt(256, X)
+    rule sgn(chop(Y)) ==K 1 => 0 <=Int Y requires #rangeSInt(256, Y)
+
+    // ### vat-subui-pass-rough
+
+    rule X -Word chop(Y) => chop(X -Int Y)
+
```
daejunpark, line 324: z3 proofs for the last three lemmas: https://rise4fun.com/Z3/LFIz

daejunpark (approved):

tests/specs/mcd/verification.k
daejunpark: As I suggested before, please try to write a z3 proof for lemmas involving the modular arithmetic.

ehildenb: I've included your proofs in this PR. I figured these ones were simple enough it wasn't necessary. I do like the idea of each lemma being attached to a Z3 query that has to be discharged for the lemma to hold.

PR #941, opened 2020-11-22
Lemma for ERC20 + Solidity 5 + calling constructor

denis-bogdanas:


tests/specs/lemmas.k
```diff
@@ -110,6 +114,8 @@ module LEMMAS-JAVA [symbolic, kast]
 
     rule #asWord(BUF) /Int 26959946667150639794667015087019630673637144422540572481103610249216 => #asWord(#take(4, BUF)) [simplification]
 
+    rule #asWord(BUF1 : BUF2) >>Int LSHIFT => BUF1                      requires #sizeByteArray(BUF2) *Int 8 ==Int LSHIFT [simplification]
+
```
ehildenb, line 118: I'm guessing that this is just added for completeness on the Java backend? Because I htoughtt the ERC20 verifications only used the Haskell backend.

tests/specs/lemmas.k
```diff
@@ -57,6 +57,9 @@ module LEMMAS
     rule #padToWidth(32, #asByteStack(V)) => #buf(32, V)
       requires #rangeUInt(256, V)                                                                         [simplification]
 
+    rule #asWord(BUF1 ++ BUF2) >>Int LSHIFT => #asWord(BUF1)
+      requires #sizeByteArray(BUF2) *Int 8 ==Int LSHIFT                                                   [simplification]
```
ehildenb, line 61: Wouldn't it be more appropriate to name this variable `RSHIFT` and not `LSHIFT`?

ehildenb:
I have this lemma for MKR proofs, which I think generalizes all the lemmas here:

```
    rule #asWord(WS) >>Int M => #asWord(WS [ 0 .. #sizeByteArray(WS) -Int (M /Int 8) ]) requires 0 <=Int M andBool M modInt 8 ==Int 0 [simplification]
```

It will only apply when `M` is a multiple of `8`, and then it works for both Java and Haskell backends (because of additional lemmas for `_[_.._]` which already exist).

Would this lemma work for you here? Maybe instead we should be promoting lemmas from the existing proof-suites?

tests/specs/lemmas.k
denis-bogdanas: To make the test pass on Java backend.

tests/specs/lemmas.k
denis-bogdanas: Right, renamed.

denis-bogdanas:
Your lemma looks good. I  created another PR for this: https://github.com/kframework/evm-semantics/pull/942

If it passes, please merge that one and delete this.

ehildenb:
Closing in favor of #942

PR #945, opened 2020-12-09
Updates to infinite gas abstraction

ehildenb:
-   Adds lemmas for reasoning about #gas around different constructs (eg. Cmem, minInt, /Int).
-   Makes separate modules for Haskell/Java lemmas about infinite gas which requires the `concrete` attribute.
-   Adds tests of the infinite gas abstraction (inspired by making the end-pack-pass-rough and pot-join-pass-rough proofs pass).
-   Adds the end-pack-pass-rough proof.
-   #gas abstraction now counts gas down, instead of up.
-   #gas should behave like a proper positive infinity now.

### DIFF
--- a/tests/specs/functional/infinite-gas-spec.k
+++ b/tests/specs/functional/infinite-gas-spec.k
@@ -50,10 +50,7 @@ module INFINITE-GAS-SPEC
 
     claim <k> runLemma((VGas +Int -3844) -Int ((VGas -Int ((VGas +Int -4544) /Int 64)) +Int -3844)) => doneLemma((VGas +Int -4544) /Int 64) ... </k>
 
-    claim <k> runLemma(#gas((((((VGas +Int -1148) -Int G1) +Int -866) -Int G2) +Int -4337)))
-           => doneLemma(#gas(((VGas -Int G1) -Int G2) +Int -6351))
-          ...
-          </k>
+    claim <k> runLemma(#gas((((((VGas +Int -1148) -Int G1) +Int -866) -Int G2) +Int -4337))) => doneLemma(#gas(((VGas -Int G1) -Int G2) +Int -6351)) ... </k>
 
     claim <k> runLemma(#if _:Int ==Int 0 #then #gas(VGas -Int Csstore(_, _, _,_)) #else #gas(VGas +Int -344) #fi <Int 8)  => doneLemma(false) ... </k>
     claim <k> runLemma(8 <=Int #if _:Int ==Int 0 #then #gas(VGas -Int Csstore(_, _, _,_)) #else #gas(VGas +Int -344) #fi) => doneLemma(true)  ... </k>
--- a/tests/specs/functional/infinite-gas-spec.k
+++ b/tests/specs/functional/infinite-gas-spec.k
@@ -55,4 +55,10 @@ module INFINITE-GAS-SPEC
           ...
           </k>
 
+    claim <k> runLemma(#if _:Int ==Int 0 #then #gas(VGas -Int Csstore(_, _, _,_)) #else #gas(VGas +Int -344) #fi <Int 8)  => doneLemma(false) ... </k>
+    claim <k> runLemma(8 <=Int #if _:Int ==Int 0 #then #gas(VGas -Int Csstore(_, _, _,_)) #else #gas(VGas +Int -344) #fi) => doneLemma(true)  ... </k>
+
+    claim <k> runLemma(#if _:Int ==Int 0 #then #gas(VGas -Int Csstore(_, _, _,_)) #else #gas(VGas +Int -344) #fi <Int minInt(#if _ #then #gas(_) #else #gas(_) #fi, #gas(_)))  => doneLemma(false) ... </k>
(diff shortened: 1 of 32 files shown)


tests/specs/functional/infinite-gas-spec.k
```diff
@@ -33,4 +33,29 @@ module INFINITE-GAS-SPEC
...
+
+    claim <k> runLemma((VGas +Int -3844) -Int ((VGas -Int ((VGas +Int -4544) /Int 64)) +Int -3844)) => doneLemma((VGas +Int -4544) /Int 64) ... </k>
+
+    claim <k> runLemma(#gas((((((VGas +Int -1148) -Int G1) +Int -866) -Int G2) +Int -4337))) => doneLemma(#gas(((VGas -Int G1) -Int G2) +Int -6351)) ... </k>
+
+    claim <k> runLemma(#if _:Int ==Int 0 #then #gas(VGas -Int Csstore(_, _, _,_)) #else #gas(VGas +Int -344) #fi <Int 8)  => doneLemma(false) ... </k>
+    claim <k> runLemma(8 <=Int #if _:Int ==Int 0 #then #gas(VGas -Int Csstore(_, _, _,_)) #else #gas(VGas +Int -344) #fi) => doneLemma(true)  ... </k>
+
+    claim <k> runLemma(#if _:Int ==Int 0 #then #gas(VGas -Int Csstore(_, _, _,_)) #else #gas(VGas +Int -344) #fi <Int minInt(#if _ #then #gas(_) #else #gas(_) #fi, #gas(_)))  => doneLemma(false) ... </k>
+    claim <k> runLemma(minInt(#if _ #then #gas(_) #else #gas(_) #fi, #gas(_)) <=Int #if _:Int ==Int 0 #then #gas(VGas -Int Csstore(_, _, _,_)) #else #gas(VGas +Int -344) #fi) => doneLemma(true)  ... </k>
```
denis-bogdanas, line 59: Again, I don't understand how these lemmas can be proved. The `if _ == 0` must be undecided.

tests/specs/infinite-gas.k
```diff
@@ -1,35 +1,96 @@
...
+
+    rule #gas(#gas(G)) => #gas(G) [simplification]
+
+    rule #gas(G) -Int #gas(G') => #gas(G -Int G') [simplification]
+
+    rule minInt(#gas(G), #gas(G'))              => #gas(minInt(G, G'))              [simplification]
+    rule #if B #then #gas(G) #else #gas(G') #fi => #gas(#if B #then G #else G' #fi) [simplification]
+
+    rule #gas(_)  <Int #gas(_) => false [simplification]
+    rule #gas(_) <=Int #gas(_) => true  [simplification]
```
denis-bogdanas, line 86: Can you explain these 2 lemmas in a comment?

denis-bogdanas: Also, in what kind of scenarios are they useful?

tests/specs/functional/infinite-gas-spec.k
```diff
@@ -33,4 +33,29 @@ module INFINITE-GAS-SPEC
...
+    claim <k> runLemma(3 <Int minInt(#gas(VGas), 2)) => doneLemma(false) ... </k>
+    claim <k> runLemma(2 <Int minInt(#gas(VGas), 3)) => doneLemma(true ) ... </k>
+
+    claim <k> runLemma(#gas(G) -Int #gas(G')) => doneLemma(#gas(G -Int G')) ... </k>
+
+    claim <k> runLemma(#gas(G) -Int Csstore(_, _, _, _) <Int 2) => doneLemma(false) ... </k>
+    claim <k> runLemma(minInt(#gas(G), #gas(G'')) +Int -2522 <Int Csstore(_, _, _, _)) => doneLemma(false) ... </k>
+
+    claim <k> runLemma(#gas(G) <Int #gas(G' +Int 700))  => doneLemma(false) ... </k>
+    claim <k> runLemma(#gas(G' +Int 700) <=Int #gas(G)) => doneLemma(true)  ... </k>
```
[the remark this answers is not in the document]
ehildenb: See https://github.com/kframework/evm-semantics/pull/945#discussion_r541680741

tests/specs/functional/infinite-gas-spec.k
ehildenb: If both branches of the `#if` are +Inf gas, then the entire expression will always be infinite gas. Same with `minInt`.

We have to make sure that `#gas` always "bubbles up" to the top of the expression in the `<gas>` cell, so that when we compare any particular value to it during execution we never run out of gas.

tests/specs/functional/infinite-gas-spec.k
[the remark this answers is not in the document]
ehildenb: The values under the `#gas` do not matter, they are _only_ used for tracking the gas usage as metadata. `#gas(_)` with any argument is the same positive infinity as any other `#gas(_)`

tests/specs/infinite-gas.k
denis-bogdanas: Thanks! All makes sense now. I think it's useful to add your explanation as comment in k file.

denis-bogdanas (approved):

tests/specs/infinite-gas.k
ehildenb: I added a comment with this explanation to the documentation at the top of the file.

PR #952, opened 2020-12-16
Improved infinite gas reasoning

ehildenb:
This PR does several things:

-   Makes sure that `#allBut64th` is always a positive number (for simplifying the infinite gas reasoning).
-   Leaves `#allBut64th` unevaluated to keep expressions small.
-   Loosens the bounds on the infinite gas simplifications (0 <Int G => 0 <=Int G in many cases).
-   Generalizes the lemmas for re-associating gas expressions so they apply in more places.

The added tests are inspired by several MKR proofs, but as they are not necessary for making them pass (just make the gas expressions generated significantly smaller which leads to up to 10% performance improvement), I decided not to add the MKR tests themselves.

### DIFF
--- a/tests/specs/infinite-gas.k
+++ b/tests/specs/infinite-gas.k
@@ -56,6 +56,7 @@ module INFINITE-GAS-COMMON
 
     rule I -Int I => 0                                                                   [simplification]
     rule I1 -Int (I1 -Int I2) => I2                                                      [simplification]
+    rule (I1 -Int I2) +Int I2 => I1                                                      [simplification]
     rule (notBool (A andBool B)) andBool A => (notBool B) andBool A                      [simplification]
     rule #if B #then C +Int C1 #else C +Int C2 #fi => C +Int #if B #then C1 #else C2 #fi [simplification]
 
--- a/tests/specs/infinite-gas.k
+++ b/tests/specs/infinite-gas.k
@@ -68,16 +68,16 @@ module INFINITE-GAS-COMMON
  // ---------------------------------------------------------------------------------------------------------
     rule #gas(_, _, _) => #gas(0)
 
-    rule #gas(G) +Int G' => #gas(G +Int G') requires 0 <Int G' orBool 0 -Int G' <Int #gas(G)  [simplification]
-    rule G +Int #gas(G') => #gas(G +Int G') requires 0 <Int G  orBool 0 -Int G  <Int #gas(G') [simplification]
+    rule #gas(G) +Int G' => #gas(G +Int G') requires 0 <=Int G' orBool 0 -Int G' <Int #gas(G)  [simplification]
+    rule G +Int #gas(G') => #gas(G +Int G') requires 0 <=Int G  orBool 0 -Int G  <Int #gas(G') [simplification]
 
-    rule #gas(G) -Int G' => #gas(G -Int G') requires 0 <Int G' andBool G' <Int #gas(G) [simplification]
+    rule #gas(G) -Int G' => #gas(G -Int G') requires 0 <=Int G' andBool G' <Int #gas(G) [simplification]
 
-    rule #gas(G) *Int G' => #gas(G *Int G') requires 0 <Int G' [simplification]
-    rule G *Int #gas(G') => #gas(G *Int G') requires 0 <Int G  [simplification]
+    rule #gas(G) *Int G' => #gas(G *Int G') requires 0 <=Int G' [simplification]
+    rule G *Int #gas(G') => #gas(G *Int G') requires 0 <=Int G  [simplification]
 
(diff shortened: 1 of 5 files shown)


tests/specs/infinite-gas.k
```diff
@@ -114,11 +135,18 @@ module INFINITE-GAS-COMMON
...
 
     rule #gas(_) <Int Cmem(_, _)  => false [simplification]
     rule Cmem(_, _) <=Int #gas(_) => true  [simplification]
+
+    rule 0 <=Int #allBut64th(G)       => true                          [simplification]
+    rule #allBut64th(G) <Int #gas(G') => true requires G <Int #gas(G') [simplification]
+
+    rule #gas(G1) +Int (G2 +Int G3) => (#gas(G1) +Int G2) +Int G3 [simplification]
+    rule #gas(G1) +Int (G2 -Int G3) => (#gas(G1) +Int G2) -Int G3 [simplification]
+
```
ehildenb, line 151: Redundant, should be removed.

tests/specs/infinite-gas.k
```diff
@@ -44,19 +50,32 @@ module INFINITE-GAS-JAVA [kast]
...
+    rule (notBool (A andBool B)) andBool A => (notBool B) andBool A                      [simplification]
+    rule #if B #then C +Int C1 #else C +Int C2 #fi => C +Int #if B #then C1 #else C2 #fi [simplification]
+
+    rule I -Int I => 0              [simplification]
+    rule (I1 -Int I2) +Int I2 => I1 [simplification]
+
+    rule I1 +Int (I2 +Int I3) => (I1 +Int I2) +Int I3 [simplification]
+    rule I1 +Int (I2 -Int I3) => (I1 +Int I2) -Int I3 [simplification]
+    rule I1 -Int (I2 +Int I3) => (I1 -Int I2) -Int I3 [simplification]
+    rule I1 -Int (I2 -Int I3) => (I1 -Int I2) +Int I3 [simplification]
```
denis-bogdanas, line 78: this combined with lines 34-37 may lead to infinite loop

denis-bogdanas: Otherwise all good.

denis-bogdanas (changes requested):

tests/specs/infinite-gas.k
ehildenb: I've adjusted the rules to make it so that infinite loops cannot happen I think, please have another look @denis-bogdanas

tests/specs/infinite-gas.k
```diff
@@ -44,19 +52,34 @@ module INFINITE-GAS-JAVA [kast]
...
```
denis-bogdanas, line 82: If there wouldn't be rule priorities, this would still lead to infinite loop in combination with rule on L29.

In practice infinite loop probably won't happen, but for rather peculiar reasons. In Haskell backend definition rules are applied before `[simplification]`, so concrete `(I2+I3)` will evaluate first. in Java although it's not documented, builtin evaluation happens before rules are applied.

So I suggest either the following comment:
```
Without rule priorities, these rules could lead to infinite loop in combination with rule
(add a label to lemma on L29). However, Haskell backend executes definition rules 
before simplification rules, 
and Java backend executes builtin evaluations like concrete `+Int` before any rules.
```

Or alternatively, you'll have to duplicate the rules above for Java and Haskell and frame then with some concrete/symbolic conditions.

denis-bogdanas (changes requested):

tests/specs/infinite-gas.k
ehildenb: Can you come up with an example where this infinite loops with the rules on lines 29/55? I can't see it.

denis-bogdanas: `(1+1)+A => A+(1+1) (L29) => (A+1)+1 (L79) => (1+1)+A (L36)`

Besides, we have a version of such rules in verified smart contracts repo. There I used this naming convention for clarity. Concrete vars are called I1, I2 etc, symbolic vars A1, A2, ...

ehildenb:
@denis-bogdanas thanks for the detailed review, I suspected that there were infinite loop problems like you pointed out, but was having trouble finding them.

I've updated the code with your suggestions (naming concrete and symbolic variables nicer), and I think I now am avoiding infinite loops. Please have another look.

tests/specs/infinite-gas.k
ehildenb: Thanks for the example. I reviewed this example with teh updated code, and while it looks like there are multiple execution paths, none of them loop and they all arrive at the same answer.

denis-bogdanas (approved):

PR #953, opened 2021-01-05
Update plugin submodule

sskeirik:
See related PR: https://github.com/runtimeverification/blockchain-k-plugin/pull/127

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -161,10 +161,15 @@ STANDALONE_KOMPILE_OPTS := -L$(LOCAL_LIB) -I$(K_RELEASE)/include/kllvm  \
                            $(PLUGIN_SUBMODULE)/plugin-c/plugin_util.cpp \
                            $(PLUGIN_SUBMODULE)/plugin-c/crypto.cpp      \
                            $(PLUGIN_SUBMODULE)/plugin-c/blake2.cpp      \
-                           -g -std=c++14 -lff -lcryptopp -lsecp256k1
+                           -g -std=c++14 -lff -lcryptopp -lsecp256k1    \
+                           -lssl -lcrypto
 
 ifeq ($(UNAME_S),Linux)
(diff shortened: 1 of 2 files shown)


sskeirik:
@ehildenb Dwight approved the plugin changes upstream. If you approve the changes on this side, we can:

- update the plugin master branch
- update this PR so that:
  - submodule points to the new master
  - rebase on KEVM master if needed
  - merge this PR in so that KEVM is based on master

Does that work for you?

Makefile
```diff
@@ -161,11 +161,16 @@ STANDALONE_KOMPILE_OPTS := -L$(LOCAL_LIB) -I$(K_RELEASE)/include/kllvm  \
...
+                           -g -std=c++14 -lff -lcryptopp -lsecp256k1    \
+                           -lssl -lcrypto
 
 ifeq ($(UNAME_S),Linux)
     STANDALONE_KOMPILE_OPTS += -lprocps
 endif
+ifeq ($(UNAME_S),Darwin)
+    OPENSSL_ROOT := $(shell brew --prefix openssl)
+    STANDALONE_KOMPILE_OPTS += -I/usr/local/include -L/usr/local/lib -I$(OPENSSL_ROOT)/include -L$(OPENSSL_ROOT)/lib
+endif
```
ehildenb, line 173: If these changes make for any difference in system dependencies, can you make sure the README is updated accordingly? May be that we already have them listed there.

sskeirik: The only difference is that we include build flags for openssl --- openssl should already be a dependency of our dependencies. I have double-checked that is the case for Ubuntu and Arch Linux instructions.

macOS instructions are slightly different; I couldn't tell if they worked or not. In any case, the macOS instructions were actually stale. I updated them to the current version and added openssl as an explicit dependency.

ehildenb (approved):

PR #958, opened 2021-01-13
More MKR regression proofs with simplification lemmas

ehildenb:
-   Adds the following proofs: flapper-tend-guy-diff-pass-rough, gemjoin-exit-pass-rough, vow-cage-deficit-pass-rough, vat-fold-pass-rough, vat-fork-diff-pass-rough
-   Adds some rules to `infinite-gas.k` where subtractions of concrete numbers were not being moved outermost/rightmost in the term.
-   Changes the lemmas for `vat-frob-diff-zero-pass-rough` to only apply in more specific circumstances, to avoid generating larger terms (keep the `chop` instead of simplifying to `+Int pow256`).
-   Integer simplification lemmas added for some specific proofs.
-   Map update lemma where multiple writes to the same storage location are simplified out to just the latter write.
-   Lemma simplifying `chop` expressions.
-   Inline the definition of `#sgnInterp` in the handful of places that it is used.
-   Simplify the definition of `-Word` to a single rule.

### DIFF
--- a/evm-types.md
+++ b/evm-types.md
@@ -224,8 +224,7 @@ Warning: operands are assumed to be within the range of a 256 bit EVM word. Unbo
                  | Int "%Word" Int [function, functional]
  // -----------------------------------------------------
     rule W0 +Word W1 => chop( W0 +Int W1 )
-    rule W0 -Word W1 => W0 -Int W1 requires W0 >=Int W1
-    rule W0 -Word W1 => chop( (W0 +Int pow256) -Int W1 ) requires W0 <Int W1
+    rule W0 -Word W1 => chop( W0 -Int W1 )
     rule W0 *Word W1 => chop( W0 *Int W1 )
     rule  _ /Word W1 => 0            requires W1  ==Int 0
     rule W0 /Word W1 => W0 /Int W1   requires W1 =/=Int 0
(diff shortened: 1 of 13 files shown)


ehildenb:
I've additionally asked @daejunpark to review this one, because it contains a change to the semantics.

denis-bogdanas (approved):

evm-types.md
```diff
@@ -253,14 +252,10 @@ The helper `powmod` is a totalization of the operator `_^%Int__` (which comes wi
...
-
-    syntax Int ::= #sgnInterp ( Int , Int ) [function, functional]
- // --------------------------------------------------------------
-    rule #sgnInterp( W0 ,  _ ) => 0          requires W0 ==Int 0
-    rule #sgnInterp( W0 , W1 ) => W1         requires W0 >Int 0
-    rule #sgnInterp( W0 , W1 ) => 0 -Word W1 requires W0 <Int 0
+    rule [divSWord.same]: W0 /sWord W1 =>          abs(W0) /Word abs(W1)  requires sgn(W0) *Int sgn(W1) ==Int  1
+    rule [divSWord.diff]: W0 /sWord W1 => 0 -Word (abs(W0) /Word abs(W1)) requires sgn(W0) *Int sgn(W1) ==Int -1
+    rule [modSWord.pos]:  W0 %sWord W1 =>          abs(W0) %Word abs(W1)  requires sgn(W0) ==Int  1
+    rule [modSWord.neg]:  W0 %sWord W1 => 0 -Word (abs(W0) %Word abs(W1)) requires sgn(W0) ==Int -1
```
daejunpark, line 258: Don't you need the rule for the case of `sgn(W0) *Int sgn(W1) == 0` and `sgn(W0) == 0`?

Although it is an undefined case, you had a rule for that before, but not now.  Just to make sure you know this.

daejunpark:
I only reviewed the semantics changes.  It looks good and I put a comment.

evm-types.md
ehildenb: Yeah I'm aware of that change, but it seems OK because both `_/sWord_` and `_%sWord_` were not defined as `functional` before, and they still aren't defined as such.

So I figured it was OK to just define the sensible cases here, because the proofs still pass. I can add in the other cases (going to 0) later if I find that it's needed to make those `functional`.

ehildenb: In the case of `_/sWord_` and `_%sWord_` we know that the semantics should never feed it anything that is not an EVM Word, so it should always be defined at runtime.

PR #964, opened 2021-01-24
edsl: Workaround to make #buf act like [functional]

denis-bogdanas:
Fixes https://github.com/kframework/kore/issues/2358 for erc20-verification.

Tom recommends Splitting `#buf` into 2 functions: `#bufStrict` which is partial and rewrites into `#buf` which is total and evaluated modulo `2^(8*BUF_LEN)`. This way engine will deal with total `#buf` most of the time and will avoid some complications.

Changes:
- now #buf(LEN, DATA) is [functional]. When DATA overflows LEN, it is interpreted modulo 2 ^Int (LEN *Int 8).
- To preserve correctness, an additional #bufStrict function is added, non-functional, with semantics
`#bufStrict(DATA,LEN) => #buf(DATA,LEN) requires DATA in range`
- When a buffer is first created, e.g. claim LHS and rule RHS (when LHS doesn't contain a buffer) it is always created as `#bufStrict`. When both rule LHS and RHS contains a `#buf`, it is not changed as this is not a "new" buffer.
- Claim RHSs that contained `#buf`were not changed, as it is expected that any created #bufStrict eventually rewrites to `#buf`.
- Some benchmark proofs needed a range side condition and a few lemmas.

@ehildenb I added some lemmas and a test for them. The lemmas are needed for benchmark tests. Because they match low-level arithmetic instructions, even if in a rather specific pattern, I'm not sure about their performance implications. So to be safe I want to have them only under Java backend. But I'm not sure if the test is really run the way I placed it right now. The module `LEMMAS-SPEC` now imports `LEMMAS-JAVA-SPEC` which has the new lemma spec. Please check this part in particular.

### DIFF
--- a/edsl.md
+++ b/edsl.md
@@ -358,7 +358,12 @@ If the data is at most 31 bytes long, it is stored in the higher-order bytes (le
 
 ```k
 endmodule
+```
 
+Byte-Buffer Abstraction
+-----------------------
+
+```k
 module BUF-SYNTAX
     imports EVM
 ```
--- a/edsl.md
+++ b/edsl.md
@@ -11,8 +11,7 @@ requires "evm.md"
 
 module EDSL
     imports EVM
-    imports BUFSTRICT-SYNTAX
-    imports BUFSTRICT-HASKELL
+    imports BUF
 ```
 
 ### ABI Call Data
@@ -178,21 +177,6 @@ where `F1 : F2 : F3 : F4` is the (two's complement) byte-array representation of
     syntax ByteArray ::= #encBytes ( Int , ByteArray ) [function]
  // -------------------------------------------------------------
     rule #encBytes(N, BS) => #enc(#uint256(N)) ++ BS ++ #bufStrict(#ceil32(N) -Int N, 0)
-
-    //Byte array buffer. Lemmas defined in evm-data-symbolic.k
-    //Appears in this form only when when #buf is first created, as part of `#enc` logic.
-    // SIZE, DATA // left zero padding
-    // Definition in module BUFSTRICT-SYNTAX
-    //syntax ByteArray ::= #bufStrict ( Int , Int ) [function]
- // ---------------------------------------------------------------
-    rule #bufStrict(SIZE, DATA) => #buf(SIZE, DATA)
-      requires #range(0 <= DATA < (2 ^Int (SIZE *Int 8)))
-
-    syntax ByteArray ::= #buf ( Int , Int ) [function, functional, smtlib(buf)]
-    // Workaround to have #buf functional is to extend the domain of the original #buf by interpreting DATA modulo 2 ^Int (SIZE *Int 8).
-    // Terms #buf are always initially created from #bufStrict
-    rule #buf(SIZE, DATA) => #padToWidth(SIZE, #asByteStack(DATA %Int (2 ^Int (SIZE *Int 8))))
-      [concrete]
 ```
 
 ```k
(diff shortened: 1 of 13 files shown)


ehildenb:
I only see a few of the `tests/specs/...` tests updated, have you checked them all to see if any of the others need updated?

denis-bogdanas:
I reviewed all instances of `#buf`. That's all that needs updating.

edsl.md
```diff
@@ -365,4 +374,18 @@ If the data is at most 31 bytes long, it is stored in the higher-order bytes (le
...
+
+endmodule
+
+module BUFSTRICT-HASKELL [symbolic, kore]
+    imports BUFSTRICT-SYNTAX
+
+    rule #bufStrict(_, _) => #Bottom              [owise]
+
+endmodule
 ```
```
ehildenb, line 391: I want to make new modules `BUF-SYNTAX` and `BUF-HASKELL`, and `BUF-JAVA` here, and to have `EDSL` import `BUF-HASKELL` and `BUF-JAVA`. All declarations and rules for `#buf` and `#bufStrict` will go here. Also add the documentation (markdown style, not inline comments) to those modules explaining the proper way to use them, and how the backends expcet them to be used? (eg. do I use #buf on LHS or RHS of rule? Sholud I use #bufStrict on RHS of claim? etc...)

edsl.md
```diff
@@ -365,4 +358,50 @@ If the data is at most 31 bytes long, it is stored in the higher-order bytes (le
 
 ```k
 endmodule
+
+module BUF-SYNTAX
+    imports EVM
+```
```
ehildenb, line 364:
`````suggestion
```

Byte-Buffer Abstraction
-----------------------

```k
module BUF-SYNTAX
    imports EVM
```
`````

File: tests/specs/functional/lemmas-spec.k [hunk 3]
```diff
@@ -122,3 +123,10 @@ module LEMMAS-SPEC
            => doneLemma( true ) ... </k>
 
 endmodule
+
+module LEMMAS-JAVA-SPEC [symbolic, kast]
+    imports VERIFICATION
+
+    claim <k> runLemma ( #bufStrict( #ceil32(LEN) -Int LEN, 0 ) ) => doneLemma( #buf( #ceil32(LEN) -Int LEN, 0 ) ) ... </k>
+
+endmodule
```
ehildenb, line 132: Why is this only run with the Java backend? Also, are we sure that this actually even gets run as a lemma? I've never seen claims imported into the main verification module.

tests/specs/lemmas.k
```diff
@@ -267,6 +267,12 @@ module LEMMAS-JAVA [symbolic, kast]
     rule #buf( 32, DATA ) ==K BA:ByteArray     => DATA ==Int #asInteger(BA)
       requires #sizeByteArray(BA) <=Int 32                                      andBool  #isConcrete(BA) [simplification]
 
+    //For #bufStrict simplification in benchmarks
+    rule 0 <Int 2 ^Int I           => true
+      requires 0 <=Int I                                                        [simplification]
+    rule 0 <=Int I *Int 8          => 0 <=Int I                                 [simplification]
```
ehildenb, line 273: Is it required that we rewrite the actual condition? Could we get away with instead `rule 0 <=Int I *Int 8 => true requires 0 <=Int I [simplification]`, so that we don't lose the original constraint form? Maybe some other lemmas are based on the `I *Int 8` showing up on the RHS of the `<=Int`.

tests/specs/lemmas.k
```diff
@@ -267,6 +267,12 @@ module LEMMAS-JAVA [symbolic, kast]
...
+    rule 0 <=Int #ceil32(I) -Int I => true                                      [simplification]
```
ehildenb, line 274: Is this true if `I <Int 0`? I don't think so. Maybe you need a side-condition that `0 <=Int I`?

tests/specs/lemmas.k
denis-bogdanas: Rules with no side conditions are preferred for performance. Once they match the are only evaluated once. A rule with side condition that doesn't match will be tried by the engine every time term constraint changes. I can rewrite as you suggested, please confirm.

tests/specs/functional/lemmas-spec.k
denis-bogdanas: I was not sure of this as well, as I explaind in the description. Fixed by adding a new proof file `lemmas-java-spec.k`.

It is only run on Java because lemmas were only added on Java, for performance.

tests/specs/lemmas.k
denis-bogdanas: thanks, fixed

daejunpark (approved):

PR #968, opened 2021-02-12
Refactor EDSL to separate into different modules

ehildenb:
-   Split the file `edsl.md` into three files `buf.md`, `hashed-locations.md`, and `abi.md`, to make it easier to include one but not the others.
-   `edsl.md` now includes the three sub-files, and combines them into the `EDSL` module as before.
-   Updates the `concrete-rules.txt` files for the changes.
-   The `BUF-KAST` module is removed, because it contained no rules. You can just include `BUF` directly.


buf.md
```diff
@@ -0,0 +1,47 @@
...
+    rule #bufStrict(SIZE, DATA) => #buf(SIZE, DATA)
+      requires #range(0 <= DATA < (2 ^Int (SIZE *Int 8)))
+
+    rule #buf(SIZE, DATA) => #padToWidth(SIZE, #asByteStack(DATA %Int (2 ^Int (SIZE *Int 8))))
+      [concrete]
+
+endmodule
+
+module BUF-KORE [kore, symbolic]
+    imports BUF
```
denis-bogdanas, line 42: The import relation must be opposite: BUF must import BUF-KORE. This is the pattern used in domains.k . Otherwise any module that needs BUF would also have to import BUF-KORE.

denis-bogdanas (changes requested):

buf.md
denis-bogdanas: The rest looks good.

ehildenb: The problem is that when compiling with the LLVM backend (which is needed for including `abi.md` into definitions like Firefly), then you have to say `imports BUF`, which ignores the `BUF-KORE` module, and then ignores the `BUF-SYNTAX` module, so then it complains that `#buf` and `#bufStrict` do not exist. Another solution is to have `BUF` import both `BUF-KORE` and `BUF-SYNTAX`, if you're OK with that..

denis-bogdanas: Yes, that' what I meant. Relation must be BUF imports BUF-KORE,BUF-SYNTAX ; BUF-KORE imports BUF-SYNTAX.

denis-bogdanas (approved):

PR #970, opened 2021-02-18
fixed issue with kevm removing --debug flag before it is passed to the frontend

MirceaS:


ehildenb:
@MirceaS instead you can change the case block around line 224 to say something like this:

```
diff --git a/kevm b/kevm
index 47493f0aa..bea315d9c 100755
--- a/kevm
+++ b/kevm
@@ -223,7 +223,7 @@ args=()
 while [[ $# -gt 0 ]]; do
     arg="$1"
     case $arg in
-        --debug)              debug=true        ; shift   ;;
+        --debug)              debug=true ; args+=("$1") ; shift   ;;
         --dump)               dump=true                 ; shift   ;;
         --no-unparse)         unparse=false             ; shift   ;;
         --repl)               repl=true                 ; shift   ;;
```

ehildenb:
Just make sure to format it nicely (line up the semicolons and stuff)

ehildenb (approved):

PR #986, opened 2021-04-02
Focal package and Dockerhub images

ehildenb:


ehildenb:
Evidence this works: https://office.runtimeverification.com/jenkins/blue/organizations/jenkins/evm-semantics/detail/PR-986/11/pipeline/

package/docker/Dockerfile.ubuntu
```diff
@@ -0,0 +1,28 @@
+ARG BASE_IMAGE=bionic
+FROM ubuntu:${BASE_IMAGE}
+ARG BASE_IMAGE=bionic
```
gtrepta, line 3: Why is `ARG BASE_IMAGE=bionic` here twice?

gtrepta (approved):

package/docker/Dockerfile.ubuntu
ehildenb: In a `Dockerfile`, anytime there is a `FROM` clause, all `ARG` are reset. So you need the one before, and the one after, to make sure that `BASE_IMAGE` is available both for the `FROM` clause and after the `FROM` clause.

PR #989, opened 2021-04-13
KEVM runner script can kompile

ehildenb:
This PR does:

-  Add a `kevm kompile` command, which will call `kompile` with the correct arguments to include the KEVM build in.
-  Removes the `ceil` tangle tags, which are no longer used.

### DIFF
--- a/kevm
+++ b/kevm
@@ -30,7 +30,10 @@ export KLAB_OUT="${KLAB_OUT:-~/.klab}"
 # User Commands
 
 run_kompile() {
-   kompile --backend "$backend" "$run_file" -I ${INSTALL_LIB}/include/kframework "$@"
+   kompile --backend "$backend" "$run_file"    \
+        -I "${INSTALL_LIB}/include/kframework" \
+        --hook-namespaces "$hook_namespaces"   \
+        "$@"
 }
 
 run_krun() {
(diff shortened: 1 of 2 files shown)


gtrepta:
@ehildenb Can you add documentation to `kevm --help` before I review?

ehildenb:
@gtrepta good catch, done.

kevm
```diff
@@ -183,6 +190,7 @@ if [[ "$run_command" == 'help' ]] || [[ "$run_command" == '--help' ]] ; then
                $0 kast         [--backend (llvm|java|haskell)]           <KEVM_arg>* <pgm>  <output format> <K arg>*
                $0 prove        [--backend (java|haskell)]                            <spec> <def_module> <K arg>*
                $0 search       [--backend (java|haskell)]                            <pgm>  <pattern> <K arg>*
+               $0 kompile      [--backend (llvm|haskell)]                            <main> <K arg>*
```
gtrepta, line 193:
```suggestion
               $0 kompile      [--backend (java|llvm|haskell)]                       <main> <K arg>*
```

It looks like it should be like this?

ehildenb: Sure, I'm just in general trying not to spread use of the java backend, but makes sense.

gtrepta (approved):

PR #991, opened 2021-04-26
Updates to KEVM for ERC20 verification tool

ehildenb:
- Use set-union instead of set-concat for `<touchedAccounts>` cell, to avoid definedness issues when doing symbolic reason about them.
- Clears up some of the unused variable warnings in the lemmas and proofs.
- Adjusts the merkle-spec proofs to not have undefined initial states.
- Sequences the computation of `merkleExtensionBuilder` so that when we're accessing the first byte of a given path we already know it exists.
- Adds `runLemma/doneLemma` wrappers for some other functional proofs.

### DIFF
--- a/evm.md
+++ b/evm.md
@@ -1284,7 +1284,7 @@ The various `CALL*` (and other inter-contract control flow) operations will be d
          </k>
 
     rule <k> #mkCall ACCTFROM ACCTTO ACCTCODE BYTES APPVALUE ARGS STATIC:Bool
-          => #loadProgram BYTES ~> #initVM ~> #precompiled?(ACCTCODE, SCHED) ~> #execute
+          => #touchAccounts ACCTFROM ACCTTO ~> #loadProgram BYTES ~> #initVM ~> #precompiled?(ACCTCODE, SCHED) ~> #execute
          ...
          </k>
          <callDepth> CD => CD +Int 1 </callDepth>
@@ -1295,7 +1295,6 @@ The various `CALL*` (and other inter-contract control flow) operations will be d
          <callGas> GCALL => 0 </callGas>
          <caller> _ => ACCTFROM </caller>
          <static> OLDSTATIC:Bool => OLDSTATIC orBool STATIC </static>
-         <touchedAccounts> ... .Set => SetItem(ACCTFROM) SetItem(ACCTTO) ... </touchedAccounts>
          <schedule> SCHED </schedule>
 
     syntax InternalOp ::= "#precompiled?" "(" Int "," Schedule ")"
@@ -1317,6 +1316,13 @@ The various `CALL*` (and other inter-contract control flow) operations will be d
     rule <k> #loadProgram BYTES => . ... </k>
          <program> _ => BYTES </program>
          <jumpDests> _ => #computeValidJumpDests(BYTES) </jumpDests>
+
+    syntax KItem ::= "#touchAccounts" Account | "#touchAccounts" Account Account
+ // ----------------------------------------------------------------------------
+    rule <k> #touchAccounts ADDR1 ADDR2 => #touchAccounts ADDR1 ~> #touchAccounts ADDR2 ... </k>
+
+    rule <k> #touchAccounts ADDR => . ... </k>
+         <touchedAccounts> TOUCHED_ACCOUNTS => TOUCHED_ACCOUNTS |Set SetItem(ADDR) </touchedAccounts>
 
     syntax Set ::= #computeValidJumpDests(ByteArray)            [function, memo]
                  | #computeValidJumpDests(ByteArray, Int, List) [function, klabel(#computeValidJumpDestsAux)]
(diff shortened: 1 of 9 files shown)


evm.md
```diff
@@ -1326,6 +1325,13 @@ The various `CALL*` (and other inter-contract control flow) operations will be d
          <program> _ => BYTES </program>
          <jumpDests> _ => #computeValidJumpDests(BYTES) </jumpDests>
 
+    syntax KItem ::= "#touchAccounts" Account | "#touchAccounts" Account Account
+ // ----------------------------------------------------------------------------
+    rule <k> #touchAccounts ADDR1 ADDR2 => #touchAccounts ADDR1 ~> #touchAccounts ADDR2 ... </k>
+
+    rule <k> #touchAccounts ADDR => . ... </k>
+         <touchedAccounts> TOUCHED_ACCOUNTS => TOUCHED_ACCOUNTS |Set SetItem(ADDR) </touchedAccounts>
```
ana-pantilie, line 1333: Not really related to your changes, but what does the `touchedAccounts` cell represent?

tests/specs/functional/merkle-spec.k
```diff
@@ -3,20 +3,21 @@ requires "evm.md"
...
-                                       #parseByteStack( "do" )    |-> "verb"
-                                       #parseByteStack( "dog" )   |-> "puppy"
-                                       #parseByteStack( "doge" )  |-> "coin"
-                                       #parseByteStack( "horse" ) |-> "stallion"
+                                       .Map [ #parseByteStack( "de" )    <- "verb"     ]
+                                            [ #parseByteStack( "def" )   <- "puppy"    ]
+                                            [ #parseByteStack( "defe" )  <- "coin"     ]
+                                            [ #parseByteStack( "becfe" ) <- "stallion" ]
                                      )
+      [macro]
```
ana-pantilie, line 20: This just means that `#initTree` will be replaced with the `MerkleUpdateMap( ... )` after `kompile`, instead of making the backend evaluate `#initTree`, right?

serialization.md
```diff
@@ -616,22 +616,30 @@ Merkle Tree Aux Functions
...
+    rule #merkleExtensionBuilder(PATH, P1, V1, P2, V2)
+      => #merkleExtensionBuilderAux(PATH, P1, V1, P2, V2)
+      requires #sizeByteArray(P1) >Int 0
+       andBool #sizeByteArray(P2) >Int 0
+
+    rule #merkleExtensionBuilder(PATH, P1, V1, P2, V2)
+      => MerkleExtension( PATH, MerklePut( MerklePut( MerkleBranch( .Map, "" ), P1, V1 ), P2, V2 ) )
+      requires notBool ( #sizeByteArray(P1) >Int 0
+                 andBool #sizeByteArray(P2) >Int 0
+                       )
```
ana-pantilie, line 631: Could this have been an `owise` rule?

serialization.md
```diff
@@ -616,22 +616,30 @@ Merkle Tree Aux Functions
...
-       andBool P1[0] ==Int P2[0]
+      requires P1[0] ==Int P2[0]
 
-    rule #merkleExtensionBuilder( PATH, P1, V1, P2, V2 )
+    rule #merkleExtensionBuilderAux( PATH, P1, V1, P2, V2 )
       => MerkleExtension( PATH, MerklePut( MerklePut( MerkleBranch( .Map, "" ), P1, V1 ), P2, V2 ) )
-      requires notBool ( #sizeByteArray(P1) >Int 0
-               andBool   #sizeByteArray(P2) >Int 0
-               andBool   P1[0] ==Int P2[0] )
+      requires P1[0] =/=Int P2[0]
```
ana-pantilie, line 642: Same question as above.

gtrepta: I think these two can be [owise], yes. When I was writing these semantics I was making the requires clauses explicit for debugging purposes.

evm.md
ehildenb: The Firefly client needs to keep track of the `<touchedAccounts>` (which are accounts that this execution affects, either by reading or writing from their state), in order to correctly report that to the testing client (eg. TruffleSuite).

Doesn't affect KEVM, except taht these rules are not working for the haskell backend because they use set concat instead of set union.

tests/specs/functional/merkle-spec.k
ehildenb: Yeah, basically. It also uses the map update operator instead of map concat, so all the appropriate definedness checks are triggered. It also uses actual hex strings instead of random stcrings, so that `#parseByteStack(_)` is defined as well.

serialization.md
ehildenb: Made the switch.

serialization.md
ehildenb: Made the switch.

ana-pantilie (approved):

PR #1005, opened 2021-05-28
Generated Optimizations for KEVM

ehildenb:
- Adds directory `optimizer/` which contains the optimization generator (generates theorems).
- Adds `optimizations.md`, which contains the "core" optimizations that consistently improve performance for the Haskell and LLVM backends.
- Adds a test-harness which checks (with the Haskell prover) that the included core optimizations are sound.

Overall performance change:

```
* HEAD
| testset: profile - Fri Jun  4 02:35:42 UTC 2021
| 0  637.38s   1609.25s   717.32s  4179244KB   make test-conformance  TEST_CONCRETE_BACKEND=llvm     -j4
| 0  1540.22s  6228.63s   92.05s   3423608KB   make test-prove        TEST_SYMBOLIC_BACKEND=haskell  -j4
|
* master
| testset: profile - Fri Jun  4 00:52:07 UTC 2021
| 0  664.16s   1848.80s   637.97s  4180116KB   make test-conformance  TEST_CONCRETE_BACKEND=llvm    -j4  
| 0  2243.73s  7922.88s   100.68s  3350588KB   make test-prove        TEST_SYMBOLIC_BACKEND=haskell -j4
|
```

So this branch sees a 32% speedup in running the Haskell backend proofs, but not any appreciable change in running the overall LLVM conformance test-suite (5% speedup). This is because the LLVM test-suite is dominated by tests which run in ~0.01s, which are beyond the resolution of the timing. However, if I exclude all tests which:

- take _less_ than 0.2s, or
- show less than a 3.5% difference in speed (slower or faster), or
- are haskell backend functional tests (rather than actual semantic tests).

Then we see that the Haskell backend has a 30% speedup, and the LLVM backend has a 23% speedup (data available upon request). This shows that this speeds up long-running executions (both backends), but doesn't have much affect on shorter executions.

To check:

- [x] BLOCKED: #1003 
- [x] REVERT: 067510845
- [x] Separate body of trusted lemmas for optimizations
- [x] Post performance numbers


dwightguth (commented):
looks fine to me, but I'd really like more eyes than just mine on this.

optimizations.md
```diff
@@ -0,0 +1,288 @@
...
+```k
+requires "evm.md"
+
+module EVM-OPTIMIZATIONS-LEMMAS [kore, symbolic]
+    imports EVM
+
+    rule #sizeWordStack(WS           , N) => #sizeWordStack(WS, 0) +Int N requires N =/=Int 0                [simplification]
+    rule #sizeWordStack(WS [ I := _ ], N) => #sizeWordStack(WS, N)        requires I <Int #sizeWordStack(WS) [simplification]
+
+    rule #stackUnderflow(WS, N) => false requires N <=Int #sizeWordStack(WS) [simplification]
```
ehildenb, line 15: @daejunpark in particular, this is the trust base for both generating the optimizations, and proving them correct w.r.t. the semantics.

Note that we _do not_ trust any of the other lemmas is `tests/specs/**` for generating or proving these optimizations.

Makefile
```diff
@@ -447,6 +450,11 @@ test-failing-prove: $(prove_failing_tests:=.prove)
 
 test-klab-prove: $(smoke_tests_prove:=.klab-prove)
 
+test-prove-optimizations: $(prove_optimization_tests:=.prove)
+
+tests/specs/evm-optimizations-spec.md: optimizations.md
```
denis-bogdanas, line 455: can you add a comment explaining how this file is generated?

denis-bogdanas (approved):
Looks good. Optionally please add a comment as mentioned above.

Makefile
ehildenb:
```suggestion
# to generate optimizations.md, run: ./optimizer/optimize.sh &> output
tests/specs/evm-optimizations-spec.md: optimizations.md
```

PR #1006, opened 2021-06-01
Berlin schedule update

anvacaru:
Fixes: runtimeverification/iog-pm#95

 - [x] update ethereum-tests submodule to 8.0.0
 - [x] - eip-2929 https://eips.ethereum.org/EIPS/eip-2929
 - [x] - eip-2718 https://eips.ethereum.org/EIPS/eip-2718
 - [x] - eip-2930 https://eips.ethereum.org/EIPS/eip-2930
 - [x] - eip-2565 https://eips.ethereum.org/EIPS/eip-2565


evm.md
```diff
@@ -1284,7 +1290,7 @@ The various `CALL*` (and other inter-contract control flow) operations will be d
          </k>
 
     rule <k> #mkCall ACCTFROM ACCTTO ACCTCODE BYTES APPVALUE ARGS STATIC:Bool
-          => #touchAccounts ACCTFROM ACCTTO ~> #loadProgram BYTES ~> #initVM ~> #precompiled?(ACCTCODE, SCHED) ~> #execute
+          => #loadProgram BYTES ~> #initVM ~> #precompiled?(ACCTCODE, SCHED) ~> #execute
```
ehildenb, line 1293: Why aren't we touching the accounts here? Will they already be touched somewhere else?

evm.md
```diff
@@ -1857,6 +1880,48 @@ In the YellowPaper, each opcode is defined to consume zero gas unless specified
...
+                                       orBool OP ==K EXTCODEHASH
+                                       orBool OP ==K EXTCODESIZE
+                                       orBool OP ==K EXTCODECOPY
+                                       orBool OP ==K SLOAD
+                                       orBool OP ==K SSTORE
+                                       orBool OP ==K CALL
+                                       orBool OP ==K CALLCODE
+                                       orBool OP ==K DELEGATECALL
+                                       orBool OP ==K STATICCALL
+    rule #usesAccessList(_) => false [owise]
```
ehildenb, line 1900: It's unlikely this will work well with the Java backend (the `owise`). So you may need to just say that `#usesAccessList` is `functional`, and have it go directly to the large requires clause you've written out.

tests/failing.llvm
```diff
@@ -1,11 +1,3 @@
...
-tests/ethereum-tests/LegacyTests/Constantinople/BlockchainTests/GeneralStateTests/stCallCreateCallCodeTest/Call1024PreCalls_d0g0v0.json
-tests/ethereum-tests/LegacyTests/Constantinople/BlockchainTests/GeneralStateTests/stCreate2/Create2Recursive_d0g0v0.json
-tests/ethereum-tests/LegacyTests/Constantinople/BlockchainTests/GeneralStateTests/stCreate2/Create2Recursive_d0g1v0.json
-tests/ethereum-tests/LegacyTests/Constantinople/BlockchainTests/GeneralStateTests/stQuadraticComplexityTest/Create1000Byzantium_d0g1v0.json
-tests/ethereum-tests/LegacyTests/Constantinople/BlockchainTests/GeneralStateTests/stQuadraticComplexityTest/Create1000_d0g1v0.json
-tests/ethereum-tests/LegacyTests/Constantinople/BlockchainTests/GeneralStateTests/stRecursiveCreate/recursiveCreateReturnValue_d0g0v0.json
+tests/ethereum-tests/BlockchainTests/GeneralStateTests/stBadOpcode/invalidAddr.json
+tests/ethereum-tests/BlockchainTests/GeneralStateTests/stBugs/extcodehashEmpty.json
+tests/ethereum-tests/BlockchainTests/GeneralStateTests/stReturnDataTest/modexp_modsize0_returndatasize.json
```
ehildenb, line 3: I believe the red tests here were disabled because they were too slow on CI (not because they were failing), so you may need to re-disable them (if those tests haven't disappeared).

What's up with the three added green tests? Are they failing now? I guess they shouldn't be.

evm.md
anvacaru: They are already being touched in the `loadTx` rule from `driver.md`.

ehildenb: Yeah, but there are many ways to call `#mkCall`, and `loadTx` is not always called by it.

`#mkCall` is called for every `CALL*` instruction. Also, the EIP 2929 specifically says that when calls happen, the accounts are added to `<touchedAccounts>`, so I think you should leave it here.

driver.md
```diff
@@ -115,10 +139,11 @@ To do so, we'll extend sort `JSON` with some EVM specific syntax, and provide a
            <nonce> NONCE </nonce>
            ...
          </account>
-         <touchedAccounts> _ => SetItem(MINER) </touchedAccounts>
+         <touchedAccounts> _ => SetItem(ACCTFROM) SetItem(#newAddr(ACCTFROM, NONCE)) #precompiledAccounts(SCHED) </touchedAccounts>
```
ehildenb, line 142:
```suggestion
         <touchedAccounts> _ => SetItem(ACCTFROM) |Set SetItem(#newAddr(ACCTFROM, NONCE)) |Set #precompiledAccounts(SCHED) </touchedAccounts>
```

driver.md
```diff
@@ -144,7 +169,7 @@ To do so, we'll extend sort `JSON` with some EVM specific syntax, and provide a
            <nonce> NONCE => NONCE +Int 1 </nonce>
            ...
          </account>
-         <touchedAccounts> _ => SetItem(MINER) </touchedAccounts>
+         <touchedAccounts> _ => SetItem(ACCTFROM) SetItem(ACCTTO) #precompiledAccounts(SCHED) </touchedAccounts>
```
ehildenb, line 172:
```suggestion
         <touchedAccounts> _ => SetItem(ACCTFROM) |Set SetItem(ACCTTO) |Set #precompiledAccounts(SCHED) </touchedAccounts>
```


Haskell backend prefers it this way.

driver.md
```diff
@@ -84,12 +84,36 @@ To do so, we'll extend sort `JSON` with some EVM specific syntax, and provide a
...
+           <sigV>       TW   </sigV>
+           <sigR>       TR   </sigR>
+           <sigS>       TS   </sigS>
+           <data>       DATA </data>
+           <txType>     TYPE </txType>
+           <txChainID>  CID  </txChainID>
+           <txAccess>   TA   </txAccess>
+           ...
          </message>
+    requires TYPE ==Int 1
```
ehildenb, line 110:
```suggestion
           <txType>     1    </txType>
           <txChainID>  CID  </txChainID>
           <txAccess>   TA   </txAccess>
           ...
         </message>
```

driver.md
```diff
@@ -84,12 +84,36 @@ To do so, we'll extend sort `JSON` with some EVM specific syntax, and provide a
...
 
     syntax EthereumCommand ::= loadTx ( Account )
  // ---------------------------------------------
```
ehildenb, line 113: I think I would prefer if this set the `<touchedAccounts>` cell to `.Set`, and then we used `#touchAccounts` for the three addresses to populate it with. You can add a 3-argument `#touchAccounts _ _ _` to do this.

driver.md
```diff
@@ -115,10 +139,11 @@ To do so, we'll extend sort `JSON` with some EVM specific syntax, and provide a
            <nonce> NONCE </nonce>
            ...
          </account>
-         <touchedAccounts> _ => SetItem(MINER) </touchedAccounts>
+         <touchedAccounts> _ => SetItem(ACCTFROM) |Set SetItem(#newAddr(ACCTFROM, NONCE)) |Set #precompiledAccounts(SCHED) </touchedAccounts>
 
     rule <k> loadTx(ACCTFROM)
-          => #call ACCTFROM ACCTTO ACCTTO VALUE VALUE DATA false
+          => #loadAccessList(TA)
```
ehildenb, line 145: Same here.

driver.md
```diff
@@ -304,15 +360,17 @@ Note that `TEST` is sorted here so that key `"network"` comes before key `"pre"`
...
+    rule <k> loadTransaction _ { "chainID"    : (TC:String => #asWord(#parseByteStackRaw(TC))), _      } ... </k>
+    rule <k> loadTransaction _ { "gasLimit"   : (TG:String => #asWord(#parseByteStackRaw(TG))), _      } ... </k>
+    rule <k> loadTransaction _ { "gasPrice"   : (TP:String => #asWord(#parseByteStackRaw(TP))), _      } ... </k>
+    rule <k> loadTransaction _ { "nonce"      : (TN:String => #asWord(#parseByteStackRaw(TN))), _      } ... </k>
+    rule <k> loadTransaction _ { "v"          : (TW:String => #asWord(#parseByteStackRaw(TW))), _      } ... </k>
+    rule <k> loadTransaction _ { "value"      : (TV:String => #asWord(#parseByteStackRaw(TV))), _      } ... </k>
+    rule <k> loadTransaction _ { "to"         : (TT:String => #asAccount(#parseByteStackRaw(TT))), _   } ... </k>
+    rule <k> loadTransaction _ { "data"       : (TI:String => #parseByteStackRaw(TI)), _               } ... </k>
+    rule <k> loadTransaction _ { "r"          : (TR:String => #padToWidth(32, #parseByteStackRaw(TR))), _ } ... </k>
+    rule <k> loadTransaction _ { "s"          : (TS:String => #padToWidth(32, #parseByteStackRaw(TS))), _ } ... </k>
```
ehildenb, line 373:
```suggestion
    rule <k> loadTransaction _ { "type"     : (TT:String => #asWord(#parseByteStackRaw(TT))), _         } ... </k>
    rule <k> loadTransaction _ { "chainID"  : (TC:String => #asWord(#parseByteStackRaw(TC))), _         } ... </k>
    rule <k> loadTransaction _ { "gasLimit" : (TG:String => #asWord(#parseByteStackRaw(TG))), _         } ... </k>
    rule <k> loadTransaction _ { "gasPrice" : (TP:String => #asWord(#parseByteStackRaw(TP))), _         } ... </k>
    rule <k> loadTransaction _ { "nonce"    : (TN:String => #asWord(#parseByteStackRaw(TN))), _         } ... </k>
    rule <k> loadTransaction _ { "v"        : (TW:String => #asWord(#parseByteStackRaw(TW))), _         } ... </k>
    rule <k> loadTransaction _ { "value"    : (TV:String => #asWord(#parseByteStackRaw(TV))), _         } ... </k>
    rule <k> loadTransaction _ { "to"       : (TT:String => #asAccount(#parseByteStackRaw(TT))), _      } ... </k>
    rule <k> loadTransaction _ { "data"     : (TI:String => #parseByteStackRaw(TI)), _                  } ... </k>
    rule <k> loadTransaction _ { "r"        : (TR:String => #padToWidth(32, #parseByteStackRaw(TR))), _ } ... </k>
    rule <k> loadTransaction _ { "s"        : (TS:String => #padToWidth(32, #parseByteStackRaw(TS))), _ } ... </k>
```

driver.md
```diff
@@ -482,7 +540,15 @@ Here we check the other post-conditions associated with an EVM test.
...
+    rule <k> check "transactions" : ( KEY : (VALUE:ByteArray => #asWord(VALUE)))         ... </k> requires KEY in (SetItem("gasLimit") SetItem("gasPrice") SetItem("nonce") SetItem("v") SetItem("value") SetItem("chainId") SetItem("type"))
+
+    rule <k> check "transactions" : "accessList" : [ ACCESSLIST  , REST ] => check "transactions" : "accessList" : ACCESSLIST  ~> check "transactions" : "accessList" : [ REST ] ... </k>
+    rule <k> check "transactions" : "accessList" : { "address" : V1 , "storageKeys": V2 , .JSONs } => check "transactions" : "accessList" : "address" : #parseHexWord(V1) : "storageKeys" : V2  ... </k>
+
+    rule <k> check "transactions" : "accessList" : "address" : ADDR : "storageKeys" : [ KEY , REST ] => check "transactions" : "accessList" : "address" : ADDR : "storageKeys" : #parseHexWord(KEY) ~> check "transactions" : "accessList" : "address" : ADDR : "storageKeys" : [ REST ] ... </k>
+    rule <k> check "transactions" : "accessList" : "address" : ADDR : "storageKeys" : KEY => . ... </k> <txOrder> ListItem(TXID) ... </txOrder> <message> <msgID> TXID </msgID> <txAccess> TA </txAccess> ... </message> requires isInAccessList(ADDR, KEY, TA)
```
ehildenb, line 549: Put this rule below with the next block of rules. All the rules in this block just extract the individual accessLists. But the next block is for actually checking them.

evm.md
```diff
@@ -1322,6 +1328,19 @@ The various `CALL*` (and other inter-contract control flow) operations will be d
...
+         <touchedStorage> ... ACCT |-> (TS:Set => TS |Set SetItem(INDEX)) ... </touchedStorage>
+    rule <k> #touchStorage ACCT INDEX => . ... </k>
+         <touchedStorage> TS => TS[ACCT <- SetItem(INDEX)] </touchedStorage>
+      requires notBool ACCT in_keys(TS)
+
+    syntax Bool ::= inStorage ( Map , Account , Int ) [function]
+ // ------------------------------------------------------------
+    rule inStorage (TS, ACCT, _ )  => false requires notBool ACCT in_keys(TS)
+    rule inStorage (TS, ACCT, KEY) => KEY in {TS[ACCT]}:>Set requires ACCT in_keys(TS)
+
```
ehildenb, line 1343: This is only used for gas calculation, so move it to that section. Also prefix it with a hash `#inStorage`.

evm.md
```diff
@@ -1552,7 +1571,8 @@ For each `CALL*` operation, we make a corresponding call to `#call` and a state-
     syntax TernStackOp ::= "CREATE"
  // -------------------------------
     rule <k> CREATE VALUE MEMSTART MEMWIDTH
-          => #checkCall ACCT VALUE
+          => #touchAccounts #newAddr(ACCT, NONCE)
+          ~> #checkCall ACCT VALUE
```
ehildenb, line 1575: Are the accounts supposed to be touched _before_ the `#checkCall`? As in, if the call fails to go through (because not enough funds, for example), are we supposed to still touch the new account?

If it doesn't matter (which it should matter, but maybe the test-suite does not test that functionality), then we can have `#create` call `#touchAccounts` instead, which would only affect one rule, instead of both rules for `CREATE` and `CREATE2`.

evm.md
```diff
@@ -1587,7 +1608,7 @@ Self destructing to yourself, unlike a regular transfer, destroys the balance in
 ```k
     syntax UnStackOp ::= "SELFDESTRUCT"
-    rule <k> SELFDESTRUCT ACCTTO => #touchAccounts ACCT ACCTTO ~> #transferFunds ACCT ACCTTO BALFROM ~> #end EVMC_SUCCESS ... </k>
+    rule <k> SELFDESTRUCT ACCTTO => #transferFunds ACCT ACCTTO BALFROM ~> #end EVMC_SUCCESS ... </k>
          <id> ACCT </id>
          <selfDestruct> ... (.Set => SetItem(ACCT)) ... </selfDestruct>
          <account>
```
ehildenb, line 1614:
```suggestion
    rule <k> SELFDESTRUCT ACCTTO => #transferFunds ACCT ACCTTO BALFROM ~> #end EVMC_SUCCESS ... </k>
         <id> ACCT </id>
         <selfDestruct> SDS => SDS |Set SetItem(ACCT) </selfDestruct>
         <account>
```

evm.md
```diff
@@ -1598,7 +1619,7 @@ Self destructing to yourself, unlike a regular transfer, destroys the balance in
          <output> _ => .ByteArray </output>
       requires ACCT =/=Int ACCTTO
 
-    rule <k> SELFDESTRUCT ACCT => #touchAccounts ACCT ~> #end EVMC_SUCCESS ... </k>
+    rule <k> SELFDESTRUCT ACCT => #end EVMC_SUCCESS ... </k>
          <id> ACCT </id>
          <selfDestruct> ... (.Set => SetItem(ACCT)) ... </selfDestruct>
          <account>
```
ehildenb, line 1625:
```suggestion
    rule <k> SELFDESTRUCT ACCT => #end EVMC_SUCCESS ... </k>
         <id> ACCT </id>
         <selfDestruct> SDS => SDS |Set SetItem(ACCT) </selfDestruct>
         <account>
```

evm.md
```diff
@@ -1779,6 +1801,7 @@ Overall Gas
     rule <k> #gas [ OP , AOP ]
           => #if #usesMemory(OP) #then #memory [ AOP ] #else .K #fi
           ~> #gas [ AOP ]
+          ~> #if #usesAccessList(OP) #then #access [ AOP ] #else .K #fi
```
ehildenb, line 1804: Any reason to do access list account after execution gas? I guess I would put it even before the memory accounting.

evm.md
```diff
@@ -1857,6 +1880,47 @@ In the YellowPaper, each opcode is defined to consume zero gas unless specified
...
+                         orBool OP ==K SELFDESTRUCT
+                         orBool OP ==K EXTCODEHASH
+                         orBool OP ==K EXTCODESIZE
+                         orBool OP ==K EXTCODECOPY
+                         orBool OP ==K SLOAD
+                         orBool OP ==K SSTORE
+                         orBool OP ==K CALL
+                         orBool OP ==K CALLCODE
+                         orBool OP ==K DELEGATECALL
+                         orBool OP ==K STATICCALL
```
ehildenb, line 1899:
```suggestion
    syntax Bool ::= #usesAccessList ( OpCode ) [function, functional]
 // -----------------------------------------------------------------
    rule #usesAccessList(OP) => isAddr1Op(OP) orBool isAddr2Op(OP) orBool OP ==K SLOAD orBool OP ==K SSTORE
```

If you merge to master, you'll have access to these functions.

evm.md
```diff
@@ -1857,6 +1880,47 @@ In the YellowPaper, each opcode is defined to consume zero gas unless specified
...
+
+    syntax InternalOp ::= "#access" "[" OpCode "]"
+ // --------------------------------------------
+    rule <k> #access [ OP ] => #gasAccess(SCHED, OP) ~> #deductGas ... </k>
+         <schedule> SCHED </schedule>
+      requires Ghasaccesslist << SCHED >>
+    rule <k> #access [ _ ] => . ... </k> [owise]
```
ehildenb, line 1906: This condition can be inlined above where `#access` is introduced.

evm.md
ehildenb:
```suggestion
    rule <k> #gas [ OP , AOP ]
          => #if #usesMemory(OP) #then #memory [ AOP ] #else .K #fi
          ~> #gas [ AOP ]
          ~> #if Ghasaccesslist << SCHED >> andBool #usesAccessList(OP) #then #access [ AOP ] #else .K #fi
```

ehildenb: Here is where you can inline the condition which checks whether we're on an access list schedule.

evm.md
ehildenb: I left a comment above.

evm.md
```diff
@@ -2019,21 +2088,19 @@ The intrinsic gas calculation mirrors the style of the YellowPaper (appendix H).
...
 
-    rule <k> #gasExec(SCHED, EXTCODESIZE _) => Gextcodesize < SCHED > ... </k>
-    rule <k> #gasExec(SCHED, BALANCE _)     => Gbalance     < SCHED > ... </k>
-    rule <k> #gasExec(SCHED, EXTCODEHASH _) => Gbalance     < SCHED > ... </k>
-    rule <k> #gasExec(SCHED, BLOCKHASH _)   => Gblockhash   < SCHED > ... </k>
+    rule <k> #gasExec(SCHED, EXTCODECOPY _ _ _ WIDTH) => Cextcodecopy(SCHED, WIDTH) ... </k>
+    rule <k> #gasExec(SCHED, EXTCODESIZE _)           => Cextcodesize(SCHED)        ... </k>
+    rule <k> #gasExec(SCHED, BALANCE _)               => Cbalance(SCHED)            ... </k>
+    rule <k> #gasExec(SCHED, EXTCODEHASH _)           => Cextcodehash(SCHED)        ... </k>
+    rule <k> #gasExec(SCHED, BLOCKHASH _)             => Gblockhash < SCHED > ... </k>
```
ehildenb, line 2095:
```suggestion
    rule <k> #gasExec(SCHED, EXTCODECOPY _ _ _ WIDTH) => Cextcodecopy(SCHED, WIDTH) ... </k>
    rule <k> #gasExec(SCHED, EXTCODESIZE _)           => Cextcodesize(SCHED)        ... </k>
    rule <k> #gasExec(SCHED, BALANCE _)               => Cbalance(SCHED)            ... </k>
    rule <k> #gasExec(SCHED, EXTCODEHASH _)           => Cextcodehash(SCHED)        ... </k>
    rule <k> #gasExec(SCHED, BLOCKHASH _)             => Gblockhash < SCHED >       ... </k>
```

state-loader.md
```diff
@@ -230,6 +231,23 @@ The `"rlp"` key loads the block information.
           ...
           </k>
 
+    rule <k> load "transaction" : [ (T => [#rlpDecodeTransaction(#parseByteStackRaw(T))]) , _ ]
+          ...
+          </k>
```
ehildenb, line 236:
```suggestion
    rule <k> load "transaction" : [ (T => [#rlpDecodeTransaction(#parseByteStackRaw(T))]) , _ ]
          ...
          </k>
```

Put this all on one line and consider moving it somewhere else.

evm.md
anvacaru: From EIP-2929 
> When a CREATE or CREATE2 opcode is called, immediately (ie. before checks are done to determine whether or not the address is unclaimed) add the address being created to accessed_addresses, but gas costs of CREATE and CREATE2 are unchanged. Clarification: If a CREATE/CREATE2 operation fails later on, e.g during the execution of initcode or has insufficient gas to store the code in the state, the address of the contract itself remains in access_addresses (but any additions made within the inner scope are reverted).

So, to answer your first two questions, yes.
Regarding this
> If it doesn't matter (which it should matter, but maybe the test-suite does not test that functionality), then we can have `#create` call `#touchAccounts` instead, which would only affect one rule, instead of both rules for `CREATE` and `CREATE2`.

I could also add `#gasAccess` rules for both of `CREATE` and `CREATE2` and  deduct 0 gas. 
` #gasAccess(_    , CREATE _ _ _)   => #touchAccounts #newAddr(ACCT, NONCE) ~> 0`

evm.md
anvacaru: > Any reason to do access list account after execution gas? I guess I would put it even before the memory accounting. 

Having it executed before `#gas` would mess the gas calculation for `CALL*` opcodes, which take in consideration the `touchedAccounts` Set.  In the case of a cold account, `#access` would mark it as warm before the gas calculation in `#gas`. 

> Here is where you can inline the condition which checks whether we're on an access list schedule.

Applied.

evm.md
anvacaru: From EIP-2929 
> When a CREATE or CREATE2 opcode is called, immediately (ie. before checks are done to determine whether or not the address is unclaimed) add the address being created to accessed_addresses, but gas costs of CREATE and CREATE2 are unchanged. Clarification: If a CREATE/CREATE2 operation fails later on, e.g during the execution of initcode or has insufficient gas to store the code in the state, the address of the contract itself remains in access_addresses (but any additions made within the inner scope are reverted).

So, to answer your first two questions, yes.
Regarding this
> If it doesn't matter (which it should matter, but maybe the test-suite does not test that functionality), then we can have `#create` call `#touchAccounts` instead, which would only affect one rule, instead of both rules for `CREATE` and `CREATE2`.

I could also add `#gasAccess` rules for both of `CREATE` and `CREATE2` and  deduct 0 gas. 
` #gasAccess(_    , CREATE _ _ _)   => #touchAccounts #newAddr(ACCT, NONCE) ~> 0`

state-loader.md
anvacaru: I don't understand why should I move it.

evm.md
ehildenb: No worries, thanks for answering my question!

state-loader.md
ehildenb: Well, it's between the two rules which handle the decoded transaction. So this rule does the decoding, but the two rules surrounding it actually handle the decoded transaction. Just seems a little out of place I guess, but up to you.

driver.md
```diff
@@ -484,7 +543,15 @@ Here we check the other post-conditions associated with an EVM test.
...
+    rule <k> check "transactions" : "accessList" : { "address" : V1 , "storageKeys": V2 , .JSONs } => check "transactions" : "accessList" : "address" : #parseHexWord(V1) : "storageKeys" : V2  ... </k>
+    rule <k> check "transactions" : "accessList" : "address" : ADDR : "storageKeys" : [ KEY , REST ] => check "transactions" : "accessList" : "address" : ADDR : "storageKeys" : #parseHexWord(KEY) ~> check "transactions" : "accessList" : "address" : ADDR : "storageKeys" : [ REST ] ... </k>
+
+    rule <k> check "transactions" : "accessList" : "address" : ADDR : "storageKeys" : KEY        => . ... </k> <txOrder> ListItem(TXID) ... </txOrder> <message> <msgID> TXID </msgID> <txAccess> TA </txAccess> ... </message> requires isInAccessList(ADDR, KEY, TA)
+    rule <k> check "transactions" : "accessList" : "address" : _    : "storageKeys" : [ .JSONs ] => . ... </k>
+    rule <k> check "transactions" : "accessList" : [ .JSONs ] => . ... </k>
 
     rule <k> check "transactions" : ("data"     : VALUE) => . ... </k> <txOrder> ListItem(TXID) ... </txOrder> <message> <msgID> TXID </msgID> <data>       VALUE </data>       ... </message>
```
ehildenb, line 556: The first rule here should go in the second block here, not with all the code which just handles manipulating the JSONs.

The second block here does the actual checks, the first block just handles manipulating the JSONs.

evm.md
```diff
@@ -1881,6 +1914,36 @@ In the YellowPaper, each opcode is defined to consume zero gas unless specified
...
+```k
+    syntax Bool ::= #usesAccessList ( OpCode ) [function, functional]
+ // -----------------------------------------------------------------
+    rule #usesAccessList(OP) => isAddr1Op(OP) orBool isAddr2Op(OP) orBool OP ==K SLOAD orBool OP ==K SSTORE
+
+    syntax InternalOp ::= "#access" "[" OpCode "]"
+ // --------------------------------------------
+    rule <k> #access [ OP ] => #gasAccess(SCHED, OP) ~> #deductGas ... </k>
+         <schedule> SCHED </schedule>
+      requires Ghasaccesslist << SCHED >>
```
ehildenb, line 1929: We don't need `Ghasaccesslist` check here anymore, if you don't want, because we are already checking it above.

So you can have `#gasAccess` have an `[owise]` rule (which will never fire) that goes to `0`, similar to how `#memory` works above.

evm.md
```diff
@@ -2153,6 +2232,44 @@ There are several helpers for calculating gas (most of them also specified in th
 
     rule [Cmem]: Cmem(SCHED, N) => (N *Int Gmemory < SCHED >) +Int ((N *Int N) /Int Gquadcoeff < SCHED >)
 
+    rule [Caddraccess]:
```
ehildenb, line 2235: Formatting in this section can be tidied up a bit. Use one line when possible (I recommend disabling line-wrapping in your editor).

Also, line up side-conditions where it makes sense, for example, instead of:

```
requires Ghasaccesslist << SCHED >>
requires notBool Ghasaccesslist << SCHED >>
```

Do:

```
requires         Ghasaccesslist << SCHED >>
requires notBool Ghasaccesslist << SCHED >>
```

Also, you may need to add these new rule names (eg. `EVM.Caddraccess`) to the various `concrete-rules.txt` files present in `tests/specs/...` to get the proofs to pass.

ehildenb:
You likely need to add lemmas similar to this one: https://github.com/kframework/evm-semantics/blob/master/tests/specs/infinite-gas.k#L164

To get over issues liek this:

```
    {
      true
    #Equals
      #gas ( -133 ) <Int Csload ( ISTANBUL , #inStorage ( _20 , CONTRACT_ID:Int , 0 ) )
    } 
```

This should go to `false`. So add a test `tests/specs/functional/infinite-gas-spec.k`, which looks like this: `claim <k> runLemma( #gas ( -133 ) <Int Csload ( ISTANBUL , #inStorage ( _20 , CONTRACT_ID:Int , 0 ) )) => doneLemma(false) ... </k> requires CONTRACT_ID =/=Int andBool #rangeAddress(CONTRACT_ID)`.

Similarly for other unsimplified side-conditions involving `#gas`.

ehildenb:
`make test-prove --dry-run` to get the command. OR `make tests/specs/functional/infinite-gas-spec.k.prove` to run a specific one (and get the command needed by adding `--dry-run`). Also remember to run with both backends:

```
make tests/specs/functional/infinite-gas-spec.k.prove TEST_SYMBOLIC_BACKEND=haskell
make tests/specs/functional/infinite-gas-spec.k.prove TEST_SYMBOLIC_BACKEND=java
```

driver.md
```diff
@@ -117,10 +141,12 @@ To do so, we'll extend sort `JSON` with some EVM specific syntax, and provide a
            <nonce> NONCE </nonce>
            ...
          </account>
-         <touchedAccounts> _ => SetItem(MINER) </touchedAccounts>
+         <touchedAccounts> _ => .Set </touchedAccounts>
```
gtrepta, line 144: A reminder why omitting the MINER here during loadTx is ok? I don't remember the reason it was included in the first place.

anvacaru: I couldn't find anywhere in the current codebase how MINER is used in `touchedAccounts`. Also, EIP-2929 specifies that 
> When a transaction execution begins, `accessed_addresses` is initialized to include the `tx.sender`, `tx.to` (or the address being created if it is a contract creation transaction) and the set of all precompiles.

gtrepta: Hmm... Are we confusing the `accessed_addresses` set with the `touchedAccounts` set? From what I gather, `touchedAccounts` exists so empty accounts can be deleted at the end of a transaction. `accessed_addresses` exists to track gas costs for various account related opcodes. Two separate purposes, so it should probably be two separate sets? @ehildenb What do you think?

gtrepta (commented):
My understanding is that we are going to wait to merge this until #1100 gets merged, and we will then update this PR to make the node support the new transaction types.

Makefile
```diff
@@ -587,12 +588,22 @@ test-interactive-search: $(search_tests:=.search)
...
+proto_tester := $(LOCAL_BIN)/proto_tester
+proto-tester: $(proto_tester)
+$(proto_tester): tests/vm/proto_tester.cpp
+	@mkdir -p $(LOCAL_BIN)
+	$(CXX) -I $(LOCAL_LIB)/proto $(protobuf_out) $< -o $@ -lprotobuf -lpthread
+
+node_tests:=$(wildcard tests/vm/*.bin)
+test-node: $(KEVM_BIN)/kevm-vm $(proto_tester) $(node_tests:=.run-node)
+
+tests/vm/%.run-node: tests/vm/%.expected
```
ehildenb, line 600:
```suggestion
test-node: $(node_tests:=.run-node)

tests/vm/%.run-node: tests/vm/%.expected $(KEVM_BIN)/kevm-vm $(proto_tester) 
```

serialization.md
```diff
@@ -49,6 +50,10 @@ Address/Hash Helpers
       => #sender(#unparseByteStack(#parseHexBytes(#hashUnsignedTx(TN, TP, TG, TT, TV, #parseByteStackRaw(DATA), CID))), 28 -Int (TW %Int 2), #unparseByteStack(TR), #unparseByteStack(TS))
       requires TW ==Int CID *Int 2 +Int 35 orBool TW ==Int CID *Int 2 +Int 36
 
+    rule #sender(TYPE, CID, TN, TP, TG, TT, TV, TD, TA, TW, TR, TS)
```
ehildenb, line 53: Is the `CID` argument here the same as the one for the other `#sender`s? Then lets put it in the same position, at the end of the argument list.

driver.md
```diff
@@ -495,6 +566,19 @@ Here we check the other post-conditions associated with an EVM test.
...
+    syntax Bool ::= isInAccessListStorage ( Int , JSON )    [function]
+                  | isInAccessList ( Account , Int , JSON ) [function]
+ // ------------------------------------------------------------------
+    rule isInAccessList(_, _, [.JSONs]) => false
+    rule isInAccessList(ADDR, KEY, [[ACCT, [STRG:JSONs]],  REST]) => isInAccessListStorage (KEY, [STRG]) orBool isInAccessList(ADDR, KEY, [REST]) requires ADDR ==K #asAccount(#parseByteStackRaw(ACCT))
+    rule isInAccessList(ADDR, KEY, [[ACCT, [_STRG:JSONs]], REST]) => isInAccessList(ADDR, KEY, [REST])   requires notBool ADDR ==K #asAccount(#parseByteStackRaw(ACCT))
+
+    rule isInAccessListStorage(_, [.JSONs]) => false
+    rule isInAccessListStorage(KEY, [SKEY, _REST]) => true requires KEY ==Int #asWord(#parseByteStackRaw(SKEY))
+    rule isInAccessListStorage(KEY, [SKEY, REST])  => isInAccessListStorage(KEY, [REST]) requires  notBool KEY ==Int #asWord(#parseByteStackRaw(SKEY))
```
ehildenb, line 581: Format these rules a bit better.

Also, I think we are better served by `#if_#then_#else_#fi` here actually. For both `isInAccessList` and `isInAccessListStorage`, we should use `#ite` in order to make it so we only need to call `#asWord(#parseByteStackRaw(SKEY))` once to determine which branch of the `#ite` to take, instead of calling it twice to check which rule applies.

Because this module is not used by the proofs at all, it won't hurt them to use `#ite`.

serialization.md
```diff
@@ -135,6 +141,20 @@ Address/Hash Helpers
...
+                                                            +String #rlpEncodeWord(TP)
+                                                            +String #rlpEncodeWord(TG)
+                                                            +String #rlpEncodeAccount(TT)
+                                                            +String #rlpEncodeWord(TV)
+                                                            +String #rlpEncodeString(#unparseByteStack(TD))
+                                                            +String #rlpEncodeAccessList([TA], "")
+                                                          , 192
+                                                          )
+                                  )
 ```
```
ehildenb, line 158: It seems like here, again, we should just pass in `TYPE` as a parameter to the original `#hashUnsignedTx`, and then have it branch on whether `TYPE ==Int 0` or `TYPE ==Int 1`, instead of making another function symbol.

Issue #1012, opened 2021-06-02
Debug `make tests/specs/benchmarks/keccak00-spec.k.prove TEST_SYMBOLIC_BACKEND=haskell`


ehildenb:

Riib11:
#1030 fixes first SMT failure

Riib11:
solved by #1049

PR #1030, opened 2021-06-07
powbytelen abstraction

ehildenb:
`#buf => #bufStrict` lengths result in formula that look like `2 ^Int (SIZE *Int 8)`, which Z3 has trouble with when using the Haskell backend's encoding.

This wraps up the sub-formula in an uninterpreted function, and axiomatizes it as much as needed to get the proofs to go through.

This also enables the test `tests/specs/functional/lemmas-java-spec.k` to run on the Haskell backend, so the proof there is inlined into `tests/specs/lemmas.k` instead.


daejunpark (approved):

Riib11:
Fixes #1012

PR #1031, opened 2021-06-08
Reduce ceil32 to up/Int

ehildenb:
Currently we have several places which use ceil32, and several places which use up/Int, but no common lemmas because they're used in different contexts. This PR:

- Reduces `ceil32` to `up/Int` via a `macro`.
- Adds gas simplification lemmas for `up/Int` and some other gas expressions.
- Switches over the concrete-rules files to using `up/Int` instead of `ceil32`.
- Adds functional tests of the new lemmas.

### DIFF
--- a/tests/specs/benchmarks/verification.k
+++ b/tests/specs/benchmarks/verification.k
@@ -251,8 +251,6 @@ module VERIFICATION
 
     rule X <=Int #ceil32(X) => true requires X >=Int 0 [simplification]
 
-    rule #ceil32(N) => ((N +Int 31) /Int 32) *Int 32 [smt-lemma]
-
   // ########################
   // Gnosis
   // ########################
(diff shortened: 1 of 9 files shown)


Riib11 (approved):

tests/specs/benchmarks/verification.k
```diff
@@ -249,9 +249,11 @@ module VERIFICATION
 
     rule 0 <Int 1 <<Int N => true requires 0 <=Int N [simplification]
 
-    rule X <=Int #ceil32(X) => true requires X >=Int 0 [simplification]
+    rule X +Int Y <Int pow256 => true requires X <=Int pow16 andBool Y <Int pow16 [simplification]
 
-    rule #ceil32(N) => ((N +Int 31) /Int 32) *Int 32 [smt-lemma]
+    rule X <Int #ceil32(X)      => true  requires 0 <=Int X                       [simplification]
```
denis-bogdanas, line 254: I believe this is `X <=Int #ceil32(X)      => true`

denis-bogdanas (changes requested):
One issue found. The rest looks good.

tests/specs/benchmarks/verification.k
ehildenb: Fixed.

denis-bogdanas (approved):

Issue #1032, opened 2021-06-08
Pull out `infinite-gas.md`.


ehildenb:
Currently symbols `#gas(_)` and `#gas(_,_,_)` are defined _outside_ the main module, meaning it's hard to unparse them in the `kore-repl`.

- Add a file `infinite-gas.md` which defines the symbol `#gas(_)`, as well as documenting how to use it. Include in it all the lemmas which are about moving symbols inside the `#gas(_)` wrapper, and about comparing `#gas(_)` to other quantities.
- Make the EDSL import INFINITE-GAS from infinite-gas.md.
- Rename the INFINITE-GAS in tests-specs/infinite-gas.k to INFINITE-GAS-LEMMAS (and adjust the imports throughout `tests/specs`, find them with `git grep 'imports INFINITE-GAS' -- tests/specs`).
- Uses of `#gas(_,_,_)` should be simplified to `#gas(_)` in a haskell-backend only module (via a `macro`). This can be done in the `tests/specs/infinite-gas.k` module.
- Proofs which use `#gas(_,_,_)` (eg. many of the ERC20s), should make sure to import `tests/specs/infinite-gas.k`, so that they see the simplification lemma from the previous bullet point.

The file `infinite-gas.k` will contain the remaining lemmas not dealing with defining `#gas(_)` as positive infinity directly.

PR #1034, opened 2021-06-08
Pull out infinite gas

JKTKops:
Fixes: #1032

I'm currently running tests locally. Experimentation with the debugger raised one issue, the `[function]` attribute on the 3-argument `#gas`, which I removed.


ehildenb:
Yep, `macro` are not considered `function`, so you can't have both.

infinite-gas.md
```diff
@@ -0,0 +1,108 @@
+Infinite Gas
+Here we use the construct `#gas` to represent positive infinity, while tracking the gas formula through execution.
+This allows (i) computing final gas used, and (ii) never stopping because of out-of-gas.
+Note that the argument to `#gas(_)` is just metadata tracking the current gas usage, and is not meant to be compared to other values.
+such, any `#gas(G)` and `#gas(G')` are the _same_ positive infinite, regardless of the values `G` and `G'`.
+In particular, this means that `#gas(_) <Int #gas(_) => false`, and `#gas(_) <=Int #gas(_) => true`, regardless of the values contained in the `#gas(_)`.
```
ehildenb, line 6:
```suggestion
Infinite Gas
============

Here we use the construct `#gas` to represent positive infinity, while tracking the gas formula through execution.
This allows (i) computing final gas used, and (ii) never stopping because of out-of-gas.
Note that the argument to `#gas(_)` is just metadata tracking the current gas usage, and is not meant to be compared to other values.
such, any `#gas(G)` and `#gas(G')` are the _same_ positive infinite, regardless of the values `G` and `G'`.
In particular, this means that `#gas(_) <Int #gas(_) => false`, and `#gas(_) <=Int #gas(_) => true`, regardless of the values contained in the `#gas(_)`.
```

infinite-gas.md
```diff
@@ -0,0 +1,108 @@
...
+
+module INFINITE-GAS-JAVA [kast]
+    imports INFINITE-GAS-COMMON
+    imports K-REFLECTION
+
+    rule #gas(_)  <Int C => false requires #isConcrete(C) [simplification]
+    rule #gas(_) <=Int C => false requires #isConcrete(C) [simplification]
+    rule #gas(_) >=Int C => true  requires #isConcrete(C) [simplification]
+    rule C <=Int #gas(_) => true  requires #isConcrete(C) [simplification]
+endmodule
```
ehildenb, line 33: Put these modules at the end (after `INFINITE-GAS-COMMON`).

infinite-gas.md
```diff
@@ -0,0 +1,108 @@
...
+
+module INFINITE-GAS-COMMON
+    imports EVM
+
+    syntax Int ::= #gas ( Int ) [function, functional, no-evaluators, klabel(infGas), symbol, smtlib(infGas)]
+ // ---------------------------------------------------------------------------------------------------------
```
ehildenb, line 39: Move the textual documentation at the top of the file (and the note about positive infinity below) to right before this declaration.

tests/specs/infinite-gas.k
```diff
@@ -9,18 +10,18 @@ requires "evm.md"
...
+module INFINITE-GAS-LEMMAS-HASKELL [kore]
+    imports INFINITE-GAS-LEMMAS-COMMON
 
-    rule #gas(_)  <Int _C => false [concrete(_C), simplification]
-    rule #gas(_) <=Int _C => false [concrete(_C), simplification]
-    rule #gas(_) >=Int _C => true  [concrete(_C), simplification]
-    rule _C <=Int #gas(_) => true  [concrete(_C), simplification]
+    syntax Int ::= #gas ( Int , Int , Int ) // startGas, nonMemory, memory
+ //-----------------------------------------------------------------------------------
+
```
ehildenb, line 23: Leave this declaration in `INFINITE-GAS-COMMON`, and remove other declarations of this one in their respective test-suites (eg. `tests/specs/erc20/abstract-semantics-segmented-gas.k`).

Riib11:
There are extra backticks in infinite-gas.md#L90

Jenkinsfile
```diff
@@ -24,23 +24,16 @@ pipeline {
...
             timeout(time: 120, unit: 'MINUTES')
           }
           parallel {
-            stage('Java')              { steps { sh 'make test-prove -j5 TEST_SYMBOLIC_BACKEND=java'    } }
-            stage('Haskell')           { steps { sh 'make test-prove -j4 TEST_SYMBOLIC_BACKEND=haskell' } }
-            stage('Haskell (dry-run)') { steps { sh 'make test-haskell-dry-run -j3'                     } }
+            stage('Conformance (LLVM)') { steps { sh 'make test-conformance -j8 TEST_CONCRETE_BACKEND=llvm' } }
+            stage('Java')               { steps { sh 'make test-prove -j5 TEST_SYMBOLIC_BACKEND=java'    } }
+            stage('Haskell')            { steps { sh 'make test-prove -j4 TEST_SYMBOLIC_BACKEND=haskell' } }
+            stage('Haskell (dry-run)')  { steps { sh 'make test-haskell-dry-run -j3'                     } }
```
ehildenb, line 36:
```suggestion
            stage('Conformance (LLVM)') { steps { sh 'make test-conformance -j8 TEST_CONCRETE_BACKEND=llvm' } }
            stage('Java')               { steps { sh 'make test-prove -j5 TEST_SYMBOLIC_BACKEND=java'       } }
            stage('Haskell')            { steps { sh 'make test-prove -j4 TEST_SYMBOLIC_BACKEND=haskell'    } }
            stage('Haskell (dry-run)')  { steps { sh 'make test-haskell-dry-run -j3'                        } }
```

tests/specs/infinite-gas.k
```diff
@@ -81,8 +74,12 @@ module INFINITE-GAS-JAVA [kast]
     rule (S1 -Int C2) -Int C3 => S1 -Int (C2 +Int C3) requires #isConcrete(C2) andBool #isConcrete(C3) andBool notBool #isConcrete(S1) [simplification]
 endmodule
 
-module INFINITE-GAS-COMMON
+module INFINITE-GAS-LEMMAS-COMMON
     imports EVM
+    imports INFINITE-GAS
+
+    syntax Int ::= #gas ( Int , Int , Int ) // startGas, nonMemory, memory
+ // -----------------------------------------------------------------------
```
ehildenb, line 82:
```suggestion
    syntax Int ::= #gas ( Int , Int , Int ) [function, no-evaluators] // startGas, nonMemory, memory
 // ------------------------------------------------------------------------------------------------
```

ehildenb: Actually, nevermind, this isn't needed because of the `macro` rule the haskell backend sees.

JKTKops:
The CI output from 335815c has this error during the Haskell proof step:
```
[Error] Compiler: Cannot add new constructors to hooked sort Int
	Source(/home/jenkins-slave/workspace/evm-semantics_PR-1034/.build/usr/lib/kevm/include/kframework/abi.md)
	Location(192,20,192,34)
[Error] Compiler: Had 1 structural errors.
```
It's in `abi.md` which we didn't touch, so this probably (also) has something to do with changing the main module.

ehildenb:
The java backend is still set to use `driver` and `ETHEREUM-SIMULATION`. You can also switch its main file and main module over @JKTKops .

re. "adding new constructors". You can add `function, no-evaluators` to the declaration of a production to tell the Haskell backend that it's not a constructor (it's a `function`), and that you don't intend to provide rules for it (`no-evaluators`).

Makefile
```diff
@@ -223,9 +224,9 @@ $(KEVM_LIB)/$(java_kompiled): $(includes)
 # Haskell
 
 haskell_dir            := haskell
-haskell_main_module    := ETHEREUM-SIMULATION
+haskell_main_module    := EDSL
 haskell_syntax_module  := $(haskell_main_module)
-haskell_main_file      := driver.md
+haskell_main_file      := edsl.md
```
ehildenb, line 229: You can make these same changes to `java_main_module` and `java_main_file`.

JKTKops:
The specific syntax it's complaining about is the definition for `#ceil32`, which I can't make a `[function]` because there's a `[macro]` rule for it.

JKTKops:
Local tests (and also CI, as it started running) produced some warnings about "missing SMTLib translations" (e.g. `[Warning] Internal: execution phase: missing SMTLib translation for _[_]_EVM-TYPES`). My local test is still going, but on CI, the Java proofs eventually caused z3 to crash:
```
[Error] Critical: Z3 crashed on input query:
... the whole query, it's unsurprisingly long ...
result:
(error "line 90 column 34: unknown function/constant smt_seq_elem")
unknown

**Review by Riib11 — APPROVED** [review]

**Comment by Riib11** [comment]
why does this break the java tests?

**Comment by JKTKops** [comment]
See the previous comment. I don't really understand it either.

It does look like the latest run had a new error message that I don't think was in the other runs too. A particular test^1 gave up (?) after nearly 6800 steps. It's possible that I introduced a loop somewhere. That would surprise me though. Between other tasks I can keep trying to poke at the Java part of this one.

^1 tests/specs/mcd/flopper-dent-guy-diff-tic-not-0-pass-rough-spec.k

File: tests/specs/benchmarks/verification.k [hunk 8]
```diff
@@ -344,18 +344,13 @@ endmodule
...
     imports VERIFICATION-COMMON
     imports K-REFLECTION
 
-  // ########################
-  // Symbolic Gas
-  // ########################
-
-    syntax Int ::= #gas ( Int , Int , Int )  [function]  // startGas, nonMemory, memory
  // -----------------------------------------------------------------------------------
     rule #gas(_, _, _) <=Int I => false requires #isConcrete(I) [simplification]
```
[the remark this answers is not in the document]
Riib11: I'll see if it can be safely removed

PR #1035, opened 2021-06-09
Fixes `Could not find module: INFINITE-GAS-JAVA`

Riib11:
Previously, `make build -j8` yielded `[Error] Compiler: Could not find module: INFINITE-GAS-JAVA`. This was because of an extra triple-backtick.


ehildenb:
This would probably be better as a suggestion/comment on the PR itself.

JKTKops:
I think let's close this one and do it that way, so it's more visible from the PR.

PR #1037, opened 2021-06-10
Add functional spec and simplifications for allowance-spec.k

### LINKED ISSUE #1028: Debug `make tests/specs/erc20/hkg/allowance-spec.k.prove TEST_SYMBOLIC_BACKEND=haskell`

JKTKops:
fixes #1028

~~Not quite ready for review, I have to merge the Makefile changes first.~~

However I don't know if duplicating rules and adding `[simplification]` attributes is the right way to go, as it feels anti-DRY.


tests/specs/erc20/functional-spec.k
```diff
@@ -0,0 +1,21 @@
...
+    syntax Step ::= Int
+ // -------------------------
+
+endmodule
+
+module FUNCTIONAL-SPEC
+    imports FUNCTIONAL-SPEC-SYNTAX
+
+    claim <k> runLemma(#hashedLocation("Solidity", keccakIntList(OWNER 2 .IntList), SPENDER .IntList))
+           => doneLemma(keccak(#buf(32, SPENDER) ++ #buf(32, keccak(#buf(32, OWNER) ++ #buf(32, 2))))) </k>
```
ehildenb, line 19:
```suggestion
           => doneLemma(keccak(#buf(32, SPENDER) ++ #buf(32, keccak(#buf(32, OWNER) ++ #buf(32, 2))))) ... </k>
```

JKTKops: I keep forgetting this one!

tests/specs/erc20/verification.k
```diff
@@ -25,6 +25,15 @@ module VERIFICATION-JAVA [symbolic, kast]
...
+    imports EDSL
+
+    rule #hashedLocation(_COMPILER, BASE, .IntList) => BASE [simplification]
 
+    rule keccakIntList(VS) => keccak(intList2ByteArray(VS)) [simplification]
+
+    rule intList2ByteArray(.IntList) => .ByteArray [simplification]
+    rule intList2ByteArray(V VS)     => #bufStrict(32, V) ++ intList2ByteArray(VS)
+      requires 0 <=Int V andBool V <Int pow256
+      [simplification]
```
ehildenb, line 38: We probably don't want this solution (but it's definitely a valid one). The basic reason is that we have decided that the `keccakIntList` rule in `hashed-locations.md` should not be applied unless it has fully concrete arguments (we say this by adding that rule to the file `tests/specs/erc20/concrete-rules.txt`). So your second lemma here actually overrides that, and lowers us to the underlying definition of `keccakIntList`, rather than letting us reason about `keccakIntList` directly.

The issue is actually that the `hash1/hash2` abstraction here: https://github.com/kframework/evm-semantics/blob/master/tests/specs/lemmas.k#L166 is not being imported into the Haskell backend. So you could do this:

- Move the definitions of `hash1` and `hash2` to the file `hashed-locations.md`, and add in documentation about them.
- Move the lemmas about `keccakIntList` there into the common lemmas module `LEMMAS` in `tests/specs/lemmas.k`.
- Move the functional test you've added here to `tests/specs/functional/lemmas-spec.k`.

ehildenb: Once you do that, you should confirm that this proof `hkg/allowance` still passes.

JKTKops: Do we not want the `keccakIntList` rules in `hashed-locations.md` to stay there for the non-proof part of `kevm`?

tests/specs/erc20/verification.k
```diff
@@ -25,6 +25,15 @@ module VERIFICATION-JAVA [symbolic, kast]
 endmodule
 
 module VERIFICATION-HASKELL [symbolic, kore]
-    import INFINITE-GAS
+    imports INFINITE-GAS
+    imports EDSL
+
+    rule #hashedLocation(_COMPILER, BASE, .IntList) => BASE [simplification]
```
ehildenb, line 31: You shouldn't need this lemma, it's duplicate of a lemma here: https://github.com/kframework/evm-semantics/blob/master/hashed-locations.md#hashed-location-for-storage

tests/specs/erc20/verification.k
ehildenb: What I'd like to do is just mark all the `#hashedLocation` rules as `concrete`, and then remove everything about `keccakIntList` and such. Just write lemmas about `#hashedLocation` directly in the `hashed-locations.md` file.

tests/specs/erc20/verification.k
JKTKops: All of them are duplicates of lemmas there. I added them incrementally and they each made progress, but I didn't try different combinations of them. I'll try removing this one and see if it still passes.

tests/specs/erc20/functional-spec.k
```diff
@@ -0,0 +1,21 @@
...
+ // -------------------------
+
+endmodule
+
+module FUNCTIONAL-SPEC
+    imports FUNCTIONAL-SPEC-SYNTAX
+
+    claim <k> runLemma(#hashedLocation("Solidity", keccakIntList(OWNER 2 .IntList), SPENDER .IntList))
+           => doneLemma(keccak(#buf(32, SPENDER) ++ #buf(32, keccak(#buf(32, OWNER) ++ #buf(32, 2))))) ... </k>
+      requires #rangeUInt(256, SPENDER) andBool #rangeUInt(256, OWNER)
```
ehildenb, line 20: `rule keccak(#buf(32, OFFSET2) ++ #buf(32, keccak(#buf(32, OFFSET1) ++ BASE_BA))) => #hashedLocation("Solidity", #asWord(BASE_BA), OFFSET1 OFFSET2 .IntList)` and make sure to include requires clause that makes pre-condition valid (ranges on `OFFSET1`, `OFFSET2`, and `#asWord(BASE_BA)`).

JKTKops:
Update: I'm investigating exactly what's happening near the end of the proof, but adding one rewrite rule to try and collapse a hashed location into a `#hashedLocation` did not work on the first try.

JKTKops:
I'm noticing that the `#hashedLocation` terms don't always get fully rewritten to the normal form we decided on. I see several instances of `#hashedLocation("Solidity", keccakIntList(OWNER 2 .IntList), SPENDER .IntList)` which should have been put through one more rewrite to `#hashedLocation("Solidity", keccakIntList(SPENDER keccakIntList(OWNER 2 .IntList) .IntList), .IntList)`. Interestingly, I also see instances of the later.

tests/specs/erc20/verification.k
ehildenb: The new `#hashedLocation` abstraction actually _lowers_ everything to `keccak`, so you may need ot re-run the original proof to see what it's getting stuck on now, it's likely different than before.

JKTKops:
I think there's value in keeping the new `functional-spec` file, so this should be ready for review.

If we don't want to keep it, then the only change in this PR would be to enable the test. In that case, we should just close this one and the test will be enabled by the next `tests/failing-symbolic.haskell` update.

ehildenb (approved):

ehildenb:
I agree, having the file in place makes it easier to debug other ERC20 proofs.

PR #1040, opened 2021-06-11
Build updates for `kevm kompile`

ehildenb:
This PR updates several things, so that `kevm kompile` is more useful for projects that depend on KEVM.

- The `deps/plugin` build products are included in the `kevm` library build too now, so that projects that depend on KEVM have access to them.
- `kevm kompile` now does most of the heavy lifting of deciding which options to send to each backend. This ends up simplifying the `Makefile` quite a bit. It also handles the options relating to plugins which means that users of KEVM can call `kevm kompile` without knowing about the plugin dependencies.
- The `tangle_*` tags are simplified to just `bytes` backends, and `nobytes` backends. We used to need tags `symbolic` and `concrete` as well, but that distinction is mostly handled with module comments anymore.
- The installation of plugin dependencies happens in `$(KEVM_LIB)/blockchain-k-plugin` now instead of mixing those files into KEVM source files. This follows the pattern of other dependencies of installing them into `$(KEVM_LIB)/DEPENDENCY` rather than mixing the files into KEVM source files.

This PR enables building KEVM into mkr-mcd-spec for doing refinement proofs: https://github.com/makerdao/mkr-mcd-spec/pull/243. It works (and passes tests) locally, but the docker image for this branch is not deployed yet so it's not working on CI yet.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -22,7 +22,6 @@ pipeline {
         }
       }
       stages {
-        stage('Deps')  { steps { sh 'make plugin-deps'            } }
         stage('Build') { steps { sh 'make build RELEASE=true -j6' } }
         stage('Test Execution') {
           failFast true
--- a/Makefile
+++ b/Makefile
@@ -98,6 +98,20 @@ $(libff_out): $(PLUGIN_SUBMODULE)/deps/libff/CMakeLists.txt
 
 deps: k-deps plugin-deps
 
+K_MVN_ARGS :=
+ifneq ($(SKIP_LLVM),)
+    K_MVN_ARGS += -Dllvm.backend.skip
+endif
+ifneq ($(SKIP_HASKELL),)
+    K_MVN_ARGS += -Dhaskell.backend.skip
+endif
+
+ifneq ($(RELEASE),)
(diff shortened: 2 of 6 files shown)


dwightguth (approved):
This seems okay. I'm not sure how I feel about this `kevm compile script, though. You say that it makes the Makefile simpler, but that's counterbalanced by the increase in complexity to the kevm script. Overall it seems that this PR adds lines rather than removing them, which is less than ideal if your goal is simplification. But I don't really feel too strongly about it as long as it's still possible to build just one backend and then run the tests for that backend.

ehildenb:
@dwightguth  To be fair I'm not sure how I feel about it either, but I'm experimenting with the idea on this repo and the mkr-mcd-spec and erc20-verification repos. I ran a small version of the experiment already which has been in `master` for a bit, this just brings in more of the options because I was happy with how it turned out.

The major downside I would see is if you want to call `kompile` on a dependent project, but _without_ the defaults that `kevm kompile` sets for you. This does legitimately make both the erc20-verification and the mkr-mcd-spec build systems simpler though, because they don't need to repeat all the options to `kompile` that KEVM already sets.

PR #1042, opened 2021-06-13
Remove VMTests on Haskell backend stage

ehildenb:
- ~Builds both java/haskell backends against EDSL instead of ETHEREUM-SIMULATION. This means that the `--main-module` contains more of the syntax by default, to make parsing/unparsing issues in the kore-repl go away.~
- Stops running the VMTests for haskell backend. The VMTests slowed down on Haskell backend for #1005, which sped up the proofs significantly.

~BLOCKED on kframework/k#2024~

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -28,7 +28,7 @@ pipeline {
           failFast true
           options {
             lock("proofs-${env.NODE_NAME}")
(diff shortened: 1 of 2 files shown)


JKTKops (approved):
Note this is a subset of changes from #1034 which are not blocked.

Issue #1027, opened 2021-06-07
Debug `make tests/specs/benchmarks/bytes00-spec.k.prove TEST_SYMBOLIC_BACKEND=haskell`


ehildenb:

ehildenb:
`rule (#buf(DATA_LEN, DATA) ++ _BUF) [0 .. DATA_LEN] => #buf( DATA_LEN , DATA ) [simplification]`

Riib11:
rule SIZE <Int #powByteLen(SIZE) => true requires 0 <=Int SIZE [simplification]

Riib11:
solved by #1049

PR #1049, opened 2021-06-15
Tests bytes00-spec and keccak00-spec now pass

Riib11:
Fixes: #1027 
Fixes: #1012 

The following tests now pass:
- `tests/specs/benchmarks/bytes00-spec.k`
- `tests/specs/benchmarks/keccak00-spec.k`

The new lemmas introduced are:
- `rule #bufStrict(SIZE, DATA) => #buf(SIZE, DATA) requires #range(0 <= DATA < (2 ^Int (SIZE *Int 8)))`
- `rule (#buf(DATA_LEN, DATA) ++ _BUF) [0 .. DATA_LEN] => #buf( DATA_LEN , DATA ) [simplification]`


Riib11:
I will update this soon with a claim spec to test simplification of the term that I handled.

Riib11:
Fails CI. Perhaps there were changes I relied on from #1034 that I didn't account for in this PR.

ehildenb:
No it's just timing out, you can re-trigger the job.

tests/failing-symbolic.haskell
```diff
@@ -1,4 +1,3 @@
-tests/specs/benchmarks/bytes00-spec.k
```
ehildenb, line 1: Do these tests pass in <20m on your machine? If not, it's unlikely they'll make it past CI. We should definitely add the term which is simplified by this lemma to the `tests/specs/benchmarks/functional-spec.k` file, because we won't be able to enable these tests yet (and running that file first to debug updates to KEVM is much easier than running the entire proofs).

Riib11: No, they take closer to 40m. I will remove the tests from `tests/failing-symbolic.haskell` and add that spec to `tests/specs/benchmarks/functional-spec.k` next.

tests/specs/benchmarks/functional-spec.k
```diff
@@ -27,4 +27,8 @@ module FUNCTIONAL-SPEC
 
     claim <k> runLemma(#ceil32(DATA_LEN) +Int 292 <Int DATA_LEN +Int 292) => doneLemma(false) ... </k> requires #range(0 <= DATA_LEN < pow16)
 
+    claim <k> runLemma ( #buf ( 32 , keccak ( #range ( b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x80" [ 128 := ( #buf ( DATA_LEN , DATA ) ++ #buf ( ( DATA_LEN up/Int 32 ) *Int 32 -Int DATA_LEN , 0 ) ) [ 0 .. DATA_LEN ] ] , 128 , DATA_LEN ) ) ) ) =>
```
Riib11, line 30: This is surely an incorrect way of representing the "starting" byte array. How would I write this in k-friendly form? Perhaps there is a macro defined somewhere for it that I didn't find?

ehildenb: But I think you can replace the entire `b"\x...."` with just `_:ByteArray`, and the test is still testing the lemmas well??

ehildenb: Or you could replace it with `BA:ByteArray`, and add the side-condition `#sizeByteArray(BA) ==Int 32`.

Riib11:
Still fails CI because of timeout.

ehildenb (approved):

ehildenb:
Blocked on #1055

tests/specs/lemmas.k
```diff
@@ -30,6 +30,8 @@ module LEMMAS
     rule _BUF [ _ .. W ] => .ByteArray requires W ==Int 0                    [simplification]
     rule  BUF [ 0 .. W ] => BUF        requires W ==Int #sizeByteArray(BUF)  [simplification]
 
+    rule (#buf(DATA_LEN, DATA) ++ _BUF) [0 .. DATA_LEN] => #buf( DATA_LEN , DATA ) [simplification]
+
```
ehildenb, line 34: Actually, I think the three rules below this rule should subsume this one completely. Specifically the second rule should, starting on line 39.

So I'm actually surprised it doesn't work. I think we should figure out why the below lemmas are not applying. It could be that it cannot show `#range(0 <= START < #sizeByteArray(#buf(DATA_LEN, DATA)))`, or that it cannot show `START +Int WIDTH <=Int #sizeByteArray(DATA_LEN, DATA)`. You should:

- Remove this added lemma.
- Add functional claims that the two side-conditions above simplify to `true`, in addition to the one you already added above.
- Add lemmas which make those two side-conditions simplify to `true`.

Then the original claim should also simplify out, but without duplicating lemmas for `(_ ++ _) [_ .. _]`.

Riib11 (commented):
Unfortunately, adding
```
rule BUF1 ++ BUF2 => BUF2 requires #sizeByteArray(BUF1) ==Int 0 [simplification]
```
doesn't seem to be enough

tests/specs/lemmas.k
Riib11: Replacing the lemma
```
rule (BUF1 ++_BUF2)[START .. WIDTH] => (BUF1[START .. WIDTH])
      requires #range(0 <= START < #sizeByteArray(BUF1))
       andBool START +Int WIDTH <=Int #sizeByteArray(BUF1) [simplification]
```
with
```
rule (BUF1 ++ BUF2)[START .. WIDTH] => (BUF1[START .. #sizeByteArray(BUF1) -Int START]) ++ (BUF2[0 .. START +Int WIDTH -Int #sizeByteArray(BUF1)])
      requires 0 <=Int START 
        andBool 0 <=Int WIDTH
        andBool START +Int WIDTH <=Int #sizeByteArray(BUF1)
        [simplification]
```
seems to be a viable solution, since it's requirements are weaker and it makes the following claim pass:
```
claim <k> runLemma  ( (#buf (SIZE, DATA) ++ _BUF) [ 0 .. SIZE ] )
           => doneLemma ( #buf (SIZE, DATA) ) </k>
      requires #range(0 <= DATA < 2 ^Int (SIZE *Int 8))
       andBool 0 <=Int SIZE
```

tests/specs/lemmas.k
```diff
@@ -33,9 +33,10 @@ module LEMMAS
     rule (BUF1 ++ BUF2)[START .. WIDTH] => (BUF1[START .. #sizeByteArray(BUF1) -Int START]) ++ (BUF2[0 .. START +Int WIDTH -Int #sizeByteArray(BUF1)])
       requires #range(0 <= START < #sizeByteArray(BUF1))
        andBool #sizeByteArray(BUF1) <Int START +Int WIDTH [simplification]
-
-    rule (BUF1 ++_BUF2)[START .. WIDTH] => (BUF1[START .. WIDTH])
-      requires #range(0 <= START < #sizeByteArray(BUF1))
+        
+    rule (BUF1 ++ _BUF2)[START .. WIDTH] => (BUF1[START .. WIDTH])
```
ehildenb, line 37:
```suggestion
    rule (BUF1 ++ _BUF2)[START .. WIDTH] => BUF1[START .. WIDTH]
```

Can you also clean up the formatting of the other ones too?

ehildenb (approved):

tests/specs/benchmarks/functional-spec.k
```diff
@@ -28,5 +28,5 @@ module FUNCTIONAL-SPEC
     claim <k> runLemma(#ceil32(DATA_LEN) +Int 292 <Int DATA_LEN +Int 292) => doneLemma(false) ... </k> requires #range(0 <= DATA_LEN < pow16)
 
     claim <k> runLemma(#buf(4, 0) ++ (#buf(32, 6) ++ BA:ByteArray)) => doneLemma(#buf(36, 6) ++ BA:ByteArray) ... </k>
-
+
```
ehildenb, line 31:
```suggestion

```

ehildenb (approved):

PR #1057, opened 2021-06-21
Optimize some specific rules

ehildenb:
By making some of the higher-frequency rules higher priority, maybe we can optimize how the semantics operates.

Initial data shows this as a net loss, but I'm re-collecting.


ehildenb:
This pretty much has proven to be a de-optimization.

PR #1058, opened 2021-06-21
Updates to Build system for symbolic backends

ehildenb:
- Symbolic backends now use EDSL as main module.
- All backends will not rebuild (unnecessarily) if some of the lemmas files in `tests/specs` changes.

Blocked on: https://github.com/kframework/k/issues/2024

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,20 +140,20 @@ plugin-deps: $(plugin_includes) $(plugin_c_includes)
 
 KOMPILE := $(KEVM) kompile
 
-kevm_includes := abi.md              \
-                 asm.md              \
-                 buf.md              \
-                 data.md             \
-                 driver.md           \
-                 edsl.md             \
-                 evm.md              \
-                 evm-types.md        \
-                 hashed-locations.md \
-                 json-rpc.md         \
-                 network.md          \
-                 optimizations.md    \
-                 serialization.md    \
-                 state-loader.md
+kevm_files := abi.md              \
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -166,7 +166,9 @@ kevm_lemmas := infinite-gas.k                           \
 
 lemma_includes := $(patsubst %, $(KEVM_INCLUDE)/kframework/lemmas/%, $(kevm_lemmas))
 
-includes := $(patsubst %, $(KEVM_INCLUDE)/kframework/%, $(kevm_includes)) $(plugin_includes) $(lemma_includes)
+kevm_includes := $(patsubst %, $(KEVM_INCLUDE)/kframework/%, $(kevm_files))
+
+includes := $(kevm_includes) $(lemma_includes) $(plugin_includes) $(plugin_c_includes)
```
JKTKops, line 171: It doesn't look like `$(plugin_c_includes)` was _removed_ anywhere, so just want to double check that it was indeed supposed to be added here.

JKTKops (approved):

Makefile
ehildenb: Yeah, the `$(includes)` variable is just used to make sure that we re-copy the `kevm` script into the build directory whenever one of the include files changes. This is because the `kevm` script is supposed to be able to enumerate those files for various things.

Basically, the `$(includes)` isn't used for much, so I'm just putting all the includes in there.

PR #1060, opened 2021-06-21
Add --profile Option to kevm Script

JKTKops:
When analyzing the performance of rules on the Haskell backend it is useful to get a Kore profile of executed proofs. This requires modifying the `haskell-backend-command` to pass some RTS options, and an additional call to `kore-prof` after `kprove` is done.

`--profile` is similar in spirit to `--bug-report`. Both of them only have effects in the `prove` mode, and only on the `haskell` backend.  `--profile` will complain similarly to `--bug-report` if used with a non-haskell backend. Neither of them were documented in the `help` mode, so I added them as `<KEVM_arg>` options. However I'm not sure if it should mention that they only have effects in `prove` mode.

I don't have much experience with Bash so it's entirely possible there's a better way to do things here.


kevm
```diff
@@ -103,21 +103,30 @@ run_kast() {
...
     check_k_install
 
     def_module="$1" ; shift
     bug_report_name="kevm-bug-$(basename "${run_file%-spec.k}")"
+    eventlog_name="$run_file.log.json"
+
+    haskell_backend_args=(initial)
+    ! $bug_report || haskell_backend_args+=(--bug-report "$bug_report_name")
+    ! $profile    || haskell_backend_args+=(+RTS -l -ol$eventlog_name -RTS)
+    haskell_backend_args="${haskell_backend_args[@]:1}"
```
ehildenb, line 117: I would instead start the `haskell_backend_args=(kore-exec)` as the intiial array, and add to it. Then later, have a check which looks like `[[ "${#haskell_backend_args}" -le 1 ]] || proof_args+=(--haskell-backend-command "${haskell_backend_args}")`

ehildenb:
Do the options `--not-matching` and `--not-matching-children` pass through smoothly?

kevm
```diff
@@ -103,21 +103,32 @@ run_kast() {
...
 
     ! $debug || set -x
-    kprove --directory "$backend_dir" "$run_file" "${proof_args[@]}" "$@"
+    if [[ "$profile_timeout" != "0" ]]
+    then
+        timeout -s INT "$profile_timeout" kprove --directory "$backend_dir" "$run_file" "${proof_args[@]}" "$@" || true
+    else
+        kprove --directory "$backend_dir" "$run_file" "${proof_args[@]}" "$@" || true
+    fi
+    ! $profile || kore-prof "$eventlog_name" $kore_prof_args > "$eventlog_name.json"
```
ehildenb, line 131: This should likely read (formatting and logic changes):

```suggestion
    if $profile && [[ "$profile_timeout" != "0" ]]; then
        timeout -s INT "$profile_timeout" kprove --directory "$backend_dir" "$run_file" "${proof_args[@]}" "$@" || true
        kore-prof "$eventlog_name" $kore_prof_args > "$eventlog_name.json"
    else
        kprove --directory "$backend_dir" "$run_file" "${proof_args[@]}" "$@"
    fi
```

kevm
```diff
@@ -246,6 +257,10 @@ if [[ "$run_command" == 'help' ]] || [[ "$run_command" == '--help' ]] ; then
                                       [--schedule (ISTANBUL|PETERSBURG|CONSTANTINOPLE|BYZANTIUM|SPURIOUS_DRAGON|TANGERINE_WHISTLE|HOMESTEAD|FRONTIER|DEFAULT)]
                                       [--chainid NNN]
                                       [--concrete-rules-file <concrete_rules_file>]
+                                      [--bug-report]
+                                      [--profile]
+                                      [--profile-timeout]
+                                      [--kore-prof-args]
```
ehildenb, line 263:
```suggestion
                                      [--bug-report]
                                      [--profile]
                                      [--profile-timeout <timeout_seconds>]
                                      [--kore-prof-args "<kore_prof_arg>*"]
```

kevm
JKTKops: Initially I didn't have `|| true` on either branch and couldn't figure out why the timeout was killing the whole script (I'm clearly not used to bash 😅), then `$profile` in the condition didn't do anything. However this is both clearer and preserves the status code of `kprove` which is quite important.

However `kore-prof` needs to stay outside of the `if`, because it has to happen even if there is no timeout.

Riib11: I noticed that sometimes, if a `.json` file already existed at the expected place that weird things happened e.g. the new log got appended to the old `.json` content. This may be a user error, but to be safe, I suggest removing any old `.json` file that exists there before writing the one.

JKTKops: That sounds like user error because the file is overwritten with the `>` redirect. Appending would happen if `>>` were used.

Are you sure you weren't putting log entries in a `.json` file (e.g with `--log [file] --log-entries [entries]`)? `kore-exec` appends those to the file, but that's not related to this PR.

Riib11: Ah, I was using an old version of your script where it wrote the log to the same file as ended up being the `.json` e.g.
```
! $profile || { kore-prof "$eventlog_name" > "$eventlog_name.tmp" \
               &&   mv "$eventlog_name.tmp" "$eventlog_name"; }
```
it looks like the latest version doesn't do that, so probably not an issue anymore

ehildenb (approved):

Riib11 (approved):

PR #1066, opened 2021-06-23
Simplify Concrete Byte Arrays on Haskell Backend

JKTKops:
Left-associates concrete byte arrays on the haskell backend so that they can get simplified.


JKTKops:
Oops, I carried over `pull-out-infinite-gas` when I made the branch. Closed this one and I will fix that and open a new one.

PR #1067, opened 2021-06-23
Simplify Concrete ByteArrays on Haskell Backend

JKTKops:
Left-associates concrete byte arrays on the haskell backend so that they can get simplified.


ehildenb:
@JKTKops did you confirm that this ends up simplifying out the larger term in teh proof you were working with? If so, feel free to add `automerge` label.

ehildenb (approved):

JKTKops:
I have confirmed that the term simplifies.

Issue #1062, opened 2021-06-22
Remove unused `#take` lemmas


ehildenb:
We should not have any lemmas about `#take` or `#drop`, instead they should be over `_[_.._]`.

This also means that we need to make sure that all the `concrete-rules.txt` files contain BOTH rules (for the java and haskell backends).

The rules are in `evm-types.md`.

Riib11:
(As in #1024)

After these changes:
- disabled all lemmas in `tests/specs/lemmas.k` and `tests/specs/benchmark/verification.k` that use either `#take` or `#drop`
- added `ByteArray.range` and `bytesRange` to all `concrete-rules.txt` files

The result is that `make test-prove TEST_SYMBOLIC_BACKEND=haskell` still passes! This indicates the those lemmas are not being used at all after the rules that consume `#take` and `#drop` were concretized. No new lemmas over `_[_.._]` were needed.

PR #1068, opened 2021-06-23
Prune lemmas that refer to `#take` and `#drop`

Riib11:
Fixes: #1062

The functions `#take` and `#drop` are handled slightly differently by the Java and Haskell backends. So, it is bad form to have lemmas that directly refer to `#take` and `#drop`. Instead, better lemmas should refer to the function `_[_.._]` that simplifies to the appropriate use of `#take` and `#drop` for each backend.

- .bytes:
```
rule [bytesRange] : WS [ START .. WIDTH ] => substrBytes(padRightBytes(WS, START +Int WIDTH, 0), START, START +Int WIDTH)
      requires WIDTH >=Int 0 andBool START >=Int 0 andBool START <Int #sizeByteArray(WS)
```

- .nobytes:
```
rule [ByteArray.range]: WS [ START .. WIDTH ] => #take(WIDTH, #drop(START, WS))
```

This PR cleans up lemmas outside of `evm-types.md` -- mainly lemmas in `tests/specs/lemmas.k` and `tests/specs/**/verification.k` -- so that none refer directly to `#take` and `#drop`.


Riib11:
I'll remove `out_test-prove-benchmarks*` once CI is passing.

Riib11:
`tests/specs/benchmarks/ecrecoverloop02-sig1-invalid-spec.k.prove` is timing out in CI

denis-bogdanas (approved):

PR #1081, opened 2021-07-12
Generalize Haskell lemmas in `tests/specs/lemmas.k` and `tests/specs/benchmarks/verification.k`

### LINKED ISSUE #1024: Inspect lemmas in `tests/specs/lemmas.k` and `tests/specs/benchmarks/verification.k` for duplicates/generalizations

Riib11:
Fixes #1024

In `tests/specs/lemmas.k` and `tests/specs/benchmarks/verification.k`:
- reformats and reorganizes some lemmas (only within files -- nothing is moved between files)
- removes unneeded lemmas
- removes lemmas in `tests/specs/benchmarks/verification.k` that are already covered by lemmas in `tests/specs/lemmas.k`

### DIFF
--- a/tests/specs/benchmarks/verification.k
+++ b/tests/specs/benchmarks/verification.k
@@ -113,7 +113,8 @@ module VERIFICATION
   // Simplification
   // ########################
 
-    rule #asWord(#buf(32, DATA)) => DATA  requires #rangeUInt(256, DATA) [simplification]
+    // TODO: possibly generalizes via lemma from lemmas.k
+    // rule #asWord(#buf(32, DATA)) => DATA  requires #rangeUInt(256, DATA) [simplification]
 
     rule #range(_M [ N := #buf(WIDTH, DATA) ], N, WIDTH) => #buf(WIDTH, DATA) [simplification]
 
--- a/tests/specs/benchmarks/verification.k
+++ b/tests/specs/benchmarks/verification.k
@@ -50,20 +50,6 @@ module VERIFICATION
     rule          #symEcrec(_DATA) <Int pow256 => true [simplification]
 
   // ########################
-  // Symbolic Call
-  // ########################
-
-    syntax Int ::= #extCodeSize ( Int )  [function]
- // -----------------------------------------------
-
-  // ########################
-  // Rule Replacement
-  // ########################
-
-    // TODO: is this needed anymore?
-    // claim <k> EXTCODESIZE ACCT => #extCodeSize(ACCT) ~> #push ... </k>  [trusted]
-
-  // ########################
   // ABI Encoding
   // ########################
 
--- a/tests/specs/benchmarks/verification.k
+++ b/tests/specs/benchmarks/verification.k
@@ -96,7 +96,7 @@ module VERIFICATION
     rule #sizeWordStack(WS, N) <Int SIZE => #sizeWordStack(WS, 0) +Int N <Int SIZE  requires N =/=Int 0 [simplification]
 
     rule SIZELIMIT <Int #sizeWordStack(WS, N) +Int DELTA  => SIZELIMIT <Int (#sizeWordStack(WS, 0) +Int N) +Int DELTA  requires N =/=Int 0 [simplification]
-    rule SIZELIMIT <Int #sizeWordStack(WS, N)             => SIZELIMIT <Int #sizeWordStack(WS, 0) +Int N               requires N =/=Int 0 [simplification]
+    rule SIZELIMIT <Int #sizeWordStack(WS, N)             => SIZELIMIT <Int  #sizeWordStack(WS, 0) +Int N              requires N =/=Int 0 [simplification]
 
     rule #sizeWordStack(WS, N) <=Int SIZE => #sizeWordStack(WS, 0) +Int N <=Int SIZE  requires N =/=Int 0 [simplification]
 
--- a/tests/specs/benchmarks/verification.k
+++ b/tests/specs/benchmarks/verification.k
@@ -60,7 +60,8 @@ module VERIFICATION
   // Rule Replacement
   // ########################
 
-    claim <k> EXTCODESIZE ACCT => #extCodeSize(ACCT) ~> #push ... </k>  [trusted]
+    // TODO: is this needed anymore?
+    // claim <k> EXTCODESIZE ACCT => #extCodeSize(ACCT) ~> #push ... </k>  [trusted]
 
   // ########################
   // ABI Encoding
--- a/tests/specs/benchmarks/verification.k
+++ b/tests/specs/benchmarks/verification.k
@@ -182,7 +182,7 @@ module VERIFICATION
 
     rule ((X *Int Y) /Int Z) /Int Y => X /Int Z  requires Y =/=Int 0 [simplification]
 
-    // x &Int (NOT 31)
+    // X &Int (NOT 31)
     rule X &Int 115792089237316195423570985008687907853269984665640564039457584007913129639904 => (X /Int 32) *Int 32  requires 0 <=Int X [simplification]
 
     rule (X /Int 32) *Int 32 => X  requires X modInt 32 ==Int 0 [simplification]
--- a/tests/specs/benchmarks/verification.k
+++ b/tests/specs/benchmarks/verification.k
@@ -127,7 +127,8 @@ module VERIFICATION
   // Simplification
   // ########################
 
-    rule #asWord(#buf(32, DATA)) => DATA  requires #rangeUInt(256, DATA) [simplification]
+    // TODO: is generalized by `#asWord(#buf(N, BUF)) => BUF` in `lemmas.k`?
+    // rule #asWord(#buf(32, DATA)) => DATA  requires #rangeUInt(256, DATA) [simplification]
 
     rule #range(_M [ N := #buf(WIDTH, DATA) ], N, WIDTH) => #buf(WIDTH, DATA) [simplification]
 
--- a/tests/specs/benchmarks/verification.k
+++ b/tests/specs/benchmarks/verification.k
@@ -127,8 +127,7 @@ module VERIFICATION
   // Simplification
   // ########################
 
-    // TODO: is generalized by `#asWord(#buf(N, BUF)) => BUF` in `lemmas.k`?
-    // rule #asWord(#buf(32, DATA)) => DATA  requires #rangeUInt(256, DATA) [simplification]
+    rule #asWord(#buf(32, DATA)) => DATA  requires #rangeUInt(256, DATA) [simplification]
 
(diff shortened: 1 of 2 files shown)


tests/specs/benchmarks/verification.k
```diff
@@ -60,7 +60,8 @@ module VERIFICATION
   // Rule Replacement
   // ########################
 
-    claim <k> EXTCODESIZE ACCT => #extCodeSize(ACCT) ~> #push ... </k>  [trusted]
+    // TODO: is this needed anymore?
```
Riib11, line 63: Is this claim needed for something? Its odd to see a `claim` outside of a `-SPEC` module

tests/specs/benchmarks/verification.k
```diff
@@ -127,7 +127,8 @@ module VERIFICATION
   // Simplification
   // ########################
 
-    rule #asWord(#buf(32, DATA)) => DATA  requires #rangeUInt(256, DATA) [simplification]
+    // TODO: is generalized by `#asWord(#buf(N, BUF)) => BUF` in `lemmas.k`?
```
Riib11, line 130: no this doesn't work

ehildenb: Hmmmmm, that's strange. Did it consistently fail on re-runs (and on your machine?)

The LLVM conformance test-suite doesn't even see this file.

tests/specs/benchmarks/verification.k
Riib11: in this case, the `claim` is acting like a rule basically

tests/specs/lemmas.k
```diff
@@ -9,9 +9,12 @@ module LEMMAS
   // Arithmetic
   // ########################
 
-    //For #bufStrict simplification in benchmarks
-    rule 0 <=Int #ceil32(I)        => true requires 0 <=Int I [smt-lemma, simplification]
+    // For #bufStrict simplification in benchmarks
+    // TODO: why does this need to be an `smt-lemma`?
+    rule 0 <=Int #ceil32(I)        => true requires 0 <=Int I [simplification, smt-lemma]
     rule 0 <=Int #ceil32(I) -Int I => true requires 0 <=Int I [simplification]
+    // // TODO: possibly generalizes the above two rules
```
Riib11, line 16: no, needs to match structurally more specifically

tests/specs/benchmarks/verification.k
Riib11: turns out I started a new CI job before this one finished, so that's why it failed

Riib11:
I plan to have this PR ready for review as early as I can tomorrow

tests/specs/lemmas.k
```diff
@@ -335,9 +335,10 @@ module LEMMAS-HASKELL [symbolic, kore]
 
     rule 0 <=Int lengthBytes ( _ ) => true [smt-lemma, simplification]
 
-    // TODO: why do we need an explicit symmetry rule for `==K`? should this not happend automatically?
```
Riib11: why do we need an explicit symmetry rule for `==K`? I tried combining these two rules into one (on line 343 below) to see if it works without the symmetry rule

Riib11: The generalization I tried (on line 343 in the commit) does not work. I fails for a proof in `tests/specs/functional/lemmas-spec.k`, but that one only. Perhaps there is a correct way to though?

tests/specs/lemmas.k
```diff
@@ -199,32 +194,37 @@ module LEMMAS-JAVA [symbolic, kast]
...
+    rule I +Int B => B         +Int I  requires #isConcrete(I) andBool notBool #isConcrete(B) [simplification]
+    rule A -Int I => A +Int (0 -Int I) requires #isConcrete(I) andBool notBool #isConcrete(A) [simplification]
 
-    rule (A +Int I2) +Int I3 => A +Int (I2 +Int I3) when notBool #isConcrete(A) andBool #isConcrete(I2) andBool #isConcrete(I3) [simplification]
+    // associate concrete terms to the left and symbolic terms to the right
+    
+    // // OLD
+    // rule (A  +Int  J) +Int  K  =>  A +Int (J  +Int K) requires #isConcrete(J) andBool #isConcrete(K) andBool notBool #isConcrete(A) [simplification]
+    // rule  I  +Int (B  +Int  K) =>  B +Int (I  +Int K) requires #isConcrete(I) andBool #isConcrete(K) andBool notBool #isConcrete(B) [simplification]
+    // NEW (moves symbolic terms to the fight)
```
ehildenb, line 206: GitHub will show us old/new, so no need for the comment.
```suggestion
```

tests/specs/lemmas.k
```diff
@@ -199,32 +194,37 @@ module LEMMAS-JAVA [symbolic, kast]
...
+    rule (A  +Int  J) +Int  K  => (J +Int K) +Int A requires #isConcrete(J) andBool #isConcrete(K) andBool notBool #isConcrete(A) [simplification]
```
ehildenb, line 207: This rule seems to be moving symbolic terms to the right, but the rule above (line 198) move symbolic terms to the left. This doesn't make sense to me.

tests/specs/lemmas.k
```diff
@@ -194,25 +194,27 @@ module LEMMAS-JAVA [symbolic, kast]
...
-    rule  I  +Int (J  -Int  C) => (I +Int J) -Int C requires #isConcrete(I) andBool #isConcrete(J) andBool notBool #isConcrete(C) [simplification]
-
-    rule  I  -Int (J  +Int  C) => (I -Int J) -Int C requires #isConcrete(I) andBool #isConcrete(J) andBool notBool #isConcrete(C) [simplification]
-    rule  I  -Int (J  -Int  C) => (I -Int J) +Int C requires #isConcrete(I) andBool #isConcrete(J) andBool notBool #isConcrete(C) [simplification]
-
-    rule  I  &Int (J  &Int  C) => (I &Int J) &Int C requires #isConcrete(I) andBool #isConcrete(J) andBool notBool #isConcrete(C) [simplification]
+    // 3 terms
+
+    // symbolic first: A, J, K
+    rule (A  +Int  J) +Int  K  =>  A  +Int (J  +Int  K) requires #isConcrete(J) andBool #isConcrete(K) andBool (notBool #isConcrete(A)) [simplification]
```
Riib11, line 207: fixed previous issue [mentioned](https://github.com/kframework/evm-semantics/pull/1081#discussion_r670535924)

tests/specs/lemmas.k
Riib11: fixed [here](https://github.com/kframework/evm-semantics/pull/1081/commits/274ff6a354e72a6a27dd0ee3602a992073837fec#r670663973)

tests/specs/lemmas.k
```diff
@@ -75,25 +65,29 @@ module LEMMAS
...
+    rule 0 <=Int #lookup( _M:Map , _ )             => true [simplification, smt-lemma]
+    rule         #lookup( _M:Map , _ ) <Int pow256 => true [simplification, smt-lemma]
 
+    // re-order assignments
+    rule MEM [ K1 := BUF1:ByteArray ] [ K2 := BUF2:ByteArray ] => MEM [ K2 := BUF2 ] [ K1 := BUF1 ]  requires         #sizeByteArray(BUF2) <=Int K1 -Int K2                   andBool K2  <Int K1 [simplification]
+    // overwritten assignment
+    rule MEM [ K1 := BUF1           ] [ K2 := BUF2           ] => MEM [ K2 := BUF2 ]                 requires K1 +Int #sizeByteArray(BUF1) <=Int K2 +Int #sizeByteArray(BUF2) andBool K2 <=Int K1 [simplification]
     // redundant assignment
-    rule MEM [ K  := BUF1           ] [ K  := BUF2           ] => MEM [ K  := BUF2 ]                 requires #sizeByteArray(BUF1) ==Int #sizeByteArray(BUF2)                                     [simplification]
+    rule MEM [ K  := BUF1           ] [ K  := BUF2           ] => MEM [ K  := BUF2 ]                 requires         #sizeByteArray(BUF1) ==Int #sizeByteArray(BUF2)                             [simplification]
```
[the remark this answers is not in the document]
gtrepta: Here's an example where the second works but not the first:
```MEM [ 2 := "1" ] [ 1 := "123" ]```

And the first doesn't look like 'overwritten' assignment, the two assignments are disjoint and are just being reordered.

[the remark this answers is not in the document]
Comment (gtrepta) [reply tests/specs/lemmas.k:86]: Oh, I realized you may have been referring to the last two of the three and not the first two...

File: tests/specs/lemmas.k [hunk 9]
```diff
@@ -199,32 +194,35 @@ module LEMMAS-JAVA [symbolic, kast]
...
+    rule I +Int B => B +Int I          requires #isConcrete(I) andBool (notBool #isConcrete(B)) [simplification]
+    // converting -Int to +Int for concrete values.
+    rule A -Int I => A +Int (0 -Int I) requires #isConcrete(I) andBool (notBool #isConcrete(A)) [simplification]
+    
+    // 3 terms
+
+    // symbolic first: A, J, K
+    rule (A  +Int  J) +Int  K  =>  A  +Int (J  +Int  K) requires #isConcrete(J) andBool #isConcrete(K) andBool (notBool #isConcrete(A)) [simplification]
+    // symbolic second: I, B, K
+    rule  I  +Int (B  +Int  K) => (I  +Int  K) +Int  B  requires #isConcrete(I) andBool #isConcrete(K) andBool (notBool #isConcrete(B)) [simplification]
```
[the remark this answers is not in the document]
Riib11: You're right, I fixed this here so they are consistent: https://github.com/kframework/evm-semantics/pull/1081/commits/b335da4d9f8186880f4bc39b9f3e83c910da7fce

Turns out this whole section is redundant and ends up being removed in #1088.

tests/specs/lemmas.k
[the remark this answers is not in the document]
Riib11: I also think @rv-jenkins (aka @ehildenb) was probably referring to the last two.

I think you're right, if K1 ==Int K2 then the second rule implies the third rule, since then
- the first side condition of the second rule simplifies to the first side condition of the third rule:
```
K1 +Int #sizeByteArray(BUF1) <=Int K2 +Int #sizeByteArray(BUF2)
=> (since K1 ==Int K2)
K1 +Int #sizeByteArray(BUF1) <=Int K1 +Int #sizeByteArray(BUF2)
=>
#sizeByteArray(BUF1) <=Int #sizeByteArray(BUF2)
```
- the second side condition of the second rule simplifies to true (and the third rule has no second side condition)
```
K2 <=Int K1
=> (since K1 ==Int K2)
true
```

I removed the redundant rule here: https://github.com/kframework/evm-semantics/pull/1081/commits/c8b4f79e32e2fdb99a163445648633e40a69a1e8

ehildenb (approved):

PR #1088, opened 2021-07-16
[OLD] Pull out tests/specs/int-simplification.k

### LINKED ISSUE #1086: Pull out `tests/specs/int-simplification.k`
Lots of the lemmas files/verificaiton files have integer simplification routines, including things which dictate how we re-associate integer terms.

I would recommend starting by pulling out the integer simplification stuff from `tests/specs/infinite-gas.k` into `tests/specs/int-simplification.k`, and making sure tests pass with that refactoring. Then pull out the integer simplification stuff from `tests/specs/lemmas.k` (which you likely just need to delete things there, not migrate anything), and make tests pass again (by including `tests/specs/int-simplification` into `tests/specs/lemmas.k`).

Finally go through each verification file and remove integer simplification lemmas, making sure that the given test-suite passes with the update.

Ideally we would end up with just one file which contains all the integer simplifications. Note that both arithmetic operators and comparison operators (eg. `_<=Int_`) can be included in this file. Also don't forget things like `modInt` or `^Int` or any pure integer + comparison operators.

But anything specific to KEVM (eg. `#lookup(...)` or `chop(...)` should not be moved).

Riib11:
DEPRECATED: go to #1097

~Fixes #1086~

Pulls out Int-related lemmas from several modules used among the test suite, in particular `tests/specs/infinite-gas.k` and `tests/specs/lemmas.k`.

The Int-relatesd symbols conglomerated in Int-simplification are:
- `-Int`, `+Int`, `*Int`, `|Int`
- `&Int`
- `<Int`, `<=Int`, `>=Int`, `>Int`
- `maxInt`, `minInt`

Note that Int-related lemmas that use symbols that are KEVM specific (e.g. `#lookup` and `chop`) are not include in `tests/specs/int-simplification.k`. The module `INT-SIMPLIFICATION` only imports `INT` and `BOOL`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,28 +1,28 @@
 ARG K_COMMIT
 FROM runtimeverificationinc/kframework-k:ubuntu-focal-${K_COMMIT}
 
-RUN    apt-get update           \
-    && apt-get upgrade --yes    \
-    && apt-get install --yes    \
-            cmake               \
-            curl                \
-            debhelper           \
-            jq                  \
-            libboost-test-dev   \
-            libcrypto++-dev     \
-            libgflags-dev       \
-            libprocps-dev       \
-            libsecp256k1-dev    \
-            libssl-dev          \
-            libyaml-dev         \
-            maven               \
-            netcat-openbsd      \
-            openjdk-11-jdk      \
-            pkg-config          \
-            python3             \
-            python-pygments     \
-            rapidjson-dev       \
-            z3                  \
+RUN    apt-get update            \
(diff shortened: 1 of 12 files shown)


Riib11:
I expect this to pass the next CI run

Riib11:
I think I've done everything reasonable I can for this PR, so I'll mark it as ready for review once #1081 gets approved.

Riib11:
The most recent CI fail may have been because of a CI timeout problem. Lets see if it works now

PR #1090, opened 2021-07-21
Fix Makefile and kevm script for building on MacOS BigSur

spencerhaoxiao:
1. Both `Makefile` and `bin/kevm` assume that the brew installation directory on MacOS is `/usr/local` and thus use this directory to link to the include and lib folders of libraries installed by brew. However, in BigSur with Apple Silicon (M1), the brew installation directory is `/opt/homebrew` (see  [home-brew install.sh](https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh) line 35). Thus in the two scripts we use `brew --prefix` to find the installation directory of home-brew. 

2. On MacOS, both `libsecp256k1` and `libcryptopp` need to be built from source before running the commands to generate Haskell and LLVM backends. The change is to include the two dependencies in `Makefile`
3.  Script `bin/kevm` omitted the `-ccopt` option before the include and lib path of home brew and openssl in MacOS. We added this option name.


Makefile
```diff
@@ -138,7 +146,7 @@ plugin-deps: $(plugin_includes) $(plugin_c_includes)
 # Building
 # --------
 
-KOMPILE := $(KEVM) kompile
+KOMPILE := ${KEVM_BIN}/$(KEVM) kompile
```
ehildenb, line 149: This won't work, we'll need to find a better solution.

Makefile
```diff
@@ -212,6 +220,10 @@ haskell_main_file      := driver.md
 haskell_main_filename  := $(basename $(notdir $(haskell_main_file)))
 haskell_kompiled       := $(haskell_dir)/$(haskell_main_filename)-kompiled/definition.kore
 
+ifeq ($(UNAME_S),Darwin)
+$(KEVM_LIB)/$(haskell_kompiled): $(libsecp256k1_out)
+endif
```
ehildenb, line 225:
```suggestion
ifeq ($(UNAME_S),Darwin)
$(KEVM_LIB)/$(haskell_kompiled): $(libsecp256k1_out)
$(KEVM_LIB)/$(llvm_kompiled): $(libsecp256k1_out)
endif
```

Makefile
```diff
@@ -229,6 +241,10 @@ llvm_main_file     := driver.md
 llvm_main_filename := $(basename $(notdir $(llvm_main_file)))
 llvm_kompiled      := $(llvm_dir)/$(llvm_main_filename)-kompiled/interpreter
 
+ifeq ($(UNAME_S),Darwin)
+$(KEVM_LIB)/$(llvm_kompiled): $(libcryptopp_out)
+endif
```
ehildenb, line 246:
```suggestion
ifeq ($(UNAME_S),Darwin)
$(KEVM_LIB)/$(haskell_kompiled): $(libcryptopp_out)
$(KEVM_LIB)/$(llvm_kompiled): $(libcryptopp_out)
endif
```

ehildenb:
@spencerhaoxiao do we need both libraries for both haskell/llvm as I've indicated above? ^^^

Makefile
spencerhaoxiao: I can build without using both, but it does no harm to build both.

Makefile
spencerhaoxiao: I can build without depending on both, but it does no harm to build both at one go.

ehildenb (approved):

PR #1097, opened 2021-07-26
Pull out int-simplification

### LINKED ISSUE #1086: Pull out `tests/specs/int-simplification.k`

Riib11:
Fixes #1086

Pulls out Int-related lemmas from several modules used among the test suite, in particular `tests/specs/infinite-gas.k`, `tests/specs/lemmas.k`, and `tests/specs/**/verification.k`.

The Int-relatesd symbols conglomerated in Int-simplification are:

- -Int, +Int, *Int, |Int, modInt, &Int
- <Int, <=Int, >=Int, >Int
- maxInt, minInt

Note that Int-related lemmas that use symbols that are KEVM specific (e.g. #lookup and chop) are not include in tests/specs/int-simplification.k. The modules imported by INT-SIMPLIFICATION only import INT and BOOL (and K-REFLECTION for INT-SIMPLIFICATION-JAVA for the sake of #isConcrete).

Exception: Among `tests/specs/lemmas.k` and `tests/specs/infinite-gas.k` there exist two mutually-exclusive frameworks for associative normalization for expressions using just `+Int` and `-Int`. The one in `tests/specs/infinite-gas.k` is to be preferred (it is more exhaustively developed), but it turns out that some proofs that imports `LEMMAS` but not `INFINITE-GAS` break on (only) the **Java** backend when the associativity normalization rules from `INFINITE-GAS` replace the associative normalization rules in `LEMMAS` (by pulling out the rules from INFINITE-GAS` into `INT-SIMPLIFICATION`, disabling the associativity rules in `LEMMAS`, and then importing `INT-SIMPLIFICATION` in both `LEMMAS` and `INFINITE-GAS`.) I spent some time attempting to resolve this issue, but it proved to be too difficult for me to debug java backend problems. So, this PR. ignores this issue and leaves the two sets of associativity normalization lemmas in their home modules rather than pulling them out into `INT-SIMPLIFICATION`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -140,20 +140,20 @@ plugin-deps: $(plugin_includes) $(plugin_c_includes)
 
 KOMPILE := $(KEVM) kompile
 
-kevm_includes := abi.md              \
-                 asm.md              \
-                 buf.md              \
-                 data.md             \
-                 driver.md           \
-                 edsl.md             \
-                 evm.md              \
-                 evm-types.md        \
-                 hashed-locations.md \
-                 json-rpc.md         \
-                 network.md          \
-                 optimizations.md    \
-                 serialization.md    \
-                 state-loader.md
+kevm_files := abi.md              \
+              asm.md              \
+              buf.md              \
+              data.md             \
+              driver.md           \
+              edsl.md             \
+              evm.md              \
+              evm-types.md        \
+              hashed-locations.md \
+              json-rpc.md         \
+              network.md          \
+              optimizations.md    \
+              serialization.md    \
+              state-loader.md
 
 kevm_lemmas := infinite-gas.k                           \
                lemmas.k                                 \
@@ -166,7 +166,9 @@ kevm_lemmas := infinite-gas.k                           \
 
 lemma_includes := $(patsubst %, $(KEVM_INCLUDE)/kframework/lemmas/%, $(kevm_lemmas))
 
-includes := $(patsubst %, $(KEVM_INCLUDE)/kframework/%, $(kevm_includes)) $(plugin_includes) $(lemma_includes)
+kevm_includes := $(patsubst %, $(KEVM_INCLUDE)/kframework/%, $(kevm_files))
+
+includes := $(kevm_includes) $(lemma_includes) $(plugin_includes) $(plugin_c_includes)
 
 $(includes): $(KEVM_BIN)/$(KEVM)
 
@@ -193,7 +195,7 @@ java_main_file     := driver.md
 java_main_filename := $(basename $(notdir $(java_main_file)))
 java_kompiled      := $(java_dir)/$(java_main_filename)-kompiled/compiled.bin
 
-$(KEVM_LIB)/$(java_kompiled): $(includes)
+$(KEVM_LIB)/$(java_kompiled): $(kevm_includes)
 	$(KOMPILE) --backend java                  \
 	    $(java_main_file) $(JAVA_KOMPILE_OPTS) \
 	    --directory $(KEVM_LIB)/$(java_dir)    \
@@ -210,7 +212,7 @@ haskell_main_file      := driver.md
 haskell_main_filename  := $(basename $(notdir $(haskell_main_file)))
 haskell_kompiled       := $(haskell_dir)/$(haskell_main_filename)-kompiled/definition.kore
 
-$(KEVM_LIB)/$(haskell_kompiled): $(includes)
+$(KEVM_LIB)/$(haskell_kompiled): $(kevm_includes)
 	$(KOMPILE) --backend haskell                     \
 	    $(haskell_main_file) $(HASKELL_KOMPILE_OPTS) \
 	    --directory $(KEVM_LIB)/$(haskell_dir)       \
@@ -227,7 +229,7 @@ llvm_main_file     := driver.md
 llvm_main_filename := $(basename $(notdir $(llvm_main_file)))
 llvm_kompiled      := $(llvm_dir)/$(llvm_main_filename)-kompiled/interpreter
 
-$(KEVM_LIB)/$(llvm_kompiled): $(includes) $(libff_out) $(plugin_c_includes)
+$(KEVM_LIB)/$(llvm_kompiled): $(kevm_includes) $(plugin_includes) $(plugin_c_includes) $(libff_out)
 	$(KOMPILE) --backend llvm                 \
 	    $(llvm_main_file)                     \
 	    --directory $(KEVM_LIB)/$(llvm_dir)   \
@@ -271,11 +273,11 @@ $(KEVM_LIB)/release.md: INSTALL.md
 	echo                                    >> $@
 	cat INSTALL.md                          >> $@
 
-build: $(patsubst %, $(KEVM_BIN)/%, $(install_bins)) $(patsubst %, $(KEVM_LIB)/%, $(install_libs)) $(lemma_includes)
+build: $(patsubst %, $(KEVM_BIN)/%, $(install_bins)) $(patsubst %, $(KEVM_LIB)/%, $(install_libs))
 
-build-haskell: $(KEVM_LIB)/$(haskell_kompiled) $(KEVM_BIN)/$(KEVM) $(KEVM_LIB)/kore-json.py
-build-llvm:    $(KEVM_LIB)/$(llvm_kompiled)    $(KEVM_BIN)/$(KEVM) $(KEVM_LIB)/kore-json.py
-build-java:    $(KEVM_LIB)/$(java_kompiled)    $(KEVM_BIN)/$(KEVM) $(KEVM_LIB)/kast-json.py
+build-haskell: $(KEVM_LIB)/$(haskell_kompiled) $(KEVM_LIB)/kore-json.py $(lemma_includes)
+build-llvm:    $(KEVM_LIB)/$(llvm_kompiled)    $(KEVM_LIB)/kore-json.py
+build-java:    $(KEVM_LIB)/$(java_kompiled)    $(KEVM_LIB)/kast-json.py $(lemma_includes)
 
 all_bin_sources := $(shell find $(KEVM_BIN) -type f | sed 's|^$(KEVM_BIN)/||')
 all_lib_sources := $(shell find $(KEVM_LIB) -type f                                            \
--- a/Makefile
+++ b/Makefile
@@ -157,6 +157,7 @@ kevm_files := abi.md              \
 
 kevm_lemmas := infinite-gas.k                           \
                lemmas.k                                 \
+							 int-simplification.k                     \
                erc20/abstract-semantics-segmented-gas.k \
                erc20/evm-symbolic.k                     \
                mcd/bin_runtime.k                        \
--- a/Makefile
+++ b/Makefile
@@ -157,7 +157,7 @@ kevm_files := abi.md              \
 
 kevm_lemmas := infinite-gas.k                           \
                lemmas.k                                 \
-							 int-simplification.k                     \
+               int-simplification.k                     \
                erc20/abstract-semantics-segmented-gas.k \
                erc20/evm-symbolic.k                     \
                mcd/bin_runtime.k                        \
--- a/Makefile
+++ b/Makefile
@@ -195,7 +195,7 @@ java_main_file     := driver.md
 java_main_filename := $(basename $(notdir $(java_main_file)))
 java_kompiled      := $(java_dir)/$(java_main_filename)-kompiled/compiled.bin
 
-$(KEVM_LIB)/$(java_kompiled): $(kevm_includes)
+$(KEVM_LIB)/$(java_kompiled): $(kevm_includes) $(plugin_includes)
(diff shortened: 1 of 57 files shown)


tests/specs/int-simplification.k
```diff
@@ -0,0 +1,113 @@
...
+    rule 0 *Int _ => 0
+    rule _ *Int 0 => 0
+
+    rule 0 |Int N => N
+    rule N |Int 0 => N
+    rule N |Int N => N
+
+    rule  0 &Int _N => 0
+    rule _N &Int  0 => 0
+    rule  N &Int  N => N
```
ehildenb, line 25: These seem like they should be in `INT-SIMPLIFICATION-COMMON`.

tests/specs/int-simplification.k
```diff
@@ -0,0 +1,113 @@
...
+    imports INT-SIMPLIFICATION-COMMON
+
+  // ###########################################################################
+  // lemmas.k
+  // ###########################################################################
+
+  rule N <=Int maxInt(P, Q) => true requires N <=Int P orBool N <=Int Q  [simplification]
+
+  rule minInt(P, Q)         => P    requires P <=Int Q                   [simplification]
+
```
ehildenb, line 50: These should probably be in `INT-SIMPLIFICATION-COMMON`

Riib11:
I noticed there are a few different variable naming schemes for variables that are concrete/symbolic/either. Should I standardize these, or does it not matter enough?

Riib11:
The CI fail appears to be on the server side.

Riib11:
I am basically done with this PR. The only thing left is to figure out how to reconcile the two sets of associativity normalization rules in `lemmas.k` and `infinite-gas.k`, which I gave up on.

Riib11:
Passes all tests for both java and haskell backends on my container.

tests/specs/infinite-gas.k
```diff
@@ -1,4 +1,5 @@
 requires "evm.md"
+requires "./int-simplification.k"
```
ehildenb, line 2:
```suggestion
requires "lemmas/int-simplification.k"
```

Riib11: fixed by da01c73

tests/specs/int-simplification.k
```diff
@@ -0,0 +1,190 @@
...
+  // ###########################################################################
+  // inequality
+  // ###########################################################################
+
+    rule I1 +Int C2  <Int C3         => I1          <Int C3 -Int C2 [concrete(C2, C3), simplification]
+    rule C1          <Int I2 +Int C3 => C1 -Int C3  <Int I2         [concrete(C1, C3), simplification]
+    rule C1         <=Int I2 +Int C3 => C1 -Int C3 <=Int I2         [concrete(C1, C3), simplification]
+
+    rule X +Int Y <Int Z => X <Int Z -Int Y [concrete(Y), simplification]
```
ehildenb, line 68: Hmmmm, is the lemma on line 68 needed? It seems like a strict generalization of the one on line 64. Maybe the one on line 64 is not needed?

ehildenb: But I like the one on line 64 better, so maybe we can just have that one if the one on line 68 isn't needed?

tests/specs/int-simplification.k
```diff
@@ -0,0 +1,190 @@
...
+    rule  (A -Int  B) +Int B  => A [simplification]
+    rule   A -Int (A  -Int B) => B [simplification]
+    rule   A +Int (B  -Int A) => B [simplification]
+    rule  (A +Int  B) -Int A  => B [simplification]
+
+  // 4 terms
+    rule  (A +Int B) +Int (C  -Int A) => B +Int C [simplification]
+    rule  (A +Int B) -Int (A  +Int C) => B -Int C [simplification]
+    rule  (A +Int B) -Int (C  +Int A) => B -Int C [simplification]
+    rule  (A +Int B) +Int (C  -Int B) => A +Int C [simplification]
```
ehildenb, line 96: The associativity rules above should ensure that we always have terms nested like `((A + B) + C) + D`, don't they? So then do we need rules about terms like `(A + B) + (C + D)`?

ehildenb: Can we double-check that these ones are needed?

tests/specs/int-simplification.k
```diff
@@ -0,0 +1,190 @@
...
+    rule ((A -Int B) -Int  C) +Int B  => A -Int C [simplification]
+
+  // 5 terms
+    rule   (A +Int  B  +Int C)  -Int (A  +Int D) =>  B +Int (C  -Int D) [simplification]
+    rule   (C +Int (A  -Int D)) +Int (B  -Int A) =>  C +Int (B  -Int D) [simplification]
```
ehildenb, line 101: Same thing, shouldn't we always have terms that are nested to the left already? So do we need simplification lemmas over terms nested like this?

tests/specs/int-simplification.k
```diff
@@ -0,0 +1,190 @@
...
+    rule (((A -Int  B) -Int C)  -Int  D) +Int B  => (A -Int  C) -Int D  [simplification]
+
+  // >=6 terms
+    rule (A +Int B +Int C +Int D +Int E) -Int (A +Int C +Int F) => B +Int D +Int E -Int F [simplification]
```
ehildenb, line 105: Same thing, terms sholud nest left.

tests/specs/int-simplification.k
```diff
@@ -0,0 +1,190 @@
...
+
+    rule 1 *Int A => A
+    rule A *Int 1 => A
+    rule 0 *Int _ => 0
+    rule _ *Int 0 => 0
+
+    rule C *Int A +Int B *Int A => (C +Int B) *Int A [simplification]
+
+    rule E *Int A +Int B +Int C +Int D +Int F *Int A => (E +Int F) *Int A +Int B +Int C +Int D [simplification]
+
```
ehildenb, line 119: I know that `*Int` is tighter binding, but can we put parens here to make it easier to read?

tests/specs/int-simplification.k
```diff
@@ -0,0 +1,190 @@
...
+    rule notBool (A <=Int B) => B  <Int A [simplification]
+
+    rule 0 <=Int A *Int B => true requires 0 <=Int A andBool 0 <=Int B [simplification]
+
+    rule A -Int B +Int C <=Int D => false requires D <Int A -Int B andBool 0 <=Int C [simplification]
+
+    rule minInt(A, B) <Int C  => A <Int C orBool B <Int C  [simplification]
+
+    rule A  <Int minInt(B, C) => A  <Int B andBool A  <Int C [simplification]
+    rule A <=Int minInt(B, C) => A <=Int B andBool A <=Int C [simplification]
```
ehildenb, line 186: I would prefer if these were phrased as `rule LHS => true requires RHS`, for example `rule A <Int minInt(B, C) => true requires A <Int B andBool A <Int C`. That means we don't "lose information" about the fact that this came from a `minInt`.

tests/specs/opcodes/verification.k
```diff
@@ -1,7 +1,9 @@
 requires "edsl.md"
+requires "../int-simplification.k"
```
ehildenb, line 2:
```suggestion
requires "lemmas/int-simplification.k"
```

Riib11: the suggestions you made along these lines doesn't appear to actually work, I get a build error that says it can't find the file `lemmas/int-simplification.k`. I decided to use the relative path because that's what was used for `lemmas.k` and `infinite-gas.k`

Riib11:
I can remove the temporary comments I've made on placement and disabling decisions once this PR is approved.

tests/specs/int-simplification.k
```diff
@@ -0,0 +1,198 @@
...
+    rule A |Int 0 => A
+    rule A |Int A => A
+
+  // ###########################################################################    
+  // bit-and
+  // ###########################################################################    
+
+    rule 0 &Int _ => 0
+    rule _ &Int 0 => 0
+    rule A &Int A => A
```
ehildenb, line 152:
```suggestion
    rule 0 |Int A => A [simplification]
    rule A |Int 0 => A [simplification]
    rule A |Int A => A [simplification]

  // ###########################################################################    
  // bit-and
  // ###########################################################################    

    rule 0 &Int _ => 0 [simplification]
    rule _ &Int 0 => 0 [simplification]
    rule A &Int A => A [simplification]
```

tests/specs/int-simplification.k
```diff
@@ -0,0 +1,198 @@
...
+
+    rule 0 <=Int A *Int B => true requires 0 <=Int A andBool 0 <=Int B [simplification]
+
+    rule A -Int B +Int C <=Int D => false requires D <Int A -Int B andBool 0 <=Int C [simplification]
+
+    // TODO: can be safely removed in favor or rules below them (java, haskell)
+    // rule minInt(A, B) <Int C  => A <Int C orBool B <Int C  [simplification]
+    // rule A  <Int minInt(B, C) => A  <Int B andBool A  <Int C [simplification]
+    // rule A <=Int minInt(B, C) => A <=Int B andBool A <=Int C [simplification]
+
```
ehildenb, line 190:
```suggestion
```

tests/specs/int-simplification.k
```diff
@@ -0,0 +1,198 @@
...
+  // 5 terms
+    // NOTE: required for `tests/specs/functional/infinite-gas-spec.k.prove` (haskell)
+    rule   (A +Int  B  +Int C)  -Int (A  +Int D) =>  B +Int (C  -Int D) [simplification]
+    rule   (C +Int (A  -Int D)) +Int (B  -Int A) =>  C +Int (B  -Int D) [simplification]
+    rule (((A -Int  B) -Int C)  -Int  D) +Int B  => (A -Int  C) -Int D  [simplification]
+
+  // >=6 terms
+    // TODO: can be safely removed (java, haskell)
+    // rule (A +Int B +Int C +Int D +Int E) -Int (A +Int C +Int F) => B +Int D +Int E -Int F [simplification]
```
ehildenb, line 108:
```suggestion
```

tests/specs/int-simplification.k
```diff
@@ -0,0 +1,198 @@
...
+    rule (S1 +Int C2) -Int C3 => S1 +Int (C2 -Int C3) [concrete(C2, C3), symbolic(S1), simplification]
+    rule (S1 -Int C2) +Int C3 => S1 +Int (C3 -Int C2) [concrete(C2, C3), symbolic(S1), simplification]
+    rule (S1 -Int C2) -Int C3 => S1 -Int (C2 +Int C3) [concrete(C2, C3), symbolic(S1), simplification]
+
+  // ###########################################################################
+  // inequality
+  // ###########################################################################
+
+    // TODO: can be safely removed in favor of rule below it (java, haskell)
+    // rule I1 +Int C2  <Int C3         => I1          <Int C3 -Int C2 [concrete(C2, C3), simplification]
```
ehildenb, line 65:
```suggestion
```

Makefile
```diff
@@ -157,6 +157,7 @@ kevm_files := abi.md              \
 
 kevm_lemmas := infinite-gas.k                           \
                lemmas.k                                 \
+							 int-simplification.k                     \
```
ehildenb, line 160:
```suggestion
               lemmas.k                                 \
               int-simplification.k                     \
```

tests/specs/int-simplification.k
```diff
@@ -0,0 +1,198 @@
...
+    // rule (A +Int B +Int C +Int D +Int E) -Int (A +Int C +Int F) => B +Int D +Int E -Int F [simplification]
+
+  // ###########################################################################    
+  // mul
+  // ###########################################################################    
+
+    rule 1 *Int A => A
+    rule A *Int 1 => A
+    rule 0 *Int _ => 0
+    rule _ *Int 0 => 0
```
ehildenb, line 117:
```suggestion
    rule 1 *Int A => A [simplification]
    rule A *Int 1 => A [simplification]
    rule 0 *Int _ => 0 [simplification]
    rule _ *Int 0 => 0 [simplification]
```

ehildenb (approved):

PR #1099, opened 2021-07-28
Remove precompile accounts from state of MCD proofs

ehildenb:
Currently, for the MCD proofs, we put in "mock" accounts as precompile accounts, so that the `#accountNonexistent` and `#accountEmpty` are affected. This just removes mentions of the precompiles, so that we don't worry about them for the proofs at all (proofs are done independent of precompile account status).

### DIFF
--- a/evm.md
+++ b/evm.md
@@ -1170,9 +1170,6 @@ For now, I assume that they instantiate an empty account and use the empty data.
 
     syntax UnStackOp ::= "EXTCODEHASH"
  // ----------------------------------
-```
-
-```k
     rule <k> EXTCODEHASH ACCT => keccak(CODE) ~> #push ... </k>
          <account>
            <acctID> ACCT </acctID>
--- a/evm.md
+++ b/evm.md
@@ -2154,15 +2154,9 @@ There are several helpers for calculating gas (most of them also specified in th
     syntax KResult ::= Bool
     syntax BExp ::= #accountNonexistent ( Int )
  // -------------------------------------------
-    rule <k> #accountNonexistent(ACCT) => Gemptyisnonexistent << SCHED >> ... </k>
-         <schedule> SCHED </schedule>
-      requires #isPrecompiledAccount(ACCT, SCHED)
-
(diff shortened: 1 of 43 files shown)


JKTKops (approved):
As long as you're convinced this is sound, looks good to me.

ehildenb:
I'm double-checking some things now about it.

ehildenb:
From what I'm reading in the Yellow Paper: https://ethereum.github.io/yellowpaper/paper.pdf if an account has no code, balance, or nonce, then it is considered empty. 

The Yellow Paper says that precompiles _can_ be empty, which is confusing: https://ethereum.github.io/yellowpaper/paper.pdf

> Even callable precompiled contracts can have an empty
> account state. This is because their account states do not
> usually contain the code describing its behavior.
> An account is dead when its account state is non-existent
> or empty

But EIP-1052 (EXTCODEHASH) seems to also imply that precompiles _can_ be non-empty, and return the code-hash of empty-code (`c5d246...`) or of empty-account (`0`): https://eips.ethereum.org/EIPS/eip-1052

> The EXTCODEHASH of an precompiled contract is either c5d246... or 0.

It also appears that the precompiles are actually _not_ empty: https://github.com/ethereum/EIPs/issues/716#issuecomment-330824096

>  Incidentally, both RipeMD160 (0x3) and Identity (0x4) were ‘empty’ accounts created by the attacker, whereas ‘0x1’ and ‘0x2’ already contained ether and were not ‘empty’.

ehildenb:
But I think it's actually still OK to merge this (even though we can't skip the check), because now the proofs are making _no_ assumptions about the pre-compile accounts. The only way for this to be visible to the prover context is if you call `EXTCODEHASH SOME_PRECOMPILE` somewhere in your program, and then branch based on that. In the case of the proofs originally, this would simplify down to `PRECOMPILE_BAL ==Int 0 andBool PRECOMPILE_NONCE ==Int 0`. But if we don't put the precompiles in the state (like this change does), then the prover won't be allowed to assume anything about them.

PR #1100, opened 2021-07-28
Bring back Kevm Node

gtrepta:
fixes: https://github.com/runtimeverification/iog-pm/issues/96


evm.md
```diff
@@ -783,6 +793,30 @@ These are just used by the other operators for shuffling local execution state a
...
+-   `#loadAccount` queries for account data from the running client.
+-   `#lookupCode` loads the code of an account into the `<code>` cell.
+-   `#lookupStorage` loads the value of the specified storage key into the `<storage>` cell.
+
+```k
+    syntax InternalOp ::= "#loadAccount"   Int
+                        | "#lookupCode"    Int
+                        | "#lookupStorage" Int Int
+ // ----------------------------------------------
+```
```
ehildenb, line 808: I guess we can move these definitions to `node.md` now.

evm.md
```diff
@@ -783,6 +793,30 @@ These are just used by the other operators for shuffling local execution state a
...
+
+In `standalone` mode, the semantics assumes that all relevant account data is already loaded into memory.
+
+In `node` mode, the semantics are given in terms of an external call to a running client.
+
+```{.k .standalone}
+    rule <k> #loadAccount   _   => . ... </k>
+    rule <k> #lookupCode    _   => . ... </k>
+    rule <k> #lookupStorage _ _ => . ... </k>
+```
```
ehildenb, line 818: And we don't need these anymore.

kevm
```diff
@@ -51,26 +51,26 @@ run_kompile() {
...
-                     openssl_root="$(brew --prefix openssl)"
-                     brew_root="$(brew --prefix)"
-                     kompile_opts+=( -ccopt -I${brew_root}/include -ccopt -L/${brew_root}/lib -ccopt -I${openssl_root}/include -ccopt -L${openssl_root}/lib )
-                     kompile_opts+=( -ccopt -I${libcryptopp_dir}/include -ccopt -L/${libcryptopp_dir}/lib )
-                 fi
-                 ;;
+        haskell)   kompile_opts+=( --md-selector 'k & ! nobytes & ! node' --backend haskell)        ;;
+        java)      kompile_opts+=( --md-selector 'k & ! bytes   & ! node' --backend java )          ;;
+        llvm)      kompile_opts+=( --md-selector 'k & ! nobytes & ! node' )                         ;;&
+        node)      kompile_opts+=( --md-selector 'k & ! nobytes & ! standalone' --no-llvm-kompile ) ;;&
```
ehildenb, line 57: What are the `&` at the end of these lines for?

gtrepta: From the man page for bash: `Using ;;& in place of ;; causes the shell to test the next pattern list in the statement, if any, and execute any associated list on a successful match.`

It's so `llvm` and `node` get matched again for the commands that they share in common with each other.

evm.md
```diff
@@ -1588,12 +1622,15 @@ For each `CALL*` operation, we make a corresponding call to `#call` and a state-
...
+have been paid, and it may be to expensive to compute the hash of the init code.
 
 ```k
     syntax QuadStackOp ::= "CREATE2"
  // --------------------------------
     rule <k> CREATE2 VALUE MEMSTART MEMWIDTH SALT
-          => #checkCall ACCT VALUE
+          => #loadAccount #newAddr(ACCT, SALT, #range(LM, MEMSTART, MEMWIDTH))
+          ~> #checkCall ACCT VALUE
           ~> #create ACCT #newAddr(ACCT, SALT, #range(LM, MEMSTART, MEMWIDTH)) VALUE #range(LM, MEMSTART, MEMWIDTH)
```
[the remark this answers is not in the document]
gtrepta: I'm not totally sure, this is what it looked like before the node was removed, documentation included. Maybe @dwightguth can elaborate?

evm.md
gtrepta: `#lookupCode` and `#lookupStorage` can definitely go. `#loadAccount` will depend on what becomes of the CREATE2 case.

evm-node.md
```diff
@@ -0,0 +1,304 @@
...
+Contained in this file is glue code needed in order to enable the ability to use KEVM as a VM for an actual Ethereum node.
+
+```{.k .node}
+requires "evm.md"
+
+module EVM-NODE
+    imports EVM
+    imports K-REFLECTION
+    imports COLLECTIONS
+    imports BYTES
```
ehildenb, line 13: We should also import `EVM-OPTIMIZATIONS` so we get those speedups for long-running executions. It is in file `optimizations.md`

Makefile
```diff
@@ -97,7 +103,11 @@ $(libff_out): $(PLUGIN_SUBMODULE)/deps/libff/CMakeLists.txt
 	cd $(PLUGIN_SUBMODULE)/deps/libff/build                                                                     \
 	    && cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$(INSTALL_LIB)/libff $(LIBFF_CMAKE_FLAGS) \
 	    && make -s -j4                                                                                          \
-	    && make install DESTDIR=$(CURDIR)/$(BUILD_DIR)
+	    && make install DESTDIR=$(BUILD_DIR)
```
ehildenb, line 106: This change is not correct I don't think. Pretty sure it needs to be `$(CURDIR)/$(BUILD_DIR)`, or it will be installed in the submodule build directory, not in this ones build dircetory.

Makefile
```diff
@@ -124,8 +134,8 @@ endif
 
 k-deps:
 	cd $(K_SUBMODULE)                                                                                                                                                                            \
-	    && mvn --batch-mode package -DskipTests -Dllvm.backend.prefix=$(INSTALL_LIB)/kframework -Dllvm.backend.destdir=$(CURDIR)/$(BUILD_DIR) -Dproject.build.type=$(K_BUILD_TYPE) $(K_MVN_ARGS) \
-	    && DESTDIR=$(CURDIR)/$(BUILD_DIR) PREFIX=$(INSTALL_LIB)/kframework package/package
+	    && mvn --batch-mode package -DskipTests -Dllvm.backend.prefix=$(INSTALL_LIB)/kframework -Dllvm.backend.destdir=$(BUILD_DIR) -Dproject.build.type=$(K_BUILD_TYPE) $(K_MVN_ARGS) \
+	    && DESTDIR=$(BUILD_DIR) PREFIX=$(INSTALL_LIB)/kframework package/package
```
ehildenb, line 138: Same here, I think we need the full path.

evm-node.md
```diff
@@ -0,0 +1,338 @@
...
+          ~> #load [ OP ]
+          ~> #exec [ OP ]
+          ~> #pc   [ OP ]
+         ...
+         </k>
+         <wordStack> WS </wordStack>
+         <static> STATIC:Bool </static>
+      requires notBool ( #stackUnderflow(WS, OP) orBool #stackOverflow(WS, OP) )
+       andBool notBool ( STATIC andBool #changesState(OP, WS) )
+    [priority(35)]
```
ehildenb, line 42: We don't want to override the entire `#next` rule, because it will obliterate the existing optimizations. The other drawback is that we then have to keep this in sync.

Instead, we should do something like:

```
configuration <kevm-node> <kevm/> <loaded> false </loaded> </kevm-node>

rule
<k> (. => #load [ OP ]) ~> #next [ OP ] ... </k>
<loaded> false => true </loaded>
requires isAddr1Op(OP) orBool isAddr2Op(OP) orBool OP ==K CREATE orBool OP ==K SSTORE orBool OP ==K SLOAD
 [priority(35)]

rule <k> #execute ... </k> <loaded> true => false </loaded> [priority(35)]
```

evm-node.md
```diff
@@ -0,0 +1,338 @@
...
+    rule <k> BLOCKHASH N => 0                ~> #push ... </k> <mode> NORMAL </mode> requires N  <Int 0  orBool N >=Int 256
+```
+
+```{.k .node}
+    rule <k> EXTCODEHASH ACCT => HASH ~> #push ... </k>
+         <account>
+           <acctID> ACCT </acctID>
+           <code> #unloaded(HASH) </code>
+           ...
+         </account>
```
ehildenb, line 198: This can be done simpler by just adding a rule `keccak(#unloaded(HASH)) => HASH`, so that we don't need to override th eexisting rule for `EXTCODEHASH`, which has problems with synchronization.

evm-node.md
```diff
@@ -0,0 +1,338 @@
...
+```
+
+-   `#getBlockhash(N)` returns the blockhash of the Nth most recent block, up to a maximum of 256 blocks.
+    It is used in the implementation of the BLOCKHASH instruction as seen below.
+
+```{.k .node}
+    syntax Int ::= #getBlockhash ( Int ) [function, hook(BLOCKCHAIN.getBlockhash)]
+ // ------------------------------------------------------------------------------
+    rule <k> BLOCKHASH N => #getBlockhash(N) ~> #push ... </k> <mode> NORMAL </mode> requires N >=Int 0 andBool N  <Int 256
+    rule <k> BLOCKHASH N => 0                ~> #push ... </k> <mode> NORMAL </mode> requires N  <Int 0  orBool N >=Int 256
```
ehildenb, line 189: This ones are harder to avoid overriding I guess, so we can leave them.

evm-node.md
```diff
@@ -0,0 +1,338 @@
...
+
+```{.k .node}
+    rule <k> CREATE2 VALUE MEMSTART MEMWIDTH SALT
+          => #loadAccount #newAddr(ACCT, SALT, #range(LM, MEMSTART, MEMWIDTH))
+          ~> #checkCall ACCT VALUE
+          ~> #create ACCT #newAddr(ACCT, SALT, #range(LM, MEMSTART, MEMWIDTH)) VALUE #range(LM, MEMSTART, MEMWIDTH)
+          ~> #codeDeposit #newAddr(ACCT, SALT, #range(LM, MEMSTART, MEMWIDTH))
+         ...
+         </k>
+         <id> ACCT </id>
```
ehildenb, line 215: Similar trick with `<loaded>` above, we don't want to override the existing rules for create, just have it inject a `#load` statement right before the CREATE2.

Makefile
gtrepta: Did you see the change I made to `BUILD_DIR`? I needed to make it an absolute path, I think it was for the CMakeLists.txt file.

ehildenb (approved):

PR #1106, opened 2021-08-05
Makefile: include lemmas in package

ehildenb:


MirceaS (approved):
Was this a bug before?

PR #1109, opened 2021-08-08
Switch all proofs away from old segmented gas model

ehildenb:
This switches all proofs away from the execution/memory segmented gas model into the new infinite-gas model.

The affected directories or `tests/specs/benchmarks` and `tests/specs/erc20`.

I also had to disable symbolic execution of the rule `Cgascap` (by adding it to the concrete-rules file). Then lemmas are added to make sure that it behaves correctly with `#gas(_)`. Note that I had to change the definition of `Cgascap` so that on negative GCAP inputs, it goes to zero. This is to ensure that the output of `Cgascap` is always positive (which one of the lemmas added relies upon), but doesn't actually affect any real-world executions since it should be impossible to call `Cgascap` with a negative `GCAP`. Note that this change does not break the conformance tests, which confirms that it's OK.

Another change is that we now just have a generic case `#gas(_) <Int _ => false`, which should always be OK because `#gas(_)` should behave like positive infinite (aleph-0), so even if th eRHS of the `<Int` is another `#gas(_)`, this should still hold.

Pay special attention to the lemmas added/modified in `tests/specs/infinite-gas.k`. They should all have the property that if the LHS is infinite, then the RHS should be infinite.

Another small update is to remove the `HASH2` module, which is not used any longer.


JKTKops (approved):
Changes look reasonable, but as @Riib11 noted there are a couple proofs failing on the Java backend.

Riib11:
The failing tests are: 
- tests/specs/mcd/vow-cage-surplus-pass-rough-spec.k
- tests/specs/mcd/vow-cage-deficit-pass-rough-spec.k
As @ehildenb mentioned, the cause is a weird thing with the `Cgascap` changes. The relevant rule that was added was:
```
rule Cgascap(SCHED, #gas(VGAS), #gas(VGAS), GCAP) => #gas(Cgascap(SCHED, VGAS, VGAS, GCAP)) [simplification]
```

Riib11 (approved):

Riib11:
On tests/specs/infinite-has.k line 94:

Why not have `rule _ <=Int #gas(_) => true [simplification]` instead? If `#gas(_)` behaves like aleph_0. This would remove a few other rules in places where you need to give `rule <symbolic quantity> <=Int #gas(_) [simplification]` for different introduced uninterpreted functions (e.g. `Cmem`)

(I tried leaving a comment on that line, but for some reason it wouldn't let me??)

ehildenb:
Kore issue: https://github.com/kframework/kore/issues/2798

tests/specs/functional/infinite-gas-spec.k
```diff
@@ -98,8 +113,8 @@ module INFINITE-GAS-SPEC
 
     claim <k> runLemma(Gverylow < SCHED > <=Int #gas(_)) => doneLemma(true) ... </k>
 
-    claim <k> runLemma(Gexp < SCHED > +Int Gexpbyte < SCHED > *Int ( log2Int ( W1 ) /Int 8 +Int 1 ) <=Int #gas(GAVAIL)) => doneLemma(true)  ... </k>
-    claim <k> runLemma(#gas(GAVAIL) <Int Gexp < SCHED > +Int Gexpbyte < SCHED > *Int ( log2Int ( W1 ) /Int 8 +Int 1 ))  => doneLemma(false) ... </k>
+    claim <k> runLemma(Gexp < SCHED > +Int Gexpbyte < SCHED > *Int ( log2Int ( W1 ) /Int 8 +Int 1 ) <=Int #gas(GAVAIL)) => doneLemma(true)  ... </k> requires 0 <Int W1 andBool W1 <Int pow256
+    claim <k> runLemma(#gas(GAVAIL) <Int Gexp < SCHED > +Int Gexpbyte < SCHED > *Int ( log2Int ( W1 ) /Int 8 +Int 1 ))  => doneLemma(false) ... </k> requires 0 <Int W1 andBool W1 <Int pow256
```
JKTKops, line 117: Why not #rangeUInt on these requires?

JKTKops (approved):

tests/specs/functional/infinite-gas-spec.k
ehildenb: Because the bottom of the range is `0 <Int W1` instead of `0 <=Int W1`. But this reminds me that these tests can be simplified slightly now.

ehildenb: But it's not necessary to simplify them I guess.

Riib11 (approved):

tests/specs/functional/infinite-gas-spec.k
JKTKops: Sounds good. I approved.

PR #1113, opened 2021-08-12
Remove trusted claims in semantics modules

ehildenb:
The new kprovex does not allow claims in the verification definition. This removes the remaining ones, which are related to ECREC precompile and overriding arithmetic comparison operators.

A new abstraction `#ecrec` is introduced in the semantics, and axiomatized in terms of a predicate `#ecrecEmpty` (which states whether it should go to 0 or not).


daejunpark (commented):
Looks good overall, but I'd like to have @denis-bogdanas to review this, as he developed this abstraction.

denis-bogdanas (approved):
looks good

PR #1114, opened 2021-08-16
Remove Unneeded Bihu Lemmas

JKTKops:
Some lemmas for the bihu proofs that seem to help the java backend cause unecessary branching on the haskell backend that appears to lead to an exponential slowdown (exponential in the number of times these rules apply).

Making these lemmas java-only gets 4 of the 7 bihu proofs to pass on the Haskell backend.


tests/specs/bihu/abstract-semantics-direct-gas.k
```diff
@@ -1,4 +1,9 @@
 module ABSTRACT-SEMANTICS-DIRECT-GAS
+  imports ABSTRACT-SEMANTICS-DIRECT-GAS-JAVA
+
+endmodule
+
+module ABSTRACT-SEMANTICS-DIRECT-GAS-JAVA [kast]
   imports EVM
```
ehildenb, line 8: These lemmas can just be deleted. The Java backend does not need them either.

ehildenb (approved):

PR #1116, opened 2021-08-18
Switch to using kprovex

ehildenb:


Makefile
```diff
@@ -366,6 +377,19 @@ tests/%.parse: tests/%
...
+	$(KEVM) prove $< $(KPROVE_MODULE) $(TEST_OPTIONS) --backend $(TEST_SYMBOLIC_BACKEND) --format-failures $(KPROVE_OPTS) \
+	    --provex --backend-dir tests/specs/$(firstword $(subst /, ,$*))/$(KPROVE_FILE)/$(TEST_SYMBOLIC_BACKEND)
+
+tests/specs/%-kompiled/timestamp: tests/specs/$$(firstword $$(subst /, ,$$*))/$$(KPROVE_FILE).$$(KPROVE_EXT) tests/specs/$$(firstword $$(subst /, ,$$*))/concrete-rules.txt $(kevm_includes) $(plugin_includes)
+	$(KOMPILE) --backend $(TEST_SYMBOLIC_BACKEND) $<                                                 \
+	    --directory tests/specs/$(firstword $(subst /, ,$*))/$(KPROVE_FILE)/$(TEST_SYMBOLIC_BACKEND) \
+	    --main-module $(KPROVE_MODULE)                                                               \
+	    --syntax-module $(KPROVE_MODULE)                                                             \
+	    --concrete-rules-file tests/specs/$(firstword $(subst /, ,$*))/concrete-rules.txt            \
+	    $(KOMPILE_OPTS)
```
gtrepta, line 391: The secondary expansion is here for the prerequisites because some of the variables like KPROVE_FILE and $* would get evaluated to the same value for all of the targets, is that correct?

gtrepta (approved):
The changes make enough sense to me, and the build passes so I'll approve.

Makefile
ehildenb: Yeah, that's right. It's to make sure that we have access to `$*`, `$(KPROVE_FILE)`, and `$(KPROVE_EXT)`

PR #1117, opened 2021-08-18
tests/failing-symbolic.haskell: enable more tests

ehildenb:


JKTKops (approved):
Looks good. That's nearly half the formerly failing tests!

Riib11 (approved):

PR #1121, opened 2021-08-25
Demo proofs using Solidity source

ehildenb:


calin1304 (approved):
Looks good to me. 

One question I have is when running `kevm prove` on a file with multiple claims, we get a single `#Top` to `stdout`, so do we consider all those claims a single proof or each claim should be proven separately? And if so, shouldn't we get multiple `#Top` printed, one for each claim?

ehildenb:
We get a single `#Top`, which means each proof was proven successfully. The proofs are run in order they appear in the file, and if one fails then the remaining ones aren't run (and the counterexample state from the last run proof is shown).

PR #1127, opened 2021-09-03
Fix release process, uniform use of --debugger

ehildenb:
Release builds were not including `-I /usr/lib/kevm/include/kframework` in their builds, which leads to the release failing for non-provex definitions.

- Option `--debugger` is now used to enable the Java backend debugging capabilities too, instead of having a separate function.
- `Makefile` adds `-I $(INSTALL_INCLUDE)/kframework` and similarly for the blockchain-k-plugin to every build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -189,7 +189,7 @@ $(KEVM_INCLUDE)/kframework/lemmas/%.k: tests/specs/%.k
 	@mkdir -p $(dir $@)
 	install $< $@
 
-KOMPILE_OPTS = --debug
+KOMPILE_OPTS = --debug -I $(INSTALL_INCLUDE)/kframework -I $(INSTALL_LIB)/blockchain-k-plugin/include/kframework
 
 ifneq (,$(RELEASE))
     KOMPILE_OPTS += -O2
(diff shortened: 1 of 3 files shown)


ehildenb:
Evidence this works: https://office.runtimeverification.com/jenkins/blue/organizations/jenkins/KEVM/detail/PR-1127/1/pipeline

calin1304:
LGTM. I'll let @0xverif take a look at it too.

0xverif:
Looks good to me as well

Makefile
```diff
@@ -48,7 +48,7 @@ export PLUGIN_SUBMODULE
         test-vm test-rest-vm test-all-vm test-bchain test-rest-bchain test-all-bchain                                            \
         test-prove test-failing-prove                                                                                            \
         test-prove-benchmarks test-prove-functional test-prove-opcodes test-prove-erc20 test-prove-bihu test-prove-examples      \
-        test-prove-mcd test-klab-prove test-haskell-dry-run                                                                      \
+        test-prove-mcd test-klab-prove                                                                                           \
```
radumereuta, line 51: I wanted this option. Any reason you want to remove it?
It makes it very convenient to test the kprove front end and make improvements.

gtrepta (commented):
The `--debugger` stuff looks good to me.

I'm having trouble understanding what the issue with the release is. Did the recent K update PR break it somehow?

Makefile
ehildenb: Well you can recover this by just calling `make test-prove KPROVE_OPTS=--dry-run`

ehildenb:
@gtrepta it was because of the missing `-I ...` in the `Makefile`. So the semantics is built in the KEVM directory, where that `-I` doesn't exist, then everything is copied over to the global installation directory `/usr/lib/kevm/...`. Then the global install wasn't correctly searching itself for the needed files, only looking ni the local directory `./.build/usr/lib/kevm/...`. I just add these `-I` options globally, so that all builds of KEVM from the repo will consider files from the global directory as well.

gtrepta (approved):
Looks good.

PR #1129, opened 2021-09-10
Integer equality simplification for Haskell backend

ehildenb:
The Java bcakend normalizes terms to use `==K`, and the Haskell backend uses the more specific `==Int`. So semantic rules simplifying out equalities need to be specific per backend.


RaoulSchaffranek (approved):

calin1304:
Shouldn't [range rules](https://github.com/kframework/evm-semantics/blob/acfb82b4e393cb0d66b25a40a0094a6f49ffcbe2/tests/specs/mcd/verification.k#L122) for Java also use `==K` ?

calin1304 (approved):

PR #1130, opened 2021-09-13
Build provex definitions ahead of time, fix optimizer

ehildenb:
This PR adds a "best-effort" build-provex target which is called ahead of time to remove the build stage from the prove stage in Jenkins. It will not fail if a definition is missing, that definition will just be built at prove time instead. So if we add more proofs, we can notice the problem and fix it "lazily".

This PR also makes some minor adjustments to the `optimizer/` to make it work with the new way of running proofs in this repo.

### DIFF
--- a/optimizer/optimize-spec.k.tmpl
+++ b/optimizer/optimize-spec.k.tmpl
@@ -1,9 +1,10 @@
-requires "lemmas/mcd/verification.k"
+requires "lemmas/infinite-gas.k"
 
 module OPTIMIZE
     imports INT
-    imports LEMMAS-MCD
+    imports EVM-OPTIMIZATIONS-LEMMAS
     imports EVM
+    imports INFINITE-GAS
 endmodule
 
 module {MODULE}-SPEC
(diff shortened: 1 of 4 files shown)


optimizer/optimize-spec.k.tmpl
```diff
@@ -1,9 +1,10 @@
-requires "lemmas/mcd/verification.k"
+requires "lemmas/infinite-gas.k"
 
 module OPTIMIZE
     imports INT
-    imports LEMMAS-MCD
```
calin1304, line 5: Why was `LEMMAS-MCD` imported and why is it no longer needed ?

ehildenb: It was imported so we could get that body of lemmas for doing these optimizations. It's a large body of lemmas that was developed for the maker proofs.

But I realized those lemmas weren't needed for these proofs, and the smaller the trust base we can have, the better.

calin1304 (approved):
This PR looks good to me

PR #1133, opened 2021-09-15
Enable some passing bihu proofs

ehildenb:
- Switch these proofs over to using #gas (infinite gas).
- Allows the <refund> cell to change state in these specs.
- Add no-evaluators to function symbol with no rules.


0xverif:
Looks good to me, waiting on CI to finish

PR #1137, opened 2021-09-24
Bump Z3 version to 4.8.12

Baltoli:
This reflects a change in the downstream K distribution; the main PR is at https://github.com/kframework/k/pull/2225


Baltoli:
This seems to be another Z3 timeout. Will take a look at what test is triggering it

PR #1138, opened 2021-09-28
More ergonomic interaction with provex/haskell backend

ehildenb:
- Add `build-kevm` target which builds the runner script and copies include files into the correct places.
- Add option `--branching-allowed` to `kevm prove` which allows specifying the maximum number of paths to explore when proving.
- Add option `--haskell-backend-arg` for easily injecting new arguments to pass to the haskell backend when proving.

The option `--branching-allowed` can be removed once https://github.com/kframework/k/issues/2224 is addressed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -337,7 +337,7 @@ build-haskell: $(KEVM_LIB)/$(haskell_kompiled) $(KEVM_LIB)/kore-json.py $(lemma_
 build-llvm:    $(KEVM_LIB)/$(llvm_kompiled)    $(KEVM_LIB)/kore-json.py
 build-java:    $(KEVM_LIB)/$(java_kompiled)    $(KEVM_LIB)/kast-json.py $(lemma_includes)
 build-node:    $(KEVM_LIB)/$(node_kompiled)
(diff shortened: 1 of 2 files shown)


radumereuta (approved):
This looks good, but, I have a few other findings.
I want to start profiling the EVM semantics for performance improvements in the front-end, and if I run `make build-provex`, when I run `make test-prove KPROVE_OPTS=--dry-run TEST_SYMBOLIC_BACKEND=haskell` as you suggested, I still see some calls to `kompile`.

The other thing. Do you think you can update to the latest K version? I want to use `--profile-rule-parsing` which is ahead of the current K dependency.

ehildenb:
Calling `make build-provex` should take care of most of the calls to kompile, it's a best effort list though.

K submodule update goes through when auto-updater happens @radumereuta . So when K release is cut and pushed to github, PRs are automatically opened into downstream repos, and anyone can approve them. You'll have to wait for that to go through, though you can locally always peg whatever version of K you want.

radumereuta:
I got some errors with the latest K version.

PR #1142, opened 2021-10-12
tests/failing-symbolic.haskell: add some now passing specs

ehildenb:


ana-pantilie (approved):
Cool! Is this because of changes in the backend?

PR #1149, opened 2021-10-21
Add nix packaging

calin1304:
Following the nix packaging in the [iele-semantics](https://github.com/runtimeverification/iele-semantics) repository, I did the following things so far:
- Added `deps/k_release` which should store the tag of a K release that the semantic depends on. This is generated when calling the `nix-install` target from `Makefile`.
- Pinned the `nixpkgs` version to a specific commit by generating the `nix/sources.json` and `nix/sources.nix` files using [niv](https://github.com/nmattia/niv). 
- Since we let `nix` build all the dependencies, I added flags in the `Makefile` for `libsecp256k1`, `libff` and `libcryptopp` in order to use the ones provided by `nix` in the build environment, instead of building them from the `deps` directory.
- Copied `nix` expression for `libff` (`./nix/libff.nix`).

#### TODO
- [x] Build `kevm-vm` and `kevm` with `nix/kevm.nix`
  - [x] Fix missing includes in `cmake/node/CMakeLists.txt`. `cmake` is complaining about [these missing modules](https://github.com/kframework/llvm-backend/tree/master/cmake) when building `kevm`. These are found in the `llvm-backend` source, so I guess we need to have that in `deps/k/llvm-backend..` in order to build `kevm`
- [x] Update `INSTALL.md`


calin1304:
@ehildenb this PR should be ready for review.

One final thing I'm not sure about is how to get the correct K version when building with nix. I added a file `deps/k_release` that stores a tag to the K release version, based on what I saw other repos do. I'm not sure how this file should be updated whenever `deps/k` is updated. Let me know if I should do something else.

Installing with nix is just cloning the repo, init submodules and calling nix, so there's not much to do regarding CI. I see that other repos make a `release.nix` file using github actions, but I have to look into that some more, I think that could easily be done in another PR.

ehildenb:
@calin1304 you should use `cd deps/k && git tag --points-at HEAD | cut --characters=2-` to get the K release.

ehildenb:
After doing a `git submodule update --init --recursive`

nrdxp:
why symlink in kevm-vm binary when you can just run `make build` and the install phase will put both in ther proper place, including all the definition files, which as is, are not included and cause errors.

See my attempt at reusing some of this code to package as a Nix flake for an example of what I mean:
https://github.com/input-output-hk/kevm.flake/blob/6abf9d73b958e152a3bfcfd723f250a596c82a69/pkgs/kevm.nix#L51-L55

nrdxp:
@calin1304, one thing I forgot to mention about the comment above is that it also places the proper version string in by default. In general, it seems like a better approach since the default installPhase just drops everything that was created at the build path, which when running `make build` is everything that is needed

goodlyrottenapple:
completed in #1287

PR #1155, opened 2021-11-04
Make some proofs pass with Haskell backend

andreiburdusa:
Make all but two "benchmarks" proofs pass with Haskell backend

### DIFF
--- a/tests/specs/benchmarks/verification.k
+++ b/tests/specs/benchmarks/verification.k
@@ -20,6 +20,8 @@ module VERIFICATION-COMMON
     rule 0 <=Int #asWord(#ecrec(_   , _   , _   , _   ))                  => true                                                      [simplification]
     rule         #asWord(#ecrec(_   , _   , _   , _   )) <=Int maxUInt160 => true                                                      [simplification]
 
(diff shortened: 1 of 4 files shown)


tests/specs/benchmarks/verification.k
```diff
@@ -20,6 +20,8 @@ module VERIFICATION-COMMON
     rule 0 <=Int #asWord(#ecrec(_   , _   , _   , _   ))                  => true                                                      [simplification]
     rule         #asWord(#ecrec(_   , _   , _   , _   )) <=Int maxUInt160 => true                                                      [simplification]
 
+    rule maxUInt8   &Int X => X requires #rangeUInt(8 , X)      [simplification]
+
 endmodule
```
ehildenb, line 26: It looks like there are similar lemmas for this one here: https://github.com/kframework/evm-semantics/blob/master/tests/specs/lemmas.k#L288

Note that the linked lemmas are for the Haskell backend. The Java backend's version is here (same file, LEMMAS-JAVA module): https://github.com/kframework/evm-semantics/blob/master/tests/specs/lemmas.k#L202

So I think this lemma should be moved to `LEMMAS-HASKELL` in `tests/specs/lemmas.k`.

ehildenb (approved):

PR #1158, opened 2021-11-12
Fix release process

ehildenb:
The main problem was that we were using the `Makefile` to test the installed package, but the makefile automatically tries to build the semantics in the repository clone.

So I switched to just normal calls to `kevm` for testing, which is more appropriate anyway because we shouldn't be accidentally using things the Makefile helps us with to test installed versions.

I also add a `--verbose` flag to KEVM that prints out timing information when executing, and passes through to sub-commands.

### DIFF
--- a/package/test-package.sh
+++ b/package/test-package.sh
@@ -36,6 +36,14 @@ kevm kompile --backend haskell tests/specs/erc20/verification.k \
 kevm prove tests/specs/erc20/ds/transfer-failure-1-a-spec.k --backend haskell --format-failures \
     --directory tests/specs/erc20/verification/haskell
 
+kevm kompile --backend java tests/specs/erc20/verification.k \
+    --directory tests/specs/erc20/verification/java          \
+    --main-module VERIFICATION                               \
+    --syntax-module VERIFICATION                             \
+    --concrete-rules-file tests/specs/erc20/concrete-rules.txt
+kevm prove tests/specs/erc20/ds/transfer-failure-1-a-spec.k --backend java --format-failures --debugger \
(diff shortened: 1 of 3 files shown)


ehildenb:
Evidence this works: https://office.runtimeverification.com/jenkins/blue/organizations/jenkins/KEVM/detail/PR-1158/14/pipeline

package/test-package.sh
```diff
@@ -0,0 +1,51 @@
...
+    --main-module VERIFICATION                                 \
+    --syntax-module VERIFICATION                               \
+    --concrete-rules-file tests/specs/erc20/concrete-rules.txt \
+    --debug
+kevm prove tests/specs/erc20/ds/transfer-failure-1-a-spec.k --backend java --format-failures --debugger \
+    --directory tests/specs/erc20/verification/java
+
+kevm search tests/interactive/search/branching-invalid.evm "<statusCode> EVMC_INVALID_INSTRUCTION </statusCode>"  --backend haskell > tests/interactive/search/branching-invalid.evm.search-out
+git --no-pager diff --no-index --ignore-all-space -R tests/interactive/search/branching-invalid.evm.search-out tests/interactive/search/branching-invalid.evm.search-expected
+rm -rf tests/interactive/search/branching-invalid.evm.search-out
```
radumereuta, line 51: No parallelism here?

radumereuta (approved):
lgtm
Just the odd question.

package/test-package.sh
ehildenb: No it's hard without the makefile

PR #1159, opened 2021-11-15
Adding lemmas enabling Vat-flux-diff-pass-rough-spec to pass

ACassimiro:
* Added lemmas for the simplification of #buf and keccak comparison

* Added lemmas for the simplification of operations over bool2Word

* Added cases for verification of the newly added lemmas to tests/specs/mcd/functional-spec.k file

* Removing vat-flux-diff-pass-rough-spec from tests/failing-symbolic.haskell


tests/specs/mcd/verification.k
```diff
@@ -74,6 +74,22 @@ endmodule
 module LEMMAS-MCD-HASKELL [symbolic, kore]
     imports LEMMAS-MCD-COMMON
 
+  // ########################
+  // Word Reasoning
+  // ########################
+  //Simplification of bool2Word() ==Int CONCRETE, #buf() ==K CONCRETE
+  rule I                   ==Int bool2Word( B:Bool ) => bool2Word(B) ==Int I  [simplification, concrete(I)]
+  rule bool2Word( B:Bool ) ==Int I                   => B ==K word2Bool(I)    [simplification, concrete(I)]
```
ehildenb, line 82: Are these lemmas needed actually?

Also, it looks to me like the second lemma here (line 82), is simplifying a `bool2Word` to `word2Bool`, but all the following lemmas seem to prefer to reason over `bool2Word` directly. So it's probably better to leave things in `bool2Word` form if we can, and just reason about that instead.

If you remove these lemmas, does the proof pass? These ones don't look like they're exercised by your functional spec above, so if the proof doesn't pass without these lemmas we probably need another functional spec describing that case.

PR #1161, opened 2021-11-22
Make ecrecoverloop00-sig1-invalid-spec.k pass with Haskell backend

andreiburdusa:
The lemmas I added were already present in [`LEMMAS-JAVA` module](https://github.com/kframework/evm-semantics/blob/af66662b9db26f2317f073cf8b32e6218005ca46/tests/specs/lemmas.k#L246).

I think they hold because these are all the rules I found that implement `#asWord`:
```
    rule #asWord(WS) => chop(Bytes2Int(WS, BE, Unsigned)) [concrete]
```
```
    rule [#asWord.base-empty]:  #asWord( .WordStack     ) => 0
    rule [#asWord.base-single]: #asWord( W : .WordStack ) => W
    rule [#asWord.recursive]:   #asWord( W0 : W1 : WS   ) => #asWord(((W0 *Word 256) +Word W1) : WS)
```
- [`chop` interprets an integer modulo `2^256`](https://github.com/kframework/evm-semantics/blob/af66662b9db26f2317f073cf8b32e6218005ca46/evm-types.md?plain=1#L156)
- `+Word` is implemented in terms of `chop` (https://github.com/kframework/evm-semantics/blob/af66662b9db26f2317f073cf8b32e6218005ca46/evm-types.md?plain=1#L236)

I'm not completely sure that the rhs of `#asWord.base-single` (`W`) is always greather or equal to 0 and less than `2^256`, but think this is the case [because it's a word](https://github.com/kframework/evm-semantics/blob/master/evm-types.md#data-structures-over-word).

### DIFF
--- a/tests/failing-symbolic.haskell
+++ b/tests/failing-symbolic.haskell
@@ -1,4 +1,3 @@
-tests/specs/benchmarks/ecrecover00-sigvalid-spec.k
 tests/specs/benchmarks/ecrecoverloop00-sig1-invalid-spec.k
 tests/specs/bihu/collectToken-spec.k
 tests/specs/bihu/forwardToHotWallet-success-1-spec.k
--- a/tests/specs/lemmas.k
+++ b/tests/specs/lemmas.k
@@ -47,6 +47,9 @@ module LEMMAS
 
     rule #isPrecompiledAccount(#newAddr(_, _), _) => false [simplification]
 
+    rule 0 <=Int #asWord(_WS)             => true [simplification]
+    rule         #asWord(_WS) <Int pow256 => true [simplification]
+
   // ########################
   // Keccak
   // ########################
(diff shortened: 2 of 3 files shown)


tests/specs/lemmas.k
```diff
@@ -47,6 +47,9 @@ module LEMMAS
 
     rule #isPrecompiledAccount(#newAddr(_, _), _) => false [simplification]
 
+    rule 0 <=Int #asWord(_WS)             => true [simplification]
+    rule         #asWord(_WS) <Int pow256 => true [simplification]
+
   // ########################
```
ehildenb, line 53: These same lemmas should also be removed from the `LEMMAS-JAVA` module.

tests/failing-symbolic.haskell
```diff
@@ -1,4 +1,3 @@
-tests/specs/benchmarks/ecrecover00-sigvalid-spec.k
 tests/specs/benchmarks/ecrecoverloop00-sig1-invalid-spec.k
```
ehildenb, line 1: Were there any other tests that started passing with this change?

andreiburdusa: I didn't get exit code 0 for any new tests

ehildenb (approved):

PR #1166, opened 2021-12-01
Switch to Haskell backend/provex by default for symbolic execution, updates to kevm

ehildenb:
This PR:

- Switches to Haskell backend by default. Devs are realistically only using `--backend haskell`, so this just makes it more convenient, and the Haskell backend behaves well enough to make it default now.
- Make `--provex` default, and switch to providing a `--no-provex` argument (provex has not given us any problems in a while).
- Switch the option `--backend-dir => --directory`, so that tools which integrate with `kprove` won't have to translate this argument to `kevm prove` (eg. the pyk library KProve object).
- More informative logging messages (with `notif`), and add `notif_exec` to tell people the commands being run.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -415,14 +415,14 @@ tests/%.parse: tests/%
 	$(CHECK) $@-out $@-expected
 	$(KEEP_OUTPUTS) || rm -rf $@-out
 
-tests/%.prove: tests/%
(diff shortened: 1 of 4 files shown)


radumereuta (approved):
lgtm

ana-pantilie (approved):

PR #1167, opened 2021-12-04
concrete-rules.txt: add attribute to corresponding rules

vasil-sd:


utils/process-concrete-rules.sh
```diff
@@ -0,0 +1,112 @@
+#!/bin/bash
+
+root_dir=$(dirname $(dirname $(readlink -f "$0")))
+
+rules_map=$(mktemp)
+awk_script=$(mktemp)
+concrete_rules=$(mktemp)
+concrete_rules_map=$(mktemp)
+
+editor_call_file="$root_dir/editor_call.sh"
```
ehildenb, line 10: We don't need this script to stick around, let's remove it.

ehildenb:
Because we're inlining all th econcrete rules into the definition, we no longer need `--concrete-rules-file` to be passed to `kompile` by default in the `Makefile`. So you'll have to remove that call (https://github.com/kframework/evm-semantics/blob/master/Makefile#L419 and https://github.com/kframework/evm-semantics/blob/master/Makefile#L431).

vasil-sd:
> Because we're inlining all th econcrete rules into the definition, we no longer need `--concrete-rules-file` to be passed to `kompile` by default in the `Makefile`. So you'll have to remove that call (https://github.com/kframework/evm-semantics/blob/master/Makefile#L419 and https://github.com/kframework/evm-semantics/blob/master/Makefile#L431).

In last update, I've added set intersection calculation, so only shared rules are removed from `concrete-rules.txt` files. And as a result, there are no empty files, so `--concrete-rules-file` is still needed for the kompile.

PR #1168, opened 2021-12-04
Add solc-to-k command to kevm

ehildenb:
The `solc-to-k` command takes as input a Solidity file and produces the K file needed to run proofs about it.

Todo:

- [x] Make sure `solc` is installed on CI and in the Docker image.

Blocked on: https://github.com/kframework/evm-semantics/pull/1158 and https://github.com/kframework/k/pull/2338


ehildenb:
Need solc installed on CI, example of building Z3 from source (if that's the approach): https://github.com/kframework/evm-semantics/blob/master/Dockerfile#L29

Dockerfile
```diff
@@ -1,6 +1,10 @@
 ARG K_COMMIT
 FROM runtimeverificationinc/kframework-k:ubuntu-focal-${K_COMMIT}
 
+RUN    apt-get update                                   \
+    && apt-get install --yes software-properties-common \
+    && add-apt-repository ppa:ethereum/ethereum
+
 RUN    apt-get update            \
     && apt-get upgrade --yes     \
     && apt-get install --yes     \
```
tothtamas28, line 10: :sparkles: Consider merging the two `apt-get install` commands:

```dockerfile
RUN add-apt-repository ppa:ethereum/ethereum

RUN    apt-get update             \
    && apt-get upgrade --yes      \
    && apt-get install --yes      \
       ...
       software-properties-common \
       ...
```

tothtamas28 (approved):
LGTM

Dockerfile
ehildenb: We can't actually, `add-apt-repository` depends on `software-properties-common` being installed, so that has to happen first so that we can install `solc` in the second block.

PR #1170, opened 2021-12-07
Make some proofs from tests/erc20/ds pass with Haskell backend

vasil-sd:


tests/specs/functional/lemmas-spec.k
```diff
@@ -128,6 +128,9 @@ module LEMMAS-SPEC
     claim <k> runLemma ( chop( #lookup(M, KEY)   ) ) => doneLemma( #lookup(M, KEY) ) ... </k>
     claim <k> runLemma ( chop( bool2Word(B:Bool) ) ) => doneLemma( bool2Word(B)    ) ... </k>
 
+    claim <k> runLemma ( chop ( X +Int Y ) <Int X )  => doneLemma( true )  ... </k> requires pow256 <=Int X +Int Y andBool #rangeUInt(256, X) andBool #rangeUInt(256, Y)
+    claim <k> runLemma ( X <=Int chop ( X +Int Y ) ) => doneLemma( false ) ... </k> requires pow256 <=Int X +Int Y andBool #rangeUInt(256, X) andBool #rangeUInt(256, Y)
+
```
ehildenb, line 133: These lemmas should not be exactly the rule that we have written, because that's not a very good test.

Instead, disable the lemmas you added, then re-run the original proof you were working on, and post the output of it here. I will suggest better tests from the output of that.

vasil-sd: Replaced these tests with smt2 script in comments

ehildenb (approved):

vasil-sd:
@ehildenb , Ok. What is a reasonable value for the the timeout in your opinion?

PR #1172, opened 2021-12-14
Add support for generation of storage slot macros to solc-to-k

tothtamas28:
Fixes: https://github.com/runtimeverification/erc20-verification/issues/267


Makefile
```diff
@@ -426,6 +431,8 @@ tests/%.prove-legacy: tests/%
 tests/specs/examples/erc20-spec/haskell/erc20-spec-kompiled/timestamp: tests/specs/examples/erc20-bin-runtime.k
 tests/specs/examples/erc20-bin-runtime.k: tests/specs/examples/ERC20.sol
 	$(KEVM) solc-to-k $< ERC20 > $@
+tests/specs/examples/vat-bin-runtime.k: tests/specs/examples/vat.sol
+	$(KEVM) solc-to-k $< Vat > $@
```
ehildenb, line 435: Let's add a `test-kevm-pyk` target which runs all the kevm-pyk tests.

ehildenb: And let's make sure it's tested on CI.

kevm_pyk/solc_to_k.py
```diff
@@ -0,0 +1,289 @@
...
+    kevm.symbolTable['abiCallData'] = lambda fname, *args: f'#abiCallData({fname}{"".join([", " + arg for arg in args])})'
+    kevm.symbolTable['address'] = _typed_arg_unparser('address')
+    kevm.symbolTable['bytes32'] = _typed_arg_unparser('bytes32')
+    kevm.symbolTable['int256'] = _typed_arg_unparser('int256')
+    kevm.symbolTable['uint256'] = _typed_arg_unparser('uint256')
+    kevm.symbolTable['rangeAddress'] = lambda t: f'#rangeAddress({t})'
+    kevm.symbolTable['rangeUInt'] = lambda n, t: f'#rangeUInt({n, t})'
+    kevm.symbolTable['rangeSInt'] = lambda n, t: f'#rangeSInt({n, t})'
+    kevm.symbolTable['binRuntime'] = lambda s: '#binRuntime(' + s + ')'
+    kevm.symbolTable[contract_name] = lambda: contract_name
```
ehildenb, line 67: Line up the `=` signs for scanability.

kevm_pyk/solc_to_k.py
```diff
@@ -0,0 +1,289 @@
...
+
+        if encoding == 'mapping':
+            key_type_name = type_dict['key']
+            value_type_name = type_dict['value']
+            return recur_mapping(syntax, lhs, rhs, var_idx, key_type_name, value_type_name)
+
+        raise ValueError(f'Unsupported encoding: {encoding}')
+
+    def recur_value(syntax, lhs, rhs, type_label):
+        if type_label not in {'address', 'uint256', 'bytes32'}:
```
ehildenb, line 139: This same check exists in multiple places. Either it should be factored out so we have a single common list of supported types or the code sholud be refactored.

ehildenb: Also, `make tests/specs/examples/erc20-spec.md.prove` fails because it requires a `uint8`, which is not in this supported list.

kevm_pyk/solc_to_k.py
```diff
@@ -0,0 +1,289 @@
...
+            [KImport('EDSL', True)],
+            [unknownBoolProduction, contractProduction] +
+            storage_productions +
+            [storage_location_production] +
+            storage_rules +
+            function_productions +
+            [function_call_data_production] +
+            function_rules +
+            [binRuntimeProduction, contractMacro]
+        )
```
ehildenb, line 54:
```suggestion
    binRuntimeModule = KFlatModule( binRuntimeModuleName
                                  , [KImport('EDSL', True)]
                                  ,   [unknownBoolProduction, contractProduction]
                                    + storage_productions
                                    + [storage_location_production]
                                    + storage_rules
                                    + function_productions
                                    + [function_call_data_production]
                                    + function_rules
                                    + [binRuntimeProduction, contractMacro]
                                  )
```

In general I prefer the join operator for associative lists to show up at the beginning of a line, not at the end, so it's easier to see where it starts/ends. May require putting `\` at the end of the line too.

I also like, as much as possible, for things to line up horizontally (eg. below where I line up the `=` sign in the symbolTable definition). This makes it much easier to scan the structure of the file based on its "shape".

ehildenb: In this case, the `_,_` and the `_+_` are associative join operators.

kevm_pyk/solc_to_k.py
tothtamas28: Done.

Note that this formatting is against the Python [style guide](https://www.python.org/dev/peps/pep-0008/#pet-peeves). I suggest we start a discussion within the team for our own coding guidelines.

kevm_pyk/solc_to_k.py
tothtamas28: This prototype version does not support packing of multiple items into a single storage slot, thus for safety it raises an exception for small types (to be precise, for types that do not take more than half of a slot) like `uint8`.

What is your preference: should this feature be implemented in this PR, or should we merge first?

kevm_pyk/solc_to_k.py
ehildenb: Hmmm, I guess we need to either adhere to their styles or pick our own style guide I guess.

I usually prefer readability and scanability to anything else, but there is something to be said for adhering to a global style guide. I find it much easier to navigate a file which is visually "pleasing" and structured.

What is your opinion on it?

kevm_pyk/solc_to_k.py
ehildenb: Well we'll need it implemented, because `tests/specs/examples/erc20-spec.md.prove` currently works and currently uses it. But we could also have a flag `--no-storage-slots`, which keeps the python script from emitting the storage slots, and pass that flag for that test.

kevm_pyk/solc_to_k.py
tothtamas28: In my opinion, the simplest solution would be to adhere to the official style guide. This way it is easy to enforce conformance using style checkers ([pycodestyle](https://github.com/PyCQA/pycodestyle)) and code formatters (e.g. [black](https://github.com/psf/black) and [isort](https://github.com/PyCQA/isort)) that we can integrate in the CI pipeline. (We might be able to configure the style checker by disabling certain rules, but I think the same approach wouldn't work with the code formatter.)

But on the other hand, as you pointed out, the standard formatting might not be optimal w.r.t. structuring or aesthetics.

kevm_pyk/solc_to_k.py
tothtamas28: I think the `--no-storage-slots` feature is valuable in itself, and is also relatively easy to implement, so I'll start with that to unblock merging this PR.

Created an issue for the rest: https://github.com/runtimeverification/erc20-verification/issues/279

kevm
```diff
@@ -259,8 +259,8 @@ run_solc() {
     contract_name="$1" ; shift
     json_run_file="${run_file}.json"
 
-    notif_exec solc --combined-json abi,bin-runtime,storage-layout "$@" "${run_file}" > "${json_run_file}"
-    notif_exec kevm-pyk.py solc-to-k ${backend_dir}/*-kompiled "${run_file}" "${json_run_file}" "${contract_name}"
+    notif_exec solc --combined-json abi,bin-runtime,storage-layout "${run_file}" > "${json_run_file}"
+    notif_exec python3 -m kevm_pyk solc-to-k "$@" ${backend_dir}/*-kompiled "${run_file}" "${json_run_file}" "${contract_name}"
```
tothtamas28, line 263: For easy passing of keyword arguments to `kevm_pyk`, I moved `$@` from the invocation of `solc` to the invocation of `kevm_pyk`. Is it safe, or do we need to call `solc` with optional arguments as well? (In this case, we will have to distinguish between these two types of arguments.)

kevm_pyk/solc_to_k.py
ehildenb: Let's stick to the existing style actually, because it mirrors what is in upstream pyk library better, and migrating that whole library over will be a pain.

I personally don't care much for linters or style formatters, I believe aesthetics are more important (code is read more often than it's written).

Jenkinsfile
```diff
@@ -46,6 +46,11 @@ pipeline {
             stage('KEVM help')      { steps { sh './kevm help'                                                    } }
           }
         }
+        stage('Test kevm_pyk') {
+          failFast true
+          options { timeout(time: 60, unit: 'SECONDS') }
+          steps { sh 'make test-kevm-pyk' }
+        }
```
ehildenb, line 53: This should be a sub-stage of `Test Interactive` I would think, unless you needed it to run after that for a reason? It can just be one line, after the `make test-node` line.

Makefile
```diff
@@ -424,8 +430,12 @@ tests/%.prove-legacy: tests/%
 	    --no-provex --format-failures $(KPROVE_OPTS) --concrete-rules-file $(dir $@)concrete-rules.txt
 
 tests/specs/examples/erc20-spec/haskell/erc20-spec-kompiled/timestamp: tests/specs/examples/erc20-bin-runtime.k
-tests/specs/examples/erc20-bin-runtime.k: tests/specs/examples/ERC20.sol $(KEVM_LIB)/$(haskell_kompiled) $(KEVM_LIB)/kevm-pyk.py
-	$(KEVM) solc-to-k $< ERC20 > $@
+tests/specs/examples/erc20-bin-runtime.k: tests/specs/examples/ERC20.sol $(KEVM_LIB)/$(haskell_kompiled) build-kevm-pyk
```
ehildenb, line 433: The reason to have `$(KEVM_LIB)/kevm-pyk.py` instead of `build-kevm-pyk` is because on rebuilds (run this target twice), it will re-trigger a build of `bulid-kevm-pyk` because it's not a file, but `$(KEVM_LIB)/kevm-pyk.py` is a file that it can check the timestamp of.

At least, that's what I am remembering now, but it may be incorrect because `build-kevm-pyk` is marked as `.PHONY`.

You should double-check that on re-builds it doesn't end up _always_ rebuilding the `build-kevm-pyk` target, even if none of its subtargets have been modified.

Makefile
```diff
@@ -563,6 +573,13 @@ failure_tests:=$(wildcard tests/failing/*.json)
 
 test-failure: $(failure_tests:=.run-expected)
 
+# kevm_pyk Tests
+
+kevm_pyk_tests := tests/specs/examples/erc20-bin-runtime.k \
+                  tests/specs/examples/vat-bin-runtime.k
+
+test-kevm-pyk: $(kevm_pyk_tests)
```
ehildenb, line 581: I guess we should test more than the fact that we're generating them successfully, rather that:

- We generate them, and they are textually identical to `*.expected` files we see.
- They are `kompile`-able.

ehildenb: So you can modify it (to check that they are identical to expected files):

```
test-kevm-pyk: $(kevm_pyk_tests:=.gen)

tests/specs/examples/%.k.gen: tests/specs/examples/%.k
    $(CHECK) $< $<.expected
```

It may be tougher to check that they are kompilable, but we already are testing it at least for the ERC20 example because we're running some proofs with the generated file. So that may be enough for now.

tests/specs/examples/vat.sol
```diff
@@ -0,0 +1,246 @@
...
+// You should have received a copy of the GNU Affero General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+
+pragma solidity >=0.5.12;
+
+// FIXME: This contract was altered compared to the production version.
+// It doesn't use LibNote anymore.
+// New deployments of this contract will need to include custom events (TO DO).
+
+contract Vat {
```
ehildenb, line 26: I think we better pick a different example to include in the repo here, AGPL is not license compatible with BSD 3-clause, which this repo basically is.

ehildenb: Any contract here is a good choice: https://github.com/OpenZeppelin/openzeppelin-contracts

For example the tokens here: https://github.com/OpenZeppelin/openzeppelin-contracts/tree/7f6a1666fac8ecff5dd467d0938069bc221ea9e0/contracts/token

Any of these can be collapsed to a single file, and as long as the MIT license identifier remains at the top we remain license compatible.

ehildenb: You can also just google for any MIT-ish licensed ERC-777 or ERC-721 (if you want something more complicated than an ERC-20). ERC-20 is also probably a fine choice.

Makefile
tothtamas28: You are right.

> A phony target should not be a prerequisite of a real target file; if it is, its recipe will be run every time make goes to update that file. As long as a phony target is never a prerequisite of a real target, the phony target recipe will be executed only when the phony target is a specified goal (see [Arguments to Specify the Goals](https://www.gnu.org/software/make/manual/html_node/Goals.html#Goals)). 

([Phony Targets](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html))

tests/specs/examples/erc20-bin-runtime.k.expected
```diff
@@ -0,0 +1,90 @@
...
+    syntax ERC20Function ::= "transfer" "(" Int "," Int ")" 
+    
+    syntax ERC20Function ::= "transferFrom" "(" Int "," Int "," Int ")" 
+    
+    syntax ByteArray ::= ERC20Contract "." ERC20Function [klabel(function_ERC20), macro()]
+    
+    rule  ( ERC20.allowance(OWNER, SPENDER) => #abiCallData("allowance", #address(OWNER), #address(SPENDER)) )
+      requires ( #rangeAddress(OWNER)
+       andBool ( #rangeAddress(SPENDER)
+               ))
```
ehildenb, line 33: Hmmm, `macro` rules aren't allowed to have `requires` clauses on them actually.

I would actually prefer if we could make these `ensures` clauses, instead of `requires` clauses. Specifically because then we will be adding the needed constraints `#rangeAddress` to those variables for the user, instead of requiring that the user adds them themselves. I'm not sure it's possible or sound though to make `ERC20Contract "." ERC20Function` production a `function` and have the rule read:

```
    rule  ( ERC20.allowance(OWNER, SPENDER) => #abiCallData("allowance", #address(OWNER), #address(SPENDER)) )
      ensures ( #rangeAddress(OWNER)
       andBool ( #rangeAddress(SPENDER)
               ))
```

tests/specs/examples/vat.sol
tothtamas28: Added example for ERC-721.

tests/specs/examples/erc20-bin-runtime.k.expected
```diff
@@ -0,0 +1,90 @@
...
+    
+    syntax ERC20Function ::= "transferFrom" "(" Int "," Int "," Int ")" 
+    
+    syntax ByteArray ::= ERC20Contract "." ERC20Function [klabel(function_ERC20), macro()]
+    
+    rule  ( ERC20.allowance(OWNER, SPENDER) => #abiCallData("allowance", #address(OWNER), #address(SPENDER)) )
+      
+       ensures ( #rangeAddress(OWNER)
+       andBool ( #rangeAddress(SPENDER)
+               ))
```
ehildenb, line 34: I don't think this can be a `macro`, the `ensures` clause won't be passed through correctly.

You can either make it a `function` instead, or you can inpsect the generated Kore for a claim that uses this and see if it passes through the ensures clause correctly.

ehildenb:
Hmmmm, the magic string appended to the end of the bytecode is different between different machines, which means that we realistically can't test this by comparing the actual file output.

We have two options:

- Don't compare output yet, make an issue for how to test it.
- Test it by running some actual proofs against the results. This can be done by migrating `tests/specs/examples/erc20-spec.md` to use more of the sugared notations (eg. the `abiCallData` can become th enice sugared `ERC20.functionName(...)` notation), and writing a few proofs in a file like `tests/specs/examples/erc721-spec.md` in the same style, also using the sugared notation.

Arguably, the second option is better, and it tests what we actually want out of this: that is, that the functionality is usable in proofs. Let's give that a try, just 1 or 2 proofs about the ERC721 should be enough (one for a simple getter, another for something that affects some storage slot so we can test the storage slot sugar). The ERC20 example can just be migrated to use the sugared `<callData>` (no need to try and use the sugared storage slots, since that doesn't work yet I guess).

tests/specs/examples/erc20-bin-runtime.k.expected
```diff
@@ -0,0 +1,134 @@
...
+    syntax ERC20Storage ::= "_symbol" 
+    
+    syntax Int ::= ERC20Contract "." ERC20Storage [klabel(storage_ERC20), macro()]
+    
+    rule  ( ERC20._balances[V0] => #hashedLocation("Solidity", 0 +Int 0, V0) )
+      
+       
+      
+    
+    rule  ( ERC20._allowances[V0][V1] => #hashedLocation("Solidity", #hashedLocation("Solidity", 0 +Int 1, V0), V1) )
```
ehildenb, line 29: Hmmm, we're not really taking advantage of the already-recursive definition of `#hashedLocation` here...

Any reason you chose to do it this way instead of say:

```suggestion
    rule  ( ERC20._allowances[V0][V1] => #hashedLocation("Solidity", 0 +Int 1, V0 V1 .IntList)
```

tests/specs/examples/erc721-spec.md
```diff
@@ -0,0 +1,205 @@
+ERC721-ish Verification
+======+================
```
ehildenb, line 2: Weird formatting bug?

tests/specs/examples/erc20-bin-runtime.k.expected
tothtamas28: This is for simplicity of implementation: this way, the structure of the term encoding the base address does not have to be considered when constructing the term encoding an element at a given key.

There a couple of similar simplifications that can be made for a nicer output (these are hinted on in comments). I opened an issue: https://github.com/runtimeverification/erc20-verification/issues/302

tests/specs/examples/erc721-spec.md
```diff
@@ -0,0 +1,147 @@
...
+
+```k
+claim <k> runLemma(chop(chop(chop(chop(chop(chop((#lookup(ACCT_STORAGE, 0) /Int 2 &Int 127) +Int 31) /Int 32 *Int 32) +Int 32) +Int 128) +Int 32) +Int 32))
+       => doneLemma((((((((#lookup(ACCT_STORAGE, 0) /Int 2 &Int 127) +Int 31) /Int 32 *Int 32) +Int 32) +Int 128) +Int 32) +Int 32)) </k>
+```
+
+```k
+claim <k> runLemma(#lookup(ACCT_STORAGE, 0) /Int 2 <Int 32)
+       => doneLemma(false) </k>
+ requires 32 <=Int #lookup(ACCT_STORAGE, 0) /Int 2 &Int 127
```
ehildenb, line 69:
`````suggestion
claim <k> runLemma(chop(chop(chop(chop(chop(chop((#lookup(ACCT_STORAGE, 0) /Int 2 &Int 127) +Int 31) /Int 32 *Int 32) +Int 32) +Int 128) +Int 32) +Int 32))
       => doneLemma((((((((#lookup(ACCT_STORAGE, 0) /Int 2 &Int 127) +Int 31) /Int 32 *Int 32) +Int 32) +Int 128) +Int 32) +Int 32)) ... </k>
```

```k
claim <k> runLemma(#lookup(ACCT_STORAGE, 0) /Int 2 <Int 32)
       => doneLemma(false) ... </k>
 requires 32 <=Int #lookup(ACCT_STORAGE, 0) /Int 2 &Int 127
`````

Comment (ehildenb) [reply tests/specs/examples/erc721-spec.md:69]: Also, I'm surprised that `... +Int 32) +Int 128) +Int 32) ...` doesn't get simplified out. Maybe you need a lemma like `(X +Int Y) +Int Z) => X +Int (Y +Int Z) [simplification, concrete(Y, Z)]`? But I guess we should already have lemmas for dealing with this in `INT-SIMPLIFICATION` in `tests/specs/int-simplification.k`, which should be included via `INFINITE-GAS`.

Or maybe this simplification is actually happenning, but your proof claim RHS isn't as simplified as it can be?

File: tests/specs/examples/erc721-spec.md [hunk 15]
```diff
@@ -0,0 +1,147 @@
...
+```
+
+### Calling name() works
+
+-   Everything from `<mode>` to `<callValue>` is boilerplate.
+-   We are setting `<callData>` to `name()`.
+-   We claim that in cases the account storage is well-formed, we will end in `EVMC_SUCCESS`, and in `EVMC_REVERT` otherwise (TODO: add claims for long strings).
+-   TODO: The `<output>` cell specifies that we correctly lookup the `NAME` value from the account storage.
```
ehildenb, line 77: Should we add a TODO for the `long` variant of these specs too? (the name does not fit in 32 bytes)

ehildenb (approved):

ehildenb:
Have a look at my comments, and make a decision about whether to ignore them or not. But feel free to add automerge if you're happy with things.

PR #1176, opened 2022-01-11
Make sure solc is present to test packages

ehildenb:
Our CI for testing the built Ubuntu package is not currently installing solc ahead of time, which means that the testing of `kevm solc-to-k ...` is not working on package testing.

This installs `solc` first, then does the package testing.

Note that I don't want to add `solc` as a dependency in the Ubuntu package directly `package/debian/control.focal`, because it requires adding a PPA, and will make it harder for a user to specify the version of solc they want.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -86,7 +86,7 @@ pipeline {
               }
               options {
                 skipDefaultCheckout()
-                timeout(time: 25, unit: 'MINUTES')
+                timeout(time: 35, unit: 'MINUTES')
               }
               steps {
                 dir('focal-test') {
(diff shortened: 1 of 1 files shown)


ehildenb:
Evidence this works: https://office.runtimeverification.com/jenkins/blue/organizations/jenkins/KEVM/detail/PR-1176/1/pipeline

Jenkinsfile
```diff
@@ -82,7 +82,7 @@ pipeline {
               }
               options {
                 skipDefaultCheckout()
-                timeout(time: 25, unit: 'MINUTES')
+                timeout(time: 35, unit: 'MINUTES')
```
anvacaru, line 85: Does the stage time out without this one?

tothtamas28 (approved):
LGTM

PR #1180, opened 2022-01-19
kevm-pyk.py: remove solidity compiler version from EVM bytecode binary

vasil-sd:


vasil-sd:
This PR was dropped after discussion with @ehildenb.

PR #1187, opened 2022-01-31
London schedule update

gtrepta:
fixes: https://github.com/runtimeverification/firefly/issues/1226

https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/london.md


serialization.md
```diff
@@ -732,5 +717,21 @@ Tree Root Helper Functions
     rule #asmTxPrefix (0) => Legacy
     rule #asmTxPrefix (1) => AccessList
 
+    syntax TxData ::= LegacyTxData         ( nonce: Int, gasPrice: Int, gasLimit: Int, to: Account, value: Int, data: ByteArray )
+                    | LegacyProtectedTxData( nonce: Int, gasPrice: Int, gasLimit: Int, to: Account, value: Int, data: ByteArray, chainId: Int )
+                    | AccessListTxData     ( nonce: Int, gasPrice: Int, gasLimit: Int, to: Account, value: Int, data: ByteArray, chainId: Int, accessLists: JSONs )
+ // ---------------------------------------------------------------------------------------------------------------------------------------------------------------
```
gtrepta, line 723: This is to help with implementing the new transaction type in EIP-1559 and future transaction types. The parameters for symbols like `#sender` and `#hashUnsignedTx` were becoming quite numerous and were unused in certain cases, so this also tries to cut down on that.

driver.md
```diff
@@ -72,9 +72,49 @@ To do so, we'll extend sort `JSON` with some EVM specific syntax, and provide a
...
+    rule <k> startTx => loadTx( #TxSender(TXID) ) ... </k>
          <txPending> ListItem(TXID:Int) ... </txPending>
+
+    syntax AcctExp ::= Account
+    syntax KResult ::= Account
+    syntax AcctExp ::= #TxSender( Int )
+ // -----------------------------------
+    rule <k> #TxSender( TXID ) => #sender( LegacyTxData(TN, TP, TG, TT, TV, DATA), TW, TR, TS ) ... </k>
+         <message>
+           <msgID>      TXID </msgID>
```
gtrepta, line 84: I've added this `#TxSender` cooling rule to `loadTx` to use the new `TxData` symbols.

It's here in `driver.md`, right next to the only place it's needed, but Firefly would benefit from it and it doesn't use `driver.md`. I guess it can go somewhere in `evm.md` or `state-loader.md`, but I'm not sure which.

serialization.md
```diff
@@ -37,20 +37,16 @@ Address/Hash Helpers
     rule [#newAddr]:        #newAddr(ACCT, NONCE) => #addr(#parseHexWord(Keccak256(#rlpEncode([#addrBytes(ACCT), NONCE]))))
     rule [#newAddrCreate2]: #newAddr(ACCT, SALT, INITCODE) => #addr(#parseHexWord(Keccak256("\xff" +String #unparseByteStack(#addrBytes(ACCT)) +String #unparseByteStack(#wordBytes(SALT)) +String #unparseByteStack(#parseHexBytes(Keccak256(#unparseByteStack(INITCODE)))))))
 
-    syntax Account ::= #sender ( Int , Int , Int , Int , Account , Int , String , Int , JSONs , Int , ByteArray, ByteArray )    [function]
-                     | #sender ( String , Int , String , String )                                                               [function, klabel(#senderAux) ]
-                     | #sender ( String )                                                                                       [function, klabel(#senderAux2)]
- // -----------------------------------------------------------------------------------------------------------------------------------------------------------
-    rule #sender(_, _, _, _, _, _, _, _, _, TW => TW +Int 27, _, _)
+    syntax Account ::= #sender ( TxData , Int , ByteArray , ByteArray ) [function, klabel(#senderTxData)]
```
gtrepta, line 40: This type of pattern is the reason for the new transaction symbols. While the new symbols may be just as verbose if not more, hopefully they're more clear whereas what was happening here is kind of opaque.

gtrepta (commented):
@ehildenb I left some comments on this WIP, I'd like your feedback if you have the chance.

serialization.md
ehildenb: Oh wow, this is really nice, makes it much more easy to compare to the EIP RLP spec :)

driver.md
ehildenb: I think because we are just using `strict` to access state, but not to evaluate some expression language (which would require nested unfolding of the expressions using the heating/cooling rules), we can probably use contextual functions here instead:

```
rule [[ #TxSender(TXID) => #sender(...) ]]
       <message>
       ...
       </message>   
```

driver.md
```diff
@@ -86,14 +126,14 @@ To do so, we'll extend sort `JSON` with some EVM specific syntax, and provide a
...
            <txChainID>  CID  </txChainID>
            <txAccess>   TA   </txAccess>
+           <txType> AccessList </txType>
            ...
          </message>
 
-    syntax EthereumCommand ::= loadTx ( Account )
- // ---------------------------------------------
+    syntax EthereumCommand ::= loadTx ( AcctExp ) [strict]
+ // ------------------------------------------------------
```
ehildenb, line 136: With `#TxSender` as a contextual function for sort `Account`, then we don't need this to be `strict`.

evm.md
```diff
@@ -2385,7 +2403,8 @@ A `ScheduleFlag` is a boolean determined by the fee schedule; applying a `Schedu
     syntax ScheduleFlag ::= "Gselfdestructnewaccount" | "Gstaticcalldepth" | "Gemptyisnonexistent" | "Gzerovaluenewaccountgas"
                           | "Ghasrevert"              | "Ghasreturndata"   | "Ghasstaticcall"      | "Ghasshift"
                           | "Ghasdirtysstore"         | "Ghascreate2"      | "Ghasextcodehash"     | "Ghasselfbalance"
-                          | "Ghassstorestipend"       | "Ghaschainid"      | "Ghasaccesslist"
+                          | "Ghassstorestipend"       | "Ghaschainid"      | "Ghasaccesslist"      | "Ghasbasefee"
+                          | "Ghasrejectedfirstbyte"
  // -----------------------------------------------------------------------------------------
```
anvacaru, line 2408:
```suggestion
 // -----------------------------------------------
```

driver.md
```diff
@@ -350,6 +348,8 @@ Note that `TEST` is sorted here so that key `"network"` comes before key `"pre"`
     rule <k> loadTransaction _ { "data"       : (TI:String => #parseByteStackRaw(TI)), _                  } ... </k>
     rule <k> loadTransaction _ { "r"          : (TR:String => #padToWidth(32, #parseByteStackRaw(TR))), _ } ... </k>
     rule <k> loadTransaction _ { "s"          : (TS:String => #padToWidth(32, #parseByteStackRaw(TS))), _ } ... </k>
+    rule <k> loadTransaction _ { "maxPriorityFeePerGas" : (V:String => #asWord(#parseByteStackRaw(V))), _ } ... </k>
+    rule <k> loadTransaction _ { "maxFeePerGas"         : (V:String => #asWord(#parseByteStackRaw(V))), _ } ... </k>
```
ehildenb, line 352: formatting

driver.md
```diff
@@ -540,6 +543,10 @@ Here we check the other post-conditions associated with an EVM test.
     rule <k> check "transactions" : ("value"    : VALUE) => . ... </k> <txOrder> ListItem(TXID) ... </txOrder> <message> <msgID> TXID </msgID> <value>      VALUE </value>      ... </message>
     rule <k> check "transactions" : ("chainId"  : VALUE) => . ... </k> <txOrder> ListItem(TXID) ... </txOrder> <message> <msgID> TXID </msgID> <txChainID>  VALUE </txChainID>  ... </message>
     rule <k> check "transactions" : ("type"     : VALUE) => . ... </k> <txOrder> ListItem(TXID) ... </txOrder> <message> <msgID> TXID </msgID> <txType>     TYPE  </txType>     ... </message> requires VALUE ==Int #dasmTxPrefix(TYPE)
+    rule <k> check "transactions" : ("maxFeePerGas"
+                                                : VALUE) => . ... </k> <txOrder> ListItem(TXID) ... </txOrder> <message> <msgID> TXID </msgID> <txMaxFee>   VALUE </txMaxFee>   ... </message>
+    rule <k> check "transactions" : ("maxPriorityFeePerGas"
+                                                : VALUE) => . ... </k> <txOrder> ListItem(TXID) ... </txOrder> <message> <msgID> TXID </msgID> <txPriorityFee> VALUE </txPriorityFee>  ... </message>
```
ehildenb, line 549: Formatting here is kinda weird too.

evm.md
```diff
@@ -144,19 +145,21 @@ In the comments next to each cell, we've marked which component of the YellowPap
...
+                <value>         0          </value>      // T_v
+                <sigV>          0          </sigV>       // T_w
+                <sigR>          .ByteArray </sigR>       // T_r
+                <sigS>          .ByteArray </sigS>       // T_s
+                <data>          .ByteArray </data>       // T_i/T_e
+                <txType>        .TxType    </txType>
+                <txAccess>      [ .JSONs ] </txAccess>
+                <txChainID>     0          </txChainID>
+                <txPriorityFee> 0          </txPriorityFee>
+                <txMaxFee>      0          </txMaxFee>
```
ehildenb, line 162: Do these fields have names in the Yellow paper (like the `T_s` and `T_i` and such above)? Or do they have names on the EIP itself? If so, let's get those added as comments.

driver.md
gtrepta: These lines were pathologically long, so I split them to get some portion of them to line up with the line above...

evm.md
```diff
@@ -1543,6 +1551,11 @@ For each `CALL*` operation, we make a corresponding call to `#call` and a state-
            ...
          </account>
 
+    syntax Bool ::= #isValidCode ( ByteArray , Schedule ) [function]
+ // ----------------------------------------------------------------
+    rule #isValidCode( OUT ,  SCHED) => notBool Ghasrejectedfirstbyte << SCHED >> orBool OUT[0] =/=Int 239 requires #sizeByteArray(OUT) >Int 0
+    rule #isValidCode(_OUT , _SCHED) => true [owise]
```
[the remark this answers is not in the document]
ehildenb: Also, can `notBool A orBool B` be written instead as `A impliesBool B` here? That makes more sense to me at least.

evm.md
gtrepta: The yellow paper has `T_a` and `T_c` for the access list/chainid, so I'll add those. It doesn't look like a london version of the paper exists yet, and the EIP doesn't give any similar looking names to the new parameters.

evm.md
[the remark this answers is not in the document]
gtrepta: oh, that's cool. I wasn't aware of `impliesBool`.

anvacaru (approved):
Looks good to me

state-utils.md
```diff
@@ -343,6 +394,43 @@ The `"rlp"` key loads the block information.
...
+           <txType> DynamicFee </txType>
+           ...
+         </message>
+
+    syntax Int ::= #effectiveGasPrice( Int ) [function]
+ // ---------------------------------------------------
+    rule [[ #effectiveGasPrice( TXID )
+         => #if notBool Ghasbasefee << SCHED >>
+                orBool TXTYPE ==K Legacy
+                orBool TXTYPE ==K AccessList
```
ehildenb, line 419: I would recommend using parens around `notBool Ghasbasefee << SCHED >>`

Also just want to mention that `#if_#then_#else_#fi` doesn't behave very well for symbolic execution, so often it can be better to split the rule into two instead if there is any chance that the condition in the `#if_...` is going to be symbolic. There is no risk of that happening here though, because this is `driver.md` which isn't really used for symbolic execution.

ehildenb (approved):

state-utils.md
gtrepta: Noted. Although I do see quite a lot of `#if_#then...` being used in evm.md. The main concern is that each branch of the statement has to be well defined, is that right?

anvacaru: This is just a suggestion.
I would have a rule `#computeGasPrice` like 
```
    syntax EthereumCommand ::= "#computeGasPrice"
 // ---------------------------------------------
    rule <k> #computeGasPrice => . ... </k> <schedule> SCHED </schedule> requires notBool Ghasbasefee << SCHED >>

    rule <k> #computeGasPrice => . ... </k>
         <schedule> SCHED </schedule>
         <baseFeePerGas> BF </baseFeePerGas>
         <txPending> ListItem(TXID:Int) ... </txPending>
         <message>
           <msgID>         TXID </msgID>
           <txMaxFee>      MF   </txMaxFee>
           <txPriorityFee> PF   </txPriorityFee>
           <txGasPrice> _ => BF +Int minInt(PF, MF -Int BF) </txGasPrice>
           ...
         </message>
      requires Ghasbasefee << SCHED >>
```

and have it called before `loadTx` 
```
    rule <k> startTx => #computeGasPrice ~> loadTx(...) ... </k>
```

`loadTx` would then remain unchaged, and update the `<gasPrice>` cell with the value from `<txGasPrice>`, which is the `effectiveGasPrice`

PR #1207, opened 2022-03-16
Update repo location

ehildenb:


web/package.json
```diff
@@ -21,7 +21,7 @@
     "k-web-theme": "file:k-web-theme"
   },
   "bugs": {
-    "url": "https://github.com/kframework/evm-semantics/issues"
+    "url": "https://github.com/kframework/runtimeverification/issues"
```
gtrepta, line 24:
```suggestion
    "url": "https://github.com/runtimeverification/evm-semantics/issues"
```

Is this what you meant?

web/package.json
```diff
@@ -21,7 +21,7 @@
...
   },
-  "homepage": "https://github.com/kframework/evm-semantics/#readme"
+  "homepage": "https://github.com/kframework/runtimeverification/#readme"
```
gtrepta, line 26:
```suggestion
  "homepage": "https://github.com/runtimeverification/evm-semantics/#readme"
```

gtrepta (changes requested):
This is probably a good time to update the remote URL for K.

https://github.com/runtimeverification/evm-semantics/blob/7c614ba29536db02cc03ae431435f2de3d4b7fbf/.gitmodules#L6-L9

There will also be outdated urls in the presentations under the `media/` folder, just pointing that out. I assume it's ok to leave those alone.

gtrepta (approved):

PR #1219, opened 2022-03-29
Reorganize kevm_pyk distribution

### LINKED ISSUE #1216: Add pyproject.toml to kevm_pyk

tothtamas28:
Fixes #1216


kevm_pyk/setup.cfg
```diff
@@ -0,0 +1,28 @@
...
+name = kevm_pyk
+version = 0.0.1
+
+[options]
+python_requires = >= 3.8.*
+packages = find:
+package_dir =
+    =src
+install_requires =
+    pyk @ git+https://github.com/runtimeverification/k.git@v5.3.4#subdirectory=pyk
```
ehildenb, line 11:
```suggestion
    pyk @ git+https://github.com/runtimeverification/k.git@${somehow get this from deps/k}#subdirectory=pyk
```

package/test-package.sh
```diff
@@ -28,7 +28,7 @@ kevm run tests/failing/static_callcodecallcodecall_110_OOGMAfter_2_d0g0v0.json -
     || git --no-pager diff --no-index --ignore-all-space -R tests/failing/static_callcodecallcodecall_110_OOGMAfter_2_d0g0v0.json.llvm-out tests/failing/static_callcodecallcodecall_110_OOGMAfter_2_d0g0v0.json.expected
 rm -rf tests/failing/static_callcodecallcodecall_110_OOGMAfter_2_d0g0v0.json.llvm-out
 
-kevm solc-to-k tests/specs/examples/ERC20.sol ERC20 > tests/specs/examples/erc20-bin-runtime.k
+kevm_pyk solc-to-k /usr/bin/lib/haskell/driver-kompiled tests/specs/examples/ERC20.sol ERC20 > tests/specs/examples/erc20-bin-runtime.k
```
ehildenb, line 31: We should still do `kevm solc-to-k ...` instead, so that this argument is not necessary.

The whole point of the `kevm` script is to "know where it's installed" and do any environment setup needed to make other scripts work with fewer arguments.

So this added argument is unecessary.

kevm
```diff
@@ -252,16 +251,6 @@ run_interpret() {
...
-run_solc() {
-    local contract_name json_run_file
-
-    contract_name="$1" ; shift
-    json_run_file="${run_file}.json"
-
-    execute solc --combined-json abi,bin-runtime,storage-layout,hashes "${run_file}" > "${json_run_file}"
-    execute python3 -m kevm_pyk solc-to-k "$@" ${backend_dir}/*-kompiled "${run_file}" "${json_run_file}" "${contract_name}"
-}
-
```
ehildenb, line 264: We should still have this command present in `kevm` runner script, in anything just as a wrapper to `kevm_pyk ...` which invokes it with the correct arguments (like passing it the `*_kompiled` directory appropriately).

kevm_pyk/setup.cfg
tothtamas28: This is not trivial. Created an issue to log this for later: https://github.com/runtimeverification/erc20-verification/issues/389

Makefile
```diff
@@ -449,15 +429,21 @@ tests/%.prove-legacy: tests/%
...
 tests/specs/examples/erc721-spec/haskell/erc721-spec-kompiled/timestamp: tests/specs/examples/erc721-bin-runtime.k
-tests/specs/examples/erc721-bin-runtime.k: tests/specs/examples/ERC721.sol $(KEVM_LIB)/$(haskell_kompiled) $(kevm_pyk_includes) $(PYK_VENV)
-	$(ACTIVATE_PYK) && $(KEVM) solc-to-k $< ERC721 > $@
-
-tests/specs/examples/empty-bin-runtime.k: tests/specs/examples/Empty.sol $(KEVM_LIB)/$(haskell_kompiled) $(kevm_pyk_includes) $(PYK_VENV)
-	$(ACTIVATE_PYK) && $(KEVM) solc-to-k $< Empty > $@
+tests/specs/examples/erc721-bin-runtime.k: tests/specs/examples/ERC721.sol $(KEVM_LIB)/$(haskell_kompiled) $(VENV)
+	$(MAKE) -C ./kevm_pyk venv-prod &&     \
+	. ./kevm_pyk/venv-prod/bin/activate && \
+	$(KEVM) solc-to-k $< ERC721 > $@
```
ehildenb, line 441: I'm pretty sure the first command doesn't need to be run in the same subshell as the later ones? So we should be able to do:

```suggestion
	$(MAKE) -C ./kevm_pyk venv-prod
	. ./kevm_pyk/venv-prod/bin/activate && $(KEVM) solc-to-k $< ERC721 > $@
```

ehildenb: Also, the first command could be it's own target, right?

So that each other target can depend on it directly, and make will only run it once?

tothtamas28: Yes, that sounds right. I'll change this accordingly.

PR #1223, opened 2022-04-08
Fix vat-subui-fail spec

gtrepta:


ehildenb:
@gtrepta you likely need to increases the amount of time that we allocate for testing on this branch.

tests/failing-symbolic.haskell
```diff
@@ -18,9 +18,7 @@ tests/specs/mcd/pot-join-pass-rough-spec.k
 tests/specs/mcd/vat-fold-pass-rough-spec.k
 tests/specs/mcd/vat-fork-diff-pass-rough-spec.k
 tests/specs/mcd/vat-frob-diff-zero-dart-pass-rough-spec.k
-tests/specs/mcd/vat-move-diff-rough-spec.k
 tests/specs/mcd/vat-slip-pass-rough-spec.k
```
ehildenb, line 21:
```suggestion
tests/specs/mcd/vat-move-diff-rough-spec.k
tests/specs/mcd/vat-slip-pass-rough-spec.k
```

tests/specs/mcd/verification.k
```diff
@@ -93,6 +93,43 @@ module LEMMAS-MCD-HASKELL [symbolic, kore]
...
+    // (declare-const x Int)
+    // (declare-const y Int)
+    //
+    // (assert (uint256 x))
+    // (assert (sint256 y))
+    // (assert (<= y 0))
+    // (assert (not (uint256 (- x y))))
+    //
+    // (assert (not (= (chop (- x y)) (- x (+ y pow256)))))
+    // (check-sat)
```
ehildenb, line 126: What was the output this supplied? It should say `unsat`, then we know that the lemma is true.

ehildenb (approved):

tests/specs/mcd/verification.k
ehildenb: I double-checked that it gave unsat on my machine.

gtrepta: Yep, it was unsat.

Issue #1165, opened 2021-11-29
Execution of all test proofs in KEVM (Profiling)


ACassimiro:
To perform the execution of all test proofs available in the KEVM repository (a process referred to as 'profiling'), first you should follow the instructions available in the [README.md file of this repository](https://github.com/kframework/evm-semantics#installingbuilding) for the installation of the KEVM project in your machine. 

Once you're done with the installation of KEVM, install [GNU-Time](https://www.gnu.org/software/time/). Please, test your installation of GNU-Time to avoid silent crashes when running the profiling.

With KEVM and GNU-time ready, create the following scripts in the root of your KEVM repository:

- Script name: `profile`

```
#!/usr/bin/env bash
        
set -euo pipefail

test_log="$1" ; shift
mkdir -p $(dirname "$test_log")
exit_status='0'
`which time` --quiet --format '%x %es %Us %Ss %MKB %C' --output "$test_log" --append \
    "$@" &>/dev/null \
    || exit_status="$?"
exit "$exit_status"
```

When running the profiling with haskell backend, to execute all proofs, including the ones that are currently failing in this backend, please erase all the contents of the file `evm-semantics/tests/failing-symbolic.haskell`.

Finally, to run the profiling with the haskell backend, run the following command in your KEVM root directory:

`make test-prove KEVM='./profile log timeout 3000 kevm' -j6 -k`

This command will run all proofs in the nested directories of `test/specs/`, running at most 6 proofs at a time and with a maximum execution time per proof of 3000 seconds. It is important to highlight that the profiling is a memory intensive and time consuming process. For machines with less than 64GB of RAM, it is recommended to run a lesser number of simultaneous jobs, which can be achieved by changing the value 6 in the parameter `-j6` to the desired amount of maximum simultaneous jobs to be executed during the profiling. Also, as to avoid losing the output the profiling, you can redirect the output of the command above to a file named `output.txt` by adding `&> output.txt` at the end of the command.

For more information about the inner workings of the profiling execution, please read [this project's Makefile](https://github.com/kframework/evm-semantics/blob/master/Makefile).

-----

Setting up `kore-repl`:

- Script name: `print`

```
kore-print --definition tests/specs/mcd/verification/haskell /dev/stdin
echo
```

- Script name: `kscript`

```
alias konfig = config | bash print
```

Run `kevm prove` with `--debug-script kscript` (or just run `kprove` with thi sargument).

ehildenb:
[...]

PR #1225, opened 2022-04-13
Skip slow-running proofs on CI

ehildenb:
Currently we are taking close to 3.5 hours on CI for the proofs, and taking quite a bit of CI resources to do so.

This removes the longest running proofs (> 900s on my machine), and decreases parallelism.


radumereuta:
Do you know if this is time spent in the front end or the backend?

ehildenb:
Mostly backend @radumereuta . Most proofs only spend up to 60s in the frontend.

ehildenb:
Feel free to profile the tests yourself @radumereuta : https://github.com/runtimeverification/evm-semantics/issues/1165

You'll have to adjust those instructions to include the verbose information which you want, but that should be possible.

ehildenb:
I've now disabled all that take >600s on my machine.

radumereuta:
Can I still run them with `$ make test-prove`? I would like to still be able to test the front end with them.

ehildenb:
Only if you delete them from `tests/slow.haskell` and `tests/failing-symbolic.haskell`

gtrepta (approved):

PR #1241, opened 2022-05-04
Switch --directory => --definition flag

ehildenb:
~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/1240~

Fortunately KEVM was already using a unique wrapper directory for every definition it `kompile`d anyway, so this only needs us to remove references to teh `*-kompiled` directories (instead preferring the unique wrapper directory for every definition), and to switch to using the `--definition` flag everywhere.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -445,12 +445,12 @@ tests/specs/examples/erc721-bin-runtime.k: tests/specs/examples/ERC721.sol $(KEV
 tests/specs/examples/empty-bin-runtime.k: tests/specs/examples/Empty.sol $(KEVM_LIB)/$(haskell_kompiled) kevm-pyk-venv
 	. ./kevm_pyk/venv-prod/bin/activate && $(KEVM) solc-to-k $< Empty > $@
 
-tests/gen-spec/mcd-spec.k.check: tests/gen-spec/timestamp kevm-pyk-venv
-	. ./kevm_pyk/venv-prod/bin/activate && $(KEVM) gen-spec tests/gen-spec MCD-SPEC > $@.out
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -528,36 +527,36 @@ prove_mcd_tests          := $(filter-out $(prove_skip_tests), $(wildcard $(prove
...
-                      tests/specs/functional/merkle-spec/haskell/merkle-spec-kompiled/timestamp                    \
-                      tests/specs/functional/storageRoot-spec/haskell/storageRoot-spec-kompiled/timestamp          \
-                      tests/specs/mcd/functional-spec/haskell/functional-spec-kompiled/timestamp                   \
-                      tests/specs/mcd/functional-spec/java/functional-spec-kompiled/timestamp                      \
-                      tests/specs/mcd/verification/haskell/verification-kompiled/timestamp                         \
-                      tests/specs/mcd/verification/java/verification-kompiled/timestamp                            \
-                      tests/specs/opcodes/evm-optimizations-spec/haskell/evm-optimizations-spec-kompiled/timestamp \
-                      tests/specs/opcodes/verification/java/verification-kompiled/timestamp
+provex_definitions :=                                                              \
+                      tests/specs/benchmarks/functional-spec/haskell/timestamp     \
```
radumereuta, line 531: Maybe:
`tests/specs/benchmarks/functional-spec/haskell-kompiled/timestamp`

radumereuta (approved):
I would like it if we kept the `*-kompiled` termination as a convention.
- It has a strong visual cue to tell you that it's the directory generated by kompile.
- it's easy to add to .gitignore
- it's easy to remove/make clean

But feel free to merge if you have reasons not to though.

PR #1246, opened 2022-05-17
vow-flog-fail-rough-spec passing

goodlyrottenapple:


ehildenb:
@goodlyrottenapple how long does this proof take on your machine? If it's well over 600s, then we may want to add it to `tests/slow*` list so it's not run on CI.

goodlyrottenapple:
@ehildenb no it was actually pretty fast... "only" around 350s

ehildenb (approved):

PR #1250, opened 2022-05-18
Modify solc_compile to use the solc standard JSON interface

tothtamas28:
Fixes https://github.com/runtimeverification/erc20-verification/issues/401

### DIFF
--- a/kevm_pyk/src/kevm_pyk/solc_to_k.py
+++ b/kevm_pyk/src/kevm_pyk/solc_to_k.py
@@ -1,9 +1,11 @@
 import functools
 import json
-import subprocess
+import logging
 from pathlib import Path
-from typing import Any, Dict, List
+from subprocess import CalledProcessError
+from typing import Any, Dict, Final, List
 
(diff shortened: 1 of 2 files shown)


tests/slow.haskell
```diff
@@ -353,6 +353,7 @@ tests/specs/erc20/ds/transfer-success-2-spec.k
 tests/specs/erc20/hkg/transferFrom-failure-1-spec.k
 tests/specs/erc20/hkg/transferFrom-success-1-spec.k
 tests/specs/examples/erc20-spec.md
+tests/specs/examples/erc721-spec.md
```
ehildenb, line 356: This should probably just be in `tests/failing.haskell` instead. Is it actually slow? (passing on your machine, timing out on CI) If it's failing on CI, let's just mark it failing.

ehildenb (approved):

tests/slow.haskell
tothtamas28: Done.

PR #1251, opened 2022-05-19
flipper-bids-pass-rough-spec.k passing

goodlyrottenapple:
Due to https://github.com/runtimeverification/haskell-backend/issues/3060, the definition and simplification rule for `#WordPackAddrUInt48UInt48` was tweaked slightly to allow the proof in `flipper-bids-pass-rough-spec.k` to pass.

### DIFF
--- a/tests/specs/mcd/word-pack.k
+++ b/tests/specs/mcd/word-pack.k
@@ -53,6 +53,15 @@ module WORD-PACK-HASKELL [kore]
     andBool #rangeUInt(256, ADDR_UINT48_UINT48)
       [simplification]
 
+
+    rule     #WordPackAddrUInt48UInt48Prop(ADDR_UINT48_UINT48, ADDR, UINT48_1, UINT48_2)
+          => true
+     ensures ADDR     ==Int maxUInt160 &Int  ADDR_UINT48_UINT48
+     andBool UINT48_1 ==Int maxUInt48  &Int (ADDR_UINT48_UINT48 /Int pow160)
+     andBool UINT48_2 ==Int maxUInt48  &Int (ADDR_UINT48_UINT48 /Int pow208)
+     andBool #rangeUInt(256, ADDR_UINT48_UINT48)
+       [simplification]
+
     rule    UINT48_UINT48 ==Int #WordPackUInt48UInt48(UINT48_1, UINT48_2)
          => UINT48_1 ==Int maxUInt48 &Int  UINT48_UINT48
     andBool UINT48_2 ==Int maxUInt48 &Int (UINT48_UINT48 /Int pow48)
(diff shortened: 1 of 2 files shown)


tests/specs/mcd/word-pack.k
```diff
@@ -78,6 +87,9 @@ module WORD-PACK-COMMON
     syntax Int ::= #WordPackUInt48UInt48     (       Int , Int ) [function, no-evaluators, smtlib(WordPackUInt48UInt48)]
                  | #WordPackAddrUInt48UInt48 ( Int , Int , Int ) [function, no-evaluators, smtlib(WordPackAddrUInt48UInt48)]
                  | #WordPackAddrUInt8        (       Int , Int ) [function, no-evaluators, smtlib(WordPackAddrUInt8)]
+
+    syntax Bool ::= #WordPackAddrUInt48UInt48Prop ( Int , Int , Int, Int ) [function, smtlib(WordPackAddrUInt48UInt48Bool)]
+
  // -----------------------------------------------------------------------------------------------------------------
```
ehildenb, line 93:
```suggestion
                 | #WordPackAddrUInt8        (       Int , Int ) [function, no-evaluators, smtlib(WordPackAddrUInt8)]
 // -----------------------------------------------------------------------------------------------------------------
 
    syntax Bool ::= #WordPackAddrUInt48UInt48Prop ( Int , Int , Int, Int ) [function, smtlib(WordPackAddrUInt48UInt48Bool)]
 // -----------------------------------------------------------------------------------------------------------------------
```

tests/specs/mcd/word-pack.k
```diff
@@ -53,6 +53,15 @@ module WORD-PACK-HASKELL [kore]
...
 
+
+    rule     #WordPackAddrUInt48UInt48Prop(ADDR_UINT48_UINT48, ADDR, UINT48_1, UINT48_2)
+          => true
+     ensures ADDR     ==Int maxUInt160 &Int  ADDR_UINT48_UINT48
+     andBool UINT48_1 ==Int maxUInt48  &Int (ADDR_UINT48_UINT48 /Int pow160)
+     andBool UINT48_2 ==Int maxUInt48  &Int (ADDR_UINT48_UINT48 /Int pow208)
+     andBool #rangeUInt(256, ADDR_UINT48_UINT48)
+       [simplification]
+
```
ehildenb, line 64: Can you transform all the other uses of `#WordPack...` to this way as well?

ehildenb:
You likely will have to disable this proof on the java backend (via `tests/failing-symbolic.java`), because the new abstraction only works for Haskell backend.

goodlyrottenapple:
Should we keep both versions of `#WordPack...` and `#WordPack...Prop` and have `ADDR ==Int #WordPack... andBool #WordPack...Prop(ADDR, ...)` in all the claims that mention it?

ehildenb:
Does the spec pass locally (on both backends, with `make tests/specs/mcd/flipper-bit-pass-rough-spec.k.prove TEST_SYMBOLIC_BACKEND=java` and `... TEST_SYMBOLIC_BACKEND=haskell`) if you do this? If so, then yes, let's go with that solution for now.

ehildenb:
Also, you should make the same adjustments to all the occurances of `WordPack*` (there are 3 types of `WordPack*`), and see if any more proofs pass because of that.

PR #1257, opened 2022-05-25
Drop kprovex

ehildenb:
This PR:

- Removes references to the old `kprovex`.
- Switches `erc721` spec to be proven with Haskell backend, not Java backend, because now Java backend is consistently failing on that proof on CI.


Baltoli:
The `prove` / `provex` changes here look fine to me.

Baltoli (approved):

PR #1258, opened 2022-05-25
Add pyk runner for prover and kompile, rule profiling

ehildenb:
~Blocked on: https://github.com/runtimeverification/k/pull/2624~

This adds an argument `--pyk` which switches to use `pyk` to do proving/kompiling with.

- A new class `KEVM` is factored out, which has function `_patch_symbol_table(...)` and `empty_config(...)` built in.
- Option `--pyk` uses `kevm_pyk` to do proving, should pass a test on CI, and produce rule profile reports.
- Option `--pyk` uses `kevm_pyk` to do kompiling, and should pass a test on CI.
- `kevm_*` helper methods are moved to `kevm.py` file.
- Adds a test of `--pyk` for both `kompile` and `kprove` by building the Haskell backend and Bihu proofs definitions using kevm_pyk, and checking that the Bihu functional spec correctly generates a rule profile with --pyk.

This enables using kevm_pyk library directly for kprove/kompile in the summarizer.

### DIFF
--- a/kevm
+++ b/kevm
@@ -49,6 +49,7 @@ export KLAB_OUT="${KLAB_OUT:-~/.klab}"
 # -------
 
 run_kevm_pyk() {
+    local pyk_args
     pyk_args=(--definition "${backend_dir}")
     pyk_args+=(-I "${INSTALL_INCLUDE}/kframework" -I "${plugin_include}/kframework")
     ! ${debug} || pyk_args+=(--debug)
@@ -139,7 +140,7 @@ run_kast() {
 
 run_prove() {
     local def_module run_dir proof_args bug_report_name \
-          eventlog_name kprove omit_cells omit_labels klab_log pyk_args
+          eventlog_name kprove omit_cells omit_labels klab_log
 
     check_k_install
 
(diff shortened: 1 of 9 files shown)


kevm
```diff
@@ -52,8 +53,8 @@ run_kevm_pyk() {
     local pyk_args
     pyk_args=(--definition "${backend_dir}")
     pyk_args+=(-I "${INSTALL_INCLUDE}/kframework" -I "${plugin_include}/kframework")
-    ! ${debug} || pyk_args+=(--debug)
-    execute python3 -m kevm_pyk "${pyk_args[@]}" "$@"
```
nishantjr: Did we drop `debug` intentionally?

kevm_pyk/src/kevm_pyk/kevm.py
```diff
@@ -0,0 +1,40 @@
...
+from .utils import build_empty_configuration_cell
+
+
+class KEVM(KProve):
+
+    def __init__(self, kompiled_directory, main_file_name=None, use_directory=None):
+        super().__init__(kompiled_directory, main_file_name=None, use_directory=use_directory)
+        KEVM._patch_symbol_table(self.symbol_table)
+
+    def empty_config(self, top_cell: KSort = KSort('GeneratedTopCell')) -> KInner:
```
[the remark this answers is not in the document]
ehildenb: done, chose `config` since it's used all over

kevm
ehildenb: Yes. Use `--verbose` for the timing information. `--debug` isn't used by the python tools anymore.

nishantjr: Aah. I had missed the `set -x` line and thought `debug` was unused.

PR #1262, opened 2022-06-07
Fixes for Clang/LLVM warnings

goodlyrottenapple:
* Add `-Wno-unused-result` to `CMakeLists.txt` to suppress error due to https://bugs.llvm.org/show_bug.cgi?id=51228
* Rename `accounts` to `accounts_ptr` and `runvm` to `runvm_ptr` due to 
```
/home/sam/git/evm-semantics-nix-flake/node/vm/vm.h:55:10: error: declaration of 'runvm* accesslist::runvm' changes meaning of 'runvm' [-fpermissive]
   55 |   runvm* runvm;
      |          ^~~~~
/home/sam/git/evm-semantics-nix-flake/node/vm/vm.h:33:8: note: 'runvm' declared here as 'struct runvm'
   33 | struct runvm {
      |        ^~~~~
In file included from /home/sam/git/evm-semantics-nix-flake/node/vm/vm.cpp:7:
/home/sam/git/evm-semantics-nix-flake/node/vm/kevm/semantics.h:44:13: error: declaration of 'accounts* tx_result::accounts' changes meaning of 'accounts' [-fpermissive]
   44 |   accounts* accounts;
      |             ^~~~~~~~
In file included from /home/sam/git/evm-semantics-nix-flake/node/vm/vm.cpp:5:
/home/sam/git/evm-semantics-nix-flake/node/vm/vm.h:28:8: note: 'accounts' declared here as 'struct accounts'
   28 | struct accounts {
      |        ^~~~~~~~
```

### DIFF
--- a/cmake/node/CMakeLists.txt
+++ b/cmake/node/CMakeLists.txt
@@ -76,7 +76,7 @@ target_include_directories(kevm-vm
 	PUBLIC ${K_LIB}/../include/kllvm)
 target_compile_options(kevm-vm
 	PUBLIC $ENV{LLVM_KOMPILE_OPTS}
-	PUBLIC -Wall -Werror)
+	PUBLIC -Wall -Werror -Wno-unused-result)
 target_link_libraries(kevm-vm
 	ff protobuf gmp ${LINK_PROCPS} cryptopp secp256k1 crypto)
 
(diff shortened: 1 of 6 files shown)


gtrepta:
What's the error and what version of llvm is this happening on? It seems kind of strange that we need to cast fread to void but fwrite is fine, this probably needs to be addressed in some other way.

goodlyrottenapple:
> What's the error and what version of llvm is this happening on? It seems kind of strange that we need to cast fread to void but fwrite is fine, this probably needs to be addressed in some other way.

The error is the same as here https://bugs.llvm.org/show_bug.cgi?id=51228

Actually there are a few other changes that I've had to make to make the code compile in LLVM/Clang 11.1
I suspect there might be some config defaults that are different tho, because I get similar errors from other versions of clang too

gtrepta (approved):
LGTM. Feel free to let me know if you add any more changes.

PR #1266, opened 2022-06-22
flipper-bids-pass-rough-spec fix

PetarMax:
Alternative way of getting flipper-bids-pass-rough-spec.k to work (cf. #1251), using simplifications only.


PetarMax:
The CI does not seem to like the simplification at the ML-level. In particular the Java backend is saying:

`[Error] Compiler: Simplification rules need to be function/functional like.` in reference to the simplification

```k
   // Deconstruction of #WordPackAddrUInt48UInt48 at the meta-logic level
    rule    { #WordPackAddrUInt48UInt48(A:Int, B:Int, C:Int) #Equals D:Int }
         => { A #Equals maxUInt160 &Int  D              }
       #And { B #Equals maxUInt48  &Int (D /Int pow160) }
       #And { C #Equals maxUInt48  &Int (D /Int pow208) }
       #And { #rangeUInt(256, D) }
      [simplification,anywhere]
```

 in `tests/specs/mcd/flipper-bids-pass-rough-spec.k`.

PR #1268, opened 2022-06-23
additional simplifications for buffers and buffer concatenation

PetarMax:
Several basic simplifications for buffers and buffer concatenation that allow [end-cash-pass-rough-spec.k](https://github.com/runtimeverification/evm-semantics/blob/master/tests/specs/mcd/end-cash-pass-rough-spec.k) and
[end-pack-pass-rough-spec.k](https://github.com/runtimeverification/evm-semantics/blob/master/tests/specs/mcd/end-pack-pass-rough-spec.k) to pass:

### DIFF
--- a/tests/specs/bihu/verification.k
+++ b/tests/specs/bihu/verification.k
@@ -65,7 +65,7 @@ module VERIFICATION
   // #accumulatedReleasedTokens
   // ###############################
 
-    syntax Int ::= "#accumulatedReleasedTokens" "(" Int "," Int "," Int "," Int ")"  [function]
+    syntax Int ::= "#accumulatedReleasedTokens" "(" Int "," Int "," Int "," Int ")"  [function, no-evaluators]
  // -------------------------------------------------------------------------------------------
     // proved manually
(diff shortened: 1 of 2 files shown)


tests/specs/bihu/verification.k
```diff
@@ -65,7 +65,7 @@ module VERIFICATION
   // #accumulatedReleasedTokens
   // ###############################
 
-    syntax Int ::= "#accumulatedReleasedTokens" "(" Int "," Int "," Int "," Int ")"  [function]
+    syntax Int ::= "#accumulatedReleasedTokens" "(" Int "," Int "," Int "," Int ")"  [function, no-evaluators]
  // -------------------------------------------------------------------------------------------
```
ehildenb, line 69:
```suggestion
    syntax Int ::= "#accumulatedReleasedTokens" "(" Int "," Int "," Int "," Int ")"  [function, no-evaluators]
 // ----------------------------------------------------------------------------------------------------------
```

ehildenb:
@PetarMax please also remove the now passing tests from the `tests/failing-symbolic.haskell` list. Either they need to be (i) removed from that list so they are run on CI, or (ii) added to the `tests/slow.haskell` list so that they are not run on CI (if they take longer than 10 minutes on your machine).

PetarMax:
Done, but CI was failing because it was not recognising `#And`. I've tried to address that in the latest commit.

PR #1277, opened 2022-07-09
Handle more arguments with pyk prover

ehildenb:
This allows the pyk prover to handle the `--bug-report` and `--debug-equations ...` convenience arguments.


nishantjr (approved):
Looks good!

PR #1284, opened 2022-07-15
Include INFINITE-GAS into EDSL

ehildenb:
Module `INFINITE-GAS` is used so often for any verification, that it is promoted to part of `EDSL`, so that it comes with the default Haskell backend that KEVM ships.

### DIFF
--- a/tests/specs/foundry/foundry-spec.k.check.expected
+++ b/tests/specs/foundry/foundry-spec.k.check.expected
@@ -66,7 +66,7 @@ module FOUNDRY-SPEC
                    ( 0 => ?_PC_CELL_acd26fc5 )
                  </pc>
                  <gas>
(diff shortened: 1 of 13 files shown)


tests/specs/foundry/foundry-spec.k.check.expected
```diff
@@ -66,7 +66,7 @@ module FOUNDRY-SPEC
                    ( 0 => ?_PC_CELL_acd26fc5 )
                  </pc>
                  <gas>
-                   ( infGas ( _VGAS ) => ?_GAS_CELL_acd26fc5 )
+                   ( #gas ( _VGAS ) => ?_GAS_CELL_acd26fc5 )
```
ehildenb, line 69: You can see here that unparsing of the generated specification is fixed now, because `#gas` definition is included in default builds of KEVM now.

tothtamas28 (approved):

PR #1285, opened 2022-07-18
Fix up unparsing when generating specifications

ehildenb:
This PR enables re-building the unparser after adding the productions with the K helpers defined for a contract.
That way, we are not manually patching the symbol table to unparse these generated productions.

- `contract_to_k` and `gen_claims_for_contract` are lowered to take an `empty_config` as input, instead of a entire `KEVM`, which makes it easier to use them in more places (for a later PR).
- The typing around some of the generator functions is tightened up, to help catch errors earlier: `_extract_function_sentences`, `generate_function_sentences`.
- `_parseByteStack` is moved to a static method `KEVM.parse_bytestack`.
- Factors out functions which should be upstreamed in `utils.py`: `KPrint_make_unparsing` and `KDefinition_module_names` and `KDefinition_empty_config`.
- Adds `klabel`s to some of the generated functions in the specification, for proper unparsing.
- Removes the symbol table updates that were used for unparsing, in favor of rebuilding an unparser after adding the generated productions.


ehildenb:
Going to break this into several PRs instead.

PR #1287, opened 2022-07-19
Nix flake

### LINKED ISSUE #1248: Create Nix derivation for KEVM
Currently blocked https://github.com/runtimeverification/ksummarize/issues/33

goodlyrottenapple:
This PR adds a nix expression to build kevm and the associated kevm python library.

Closes #1248


ehildenb:
@goodlyrottenapple  can you set up a GitHub workflow to actually build/run this Nix target on CI? Should be pretty similar to the K repo one, or Haskell backend repo one.

ehildenb:
What about `flake.lock` and `flake.nix`? Can we move those to subdirectory too? If not no worries, but if it's possibel I would prefer it.

goodlyrottenapple:
no, those have to be in the root folder, otherwise nix wont find them

flake.nix
```diff
@@ -0,0 +1,225 @@
...
+            enableParallelBuilding = true;
+
+            buildInputs = [ final.kevm prev.which prev.git ];
+
+            buildPhase = ''
+              mkdir -p tests/ethereum-tests/LegacyTests
+              cp -rv ${ethereum-tests}/* tests/ethereum-tests/
+              cp -rv ${ethereum-legacytests}/* tests/ethereum-tests/LegacyTests/
+              chmod -R u+w tests
+              APPLE_SILICON=${if prev.stdenv.isAarch64 && prev.stdenv.isDarwin then "true" else "false"} NIX=true package/test-package.sh
```
ehildenb, line 151: This is pretty weird, what's going on here? Clone of this submodule happens out of tree and needs to  be copied in? Note that this submodule  is only used for testing, not building.

package/nix/kevm.patch
```diff
@@ -0,0 +1,14 @@
...
++++ b/kevm
+@@ -96,9 +96,6 @@ run_kompile() {
+                    if [[ "$(uname -s)" == 'Linux' ]]; then
+                        kompile_opts+=( -ccopt -lprocps )
+                    elif [[ "$(uname -s)" == 'Darwin' ]]; then
+-                       openssl_root="$(brew --prefix openssl)"
+-                       brew_root="$(brew --prefix)"
+-                       kompile_opts+=( -ccopt -I${brew_root}/include -ccopt -L/${brew_root}/lib -ccopt -I${openssl_root}/include -ccopt -L${openssl_root}/lib )
+                        kompile_opts+=( -ccopt -I${libcryptopp_dir}/include -ccopt -L/${libcryptopp_dir}/lib )
+                    fi
```
ehildenb, line 13: These lines are there to get around some issues with MacOS. Are those issues not present on Nix MacOS?

package/test-package.sh
```diff
@@ -16,22 +16,29 @@ kevm kast tests/interactive/log3_MaxTopic_d0g0v0.json kast  --backend llvm > tes
 git --no-pager diff --no-index --ignore-all-space -R tests/interactive/log3_MaxTopic_d0g0v0.json.parse-out tests/interactive/log3_MaxTopic_d0g0v0.json.parse-expected
 rm -rf tests/interactive/log3_MaxTopic_d0g0v0.json.parse-out
 
-kevm run tests/failing/static_callcodecallcodecall_110_OOGMAfter_2_d0g0v0.json --backend llvm \
-    --mode NORMAL --schedule BERLIN --chainid 1                                               \
-    > tests/failing/static_callcodecallcodecall_110_OOGMAfter_2_d0g0v0.json.llvm-out          \
-    || git --no-pager diff --no-index --ignore-all-space -R tests/failing/static_callcodecallcodecall_110_OOGMAfter_2_d0g0v0.json.llvm-out tests/failing/static_callcodecallcodecall_110_OOGMAfter_2_d0g0v0.json.expected
-rm -rf tests/failing/static_callcodecallcodecall_110_OOGMAfter_2_d0g0v0.json.llvm-out
+# This test currently segfaults on M1 Macs
+if [ "${APPLE_SILICON:-false}" == "false" ]; then
```
ehildenb, line 20:
```suggestion
if ${APPLE_SILICON:-false}; then
```

`false` is a command on most systems that can be used directly for conditionals. Same with `true`.

package/test-package.sh
```diff
@@ -16,22 +16,29 @@ kevm kast tests/interactive/log3_MaxTopic_d0g0v0.json kast  --backend llvm > tes
...
+
+kevm solc-to-k tests/specs/examples/ERC20.sol ERC20 --main-module ERC20-VERIFICATION > tests/specs/examples/erc20-bin-runtime.k
 kevm kompile --backend haskell tests/specs/examples/erc20-spec.md \
     --definition tests/specs/examples/erc20-spec/haskell          \
     --main-module VERIFICATION                                    \
     --syntax-module VERIFICATION                                  \
     --concrete-rules-file tests/specs/examples/concrete-rules.txt \
     --verbose
+# This test is probably too long for public Github runner and currently seems broken
+if [ "${NIX:-false}" == "false" ]; then
```
ehildenb, line 37:
```suggestion
# This test is probably too long for public Github runner and currently seems broken
if ! ${NIX:-false}; then
```

We shouldn't be using public github runners either way though.

package/nix/kevm.patch
goodlyrottenapple: no, nix does not use any globally installed stuff from other package managers like brew, so we have to remove those lines otherwise the compiler/linker complains that `brew_root` is missing...

ehildenb (approved):

PR #1288, opened 2022-07-19
Added some foundry simple tests

lisandrasilva:


ehildenb:
Can you back out the change to `k-web-theme` submodule? I don't think that needs to be updated.

Also, do you actually need the `foundry-std` submodule either? I think probably not... But I'm not sure. If you do need it, then `foundry.toml` probably needs to know about where it lives?

lisandrasilva:
I didn't change anything in `k-web-theme` submodule.

Regarding the `forge-std` submodule, the command `forge build` was not working because it couldn't find the `Test.sol`. I had to run `forge install foundry-rs/forge-std` and then it worked.

JuanCoRo:
@lisandrasilva if you do `forge init yourProject`, foundry should automatically install `forge-std` for you. Was this not the case for you? Maybe you are using an older version in which this didn't happen? To update it, you need to follow the installation instructions again.

@ehildenb the `forge-std` is an improvement on top of the `DSTest` contract. Improves the experience when using cheat codes and error messages. So, as far as I know, it's the intended way of using foundry's capabilities when writing tests.
Also, whenever you install a public repo via `forge install repo`, foundry knows where to get it if you import it as `repo/contracts/foo.sol`, for instance. I don't know where this info is stored, but it does not need to be in the `foundry.toml` nor the `remappings.txt` files. Maybe foundry just looks in the `lib` folder to see if it finds the file being imported.

lisandrasilva:
You are right! I tried again on a freshly installed KEVM in Ubuntu and the `forge build` and `forge test` worked without having to install `forge-std`. I will remove the submodule.

ehildenb:
Let's call `forge test` as well:

```
tests/foundry/out:
	cd $(dir $@) && forge build --extra-output storageLayout --extra-output abi --extra-output evm.methodIdentifiers --extra-output evm.deployedBytecode.object
```

Here we can have another line `cd $(dir $@) && forge test`.

tests/foundry/test/FfiTest.t.sol
```diff
@@ -0,0 +1,21 @@
...
+        string[] memory inputs = new string[](3);
+        inputs[0] = "echo";
+        inputs[1] = "-n";
+        // ABI encoded "gm", as a string
+        inputs[2] = "0x00000000000000000000000000000000000000000000000000000000000000200000000000000000000000000000000000000000000000000000000000000002676d000000000000000000000000000000000000000000000000000000000000";
+
+        bytes memory res = vm.ffi(inputs);
+        string memory output = abi.decode(res, (string));
+        assertEq(output, "gm");
+    }
```
ehildenb, line 20:
```
vm.envSet('FOO', 'true');
result = ffi.callScript('./myscript');
assertEq(result, 'elephant');

vm.envSet('FOO', false);
result = ffi.callScript('./myscript');
assertEq(result, 'horse');
```

myscript 

```
if ${FOO}; then
   echo elephant
else;
  echo horse
  fi
```

tests/foundry/foundry.toml
```diff
@@ -2,3 +2,5 @@
 src = 'src'
 out = 'out'
 extra_output = ['storageLayout', 'abi', 'evm.methodIdentifiers', 'evm.deployedBytecode.object']
+rpc_endpoints = { optimism = "https://optimism.alchemyapi.io/v2/...", mainnet = "${RPC_MAINNET}" }
+
```
ehildenb, line 6: Do we need these rpc endpoints set for the tests, or just some of them? If it's just some of them, maybe we can somehow set it only for those tests?

lisandrasilva: We can delete the `mainnet` entry but we have to update the test. Do you want me to do that?

ehildenb (approved):

PR #1290, opened 2022-07-20
Less hacky unparsing

ehildenb:
This PR introduces functionality to adjust the symbol table automatically after we have built up an extended definition in `solc-to-k` and `foundry-to-k`.

- `EDSL` becomes the main entry point for the haskell backend, and includes module `BIN-RUNTIME` by default. This reflect the fact that we are never using the Haskell backend without this functionality anyway.
- Some formatting improvements to `solc_to_k.py`.
- Method `KDefinition_module_names` is factored out to get all the module names of a given KDefinition.
- Method `KPrint_make_unparsing` is factored out, which makes the most generic unparsing module possible for a given definition, and even allows extending it with new modules.
- The symbol-table patching we were using after `contract_to_k` and `gen_spec_modules` can now be removed in favor of principled unparser generation using `KPrint_make_unparsing`.
- Many places where we were building terms directly using their KLabels are factored out into static methods of KEVM class for more rigor and easier to read code.

### DIFF
--- a/kevm_pyk/src/kevm_pyk/kevm.py
+++ b/kevm_pyk/src/kevm_pyk/kevm.py
@@ -161,6 +161,10 @@ class KEVM(KProve):
         return KApply('#binRuntime', [c])
 
     @staticmethod
+    def hashed_location(compiler: str, base: KInner, offset: KInner) -> KApply:
+        return KApply('#hashedLocation(_,_,_)_HASHED-LOCATIONS_Int_String_Int_IntList', [stringToken(compiler), base, offset])
+
+    @staticmethod
     def abi_calldata(name: str, args: List[KInner]) -> KApply:
         token: KInner = stringToken(name)
         return KApply('#abiCallData(_,_)_EVM-ABI_ByteArray_String_TypedArgs', [token] + args)
@@ -176,6 +180,10 @@ class KEVM(KProve):
     @staticmethod
     def abi_bool(b: KInner) -> KApply:
         return KApply('#bool(_)_EVM-ABI_TypedArg_Int', [b])
+
+    @staticmethod
+    def abi_type(type: str, value: KInner) -> KApply:
+        return KApply('abi_type_' + type, [value])
 
     @staticmethod
     def empty_typedargs() -> KApply:
--- a/kevm_pyk/src/kevm_pyk/kevm.py
+++ b/kevm_pyk/src/kevm_pyk/kevm.py
@@ -113,6 +113,10 @@ class KEVM(KProve):
         return KApply('pow256_EVM-TYPES_Int', [])
 
     @staticmethod
+    def range_uint8(i: KInner) -> KApply:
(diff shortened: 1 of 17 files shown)


kevm_pyk/src/kevm_pyk/utils.py
```diff
@@ -17,6 +21,24 @@
     splitConfigFrom,
     substitute,
 )
+from pyk.ktool import KPrint
+from pyk.ktool.kprint import build_symbol_table
+from pyk.utils import hash_str
+
+
+def KPrint_make_unparsing(_self: KPrint, extra_modules: List[KFlatModule] = []) -> KPrint:
```
nishantjr, line 29:
```suggestion
def KPrint_add_unparsing_module(_self: KPrint, extra_modules: List[KFlatModule] = []) -> KPrint:
```

nishantjr (changes requested):
Looks good. Rename suggested.

kevm_pyk/src/kevm_pyk/kevm.py
```diff
@@ -112,6 +112,10 @@ def jump_applied(pc: KInner) -> KApply:
     def pow256() -> KApply:
         return KApply('pow256_EVM-TYPES_Int', [])
 
+    @staticmethod
+    def range_uint8(i: KInner) -> KApply:
+        return KApply('#rangeUInt(_,_)_EVM-TYPES_Bool_Int_Int', [intToken(8), i])
```
tothtamas28, line 117: Consider making these static methods top-level functions of module `kevm_pyk.kevm` (or maybe some other module, say, `kevm_pyk.prelude`).

kevm_pyk/src/kevm_pyk/utils.py
```diff
@@ -17,6 +21,24 @@
...
+    modules = list(_self.definition.modules) + extra_modules
+    main_module = KFlatModule('UNPARSING', [], [KImport(_m.name) for _m in modules])
+    defn = KDefinition('UNPARSING', [main_module] + modules)
+    kprint = KPrint(str(_self.kompiled_directory))
+    kprint.definition = defn
+    kprint.symbol_table = build_symbol_table(kprint.definition, opinionated=True)
```
tothtamas28, line 35: We should probably change `KPrint` to take a `KDefinition` in the first place to avoid rebuilding the symbol table (and to factor out IO). What do you think?

tothtamas28 (approved):

kevm_pyk/src/kevm_pyk/utils.py
ehildenb: It can be renamed when it's upstreamed. I'm sure it will change and get less hacky over time.

kevm_pyk/src/kevm_pyk/utils.py
ehildenb: Issue made: https://github.com/runtimeverification/k/issues/2740

PR #1297, opened 2022-07-25
Add storage and methods to foundry test

ehildenb:
Turns out Foundry returns less information than the Solidity compiler by default: https://github.com/foundry-rs/foundry/issues/2462

This adjusts the generation of `contract_to_k` to compensate.

- Inline `forge build` options from `Makefile` to `foundry.toml`.
- Allow `forge config --fix` to adjust the `foundry.toml`.
- Separate `build_claim` which returns a `KClaim` instead of a `KRule` and specializes arguments (from `build_rule`).
- Removes the option `--no-storage-slots` from `solc-to-k` and `foundry-to-k`.
- Adds some fields and methods to the contract used for testing `foundry-to-k`.


ehildenb:
Still need to figure out how to extract the storage slot information given Foundry's format.

kevm_pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -182,8 +180,10 @@ def recur(syntax, lhs, rhs, var_idx, type_name):
             return recur_value(syntax, lhs, rhs, type_label)
 
         if encoding == 'mapping':
-            key_type_name = type_dict['key']
-            value_type_name = type_dict['value']
+            # TODO: Make less hacky once this is addressed: https://github.com/foundry-rs/foundry/issues/2462
+            type_contents = '('.join(type_name.split('(')[1:])[0:-1]
+            key_type_name, *rest = type_contents.split(',')
+            value_type_name = ','.join(rest)
             return recur_mapping(syntax, lhs, rhs, var_idx, key_type_name, value_type_name)
```
ehildenb, line 187: This is to accomodate both `solc` output (which does already compute this for us as `key` and `value`), and `forge build` output, which discards this information for some reason.

tothtamas28 (approved):

PR #1301, opened 2022-07-27
Refactor solc_to_k generation

ehildenb:
This loses some functionality, but allows us to build it back easily. It introduces some new classes called `Contract`, `Contract.Method`, and `Contract.Field`. In particular:

- Places we were using `KToken` before to build up terms in a hacky way are replaced with actual structure now.
- The classes introduced are used to encapsulate some needed data, which makes them a little cleaner in implementation.
- Generation of storage slot rules results in slightly prettier K.
- Adds a test of `string` type in `Token.sol` test for foundry.


tests/foundry/foundry.k.check.expected
```diff
@@ -8,34 +8,39 @@ module TOKEN-BIN-RUNTIME
...
-    rule  ( Token.x => ( 0 +Int 0 ) )
+    rule  ( Token . x => #hashedLocation ( "Solidity" , 0 , .IntList ) )
       
     
-    rule  ( Token.balances[V0] => #hashedLocation ( "Solidity" , ( 0 +Int 1 ) , V0 ) )
+    rule  ( Token . balances [ V0 ] => #hashedLocation ( "Solidity" , 1 , V0  .IntList ) )
       
     
-    rule  ( Token.allowances[V0][V1] => #hashedLocation ( "Solidity" , #hashedLocation ( "Solidity" , ( 0 +Int 2 ) , V0 ) , V1 ) )
+    rule  ( Token . allowances [ V0 ] [ V1 ] => #hashedLocation ( "Solidity" , 2 , V0  V1  .IntList ) )
```
ehildenb, line 30: @RaoulSchaffranek these ones in particular should be checked, by looking at definition of `#hashedLocation` in `hashed-locations.md`, and making sure that the before and the after both unfold to the same thing.

kevm_pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -37,6 +39,198 @@
...
+@dataclass
+class Contract():
+
+    @dataclass
+    class Method:
+        name: str
+        id: int
+        sort: KSort
+        arg_names: List[str]
+        arg_types: List[str]
```
tothtamas28, line 51:
```suggestion
        arg_names: Tuple[str, ...]
        arg_types: Tuple[str, ...]
```

kevm_pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -37,6 +39,198 @@
...
+        name: str
+        slot: int
+        type: str
+        sort: KSort
+        klabel: KLabel
+        types: Dict
+        _base_types: Set[str]
+
+        def __init__(self, name: str, slot: int, type: str, contract_name: str, field_sort: KSort, types: Dict) -> None:
+            self._base_types = {'t_address', 't_bool', 't_bytes4', 't_bytes32', 't_uint256', 't_int256', 't_uint8', 't_string_storage'}
```
tothtamas28, line 100:
```suggestion
    @dataclass
    class Field:
        _BASE_TYPES: Final = {'t_address', 't_bool', 't_bytes4', 't_bytes32', 't_uint256', 't_int256', 't_uint8', 't_string_storage'}

        name: str
        slot: int
        type: str
        sort: KSort
        klabel: KLabel
        types: FrozenDict

        def __init__(self, name: str, slot: int, type: str, contract_name: str, field_sort: KSort, types: Dict) -> None:
```

To use `FrozenDict`:

```python3
from pyk.utils import FrozenDict
```

kevm_pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -37,6 +39,198 @@
...
+            var_names: List[KInner] = [KVariable(f'V{i}') for i, _ in enumerate(non_terminal_prods)]
+            lhs = KApply(application_label, [contract, KApply(self.klabel, var_names)])
+            rhs = KEVM.hashed_location('Solidity', intToken(self.slot), KEVM.intlist(var_names))
+            return KRule(KRewrite(lhs, rhs))
+
+    name: str
+    storage: Dict
+    bytecode: str
+    methods: List[Method]
+    fields: List[Field]
```
tothtamas28, line 141:
```suggestion
    methods: Tuple[Method, ...]
    fields: Tuple[Field, ...]
```

kevm_pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -37,6 +39,198 @@
...
+        def rule(self, contract: KInner, application_label: KLabel, contract_name: str) -> KRule:
+            arg_vars = [KVariable(aname) for aname in self.arg_names]
+            prod_label = self.production.klabel
+            assert prod_label is not None
+            lhs = KApply(application_label, [contract, KApply(prod_label, arg_vars)])
+            args: List[KInner] = [KEVM.abi_type(input_type, KVariable(input_name)) for input_type, input_name in zip(self.arg_types, self.arg_names)]
+            rhs = KEVM.abi_calldata(self.name, args)
+            opt_conjuncts = [_range_predicate(KVariable(input_name), input_type) for input_name, input_type in zip(self.arg_names, self.arg_types)]
+            conjuncts = [opt_conjunct for opt_conjunct in opt_conjuncts if opt_conjunct is not None]
+            if len(conjuncts) == 0:
```
tothtamas28, line 83:
```suggestion
            if not conjuncts:
```

kevm_pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -37,6 +39,198 @@
...
+    def macro_bin_runtime(self) -> KRule:
+        return KRule(KRewrite(KEVM.bin_runtime(KApply(self.klabel)), KEVM.parse_bytestack(stringToken(self.bytecode))))
+
+    @property
+    def method_sentences(self) -> List[KSentence]:
+        method_application_production: KSentence = KProduction(KSort('ByteArray'), [KNonTerminal(self.sort), KTerminal('.'), KNonTerminal(self.sort_method)], klabel=self.klabel_method, att=KAtt({'function': ''}))
+        res = [method_application_production]
+        for mprod in [method.production for method in self.methods]:
+            assert isinstance(mprod, KSentence)
+            res.append(mprod)
```
tothtamas28, line 208: This is a recurring pattern, consider extracting a function like this:

```python3
def _check_and_append_sentences(sents: List[KSentence], iterable: Iterable) -> None:
    for sent in iterable:
        assert isinstance(sent, KSentence)
        sents.append(sent)
```

tothtamas28 (commented):
Great refactoring, added some comments.

kevm_pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -351,17 +362,8 @@ def recur_mapping(syntax, lhs, rhs, var_idx, key_type_name, value_type_name):
...
-
-    function_productions: List[KSentence] = []
-    function_rules: List[KSentence] = []
-    for f_prod, f_rule in function_sentence_pairs:
-        function_productions.append(f_prod)
-        function_rules.append(f_rule)
-
-    function_sentences = function_productions + function_rules
+    function_sentences = contract.method_sentences
     return [function_call_data_production] + function_sentences if function_sentences else []
```
nishantjr: `if function_sentences else []` looks redundant.

tothtamas28: This is tricky, but operator precedence is:

```python3
([function_call_data_production] + function_sentences) if function_sentences else []
```

ehildenb: Yeah this is to keep from generating subsort productions if the supersort doesn't exist. Otherwise we end up with something like:

```
syntax Int ::= TokenContract "." TokenField
```

But where `TokenField` has not been declared, for example.

nishantjr: OK, lets use parens around the first part to make this obvious. I can see someone else making the same mistake.

kevm_pyk/src/kevm_pyk/solc_to_k.py
ehildenb: Why do we prefer `Tuple[str, ...]` to `List[str]`?

ehildenb: Is it because these are going to be set only once, so immutable representation is better? Also, does it type-check better?

kevm_pyk/src/kevm_pyk/solc_to_k.py
ehildenb: Done.

kevm_pyk/src/kevm_pyk/solc_to_k.py
ehildenb: Done, it reads much nicer.

kevm_pyk/src/kevm_pyk/solc_to_k.py
ehildenb: This code got removed anyway.

kevm_pyk/src/kevm_pyk/solc_to_k.py
ehildenb: Done.

kevm_pyk/src/kevm_pyk/solc_to_k.py
ehildenb: Done.

tothtamas28 (approved):

kevm_pyk/src/kevm_pyk/utils.py
```diff
@@ -29,6 +30,12 @@
 from pyk.utils import hash_str
 
 
+def check_and_append_sentences(sents: List[KSentence], _sents: Iterable) -> None:
+    for sent in _sents:
+        assert isinstance(sent, KSentence)
+        sents.append(sent)
```
RaoulSchaffranek, line 36: As a Python-noob, this looks strange to me. Why do we need this dynamic type-check?

Is this any different from:

```python
check_and_append_sentences(sents: List[KSentence], _sents: Iterable[KSentence]) -> None:
    sents += list(_sents)
```

kevm_pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -28,15 +30,195 @@
...
+                        raise ValueError(f'Unsupported key type for mapping in field {self.sort}: {key_type}')
+                else:
+                    raise ValueError(f'Unsupported type for encoding in field {self.sort}: {self.type}')
+            return KProduction(self.sort, syntax, klabel=self.klabel)
+
+        def rule(self, contract: KInner, application_label: KLabel) -> KRule:
+            non_terminal_prods = [pitem for pitem in self.production.items if type(pitem) is KNonTerminal]
+            var_names: List[KInner] = [KVariable(f'V{i}') for i, _ in enumerate(non_terminal_prods)]
+            lhs = KApply(application_label, [contract, KApply(self.klabel, var_names)])
+            rhs = KEVM.hashed_location('Solidity', intToken(self.slot), KEVM.intlist(var_names))
```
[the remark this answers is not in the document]
ehildenb: No, this should be the actual correct location. It may be that we need to trim it to the subset of the bytes needed for that storage slot (for example, if it's actually a uint48 it's reading, we may need to do something like `#hashedLocation(...) &Int maxUInt48`. But that would be the remaining adjustment that is made.

`#hashedLocation` is defined in `hashed-locations.md`. Also, see `tests/specs/mcd/storage.k` for examples of using these macros.

kevm_pyk/src/kevm_pyk/utils.py
ehildenb: It's because the things being passed in as `_sents` are often coming from functions that produce `List[KProduction]`, not `List[KSentence]`. But we have that `KProduction` is a subtype of `KSentence` (via subclassing), _but_ that _does not_ in Python mean that `List[KProduction]` is a subtype of `List[KSentence]`. So we have to actually iterate through the elements of the list and cast them each to the more general type explicitly. 

https://stackoverflow.com/questions/62814180/is-there-a-covariant-mutable-version-of-list and https://blog.daftcode.pl/covariance-contravariance-and-invariance-the-ultimate-python-guide-8fabc0c24278

kevm_pyk/src/kevm_pyk/solc_to_k.py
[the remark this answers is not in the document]
RaoulSchaffranek: Correct me if I'm wrong. My understanding of `#hashedLocation` is that it's not a complete implementation of Solidity's storage-layout algorithm but only a helper to deal with dynamic arrays and mappings. Integers, addresses, fixed arrays, and strings require different handling not covered by `#hashedLocation`.

For example, consider the following contract:

~~~Solidity
contract Foo {
    uint256 x;
}
~~~

Here, `x` would have storage location `0` and not `#hashedLocation("Solidity", 0, 0 .IntList)`.
Before the refactoring, the corresponding generated K-rule looked as follows, which seems correct:

~~~
rule  ( Token . x => ( 0 +Int 0 ) )
~~~

But now we get:

~~~
rule  ( Token . x => #hashedLocation ( "Solidity" , 0 , .IntList ) )
~~~

For context, the relevant defining rule of `#hashedLocation` looks as follows:

~~~k
    rule #hashedLocation("Solidity", BASE, OFFSET .IntList) => keccak(#bufStrict(32, OFFSET) ++ #bufStrict(32, BASE))   requires #rangeUInt(256, BASE) andBool #rangeUInt(256, OFFSET)
~~~

---

Moreover, something with `#hashedLocation` doesn't seem right. Generally speaking, I don't think  the rule above captures the correct semantics of storage locations in nested dynamic arrays/mappings. In other words, I don't believe the following equation holds:

~~~
#hashedLocation ( "Solidity" , #hashedLocation ( "Solidity" , BASE , V0 ) , V1 ) ) =
#hashedLocation ( "Solidity" , BASE, V0 V1)
~~~

kevm_pyk/src/kevm_pyk/utils.py
RaoulSchaffranek: Thanks that is helpful.

kevm_pyk/src/kevm_pyk/solc_to_k.py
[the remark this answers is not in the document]
ehildenb: I think the second equality does hold, I think we also have that `#hashedLocation(_, BASE, .IntList) == BASE`, so we have that `0 +Int 0 == #hashedLocation("Solidity", 0, .IntList)`.

I am using the rules defined in `hashed-locations.md` to do this reasoning, specifically https://github.com/runtimeverification/evm-semantics/blob/master/hashed-locations.md

In particular:

```
    rule #hashedLocation(_LANG, BASE, .IntList      ) => BASE
    rule #hashedLocation( LANG, BASE, OFFSET OFFSETS) => #hashedLocation(LANG, #hashedLocation(LANG, BASE, OFFSET .IntList), OFFSETS) requires OFFSETS =/=K .IntList 
```

These are the generic rules that handle 0 offsets or 2+ offsets, and then there are the language specific rules for handling exactly 1 offset:

```
    rule #hashedLocation("Vyper",    BASE, OFFSET .IntList) => keccak(#bufStrict(32, BASE)   ++ #bufStrict(32, OFFSET)) requires #rangeUInt(256, BASE) andBool #rangeUInt(256, OFFSET) 
    rule #hashedLocation("Solidity", BASE, OFFSET .IntList) => keccak(#bufStrict(32, OFFSET) ++ #bufStrict(32, BASE))   requires #rangeUInt(256, BASE) andBool #rangeUInt(256, OFFSET) 
    rule #hashedLocation("Array",    BASE, OFFSET .IntList) => keccak(#bufStrict(32, BASE)) +Word OFFSET                requires #rangeUInt(256, BASE) andBool #rangeUInt(256, OFFSET) 
```

PR #1304, opened 2022-07-28
Bump ethereum-tests to v10.4

anvacaru:


evm.md
```diff
@@ -1280,19 +1280,21 @@ The various `CALL*` (and other inter-contract control flow) operations will be d
          <account>
            <acctID> ACCT </acctID>
            <balance> BAL </balance>
+           <nonce> NONCE </nonce>
            ...
          </account>
          <callGas> GCALL </callGas>
-      requires VALUE >Int BAL orBool CD >=Int 1024
+      requires VALUE >Int BAL orBool CD >=Int 1024 orBool notBool #rangeNonce(NONCE)
```
anvacaru, line 1287: Here I could create a new exceptional status code  like `EVMC_NONCE_EXCEEDED` for the case in which `#rangeNonce` is `False`

ehildenb: Does Geth make a new code like this? If so, we should as well.

It goes in `network.md` to do so.

Also, you may have to change some other parts of the semantics too. Do `git grep EVMC_SUCCESS` to see all the places that might need to be changed.

ehildenb:
The Solidity code for the `benchmarks` proofs are directly in the repository next to the specs themselves.

The Solidity code for the `mcd` proofs is here (though this is a newer version): https://github.com/makerdao/dss/tree/master/src

Note that if any contracts makes more than 1 call, then just `#rangeNonce(NONCE)` won't work, because that will allow the contract to make 1 call only. Instead, if it makes 2 calls for example, you'll need `#rangeNonce(NONCE +Int 1) andBool 0 <=Int NONCE`, which is a bit clunky, but should allow the contract to make 2 calls. If it makes 3 calls, you need `NONCE +Int 2`.

evm-types.md
```diff
@@ -142,7 +149,7 @@ These can be used for pattern-matching on the LHS of rules as well (`alias` attr
     rule #rangeUFixed  ( 128 , 10 , X ) => #range ( minUFixed128x10 <= X <= maxUFixed128x10 )
     rule #rangeAddress (            X ) => #range ( minUInt160      <= X <  pow160          )
     rule #rangeBytes   (   N ,      X ) => #range ( 0               <= X <  1 <<Byte N      )
-
+    rule #rangeNonce   (   X          ) => #range ( 0               <= X < maxUInt64        )
     syntax Bool ::= "#range" "(" Int "<"  Int "<"  Int ")" [macro]
```
ehildenb, line 153:
```suggestion
    rule #rangeNonce   (   X          ) => #range ( 0               <= X < maxUInt64        )

    syntax Bool ::= "#range" "(" Int "<"  Int "<"  Int ")" [macro]
```

tests/specs/mcd/vow-cage-surplus-pass-rough-spec.k
```diff
@@ -73,31 +73,31 @@ module VOW-CAGE-SURPLUS-PASS-ROUGH-SPEC
...
+              <nonce> Nonce_Flapper => ?_ </nonce>
             </account>
             <account>
               <acctID> Flopper </acctID>
               <balance> Flopper_balance </balance>
               <code> Flopper_bin_runtime </code>
               <storage> Flopper_STORAGE => Flopper_STORAGE [ #Flopper.live <- 0 ] [ #Flopper.vow <- ACCT_ID ] </storage>
               <origStorage> Flopper_ORIG_STORAGE </origStorage>
-              <nonce> _Nonce_Flopper => ?_ </nonce>
+              <nonce> Nonce_Flopper => ?_ </nonce>
```
ehildenb, line 100: Is there any reason that we're not using the more specific `<nonce> Nonce_Flopper </nonce>` instead of saying taht it could go to anything? The other nonces in this file do the same.

ehildenb: For example, does `Nonce_Flopper => Nonce_Flopper +Int 1` work?

ehildenb (approved):

PR #1306, opened 2022-07-29
Support Solidity `struct` type in generated specifications

ehildenb:
This adds support for the `struct` type in Solidity for the K code generation, and adds a test in the Foundry example of this type.

Notice that we only handle `struct`s that are not complaining except base-types (cannot contain a list in the `struct`, for example). I don't know if Solidity actually has this restriction or not.

- Extend `hashed_location` helper to actually handle what's called a `member_offset`, which is used for recording the offset inside the struct.
- `Contract.Field.{productions,rules}` are extended to return lists of productions/rules, so that we can generate all the productions and rules needed for struct members at the same time.
- An example of a `struct` is added to the Foundry test, and the expected generated output of K is updated.
- Removes the `check_and_append_sentences` with a nicer way of appeasing the type-checker that @tothtamas28 suggested.

### DIFF
--- a/kevm_pyk/src/kevm_pyk/solc_to_k.py
+++ b/kevm_pyk/src/kevm_pyk/solc_to_k.py
@@ -34,7 +34,7 @@ from pyk.prelude import Sorts, intToken, stringToken
 from pyk.utils import FrozenDict, intersperse
 
 from .kevm import KEVM
-from .utils import abstract_cell_vars, build_claim, check_and_append_sentences
+from .utils import abstract_cell_vars, build_claim
 
 _LOGGER: Final = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


kevm_pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -210,8 +221,8 @@ def method_sentences(self) -> List[KSentence]:
     def field_sentences(self) -> List[KSentence]:
         field_access_production: KSentence = KProduction(KSort('Int'), [KNonTerminal(self.sort), KTerminal('.'), KNonTerminal(self.sort_field)], klabel=self.klabel_field, att=KAtt({'macro': ''}))
         res = [field_access_production]
-        check_and_append_sentences(res, [field.production for field in self.fields])
-        check_and_append_sentences(res, [field.rule(KApply(self.klabel), self.klabel_field) for field in self.fields])
+        check_and_append_sentences(res, [prod for field in self.fields for prod in field.productions])
+        check_and_append_sentences(res, [rule for field in self.fields for rule in field.rules(KApply(self.klabel), self.klabel_field)])
```
tothtamas28, line 225: Does this work instead?

```python3
res: List[KSentence] = [field_access_production]
res.extend(prod for field in self.fields for prod in field.productions)
res.extend(rule for field in self.fields for rule in field.rules(KApply(self.klabel), self.klabel_field))
```

tothtamas28 (approved):

kevm_pyk/src/kevm_pyk/solc_to_k.py
ehildenb: It seems to, made the change.

RaoulSchaffranek (approved):
Looks good to me.

PR #1309, opened 2022-08-03
Foundry loop tests

RaoulSchaffranek:
Relevant changes are in: tests/foundry/test/Loops.t.sol


tests/foundry/test/Loops.t.sol
```diff
@@ -0,0 +1,129 @@
...
+    function setUp() public {}
+
+    function sumToN(uint256 n) internal pure returns (uint256) {
+        uint256 result = 0;
+        for (uint256 i = 0; i <= n; i++) {
+            result += i;
+        }
+        return result;
+    }
+
```
ehildenb, line 16: These are good looking tests. Can you rebase them on master cleanly so we can see if it's working on CI?

Issue #1307, opened 2022-08-02
Tests of storage-slot mappings


ehildenb:
We need to test our `solc_to_k` storage slot mapping rule generation code better. One thing we don't currently handle, for example, is a `struct` which has a member that holds a `mapping`. This particular case is needed to handle Foundry's standard library.

Currently, the only real example of tests are here: https://github.com/runtimeverification/evm-semantics/tree/master/tests/specs/examples

These tests generate the needed `tests/specs/examples/*-bin-runtime.k` files using `contract_to_k`, then run hand-written proofs about them which exercise the generated storage-slot rules.

We need to write proofs which exercise more of the rule cases.

- First, the Solidity file must be extended to have more of the types we care about (`tests/specs/examples/*.sol`).
- Then the `contract_to_k` code must be extended to handle these new types when generating helper syntax (`kevm_pyk/src/kevm_pyk/solc_to_k.py`).
- Then a proof must be written using the helper syntax, which will not pass unless the helper syntax is correct (`tests/specs/examples/*-spec.md`).

Note that as you are running these tests, if something goes wrong and it generates the file `tests/specs/examples/*-bin-runtime.k`, then you need to remove this file before the next run of the test. This is due to poor dependency management in `Makefile`.

ehildenb:
Setting up test harness:

- https://github.com/runtimeverification/evm-semantics/blob/master/Makefile#L383
- https://github.com/runtimeverification/evm-semantics/blob/master/Makefile#L456
- https://github.com/runtimeverification/evm-semantics/blob/master/Makefile#L552

PR #1310, opened 2022-08-03
Storage tests

RaoulSchaffranek:
See #1307


tests/gen-spec/foundry/foundry.toml
```diff
@@ -0,0 +1,5 @@
+[profile.default]
+src = 'src'
+out = 'out'
+rpc_endpoints = { optimism = "https://optimism.alchemyapi.io/v2/...", mainnet = "${RPC_MAINNET}" }
+
```
ehildenb, line 5: This whole directory goes away.

ehildenb:
Can you back out Lisandra's changes, and only keep the `tests/specs/examples/Storage.sol` file and `tests/specs/examples/storage-spec.md`.

tests/specs/examples/Storage.sol
```diff
@@ -0,0 +1,84 @@
...
+    string public myString;
+    bytes public myBytes;
+    constructor () {
+        myBool = true;
+        myUint128 = 100;
+        myUint256 = 200;
+        myString = "foo";
+    }
+
+}
```
ehildenb, line 82: This looks good too, probably already this is revealing some storage slot types we don't support. Can we rebase this on master and get a clean diff?

ehildenb:
We should also add nested storage types and such, for example a `Foo[]`, here `Foo` is a `struct`. Or a `struct` member which is a mapping of arrays of other struct members.

RaoulSchaffranek:
New PR: #1319

PR #1311, opened 2022-08-04
CI workflow to sync nix flake inputs to submodules

goodlyrottenapple:
@ehildenb This should fix the sync issue. However, before this can be merged, we need to update `deps/plugin` to the latest version, as the update operation expects this repo to contain a nix flake

### DIFF
--- a/.github/workflows/update-version.yml
+++ b/.github/workflows/update-version.yml
@@ -12,13 +12,25 @@ concurrency:
 jobs:
   test-pr:
     name: 'Update kevm_pyk Version'
-    runs-on: [self-hosted, linux]
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


ehildenb:
@goodlyrottenapple you'll have to run it manually thefirst time, this update workflow _only_ triggers on the `_update-deps` branch, to avoid weirdness.

goodlyrottenapple:
Oh I didn't realise we update the k dep and plugins dep on two different branches. I will move the nix job into its own yaml file and set it to trigger on both branches. Then we can merge that and test by updating deps/plugin?

ehildenb (approved):

ehildenb:
Let's try it!

PR #1323, opened 2022-08-11
Add Foundry tests for programs with loops

tothtamas28:


ehildenb (approved):

ehildenb:
CI is probably hanging because `forge test` is run on CI.

I would recommend to pull the `forge test` step out as a separate target, not run on CI. Make a new target called `foundry-forge-test`, which triggers the `forge build` step then also calls `forge test`, but only run the `forge build` on CI.

ehildenb (approved):

PR #1329, opened 2022-08-16
Run Foundry property tests with KEVM on CI

ehildenb:
This PR sets up the testing harness (and initial tests) for Foundry property tests on CI:

- Do not run `forge test ...` on CI by default, so that we can test failing foundry proofs. Separate out new target `test-foundry-forge` for that.
- Setup actually invoking the prover on the generated specification `SPEC` module in `tests/foundry/foundry.k` in the `Makefile`.
- Add option `--exclude-tests` to `kevm foundry-to-k ...`, which enables skipping the generation of those proof obligations by listing the tests in a file as `ContractName.testMethod`.
- Adds initial exclusion list to `tests/foundry/exclude`, which avoids running all tests except the trivially passing one in `Simple.t.sol`, which is `AssertTest.test_assert_true`, which just does `assert(true)`.

So now the goal is to get the list `tests/foundry/exclude` down to empty.


kevm_pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -304,26 +299,29 @@ def gen_claims_for_contract(empty_config: KInner, contract_name: str, calldata_c
...
-        assert klabel is not None
-        args = [abstract_term_safely(KVariable('_###SOLIDITY_ARG_VAR###_'), base_name=f'V{name}') for name in tm.arg_names]
-        calldata: KInner = KApply(contract_function_application_label, [KApply(contract.klabel), KApply(klabel, args)])
-        callvalue: KInner = intToken(0) if not tm.payable else abstract_term_safely(KVariable('_###CALLVALUE###_'), base_name='CALLVALUE')
-        function_test_calldatas.append((calldata, callvalue))
+    for tm in contract.methods:
+        if f'{contract.name}.{tm.name}' in exclude_tests:
+            _LOGGER.warning(f'Excluding test from contract {contract.name}: {tm.name}')
+        elif tm.name.startswith('testFail'):
+            _LOGGER.warning(f'Ignoring test from contract {contract.name}: {tm.name}')
```
JuanCoRo, line 315: Is this preferable to not ignoring the `testFail` cases and just having them in the `exclude` file until the behavior for `testFail` is added? Maybe not, but just wanted to suggest this.

ehildenb: I'm ignoring testFail for now because I think it needs a different generation process, so I just didn't want to commit to a particular way of encoding it until we actually know what it should look like.

ehildenb: It's also just harder right now to hook up the `testFail` functionality. So I wanted to get the test harness in place first, and later iterations will worry about that.

Currently, all the `testFail` tests are listed in the `tests/foundry/exclude` file anyway, so we won't forget about them.

kevm_pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -3,7 +3,7 @@
 from dataclasses import dataclass
 from pathlib import Path
 from subprocess import CalledProcessError
-from typing import Any, Dict, Final, List, Optional, Tuple
+from typing import Any, Dict, Final, Iterable, List, Optional, Tuple
```
anvacaru, line 6: What is `Iterable` used for?

anvacaru (approved):
LGTM 👍

kevm_pyk/src/kevm_pyk/__main__.py
```diff
@@ -65,12 +65,18 @@ def main():
             if args.command == 'solc-to-k':
```
tothtamas28, line 65: This function needs a clean up, please create an issue for that.

tothtamas28 (approved):

kevm_pyk/src/kevm_pyk/__main__.py
ehildenb: Recorded: https://github.com/runtimeverification/k/issues/2801

kevm_pyk/src/kevm_pyk/solc_to_k.py
ehildenb: It's used in the type signature of `contract_to_k`. You can do `git grep Iterable`.

PR #1330, opened 2022-08-16
foundry.md: add deal cheatcode

anvacaru:
Part of: https://github.com/runtimeverification/foundry/issues/12


kevm_pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -259,7 +265,10 @@ def gen_claims_for_contract(empty_config: KInner, contract_name: str, calldata_c
         'ORIGIN_CELL': KVariable('ORIGIN_ID'),
         'ID_CELL': Foundry.address_TEST_CONTRACT(),
         'CALLER_CELL': KVariable('CALLER_ID'),
-        'LOCALMEM_CELL': KApply('.Memory_EVM-TYPES_Memory'),
+        'ACCESSEDSTORAGE_CELL': KApply('.Map'),
+        'ACTIVEACCOUNTS_CELL': KApply('_Set_',[ KApply('_Set_',[ KApply('SetItem',[Foundry.address_TEST_CONTRACT()]), KApply('SetItem',[Foundry.address_CHEATCODE()])]), KApply('_Set_',[ KApply('SetItem',[Foundry.address_CALLER()]), KApply('SetItem',[Foundry.address_HARDHAT_CONSOLE()])])]),
```
anvacaru, line 269: This should be generated using something like
```python
build_assoc(KApply('_Set_'), KLabel('SetItem'), [intToken(0), intToken(100), intToken(2), intToken(3)])
```
but I couldn't get it right so far.

ehildenb: It should be something like this:

```
build_assoc(KApply('.Set'), KLabel('_Set_'), map(KLabel('SetItem'), [intToken(0), intToken(100), ...]))
```

foundry.md
```diff
@@ -96,6 +96,28 @@ First we have some helpers in K which can:
...
+         <output> _ => .ByteArray </output>
+         <localMem> LM </localMem>
+         <account>
+           <acctID> ACCT_ID </acctID>
+           <balance> _ => #asWord(#range(LM, ARGSTART +Int 36, 32)) </balance>
+           ...
+         </account>
+      requires CHEAT_ADDR ==Int #address(FoundryCheat)
+       andBool #asWord(#range(LM, ARGSTART, 4)) ==Int 3364511341 // selector ( "deal" )
+      [priority(40)]
```
ehildenb, line 119: Why are these in two separate rules+ I think they can be just one rule that directly sets `<balance>`?

anvacaru: Well, to have them in a single rule I would need to find the account ID by having something like
```
 <acctID> ACCTID </acctID>
...
requires ACCTID ==Int #asInteger(#range(LM, ARGSTART +Int 4, 32))
```

which afaik is not very efficient; and I cannot use directly `<acctID>#asInteger(#range(LM, ARGSTART +Int 4, 32)) </acctID>` because it's a lhs.

ehildenb: Did you try putting that on the LHS? Hmmm, I think for concrete execution it should be smart enough to first compute that key, then try finding the map element. For symbolic execution it may be faster to do have you have done.

ehildenb: I would still do something like this (single rule for bypassing `CALL`, separate rule for handling `#setBalance`):

```suggestion
    rule [call.deal]:
         <k> #asInteger(#range(LM, ARGSTART +Int 4, 32)) ) ~> CALL _ CHEAT_ADDR 0 ARGSTART _ARGWIDTH 0 0 ... </k>
         <output> _ => .ByteArray </output>
         <localMem> LM </localMem>
      requires CHEAT_ADDR ==Int #address(FoundryCheat)
       andBool #asWord(#range(LM, ARGSTART, 4)) ==Int 3364511341 // selector ( "deal" )
      [priority(40)]

    rule [call.dealFinal]:
         <k> CALL _ CHEAT_ADDR 0 ARGSTART _ARGWIDTH 0 0 => #setBalance(#asWord(#range(LM, ARGSTART +Int 4, 32)), #asWord(#range(LM, ARGSTART +Int 36, 32))) 1 ~> #push ... </k>
         <output> _ => .ByteArray </output>
         <localMem> LM </localMem>
      requires CHEAT_ADDR ==Int #address(FoundryCheat)
       andBool #asWord(#range(LM, ARGSTART, 4)) ==Int 3364511341 // selector ( "deal" )
      [priority(40)]
      
      rule <k> #setBalance(ACCT_ID, NEW_BAL) => . ... </k>
              <account> <acctID> ACCT_ID </acctID> <balance> _ => NEW_BAL </balance> ... </account>
```

ehildenb: (obviously my formatting needs to be improved)

anvacaru: I will check again with a single rule for `CALL`, but I have doubts since `LM` is not concrete due to the symbolic `value` function arg.

ehildenb: Well, the solution above is 2 rules, but more clear than the original solution.

tothtamas28 (approved):

foundry.md
```diff
@@ -96,6 +96,29 @@ First we have some helpers in K which can:
              ...
          </account>
 ```
+```k
+
+    rule [call.deal]:
+         <k> CALL _ CHEAT_ADDR 0 ARGSTART _ARGWIDTH 0 0
```
ehildenb, line 102: formatting

foundry.md
```diff
@@ -96,6 +96,29 @@ First we have some helpers in K which can:
...
+
+    syntax KItem ::= "#setBalance" "(" Int "," Int ")" [klabel(foundry_setBalance)]
+ // -------------------------------------------------------------------------------
+    rule <k> #setBalance(ACCTID, NEWBAL) => 1 ... </k>
+         <account>
+             <acctID> ACCTID </acctID>
+             <balance> _ => NEWBAL </balance>
+             ...
+         </account>
+```
```
ehildenb, line 121: Wait a second, why are we pushing `1` twice? We already do this once above, then again here. This `1` is probably just discarded. We should just do `.` on the RHS here.

tests/foundry/foundry.k.check.expected
```diff
@@ -12,14 +12,14 @@ module BLOCKPARAMSTEST-BIN-RUNTIME
...
       
     
     syntax Field ::= BlockParamsTestField
     
     syntax BlockParamsTestField ::= "IS_TEST" [symbol(), klabel(field_BlockParamsTest_IS_TEST)]
     
-    syntax BlockParamsTestField ::= "_failed" [symbol(), klabel(field_BlockParamsTest__failed)]
+    syntax BlockParamsTestField ::= "failed" [symbol(), klabel(field_BlockParamsTest_failed)]
```
ehildenb, line 23: This diff, I am not confident is correct. We need to investigate _why_ this happened.

PR #1332, opened 2022-08-17
Cleanups

ehildenb:
This is an assortment of cleanups and improvements that have come from other efforts, and haven't found a good PR to be merged with.

- More explicit about optional argument `foundry` in `contract_to_k` call.
- Small refactoring of computation of methods for a given contract, which makes it more clear how it handles the solc vs. foundry cases.
- Actually check if `solc` produced any errors/warnings and propogate them. Additionally, fail compilation if there are errors.
- Update `slow` and `failing` lists by running on my machine and cutting the slow ones off at 900s. I timed out all tests at 3600s (if it timed out, it's on the failing list, not the slow list).
- Removes redundant inclusion of ds-test via ds-math submodule, which caused weird solc version issues.


kevm_pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -107,15 +107,18 @@ def _get_method_abi(_mname: str) -> Dict:
         self.name = contract_name
         self.bytecode = (contract_json['evm']['deployedBytecode']['object'] if not foundry else contract_json['deployedBytecode']['object'])
         _methods = []
-        if 'methodIdentifiers' not in contract_json or not(foundry or 'methodIdentifiers' in contract_json['evm']):
-            _LOGGER.warning(f'Could not find member \'methodIdentifiers\' while processing contract: {self.name}')
+        if (not foundry) and 'evm' in contract_json and 'methodIdentifiers' in contract_json['evm']:
```
tothtamas28, line 110:
```suggestion
        if not foundry and 'evm' in contract_json and 'methodIdentifiers' in contract_json['evm']:
```

tothtamas28 (approved):

kevm_pyk/src/kevm_pyk/solc_to_k.py
ehildenb: Done.

ehildenb:
I am not getting the same behavior on my machine as CI is seeing, so I'm going to try one commit at a time to understand which one makes the different behavior.

ehildenb:
Turns out this failure is affecting _all_ KEVM builds: https://github.com/runtimeverification/evm-semantics/pull/1331

Going to try nad fix it in a separate PR.

PR #1334, opened 2022-08-23
foundry.md: draft startPrank and stopPrank

anvacaru:
part of: https://github.com/runtimeverification/foundry/issues/12


foundry.md
```diff
@@ -120,6 +131,46 @@ First we have some helpers in K which can:
...
 ```
 
+Prank
+-----
+```k
+    rule [call.startPrank]:
+        <k> CALL _ CHEAT_ADDR 0 ARGSTART _ARGWIDTH 0 0 => 1 ~> #push ... </k>
+         <caller> CL => #asAccount(#range(LM, ARGSTART +Int 4, 32)) </caller>
+         <origin> OG </origin>
```
ehildenb, line 141: The formatting here is strange.

ehildenb: Let's at least make sure the cells line up.

foundry.md
```diff
@@ -97,7 +109,6 @@ First we have some helpers in K which can:
          </account>
 ```
 ```k
-
     rule [call.deal]:
```
ehildenb, line 112: Here as well.

foundry.md
```diff
@@ -120,6 +131,46 @@ First we have some helpers in K which can:
...
+    rule [call.startPrankWithOrigin]:
+        <k> CALL _ CHEAT_ADDR 0 ARGSTART _ARGWIDTH 0 0 => 1 ~> #push ... </k>
+         <output> _ => .ByteArray </output>
+         <caller> CL => #asAccount(#range(LM, ARGSTART +Int 4, 32)) </caller>
+         <origin> OG => #asAccount(#range(LM, ARGSTART +Int 36, 32)) </origin>
+         <prevCaller> _ => CL </prevCaller>
+         <prevOrigin> _ => OG </prevOrigin>
+         <localMem> LM </localMem>
+      requires CHEAT_ADDR ==Int #address(FoundryCheat)
+       andBool #asWord(#range(LM, ARGSTART, 4)) ==Int 1169514616 // selector ( "startPrank(addr)" )
```
ehildenb, line 158: These two cheatcodes have the same signature, but different selector integers, which doesn't make sense to me.

foundry.md
```diff
@@ -145,6 +157,70 @@ This rule then takes the account using `#asWord(#range(LM, ARGSTART +Int 4, 32)`
...
+    rule [call.stopPrank]:
+        <k> CALL _ CHEAT_ADDR 0 ARGSTART _ARGWIDTH 0 0 ~> 1 ~> #push ... </k>
+         <output> _ => .ByteArray </output>
+         <caller> _ => CL </caller>
+         <origin> _ => OG </origin>
+         <prevCaller> CL </prevCaller>
+         <prevOrigin> OG </prevOrigin>
+         <localMem> LM </localMem>
+      requires CHEAT_ADDR ==Int #address(FoundryCheat)
+       andBool CL =/=K .Account andBool OG =/=K .Account
```
lisandrasilva, line 219: Why do we need this? Is this a possible scenario?

anvacaru: I was thinking about corner case scenarios in which `stopPrank()` is called before a `startPrank`, but I have to double check how foundry actually handles those tests.

ehildenb: @anvacaru please write a test which calls stopPrank before startPrank, so we can make sure we handle it similarly to Foundry.

foundry.md
```diff
@@ -145,6 +157,69 @@ This rule then takes the account using `#asWord(#range(LM, ARGSTART +Int 4, 32)`
          </account>
 ```
 
+Pranks
+------
+
+#### `startPrank` - Sets `msg.sender` for all subsequent calls until `stopPrank` is called.
```
anvacaru, line 163:
```suggestion
#### `startPrank` - Sets `msg.sender` and `tx.origin` for all subsequent calls until `stopPrank` is called.
```

foundry.md
anvacaru: Done. Looks like Foundry accepts tests in which `stopPrank` is called before `startPrank`. I've updated the rules accordingly.

lisandrasilva (approved):

tests/foundry/exclude
```diff
@@ -68,6 +68,9 @@ NonceTest.testNonce
 OwnerUpOnlyTest.testFailIncrementAsNotOwner
 OwnerUpOnlyTest.testIncrementAsNotOwner
 OwnerUpOnlyTest.testIncrementAsOwner
+PlainPrankTest.test_startPrankWithOrigin_true
+PlainPrankTest.test_stopPrank_true2
```
RaoulSchaffranek, line 72: Why are these tests excluded? Shouldn't they pass?

tests/foundry/exclude
```diff
@@ -68,6 +68,9 @@ NonceTest.testNonce
...
+PlainPrankTest.testFail_startPrank_false
```
RaoulSchaffranek, line 73: Hmmm... We should find a way to exercise this test and confirm that it fails instead of just disabling it.

tests/foundry/test/PlainPrankTest.t.sol
```diff
@@ -0,0 +1,63 @@
...
+    }
+
+    function test_stopPrank_true2() public {
+        AdditionalToken token = new AdditionalToken();
+        vm.startPrank(address(token));
+        token.incrementCount();
+        assert(token.count() == 1);
+    }
+
+    function testFail_startPrank_false() public {
```
RaoulSchaffranek, line 52: I don't understand what is tested here and in which line the test case is expected to fail.

foundry.md
```diff
@@ -145,6 +157,69 @@ This rule then takes the account using `#asWord(#range(LM, ARGSTART +Int 4, 32)`
...
+The `#if` statements in cells `<id>` and `<origin>` are used in order to avoid the rewriting of the cells to `.Account` if there is no prank active.
+
+```k
+    rule [call.stopPrank]:
+        <k> CALL _ CHEAT_ADDR 0 ARGSTART _ARGWIDTH _ 0 => 1 ~> #push ... </k>
+         <output> _ => .ByteArray </output>
+         <id> CL => #if OLD_CL =/=K .Account #then OLD_CL #else CL #fi </id>
+         <origin> OG => #if OLD_OG =/=K .Account #then OLD_OG #else OG #fi </origin>
+         <prevId> OLD_CL => .Account </prevId>
+         <prevOrigin> OLD_OG => .Account </prevOrigin>
```
RaoulSchaffranek, line 216: Is this identical to what Foundry is doing? I can imagine that calls to `startPrank` stack, however, the public documentation is not clear about that.

tests/foundry/exclude
JuanCoRo: I think so. Maybe they are there to be removed from the file when the last failing test passes. I'll try them out tomorrow.

tests/foundry/exclude
JuanCoRo: As far as I know, we haven't yet implemented the mechanisms for `testFail`, which I believe is the reason for the last test to be excluded. I will double-check that.

tests/foundry/test/PlainPrankTest.t.sol
JuanCoRo: My understanding is that the test will fail the second time `startPrank` is called since `stopPrank` has not been called before. But I don't know if this is the sole purpose of this test. In any case, I'll isolate this behavior into a new one.

foundry.md
```diff
@@ -76,6 +82,12 @@ Foundry Cheat Codes
 module FOUNDRY-CHEAT-CODES
     imports EVM
     imports FOUNDRY-ACCOUNTS
+
+    configuration
+    <cheatcodes>
+        <prevId> .Account </prevId>
+        <prevOrigin> .Account </prevOrigin>
+    </cheatcodes>
```
JuanCoRo, line 90: These cells will add to those in #1349 and could be conceptually confusing. Maybe it's better to create a cell that represents the [`Prank`](https://github.com/foundry-rs/foundry/blob/master/evm/src/executor/inspector/cheatcodes/env.rs#L36) struct. Should come in handy for adding the `prank` cheatcode in the future.
Also, notice that the `depth` is missing, which would cause a `startPrank` to be active for external calls of more depth.

PR #1340, opened 2022-08-26
Implement etch cheatcode

anvacaru:
part of: https://github.com/runtimeverification/foundry/issues/12


foundry.md
```diff
@@ -129,6 +129,39 @@ This rule then takes the account using `#asWord(#range(LM, ARGSTART +Int 4, 32)`
...
+function etch(address who, bytes calldata code) external;
+```
+
+`call.etch` will match when the `etch` function is called at the [Foundry cheatcode address](https://book.getfoundry.sh/cheatcodes/#cheatcodes-reference).
+This rule then takes the account using `#asWord(#range(LM, ARGSTART +Int 4, 32)` and the new bytecode using `#range(LM, ARGSTART +Int 36, 32)` and forwards them to the `#setCode` marker which updates the account accordingly.
+
+```k
+    rule [call.etch]:
+         <k> CALL _ CHEAT_ADDR 0 ARGSTART _ARGWIDTH 0 0
+          => #setCode(#asWord(#range(LM, ARGSTART +Int 4, 32)), #range(LM, ARGSTART +Int 36, 32))
```
anvacaru, line 145: The `code` should be read from the `calldata` cell instead of the `localmem`. The length can be `>32` bytes

foundry.md
```diff
@@ -145,6 +145,39 @@ This rule then takes the account using `#asWord(#range(LM, ARGSTART +Int 4, 32)`
...
+      requires CHEAT_ADDR ==Int #address(FoundryCheat)
+       andBool #asWord(#range(LM, ARGSTART, 4)) ==Int 3033974658 // selector ( "etch(address,bytes)" )
+      [priority(40)]
+
+    syntax KItem ::= "#setCode" "(" Int "," ByteArray ")" [klabel(foundry_setCode)]
+ // -------------------------------------------------------------------------------
+    rule <k> #setCode(ACCTID, CODE) => . ... </k>
+         <account>
+             <acctID> ACCTID </acctID>
+             <code> _ => #if #asWord(CODE) ==Int 0 #then .ByteArray #else {CODE}:>AccountCode #fi </code>
```
lisandrasilva, line 176: Why this `#if` is needed?

anvacaru: I'm using the `#if` to avoid storing a `ByteArray` full of zeros in the `<code>` cell. Also, I am using `:>AccountCode` to cast the `ByteArray` to the `AccountCode`, which is the sort of the cell.

Issue #1345, opened 2022-08-29
Implement assume and expectRevert cheatcodes


ehildenb:
Add tests of `Vm.assume` and `Vm.expectRevert` (all three forms of it) in `test/SimpleTest.t.sol` in a new module.

Then implement these cheat codes.

We need to have 4 different types of tests:

- Tests marked as passing (eg. `test_example`), that are actually passing.
- Tests marked as passing (eg. `test_exampel`), that are actually failing.
- Tests marked as failing (eg. `testFail_example`) that are actually failing.
- Tests marked as failing (eg. `testFail_example`) that are actually passing.

anvacaru:
Assume is already implemented. No concrete work on `expectRevert` yet.

PR #1348, opened 2022-08-30
Implement assume cheatcode

anvacaru:
Part of: https://github.com/runtimeverification/evm-semantics/issues/1345 and https://github.com/runtimeverification/foundry-tests/issues/12

### DIFF
--- a/tests/foundry/test/AssumeTest.t.sol
+++ b/tests/foundry/test/AssumeTest.t.sol
@@ -4,7 +4,6 @@ pragma solidity =0.8.13;
 import "forge-std/Test.sol";
 
 contract AssumeTest is Test {
-    uint256 MAX_INT = 2**256 - 1;
 
     function test_assume_true(uint256 a, uint256 b) public {
         vm.assume(a == b);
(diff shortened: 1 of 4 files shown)


tests/foundry/test/AssumeTest.t.sol
```diff
@@ -0,0 +1,27 @@
...
+        vm.assume(a == b);
+        assertEq(a, b);
+    }
+
+    function test_assume_false(uint256 a, uint256 b) public {
+        vm.assume(a != b);
+        assertEq(a, b);
+    }
+
+    function testFail_assume_true(uint256 a, uint256 b) public {
```
tothtamas28, line 18: Does the prefix `testFail_` mean the test is successful if some assertion is violated?

tothtamas28 (approved):

tests/foundry/test/AssumeTest.t.sol
anvacaru: Yes, from documentation: `testFail` is the inverse of the test `prefix` - if the function does not revert, the test fails [(see here)](https://book.getfoundry.sh/forge/writing-tests?highlight=testfail#writing-tests)

PR #1349, opened 2022-08-30
Implement expectRevert cheatcode

anvacaru:
Part of: https://github.com/runtimeverification/foundry-tests/issues/12


foundry.md
```diff
@@ -145,6 +157,73 @@ This rule then takes the account using `#asWord(#range(LM, ARGSTART +Int 4, 32)`
...
+         <expectedBytes> _ => .ByteArray </expectedBytes> [owise]
+```
+
+```k
+    rule [call.ignoreCheatcodes]:
+         <k> CALL _ CHEAT_ADDR 0 _ARGSTART _ARGWIDTH 0 0 => 1 ~> #push ... </k>
+         <output> _ => .ByteArray </output>
+         <expectRevert> true </expectRevert>
+      requires CHEAT_ADDR ==Int #address(FoundryCheat)
+      [priority(45)]
```
lisandrasilva, line 224: This rule is to make sure that "After calling expectRevert, calls to other cheatcodes before the reverting call are ignored." right? Shouldn't the priority be lower than 40?

foundry.md
```diff
@@ -145,6 +157,73 @@ This rule then takes the account using `#asWord(#range(LM, ARGSTART +Int 4, 32)`
...
+       andBool #asWord(#range(LM, ARGSTART, 4)) ==Int 3273568480 // selector ( "expectRevert(bytes4)" )
+      [priority(40)]
+
+    rule [call.setEndExpectRevert]:
+         <k> CALL _ ADDR _ _ _ _ _ ~> (. => #endExpectRevert) ~> KITEM ... </k>
+         <output> _ => .ByteArray </output>
+         <expectRevert> true </expectRevert>
+      requires ADDR =/=Int #address(FoundryCheat)
+       andBool KITEM =/=K #endExpectRevert
+      [priority(40)]
```
lisandrasilva, line 201: I am not understanding this rule very well. This is not a rule for a cheatcode right? 
From my understanding this rule will be applied after the first revert call happens? But how? I am not getting the flow. Can you please help me to understand?

foundry.md
```diff
@@ -76,6 +82,12 @@ Foundry Cheat Codes
 module FOUNDRY-CHEAT-CODES
     imports EVM
     imports FOUNDRY-ACCOUNTS
+
+    configuration
+    <cheatcodes>
+        <expectRevert> false </expectRevert>
+        <expectedBytes> .ByteArray </expectedBytes>
+    </cheatcodes>
```
JuanCoRo, line 90: The `expectRevert` cell seems to come from the [`Cheatcodes`](https://github.com/foundry-rs/foundry/blob/master/evm/src/executor/inspector/cheatcodes/mod.rs#L77) struct, but the `expectedBytes` is from the [`ExpectedRevert`](https://github.com/foundry-rs/foundry/blob/master/evm/src/executor/inspector/cheatcodes/expect.rs#L28) struct, and there is no representation for the depth. Perhaps encoding these different structs as different cells may be the way to go.

PR #1350, opened 2022-08-30
Updates to foundry commands

ehildenb:
This adds two commands, and removes one:

- Command `foundry-kompile` is added, which takes care of both doing `foundry-to-k` to generate the K sources, and building them immediately using `KEVM.kompile`. This takes only an `out` directory from a Foundry project as input, and sets all other arguments to defaults. Because generation and kompilation happen at the same time, and there are many parameters that only need to be kept live between those two steps (`main_module`, `syntax_module`, requires, imports), those options are now set as defaults and are ignored on the command-line.
- Command `foundry-prove` is added, which optionally can take a `--contract` argument to limit which contracts have their proofs being run. This will allow us to parallelize things on CI. This takes also only the `out` directory form a Foundry project as input.
- Command `foundry-to-k` is removed, in favor of `foundry-kompile`.
- Some initial documentation is added to `foundry.md` about how to use the tool.


kevm_pyk/src/kevm_pyk/__main__.py
```diff
@@ -19,6 +19,12 @@
 _LOG_FORMAT: Final = '%(levelname)s %(asctime)s %(name)s - %(message)s'
 
 
+def _ignoring_arg(args: Dict[str, Any], arg: str, cli_option: str):
+    if arg in args and args[arg]:
+        _LOGGER.warning(f'Ignoring command-line option: {cli_option}')
+    args.pop(arg)
```
tothtamas28, line 25: If `arg not in args`, statement `args.pop(arg)` raises `KeyError`.

If `args[arg]` is falsy, the log message will not be emitted. 

```suggestion
def _ignore_arg(args: Dict[str, Any], arg: str, cli_option: str) -> None:
    if arg in args:
        if args[arg] is not None:
            _LOGGER.warning(f'Ignoring command-line option: {cli_option}')
        args.pop(arg)
```

kevm_pyk/src/kevm_pyk/__main__.py
```diff
@@ -92,7 +98,7 @@ def exec_solc_to_k(
         with open(exclude_tests, 'r') as el:
             _exclude_tests = el.read().strip().split('\n')
     contract = Contract(contract_name, contract_json, foundry=False)
-    contract_module, contract_claims_module = contract_to_k(contract, empty_config, foundry=False, exclude_tests=_exclude_tests)
+    contract_module, contract_claims_module = contract_to_k(contract, empty_config, foundry=False, exclude_tests=_exclude_tests, imports=('VERIFICATION'))
```
tothtamas28, line 101:
```suggestion
    contract_module, contract_claims_module = contract_to_k(contract, empty_config, foundry=False, exclude_tests=_exclude_tests, imports=('VERIFICATION',))
```

The type checker didn't complain as `str` is an `Iterable[str]`.

tothtamas28 (changes requested):

kevm_pyk/src/kevm_pyk/__main__.py
ehildenb: Fixed, I actually removed the hardcoding of that name altogether.

tothtamas28 (approved):

RaoulSchaffranek (approved):
The changes look good to me! 👍

RaoulSchaffranek (approved):

RaoulSchaffranek (approved):

ehildenb:
The problem was the missing `--hook-namespaces` and `--concrete-rules` argument in the inner call to `KEVM.kompile(...)`.

I have just hardcoded these arguments directly into the KEVM object, they haven't changed in ages anyway.

But I want re-review, because it's bigger changes than I hoped since the approvals.

tothtamas28 (approved):

PR #1353, opened 2022-08-31
Added label cheatcode to foundry.md

lisandrasilva:
Added rule for the `label` cheatcode to `foundry.md` and also a test to the `foundry/tests` using the `vm.label` cheat code.


foundry.md
```diff
@@ -245,6 +245,24 @@ This rule then takes the `uint256` value using `#asWord(#range(LM, ARGSTART +Int
...
 ```
 
+#### `label` - Sets a label for a given address.
+
+```
+function label(address addr, string calldata label) external;
+```
+
+`call.label` will match when the `label` function is called at the [Foundry cheatcode address](https://book.getfoundry.sh/cheatcodes/#cheatcodes-reference).
+If an address is labelled, the label will show up in test traces instead of the address. However, there is no change on the state and therefore this rule just skips the cheatcode invocation.
```
ehildenb, line 255: 1 sentence per line please.

anvacaru: Fixed in https://github.com/runtimeverification/evm-semantics/pull/1340/.

PR #1356, opened 2022-09-01
Rule for addr cheatcode

lisandrasilva:
Implemented the rule for `addr` cheatcode in `foundry.md`. Also added a simple test using `addr`.


foundry.md
```diff
@@ -295,9 +295,28 @@ However, there is no change on the state and therefore this rule just skips the
          <localMem> LM </localMem>
       requires CHEAT_ADDR ==Int #address(FoundryCheat)
        andBool #asWord(#range(LM, ARGSTART, 4)) ==Int 3327641368 // selector ( "label(address,string)" )
+    [priority(40)]
 ```
 
+#### `addr` - Sets a label for a given address.
```
david-k, line 301: The description does not seem to match the implementation

tests/foundry/test/AddrTest.t.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity =0.8.13;
+
+import "forge-std/Test.sol";
+
+contract AddrTest is Test {
+
+    function testAddr() public {
+        address alice = vm.addr(1);
+        assertEq(alice, 0x7E5F4552091A69125d5DfCb7b8C2659029395Bdf);
```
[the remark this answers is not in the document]
JuanCoRo: What about having two tests, one with `assert` as you suggest to test out the cheat code itself, and another with `assertEq` to test the integration? I think using `assertEq` may be more idiomatic than `assert` when using Foundry, so maybe users do this more often. But maybe integration with other cheat codes is a pandora's box we don't want to open yet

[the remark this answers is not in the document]
david-k: Yeah, I agree that in general `assertEq` should be used. A compromise could be to only test the `load`/`store` cheatcodes with `assert` to make sure they work. Once they pass, we can use `assertEq` for all other cheatcodes. So maybe we should just really quickly add support for `load`/`store` so that we can use `assertEq`.

foundry.md
lisandrasilva: Fixed.

tests/foundry/test/AddrTest.t.sol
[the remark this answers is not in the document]
lisandrasilva: I left this tests and added 4 more tests using `assert`. Two of them are failing tests (similar to `AssumeTest.t.sol`).

david-k:
Could you add tests for `vm.addr(n)`, where (1) `n=0`, and (2) `n >= 115792089237316195423570985008687907852837564279074904382605163141518161494337`? These are invalid private keys so in both cases, the test should revert

foundry.md
```diff
@@ -295,9 +295,28 @@ However, there is no change on the state and therefore this rule just skips the
...
+function addr(uint256 privateKey) external returns (address);
+```
+
+`call.addr` will match when the `addr` function is called at the [Foundry cheatcode address](https://book.getfoundry.sh/cheatcodes/#cheatcodes-reference).
+This rule takes `uint256` value using `#asWord(#range(LM, ARGSTART +Int 4, 32)` and computes the address for a given private key updating the `<output>` cell.
+
 ```k
-endmodule
+    rule [call.addr]:
+          <k> CALL _ CHEAT_ADDR 0 ARGSTART _ARGWIDTH RETSTART RETWIDTH => 1 ~> #push ~> #setLocalMem RETSTART RETWIDTH #bufStrict(32, #addrFromPrivateKey(#unparseByteStack(#range(LM, ARGSTART +Int 4, 32)))) ... </k>
```
david-k, line 312: Here it is required that no ether is sent. On the other hand, Foundry *does* allow you to send ether when calling `vm.addr()`, which admittedly does not make much sense, but if we want to be compatible with Foundry we should probably behave the same.

lisandrasilva: How is it possible to send ether when calling `vm.addr()`? I just changed the rule, but would like to make a test for that.

david-k: Usually you can just do `foo.bar{value: 1 ether}(...)` to send ether with a call. However, for `vm.addr()` this does not work directly because it is not `payable`. Therefore one needs to create another contract/interface that changes the signature of `addr` accordingly:
```solidity
interface PayableVM {
    function addr(uint256 p) external payable returns(address);
}
```
In a test function you can then cast `vm` to `PayableVM` and send ether like so: `vm.addr{value: 1 ether}(...)`

lisandrasilva: @david-k I used `#checkCall` and `#transferFunds` to support the ether transfers. Please take a look and let me know your thoughts.

david-k: So I did some more tests and it turns out that while you can send ether the `vm`, it seems to get lost on the way, i.e., it does not influence `vm.balance`:
```solidity
PayableVM(address(vm)).addr{value: 4}(23);
console.log(address(vm).balance); // Outputs "0" and not "4"
```
According to this it seems we don't actually need `#transferFunds`. Thus, we can just use a wildcard `_`  for `VALUE` and ignore the ether that is sent. I'll ask on Discord how exactly they process calls to the VM, maybe there are other edge-cases. Sorry for all the back-and-forth!

lisandrasilva: Ahh ok! It was useful for me to learn anyway, so thanks!

david-k (approved):

PR #1360, opened 2022-09-06
Fixes for nix profiling

goodlyrottenapple:
Given a lot of recent changes to the nix packaging, the `#profile` stopped working. This PR re-enables the functionality


goodlyrottenapple:
@ehildenb Is there a way to enable only a single test for the `make test-prove` target? If so, maybe worth adding to the CI to avoid future breakage

PR #1362, opened 2022-09-07
Update the `pyk` dependency

tothtamas28:


ehildenb (approved):

RaoulSchaffranek (approved):
Looks good to me. Can you document somewhere how we can update pyk to a new version? I think we need to manually edit only a few files, and the rest is autogenerated.

tothtamas28:
> Can you document somewhere how we can update pyk to a new version?

It would probably be too error prone to do it manually. Automation is in progress: https://github.com/runtimeverification/devops/issues/110

RaoulSchaffranek:
Oh nice, that's even better!

PR #1363, opened 2022-09-07
Less restrictive rules for Foundry cheatcodes.

david-k:
Fixes https://github.com/runtimeverification/foundry-tests/issues/23

Previously, cheatcodes where matched against the following CALL instruction:

  CALL _ CHEAT_ADDR 0 ARGSTART _ARGWIDTH 0 0

In this commit, this is changed as follows:

  CALL _ CHEAT_ADDR _VALUE ARGSTART _ARGWIDTH _RETSTART _RETWIDTH

In particular:
1. The memory region that contains the return value may now be arbitrary (_RETSTART and _RETWIDTH)
2. An arbitrary amount of ether may be sent when calling the cheatcode (_VALUE)

Because RETSTART and RETWIDTH were initially required to be zero, the K rules where actually never applied. This is because the bytecode emitted by the Solidity compiler always sets RETSTART to a non-zero value, even for functions that do not return any value.

Allowing arbitrary ether to be sent along with cheatcodes ensures we are consistent with Foundry. Note that the ether is not actually transferred but simply ignored. This, too, is consistent with Foundry.

### DIFF
--- a/foundry.md
+++ b/foundry.md
@@ -109,7 +109,7 @@ This rule then takes a `bool` condition using `#range(LM, ARGSTART +Int 4, 32)` 
 ```k
 
     rule [call.assume]:
-         <k> CALL _ CHEAT_ADDR 0 ARGSTART _ARGWIDTH 0 0 => #assume(#range(LM, ARGSTART +Int 4, 32) ==K #bufStrict(32, 1)) ~> 1 ~> #push ... </k>
+         <k> CALL _ CHEAT_ADDR _VALUE ARGSTART _ARGWIDTH _RETSTART _RETWIDTH => #assume(#range(LM, ARGSTART +Int 4, 32) ==K #bufStrict(32, 1)) ~> 1 ~> #push ... </k>
          <output> _ => .ByteArray </output>
          <localMem> LM </localMem>
       requires CHEAT_ADDR ==Int #address(FoundryCheat)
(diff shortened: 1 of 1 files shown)


lisandrasilva (approved):

JuanCoRo (approved):
Looks good! The only thing I would change is to not include the commit from #1356 and instead merge `master` into the PR's branch to encapsulate the changes more clearly.

david-k:
Yeah, sorry, was playing around with git :-)

foundry.md
```diff
@@ -269,7 +269,7 @@ This rule then takes the `uint256` value using `#asWord(#range(LM, ARGSTART +Int
 
 ```k
     rule [call.coinbase]:
-         <k> CALL _ CHEAT_ADDR 0 ARGSTART _ARGWIDTH 0 0 => 1 ~> #push ... </k>
+         <k> CALL _ CHEAT_ADDR _VALUE ARGSTART _ARGWIDTH _RETSTART _RETWIDTH => 1 ~> #push ... </k>
          <localMem> LM </localMem>
          <coinbase> _ => #asWord(#range(LM, ARGSTART +Int 4, 32)) </coinbase>
```
ehildenb, line 275: Did we check the foundry source code for matching here? For eaxmple, does the foundry source code allow any value? or only zero value?

ehildenb:
Did we check that these cheat codes are actually being applied? This can be done in the `kore-repl`, and we should check at least a handful of them.

ehildenb:
That's a fine way to do it, exactyl how I would approach it, but I would have just turned on the logging for `kore-exec`, and avoided stepping into the repl. Something like `--log-entries DebugAttempRule --log log.file` added as arguments to `kore-exec`, then search thorugh teh log file produced.

david-k:
Just tried this with `kevm`, and it seems you can use `--haskell-backend-arg` to pass arguments to `kore-exec`. Nice!

PR #1364, opened 2022-09-08
Rule for setNonce cheatcode

lisandrasilva:


david-k:
I don't think the current rule is applied if the target account does not already exists. For example, Foundry allows `vm.setNonce(address(0), 3)` even if `address(0)` denotes an inactive account. In this case the account needs to be created first (also see https://github.com/runtimeverification/foundry-tests/issues/27).

I'm currently trying to figure out the best approach to implement this because I'm working on `load`/`store` and they work the same way. I'll report back once I have a solution, but in case you are also looking into this let me know of anything interesting you find out.

lisandrasilva:
@david-k that's exactly what I am trying to test right now. I also think that I will have to change the rule for the `getCode` cheat code. I'll update you with my progress working on this. Thanks!

> I don't think the current rule is applied if the target account does not already exists. For example, Foundry allows `vm.setNonce(address(0), 3)` even if `address(0)` denotes an inactive account. In this case the account needs to be created first (also see [runtimeverification/foundry-tests#27](https://github.com/runtimeverification/foundry-tests/issues/27)).
> 
> I'm currently trying to figure out the best approach to implement this because I'm working on `load`/`store` and they work the same way. I'll report back once I have a solution, but in case you are also looking into this let me know of anything interesting you find out.

foundry.md
```diff
@@ -373,18 +373,67 @@ This rule takes the `address` value using `#asWord(#range(LM, ARGSTART +Int 4, 3
...
           <localMem> LM </localMem>
-          <account>
+       requires CHEAT_ADDR ==Int #address(FoundryCheat)
+        andBool #asWord(#range(LM, ARGSTART, 4)) ==Int 755185067 // selector ( "getNonce(address)" )
+     [priority(40)]
+
+
+     syntax KItem ::= "#foundryGetNonce" Int Int Int
+ // ----------------------------------------------------------
+    rule <k> #foundryGetNonce ARGSTART RETSTART RETWIDTH
```
ehildenb, line 390:
```suggestion
     syntax KItem ::= "#foundryGetNonce" Int Int Int
 // ------------------------------------------------
    rule <k> #foundryGetNonce ARGSTART RETSTART RETWIDTH
```

ehildenb: I think we can just call this `#getNonce` instead of `#foundryGetNonce`.

Also, we should compute the `ACCTID` _before_ ppassing it into this function, because currently we're doing a map-lookup on the accounts cell (with key `<ACCTID>`), for a value that is set in the side-conditino. This is very computationally exensive, because the backend will have to search through every element in teh `<accounts>` cell looking for one with a key that matches.

Instead, the previous rule for `CALL` should have:

```
~> #getNonce #asWord(#range(LM, ARGSTART +Int 4, 32)) RETSTART RETWIDTH
```

So that this one can just become:

```
#getNonce ACCTID RETSTART RETWIDTH => #setLocalMem RETSTART RETWIDTH #bufStrict(32, NONCE)
```

without the side-condition `ACCTID ==Int ...` (it's precomputed in the previous rule). This is much more efficient, we can do direct lookup on the key instead of a search.

foundry.md
```diff
@@ -373,18 +373,67 @@ This rule takes the `address` value using `#asWord(#range(LM, ARGSTART +Int 4, 3
...
+    rule <k> #foundrySetNonce ARGSTART => 1 ~> #push ... </k>
+         <output> _ => .ByteArray </output>
+         <account>
+             <acctID> ACCTID </acctID>
+             <nonce>  _ => #asWord(#range(LM, ARGSTART +Int 36, 32))  </nonce>
+             ...
+         </account>
+         <localMem> LM </localMem>
+      requires ACCTID ==Int #asWord(#range(LM, ARGSTART +Int 4, 32))
 ```
```
ehildenb, line 437: Same here, pre-compute the ACCTID in th efirst rule, then use it directly in the second rule. This is the difference between `O(N)` in the size of the accounts cell, and `O(log(N))`.

foundry.md
lisandrasilva: Makes sense! I was just trying to follow the structure of the rules for `load` and `store`. Maybe it makes sense to also optimize those rules. I can do that in this PR, or alternatively, create a new one.

foundry.md
lisandrasilva: Thanks! I didn't know (or remember) that. It's fixed now. Please take a look.

foundry.md
```diff
@@ -373,18 +373,64 @@ This rule takes the `address` value using `#asWord(#range(LM, ARGSTART +Int 4, 3
...
        requires CHEAT_ADDR ==Int #address(FoundryCheat)
-        andBool ACCTID ==Int #asWord(#range(LM, ARGSTART +Int 4, 32))
-        andBool #asWord(#range(LM, ARGSTART, 4)) ==Int 755185067 // selector ( "getNonce(address)" )
+        andBool #asWord(#range(LM, ARGSTART, 4)) ==Int 4175530839 // selector ( "setNonce(address,uint64)" )
      [priority(40)]
+
+     syntax KItem ::= "#setNonce" Int Int
+ // ----------------------------------------------------------
+    rule <k> #setNonce ACCTID NONCE
+          => 1 ~> #push ... </k>
```
ehildenb, line 427:
```suggestion
     syntax KItem ::= "#setNonce" Int Int
 // -------------------------------------
    rule <k> #setNonce ACCTID NONCE => 1 ~> #push ... </k>
```

foundry.md
```diff
@@ -373,18 +373,64 @@ This rule takes the `address` value using `#asWord(#range(LM, ARGSTART +Int 4, 3
...
+        andBool #asWord(#range(LM, ARGSTART, 4)) ==Int 755185067 // selector ( "getNonce(address)" )
+     [priority(40)]
+
+
+     syntax KItem ::= "#getNonce" Int Int Int
+ // ------------------------------------------------
+    rule <k> #getNonce ACCTID RETSTART RETWIDTH
+          => #setLocalMem RETSTART RETWIDTH #bufStrict(32, NONCE)
+          ~> 1 ~> #push ...
+         </k>
```
ehildenb, line 393:
```suggestion
     syntax KItem ::= "#getNonce" Int Int Int
 // -----------------------------------------
    rule <k> #getNonce ACCTID RETSTART RETWIDTH => #setLocalMem RETSTART RETWIDTH #bufStrict(32, NONCE) ~> 1 ~> #push ...
         </k>
```

foundry.md
ehildenb: Issue created: https://github.com/runtimeverification/evm-semantics/issues/1397

foundry.md
lisandrasilva: Fixed.

foundry.md
lisandrasilva: Fixed.

ehildenb (approved):

PR #1365, opened 2022-09-08
Fix README.md to get build-prove working

fiedlr:


README.md
```diff
@@ -194,7 +195,7 @@ Running Tests
 -------------
 
 The tests are run using the supplied `Makefile`.
-First, run `make split-tests` to generate some of the tests from the markdown files.
+First, run `make venv && make build-prove` to generate some of the tests from the markdown files.
```
tothtamas28, line 198: I think `venv` is a (transitive) prerequisite of `build-prove`. Can you confirm?

fiedlr: Actually I just noticed @ehildenb added instructions to run `make venv` after `make build`, so this is no longer relevant. I'm crossing it out.

tothtamas28 (approved):

PR #1366, opened 2022-09-09
Allow proving individual claims

ehildenb:
This PR allows to discharge individual claims by name, using the new features from: https://github.com/runtimeverification/k/pull/2834. This will make it easier to test new features by naming the single claim you want to run.

- Options `--claim` and `--exclude-claim` are added to `kevm prove --pyk ...`, which allows specifying claims to include or claims to skip when running `kprove`.
- Options `--test` and `--exclude-test` are added to `kevm foundry-prove ...`, which allows specifynig Foundry tests to include/exclude when running `kprove`.
- The `kevm prove --pyk ...` and `kevm foundry-prove ...` commands are fixed to exit with non-zero return code on failing proofs.
- The `--exclude` option to `kevm foundry-kompile ...` is removed, and _all_ claims are generated for every run.
- Several tests which were commented because they were not working with `forge test` are uncommented, since we don't run `forge test` on CI anymore.
- Does not skip generating specs for `testFail` tests anymore, just asserts `notBool foundry_success` for these claims. Not clear if this is correct, but the test suite will tell us.
- Hardcode the `LEMMAS` and `INT-SIMPLIFICATION` imports into `kevm foundry-kompile`, because without them even the simplest of specs won't work.
- Update the documentation in `foundry.md` to reflect new way to run the tool.

### DIFF
--- a/foundry.md
+++ b/foundry.md
@@ -10,7 +10,7 @@ From the root of the [KEVM repository](../README.md), after having:
 
 -   Successfully built (or installed) KEVM, and
 -   Have `kevm` on `PATH`, and
--   Have stepped into the virtual environment.
+-   Have stepped into the virtual environment (see the [README](../README.md)).
 
 KEVM supports Foundry specifications via two CLI utilities, `foundry-kompile` and `foundry-prove`.
 To get help, you can do `kevm foundry-kompile --help` and `kevm foundry-prove --help`.
--- a/foundry.md
+++ b/foundry.md
@@ -3,46 +3,69 @@ Foundry Specifications
 
 **UNDER CONSTRUCTION**
 
-### Usage
+Example Usage
+-------------
 
-Given a Foundry `out` directory, call (~2-3 minutes):
+From the root of the [KEVM repository](../README.md), after having:
 
-```sh
-kevm foundry-kompile path/to/foundry/out
-```
+-   Successfully built (or installed) KEVM, and
+-   Have `kevm` on `PATH`, and
+-   Have stepped into the virtual environment.
 
-Then, you can attempt to prove your Foundry property tests with:
+KEVM supports Foundry specifications via two CLI utilities, `foundry-kompile` and `foundry-prove`.
+To get help, you can do `kevm foundry-kompile --help` and `kevm foundry-prove --help`.
+Each command takes as input the Foundry `out/` directory.
+For example, in the root of this repository, you can run:
 
-```sh
-kevm foundry-prove path/to/foundry/out [--contract TestContractName]
-```
-
-Adding optional argument `--contract TestContractName` will only attempt to run the proofs from that contract.
-
-For example, in this repository, you can run:
+*Build Foundry Project:*
 
 ```sh
-% cd tests/foundry
-% forge build
+$ cd tests/foundry
+$ forge build
+[⠊] Compiling...
+[⠑] Compiling 44 files with 0.8.13
+[⠊] Solc 0.8.13 finished in 3.58s
+Compiler run successful (with warnings)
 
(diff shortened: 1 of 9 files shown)


tests/foundry/test/Token.t.sol
```diff
@@ -8,7 +8,7 @@ contract TokenTest {
         assert(true);
     }
 
-    //function test_2() public pure {
-    //    assert(false);
-    //}
+    function test_2() public pure {
```
JuanCoRo, line 11:
```suggestion
    function testFail_2() public pure {
```
Should we be ensuring that `forge tests` succeeds?

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -251,27 +240,47 @@ def exec_prove(
...
+    for _t in tests:
+        _m, _c = _t.split('.')
+        _m = _m.upper() + '-BIN-RUNTIME-SPEC'
+        _c = _c.replace('_', '-')
+        claims.append(f'{_m}.{_c}')
+    for _t in exclude_tests:
+        _m, _c = _t.split('.')
+        _m = _m.upper() + '-BIN-RUNTIME-SPEC'
+        _c = _c.replace('_', '-')
+        exclude_claims.append(f'{_m}.{_c}')
```
RaoulSchaffranek, line 283: Do we expect more conversions from Foundry-names to KEVM-names like this? Then it would be nice to extract this logic into a separate function.

kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -347,24 +347,28 @@ def gen_claims_for_contract(
...
-            for tn, cd, cv in calldata_cells
-        ]
+        for test_name, call_data, call_value in calldata_cells:
+            failing = test_name.startswith('testFail')
+            refined_subst = {'CALLDATA_CELL': call_data, 'CALLVALUE_CELL': call_value}
+            claim_name = test_name.replace('_', '-')
+            init_terms.append((claim_name, substitute(init_term, refined_subst), failing))
     else:
-        init_terms = [(contract_name.replace('_', '-'), init_term)]
+        init_terms.append((contract_name.lower(), init_term, False))
```
RaoulSchaffranek, line 358: Again, we may want to extract this logic into its function.

RaoulSchaffranek (approved):

tests/foundry/test/Token.t.sol
ehildenb: No, then we would never be testing that our tool fails correctly on some inputs. We need such "negative" tests too, tests where we say `test_...` but it actually fails, or we say `testFail...` but it actually passes.

foundry.md
```diff
@@ -3,46 +3,69 @@ Foundry Specifications
...
+[⠑] Compiling 44 files with 0.8.13
+[⠊] Solc 0.8.13 finished in 3.58s
+Compiler run successful (with warnings)
 
-Adding optional argument `--contract TestContractName` will only attempt to run the proofs from that contract.
+$ cd ../..
+```
 
-For example, in this repository, you can run:
+*Kompile to KEVM specification (inside virtual environment):*
```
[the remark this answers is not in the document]
ehildenb: I added instructions to the README, and linked people to it in these instructions.

kevm-pyk/src/kevm_pyk/__main__.py
ehildenb: I have done this one.

RaoulSchaffranek (approved):
Thanks! Looks really good!

PR #1367, opened 2022-09-09
Update pyk to point to its own repository

goodlyrottenapple:
This should get nix working with the moved pyk repo

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -17,9 +17,16 @@
     ethereum-legacytests.url =
       "github:ethereum/legacytests/d7abc42a7b352a7b44b1f66b58aca54e4af6a9d7";
     ethereum-legacytests.flake = false;
+    pyk.url =
+      "github:runtimeverification/pyk/3cef4165e26b1662241c1b8d1d1edd5551413856";
+    pyk.inputs.flake-utils.follows = "k-framework/flake-utils";
+    pyk.inputs.nixpkgs.follows = "k-framework/nixpkgs";
+    pyk.inputs.poetry2nix.follows = "k-framework/poetry2nix";
+
   };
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -17,9 +17,16 @@
     ethereum-legacytests.url =
       "github:ethereum/legacytests/d7abc42a7b352a7b44b1f66b58aca54e4af6a9d7";
     ethereum-legacytests.flake = false;
+    pyk.url =
+      "github:runtimeverification/pyk/3cef4165e26b1662241c1b8d1d1edd5551413856";
```
goodlyrottenapple, line 21: note that this is currently hardcoded to the commit of this PR:
https://github.com/runtimeverification/pyk/pull/1

The pin to this specific commit should be removed once the above PR is merged + we will need to set up a CI mechanism to track and update the pyk dependency the same way we keep `deps/` submodules in sync

ehildenb: We can do this with a github workflow, see this example where ksummarize keeps a file in sync: https://github.com/runtimeverification/ksummarize/blob/master/.github/workflows/update-version.yml

KSummarize uses a template, but you could probably do a sed find/replace, or just a bash script that loops over every line in this file and finds the one with comment `REPLACE THIS LINE`, and emits the updated line with the version.

tothtamas28: Do tags work here? It would be more convenient to point to a tag rather than a commit hash.

tothtamas28 (approved):

flake.nix
goodlyrottenapple: I think so. Even if not, it should not be too difficult to make a script that queries git for the hash of a particular tag and then feeds that to nix.

PR #1369, opened 2022-09-12
Support for Foundry's load/store cheatcodes

david-k:
Adds support for `vm.load()` and `vm.store()` cheatcodes (tracking issue: https://github.com/runtimeverification/foundry-tests/issues/26). A side-effect of this is that `assertEq()` is now working as expected, which means there is no need to use `assert()` anymore.

This PR also introduces the `#loadAccount ACCT` that creates a new account if it doesn't already exist. This is useful for those cheatcodes that work with both existing and non-existing accounts.


foundry.md
```diff
@@ -383,6 +385,97 @@ This rule takes `uint256` value using `#asWord(#range(LM, ARGSTART +Int 4, 32)`
...
+ // ---------------------------------
+    rule <k> #loadAccount ACCT => . ... </k>
+         <activeAccounts> ACCTS:Set </activeAccounts>
+      requires ACCT in ACCTS
+
+    rule <k> #loadAccount ACCT => #newAccount ACCT ... </k>
+         <activeAccounts> ACCTS:Set </activeAccounts>
+      requires notBool ACCT in ACCTS
+```
+
```
ehildenb, line 478: Can we use the existing `#newAccount` instead? Or have this desugar to `#newAccount` somehow? It's defined in `evm.md`.

foundry.md
```diff
@@ -383,6 +385,97 @@ This rule takes `uint256` value using `#asWord(#range(LM, ARGSTART +Int 4, 32)`
...
+    syntax KItem ::= "#foundryVmStore" Int
+ // ----------------------------------------------------------
+    rule <k> #foundryVmStore ARGSTART => 1 ~> #push ... </k>
+         <output> _ => .ByteArray </output>
+         <account>
+             <acctID> ACCTID </acctID>
+			 <storage> STORAGE => STORAGE [ #asWord(#range(LM, ARGSTART +Int 36, 32))
+                                            <- #asWord(#range(LM, ARGSTART +Int 68, 32)) ] </storage>
+             ...
+         </account>
```
ehildenb, line 457: formatting here is very strange, make sure you only have spacse, and no tabs. And make sure to do self-review in browser before asking for review: https://docs.google.com/document/d/1brNoaHuTDMtXkSvF17E4eOGOiS-4cN8JJCOXsKFZz7c/edit

Just have the `<storage>` cell all on one line.

foundry.md
```diff
@@ -383,6 +385,97 @@ This rule takes `uint256` value using `#asWord(#range(LM, ARGSTART +Int 4, 32)`
...
+         </k>
+         <output>
+           _ => #bufStrict(32, #lookup(STORAGE, #asWord(#range(LM, ARGSTART +Int 36, 32))))
+         </output>
+         <account>
+             <acctID> ACCTID </acctID>
+			 <storage> STORAGE </storage>
+             ...
+         </account>
+         <localMem> LM </localMem>
```
ehildenb, line 423: Same here. Look through other files in the repo to see the formatting used elsewhere.

ehildenb: `git grep "<storage>"` will do it.

foundry.md
david-k: The second rule already desugars to `#newAccount`. The difference between the two is that `#newAccount` is only defined if the account does not already exist, while `#loadAccount` simply has no effect in this case (see first rule above). I think this works well in the context of most cheatcodes, which don't care if the specified account already exists or has to be created, but I'm open for suggestions.

foundry.md
david-k: Oh, I have a new vim config and forgot to enable the option to highlight tabs. Should now be fixed

foundry.md
ehildenb: Makes sense! thank you.

foundry.md
```diff
@@ -198,8 +198,10 @@ The values are forwarded to the `#setCode` marker which updates the account acco
 
 ```k
     rule [call.etch]:
-         <k> CALL _ CHEAT_ADDR _VALUE ARGSTART ARGWIDTH _RETSTART _RETWIDTH
-          => #setCode(#asWord(#range(LM, ARGSTART +Int 4, 32)), #range(LM, ARGSTART +Int 36, ARGWIDTH -Int 36))
+         <k> CALL _ CHEAT_ADDR _VALUE ARGSTART _ARGWIDTH _RETSTART _RETWIDTH
+          => #let CodeStart = ARGSTART +Int 100 #in
+             #let CodeLen = #asWord(#range(LM, ARGSTART +Int 68, 32)) #in
+             #setCode(#asWord(#range(LM, ARGSTART +Int 4, 32)), #range(LM, CodeStart, CodeLen))
           ~> 1 ~> #push
```
ehildenb, line 205: all-caps logical variable names (`CodeStart` and `CodeLen`). This would be the only place in the code where not all-caps variables are used.

david-k: Good point, fixed!

foundry.md
```diff
@@ -383,6 +385,96 @@ This rule takes `uint256` value using `#asWord(#range(LM, ARGSTART +Int 4, 32)`
...
+       andBool #asWord(#range(LM, ARGSTART, 4)) ==Int 1719639408 // selector ( "load(address,bytes32)" )
+      [priority(40)]
+
+
+    syntax KItem ::= "#foundryVmLoad" Int Int Int
+ // ----------------------------------------------------------
+    rule <k> #foundryVmLoad ARGSTART RETSTART RETWIDTH
+          => #setLocalMem RETSTART RETWIDTH #bufStrict(32, #lookup(STORAGE, #asWord(#range(LM, ARGSTART +Int 36, 32))))
+          ~> 1 ~> #push ...
+         </k>
```
ehildenb, line 414:
```suggestion
```k
    rule [call.load]:
         <k> CALL _ CHEAT_ADDR _VALUE ARGSTART _ARGWIDTH RETSTART RETWIDTH
          => #loadAccount #asWord(#range(LM, ARGSTART +Int 4, 32))
          ~> #foundryVmLoad ARGSTART RETSTART RETWIDTH
          ~> 1 ~> #push
          ...
         </k>
         <localMem> LM </localMem>
       andBool #asWord(#range(LM, ARGSTART, 4)) ==Int 1719639408 // selector ( "load(address,bytes32)" )
      [priority(40)]


    syntax KItem ::= "#foundryVmLoad" Int Int Int
 // ----------------------------------------------------------
    rule <k> #foundryVmLoad ARGSTART RETSTART RETWIDTH
          => #setLocalMem RETSTART RETWIDTH #bufStrict(32, #lookup(STORAGE, #asWord(#range(LM, ARGSTART +Int 36, 32))))
          ...
         </k>
```

ehildenb: It more closely resembles the same control flow of other rules, which means that it may be easier to factor out that control flow later.

david-k: Yeah, I actually moved `1 ~> #push` around a couple of times. I finally decided to put it inside `#foundryVmLoad` because this is also where we set the return data, so it seemed sensible to let that also be the place to decide whether the call was successful or not (and then either pushing 1 or 0, though in this case `load` cannot fail and we always push 1). Thus, at the moment, the complete semantics of `load` is encapsulated inside `#foundryVmLoad`. Moving `1 ~> #push` to the outer rule would kind of mix the responsibilities of the two rules.

In short, the way the rule is currently written feels "right" to me, but I'm not sure the reason is convincing enough, so let me know :smile:

tests/foundry/test/StoreTest.t.sol
```diff
@@ -12,22 +12,31 @@ contract Store {
...
         assertEq(reads.length, 1);
         assertEq(writes.length, 1);
     }
 
     function testStoreLoad() public {
+		Store myStore= new Store();
+
         vm.store(address(myStore), bytes32(uint256(0)), bytes32(uint256(31337)));
         bytes32 testNumber = vm.load(address(myStore), bytes32(uint256(0)));
         assertEq(uint256(testNumber), 31337); // 31337
```
ehildenb, line 29: What's going on with the formatting here? Make sure to do self-review on GitHub.

foundry.md
ehildenb: Honestly, I'm not sure we should be creating the account if it doesn't exist, it alters the semantics. Look at how `#accountNonexistent` and `#accountEmpty` are defined, which affect the behavior of the `CALL*` opcodes.

Maybe a test you could write is: "read from the storage slot of some non-existent account, then make a call into that account" (but on a Schedule before `SPURIOUS_DRAGON`, because `Gemptyisnonexistent` is changed on that schedule). So you'd have to set the schedule to some earlier one (`Default`) and some later one (`London')`, and run the tests with both schedules.

I guess we should match what foundry does here. Does it actually create an account, and potentially change the value of `#accountNonexistent`, and `#accountEmpty`? Or does it just return 0 immediately for the storage slot?

david-k: I followed Foundry here, which creates a new account:
https://github.com/foundry-rs/foundry/blob/014f7c4445dc815e07ae1b469ca31d266bd59fd8/evm/src/executor/inspector/cheatcodes/env.rs#L177-L187

See that they call `load_account()`, which is defined here:
https://github.com/bluealloy/revm/blob/main/crates/revm/src/journaled_state.rs#L478-L504

What I think is still missing from our rules is adding the account to `<accessedAccounts>` and/or `<touchedAccounts>`, which would affect gas usage (I haven't understood the difference between the two these two cells yet). In general, we may want to ensure that our rules use the same amount of gas as in Foundry, though I haven't focused on this yet.

tests/foundry/test/StoreTest.t.sol
david-k: Fixed

foundry.md
ehildenb: I think still the way to test it is to have an example where you load from an account that doesn't exist, then make a call into the account (maybe just an eth transfer), and then check the gas usage afterwards.

We can see if we get the same gas usage as Foundry. It can be a test with no symbolic inputs, and we can assert that the remaining gas is exactly what it should be.

ehildenb: If we don't update `touched` and `accessed` correctly, then the gas usage will be different. Having this test in place will tell us whether we are conformant to Foundry (when it changes too).

tests/foundry/test/StoreTest.t.sol
```diff
@@ -12,22 +12,30 @@ contract Store {
...
+    function testStoreLoadNonExistent() public {
+        vm.store(address(0), bytes32(uint256(3)), bytes32(uint256(31337)));
+        bytes32 testNumber = vm.load(address(0), bytes32(uint256(3)));
+        assertEq(uint256(testNumber), 31337); // 31337
+    }
+
+    function testLoadNonExistent() public {
+        bytes32 testNumber = vm.load(address(100), bytes32(uint256(23)));
+        assertEq(uint256(testNumber), 0);
+    }
```
ehildenb, line 40: If you make a copy of this test, that then does:

- Transfer to the address(100).
- Asserts the remaining gas is equal to some value (amke sure it passes on Foundry).

Then we'll have higher confidence in the correctness of this implementation.

ehildenb: You could assert the gas amount before and after the call.

david-k: Working on it. After a first test I was a bit surprised, because the gas used by the KEVM was not just a little bit off but completely off. Digging into this, the reason is that we currently override the rule for `CALL` to implement cheatcodes, and at this point, the gas for the whole call has already been deducted. Under normal execution, `CALL` would refund any unused gas, but our cheatcode rules don't do that, which results in the big difference in gas usage between Foundry and KEVM.

I believe the solution is to implement the cheatcodes by matching against `#exec [ CALL ]` instead of `CALL`. Normally, `#exec [ CALL ]` rewrites to `#gas [CALL, ...] CALL ...`, but when implementing cheatacodes we could simply remove the `#gas` from the rewrite in order to not deduct any gas. Looking at the Rust code I think this would be consistent with Foundry, but I'll double-check.

foundry.md
```diff
@@ -383,6 +385,96 @@ This rule takes `uint256` value using `#asWord(#range(LM, ARGSTART +Int 4, 32)`
...
+         <localMem> LM </localMem>
+      requires CHEAT_ADDR ==Int #address(FoundryCheat)
+       andBool #asWord(#range(LM, ARGSTART, 4)) ==Int 1892290747 // selector ( "store(address,bytes32,bytes32)" )
+      [priority(40)]
+
+
+    syntax KItem ::= "#foundryVmStore" Int
+ // ----------------------------------------------------------
+    rule <k> #foundryVmStore ARGSTART => 1 ~> #push ... </k>
```
lisandrasilva, line 450: To be consistent we should move `1 ~> #push` to the outer rule as well.

tests/foundry/test/StoreTest.t.sol
david-k: Okay, load/store should now use the same amount of gas under both Foundry and KEVM. It wasn't necessary to match against `#exec [ CALL ]`, we just need to refund all the reserved gas via `#refund`. Also, `#loadAccount` now inserts the loaded account into the access set. (There is also `<touchedAccounts>`, but it doesn't seem to be used inside the semantics, so I'm not sure if it also needs to be updated.)

To make the gas tests pass I had to slightly modify the initial configuration. In particular, I had to replace some symbolic variables with concrete values.

foundry.md
david-k: `touched` does not seem to be used by the semantics. It is only ever read in `evm-node.md`, but I'm not sure for what purpose and which addresses should be considered part of this set

tests/foundry/exclude
```diff
@@ -69,6 +69,7 @@ MockCallTest.testMockCalls
 MockCallTest.testMockCallValue
 NonceTest.testNonce
 NonceTest.test_GetNonce_false
+NonceTest.test_GetNonce_true
 NonceTest.testFail_GetNonce_true
 NonceTest.testFail_GetNonce_false
 OwnerUpOnlyTest.testFailIncrementAsNotOwner
```
lisandrasilva, line 75: Can't we remove `StoreTest.testStoreLoad` from the `exclude` file?

david-k: Absolutely, thanks!

lisandrasilva (changes requested):
Maybe we should create an issue to refund gas in the other cheat codes as well.

david-k:
Yes, definitely. I'll open one

lisandrasilva (approved):

tests/foundry/test/StoreTest.t.sol
ehildenb: Nice :)

Issue #1358, opened 2022-09-04
Construct KCFGs for each foundry spec instead of a K spec file


ehildenb:
Then we can discharge them in parallel, and use the kit-shell for them.

PR #1372, opened 2022-09-13
Construct KCFGs for Foundry proof obligations

ehildenb:
Fixes: https://github.com/runtimeverification/evm-semantics/issues/1358

Instead of storing K claims for each Foundry test, we store KCFGs with the proof obligations.

- Add options `--reparse` and `--reinit` to `foundry-kompile`, which enables re-parsing `foundry.k` to `spec.json`, and regenerating `kcfgs.json` from `spec.json`, respectively. This requires pulling in several downstream functions from ksummarize: `KCFG_from_claim` and `sanitize_config`.
- Implements `read_kast_flatmodulelist` for reading the output of `kprove --emit-json-spec ...`.
- Removes a few unused methods from `utils.py` (which have been upstreamed/replaced): `get_productions`, `get_production_for_klabel`, and `get_label_for_cell_sorts`.
- Adjusts the unparsing of `.AccountCellMap` to be as `.Bag` instead of empty. When there are 1 or more accounts, it's ok to unparse `.AccountCellMap` as empty, but when there are no accounts, you get things like `(  => ?_ACCOUNTS_FINAL)`. In both cases (no accounts, some accounts), it's safe to use `.Bag`.


kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -216,6 +223,26 @@ def exec_foundry_kompile(
...
+            spec_module_name=spec_module,
+            dry_run=True,
+            args=(['--emit-json-spec', str(parsed_spec)] + prove_args),
+        )
+    if regen or rekompile or reparse or reinit or not kcfgs_file.exists():
+        cfgs: Dict[str, Dict] = {}
+        for module in read_kast_flatmodulelist(parsed_spec).modules:
+            for claim in module.claims:
+                cfg_label = claim.att["label"]
+                _LOGGER.info(f'Producting KCFG: {cfg_label}')
```
tothtamas28, line 240:
```suggestion
                _LOGGER.info(f'Producing KCFG: {cfg_label}')
```

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -216,6 +223,26 @@ def exec_foundry_kompile(
...
         )
+    kevm = KEVM(foundry_definition_dir, main_file=foundry_main_file, profile=profile)
+    if regen or rekompile or reparse or not parsed_spec.exists():
+        prove_args = add_include_arg(includes)
+        kevm.prove(
+            foundry_main_file,
+            spec_module_name=spec_module,
+            dry_run=True,
+            args=(['--emit-json-spec', str(parsed_spec)] + prove_args),
+        )
```
tothtamas28, line 234: Ideally we should emit the JSON spec (or the CFGs) directly. Is that feasible?

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -216,6 +223,26 @@ def exec_foundry_kompile(
...
+    if regen or rekompile or reparse or reinit or not kcfgs_file.exists():
+        cfgs: Dict[str, Dict] = {}
+        for module in read_kast_flatmodulelist(parsed_spec).modules:
+            for claim in module.claims:
+                cfg_label = claim.att["label"]
+                _LOGGER.info(f'Producting KCFG: {cfg_label}')
+                cfgs[cfg_label] = KCFG_from_claim(kevm.definition, claim).to_dict()
+        with open(kcfgs_file, 'w') as kf:
+            kf.write(json.dumps(cfgs))
```
tothtamas28, line 243: What is the advantage of having a single file for all the specs? (I guess we have a single main file because kompiling all modules at once is more efficient than kompiling them separately. But for the specs it shouldn't matter, right?)

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -189,6 +193,9 @@ def exec_foundry_kompile(
     spec_module = 'FOUNDRY-SPEC'
     foundry_definition_dir = foundry_out / 'kompiled'
     foundry_main_file = foundry_definition_dir / 'foundry.k'
```
tothtamas28, line 195: Unrelated to this PR: I think we should have a separate folder for the K specs.

kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -1,8 +1,56 @@
...
+from pyk.prelude import Bool, mlAnd
+
+
+def read_kast_flatmodulelist(ifile: Path) -> KFlatModuleList:
+    _flat_module_list = read_kast(ifile)
+    assert isinstance(_flat_module_list, KFlatModuleList)
+    return _flat_module_list
+
+
+def KCFG_from_claim(defn: KDefinition, claim: KClaim) -> KCFG:  # noqa: N802
```
tothtamas28, line 34: Is this following section copy-pasted from `ksummarize`? We should probably upstream this to `pyk` (with unit tests).

tothtamas28 (approved):
LGTM. Added some comments for potential follow-up tasks.

RaoulSchaffranek:
This is the first time I have seen this "KCFG"-mode. Is it documented somewhere?

kevm-pyk/src/kevm_pyk/__main__.py
ehildenb: Yes, it's feasible. I thought of it after opening this PR, but figured it could wait.

ehildenb: I can change it though.

kevm-pyk/src/kevm_pyk/__main__.py
ehildenb: Perhaps. I don't think ti matters to much where it goes. But I do want to put it in the Foundry output directory, to keep it in snyc with the other artifacts there.

ehildenb:
@RaoulSchaffranek no it's not documented anywhere, implementation here: https://github.com/runtimeverification/pyk/blob/master/src/pyk/kcfg.py

kevm-pyk/src/kevm_pyk/utils.py
ehildenb: Well, not quite actually. Directly from KSummarize, then minor refactor.

ehildenb: It's recorded here: https://github.com/runtimeverification/ksummarize/issues/82

kevm-pyk/src/kevm_pyk/__main__.py
ehildenb: https://github.com/runtimeverification/evm-semantics/issues/1374

PR #1375, opened 2022-09-14
Allow passing --workers/-j argument to foundry-prove

ehildenb:
This adds parallel proving to `kevm foundry-prove` via the `--workers` or `-j` arguments.

- `pathos` package is added as depedency, lock files updated.
- Generation of claims is re-arranged a bit to compute as much ahead of time as possible (before stepping into parallel block).
- Option `-j/--workers` allows the user to now specify how many CPUs to use executing the proofs.
- CI is changed to run 4 foundry proofs at a time.
- Extra logging is added to `__main__` to make it more clear how `foundry-kompile` is progressing.
- `KProve_prove_claim` is factored out, a richer wrapper for `KProve.prove_claim`.

### DIFF
--- a/kevm-pyk/poetry.lock
+++ b/kevm-pyk/poetry.lock
@@ -7,10 +7,10 @@ optional = false
 python-versions = ">=3.5"
 
 [package.extras]
-dev = ["cloudpickle", "coverage[toml] (>=5.0.2)", "furo", "hypothesis", "mypy (>=0.900,!=0.940)", "pre-commit", "pympler", "pytest (>=4.3.0)", "pytest-mypy-plugins", "sphinx", "sphinx-notfound-page", "zope.interface"]
-docs = ["furo", "sphinx", "sphinx-notfound-page", "zope.interface"]
-tests = ["cloudpickle", "coverage[toml] (>=5.0.2)", "hypothesis", "mypy (>=0.900,!=0.940)", "pympler", "pytest (>=4.3.0)", "pytest-mypy-plugins", "zope.interface"]
-tests_no_zope = ["cloudpickle", "coverage[toml] (>=5.0.2)", "hypothesis", "mypy (>=0.900,!=0.940)", "pympler", "pytest (>=4.3.0)", "pytest-mypy-plugins"]
+tests_no_zope = ["cloudpickle", "pytest-mypy-plugins", "mypy (>=0.900,!=0.940)", "pytest (>=4.3.0)", "pympler", "hypothesis", "coverage[toml] (>=5.0.2)"]
+tests = ["cloudpickle", "zope.interface", "pytest-mypy-plugins", "mypy (>=0.900,!=0.940)", "pytest (>=4.3.0)", "pympler", "hypothesis", "coverage[toml] (>=5.0.2)"]
+docs = ["sphinx-notfound-page", "zope.interface", "sphinx", "furo"]
+dev = ["cloudpickle", "pre-commit", "sphinx-notfound-page", "sphinx", "furo", "zope.interface", "pytest-mypy-plugins", "mypy (>=0.900,!=0.940)", "pytest (>=4.3.0)", "pympler", "hypothesis", "coverage[toml] (>=5.0.2)"]
 
 [[package]]
 name = "autoflake"
@@ -66,6 +66,17 @@ optional = false
 python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*"
 
 [[package]]
+name = "dill"
+version = "0.3.5.1"
+description = "serialize all of python"
+category = "main"
+optional = false
+python-versions = ">=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*, !=3.4.*, !=3.5.*, !=3.6.*"
+
+[package.extras]
+graph = ["objgraph (>=1.7.2)"]
+
+[[package]]
 name = "flake8"
 version = "5.0.4"
 description = "the modular source code checker: pep8 pyflakes and co"
@@ -113,9 +124,9 @@ optional = false
 python-versions = ">=3.7"
 
 [package.extras]
-dev = ["flake8", "pep8-naming", "tox (>=3)", "twine", "wheel"]
-docs = ["sphinx (>=5)", "sphinx-autodoc-typehints", "sphinx-rtd-theme"]
-test = ["coverage", "mock (>=4)", "pytest (>=7)", "pytest-cov", "pytest-mock (>=3)"]
+test = ["coverage", "pytest-cov", "mock (>=4)", "pytest-mock (>=3)", "pytest (>=7)"]
+docs = ["sphinx-rtd-theme", "sphinx-autodoc-typehints", "sphinx (>=5)"]
+dev = ["twine", "wheel", "pep8-naming", "flake8", "tox (>=3)"]
 
 [[package]]
 name = "isort"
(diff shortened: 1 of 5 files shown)


ehildenb:
@goodlyrottenapple I see a message on CI here: 
image

Is this a recoverable error? Would we need to basically provide our own Nix derivation for tomli?

kevm-pyk/poetry.lock
```diff
@@ -257,133 +323,34 @@ python-versions = ">=3.7"
...
-    {file = "tomli-2.0.1-py3-none-any.whl", hash = "sha256:939de3e7a6161af0c887ef91b7d41a53e7c5a1ca976325f429cb46ea9bc30ecc"},
-    {file = "tomli-2.0.1.tar.gz", hash = "sha256:de526c12914f0c550d15924c62d72abc48d6fe7364aa87328337a31007fe8a4f"},
-]
-typing-extensions = [
-    {file = "typing_extensions-4.3.0-py3-none-any.whl", hash = "sha256:25642c956049920a5aa49edcdd6ab1e06d7e5d467fc00e0506c44ac86fbfca02"},
-    {file = "typing_extensions-4.3.0.tar.gz", hash = "sha256:e6d2677a32f47fc7eb2795db1dd15c1f34eff616bcaf2cfb5e997f854fa1c4a6"},
-]
+six = []
+toml = []
+tomli = []
```
goodlyrottenapple, line 355: @ehildenb i think this is your issue. Someone on this thread (https://github.com/nix-community/poetry2nix/issues/701) recommends running

```
poetry cache clear pypi --all
```
and making sure the `pyproject.toml` doesn't include empty lists like that?

ehildenb:
Thank you for the pointer @goodlyrottenapple . It turns out that I was using `poetry` 1.1.12, and there was a breakin gchange in PyPi's API. So peolpe must use at least 1.1.14 (I switch to 1.2.0) to be able to run `poetry update` correctly.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -287,6 +292,7 @@ def exec_foundry_prove(
     depth: Optional[int],
     tests: Iterable[str] = (),
     exclude_tests: Iterable[str] = (),
+    parallel: int = 1,
```
tothtamas28, line 295: What do you think?

```suggestion
    workers: int = 1,
```

Also, this input should be validated to be at least 1.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -296,43 +302,54 @@ def exec_foundry_prove(
...
+
+    kevm = KEVM(definition_dir, profile=profile, use_directory=use_directory)
+
+    def prove_it(_id_and_claim: Tuple[str, KClaim]) -> Tuple[bool, KInner]:
+        _claim_id, _claim = _id_and_claim
+        return KProve_prove_claim(kevm, _claim, _claim_id, _LOGGER)
+
+    with ProcessPool(ncpus=parallel) as process_pool:
+        results = process_pool.map(prove_it, claims)
+        process_pool.close()
```
tothtamas28, line 344: Are you sure closing this explicitly is necessary? I'd assume `ProcessPool.__exit__` takes care of this.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -296,43 +302,54 @@ def exec_foundry_prove(
...
+
+    failed_claims = [(cid, result) for ((cid, _), (failed, result)) in zip(claims, results) if failed]
+    _failed_claim_ids = [cid for cid, _ in failed_claims]
+
+    if _failed_claim_ids:
+        _LOGGER.error(f'Failed to prove KCFGs: {_failed_claim_ids}')
```
tothtamas28, line 350: I think this is part of ordinary usage of the program, consider using `print` instead ([docs](https://docs.python.org/3/howto/logging.html#when-to-use-logging)).

kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -21,10 +22,24 @@
     undo_aliases,
 )
 from pyk.kcfg import KCFG
-from pyk.ktool import KPrint
+from pyk.ktool import KPrint, KProve
 from pyk.prelude import Bool, mlAnd
 
 
+def KProve_prove_claim(  # noqa: N802
```
tothtamas28, line 29: It is safe to replace `KProve.prove_claim` by this, right?

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -296,43 +302,54 @@ def exec_foundry_prove(
...
-    kcfgs: Dict[str, KCFG] = {}
+
     _LOGGER.info(f'Reading file: {kcfgs_file}')
     with open(kcfgs_file, 'r') as kf:
         kcfgs = {k: KCFG.from_dict(v) for k, v in json.loads(kf.read()).items()}
+
     tests = [Contract.contract_test_to_claim_id(_t) for _t in tests]
     exclude_tests = [Contract.contract_test_to_claim_id(_t) for _t in exclude_tests]
-    claims = list(kcfgs.keys())
+    claim_names = list(kcfgs.keys())
```
tothtamas28, line 313:
```suggestion
    claim_names = set(kcfgs)
```

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -296,43 +302,54 @@ def exec_foundry_prove(
...
-    if _failed_claims:
-        _LOGGER.error(f'Failed to prove KCFGs: {_failed_claims}')
-        sys.exit(1)
+    claims = [
+        (
+            kcfg_name.replace('.', '-'),
+            kcfg.create_edge(kcfg.get_unique_init().id, kcfg.get_unique_target().id, mlTop(), depth=-1).to_claim(),
+        )
+        for kcfg_name, kcfg in kcfgs.items()
+    ]
```
tothtamas28, line 334: Consider extracting a function.

tothtamas28 (approved):
Exciting feature. :rocket:

kevm-pyk/pyproject.toml
```diff
@@ -12,6 +12,7 @@ authors = [
 
 [tool.poetry.dependencies]
 python = "^3.8"
+pathos = "*"
```
tothtamas28, line 15: Did you consider using standard library features (e.g. [multiprocessing](https://docs.python.org/3/library/multiprocessing.html))?

kevm-pyk/src/kevm_pyk/__main__.py
ehildenb: It's not necessary in terms of the exit codes produced, but if you don't do the explicit close then a bunch of ugly (and harmless) warnings are posted, so I wanted to keep the output cleaner.

kevm-pyk/src/kevm_pyk/utils.py
ehildenb: Well, not really. This is just a wrapper around KProve.prove_claim, so it assumes all the functionality of that one.

kevm-pyk/src/kevm_pyk/__main__.py
ehildenb: Done.

kevm-pyk/src/kevm_pyk/__main__.py
ehildenb: Done.

kevm-pyk/pyproject.toml
ehildenb: Yes, I tried multiprocessing first. Most things I searched said that pathos is better behaved. multiprocessing using default Python pickling to create copies of the Python interpreter, which fails in many cases (and failed in this case). pathos.multiprocessing using dill instead, which handles pickling almost everything (and works in our case).

https://stackoverflow.com/questions/29121979/python-pathos-multiprocessing-vs-class-methods

kevm-pyk/src/kevm_pyk/__main__.py

kevm-pyk/src/kevm_pyk/__main__.py

kevm-pyk/src/kevm_pyk/utils.py
tothtamas28: I mean after inlining (so that there is no dependency to the original method), could we exclusively use this one instead? If so, we should do the refactoring.

ehildenb: Hmmm, not quite sure. This one assumes you won't do `--dry-run`. But yes, it should be refactored/upstreamed. It's listed: https://github.com/runtimeverification/ksummarize/issues/82

PR #1376, opened 2022-09-14
Add all uint types for foundry test parameters

JuanCoRo:
Fixes: https://github.com/runtimeverification/evm-semantics/issues/1378

This PR makes it possible to have test parameters of all unit types.


tests/foundry/test/UintTypeTest.t.sol
```diff
@@ -0,0 +1,52 @@
...
+        assert(x == x);
+        assert(type(uint144).max >= x);
+    }
+
+    function testFail_uint144(uint144 x) public pure {
+        assert(type(uint144).max > x);
+    }
+
+    function test_uint112(uint112 x) public pure {
+        assert(x == x);
```
lisandrasilva, line 44: What is the reason to have these `assert(x == x)`?

JuanCoRo: To use `x` for something, rather than just being `assert(true)`. Actually, these tests passed with just `assert(true)` before making any changes to `lemmas.k`

JuanCoRo: The point of the first assert is to just use the variable, and the second assert is to test out `type(uintX).max` for something. I did it like this so that it's easier to debug/isolate what "components" of the types we're testing. For instance, I made a typo on one of the `type(uintX).max` and the first assert was passing but not the second. But maybe we want to subsume these into just one assert.

lisandrasilva: Ahh ok makes sense! Thanks! Other alternative is to put them into two different tests, one to test just the type with the first `assert` and other to test `type(uintX).max` with the second `assert`.

JuanCoRo: After thinking about this, since both `assert`s are reasonably trivial, I'd leave them together so that the time to build and execute the tests doesn't increase too much. What do you think about this?

david-k:
Would it be a lot of work to support all of the 32 uint types? I'm just wondering if we could solve this once and for all, so that if someone needs e.g. a `uint120` they don't have to create another PR first.
Alternatively, we could create an issue so that we don't forget about it.

JuanCoRo:
That can be done with no problem. Could you post a reference to the 32 types you're mentioning? Thanks!

david-k:
Cool! The uint types go from `uint8` up to `uint256` in 8 bit steps, see <https://docs.soliditylang.org/en/v0.8.17/types.html#integers>.

tests/foundry/test/UintTypeTest.t.sol
lisandrasilva: You are right! It's better to leave them together.

tests/foundry/test/UintTypeTest.t.sol
```diff
@@ -0,0 +1,52 @@
...
+    function setUp() public {}
+
+    function test_uint256(uint256 x) public pure {
+        assert(x == x);
+        assert(type(uint256).max >= x);
+    }
+
+    function testFail_uint256(uint256 x) public pure {
+        assert(type(uint256).max > x);
+    }
```
ehildenb, line 14: Remember that there are 4 types of tests we should have:

- Tests that _should not_ fail and _do not_ fail.
- Tests that _should not_ fail and _do_ fail.
- Tests that _should_ fail and _do not_ fail.
- Tests that _should_ fail and _do_ fail.

Please make sure you have some tests of each type.

ehildenb: Remember to add failing tests to the `tests/foundry/exclude` list.

ehildenb:
Issue made: https://github.com/runtimeverification/evm-semantics/issues/1378

Makefile
```diff
@@ -184,7 +184,8 @@ kevm_files := abi.md              \
               network.md          \
               optimizations.md    \
               serialization.md    \
-              state-utils.md
+              state-utils.md      \
+	            word.md
```
ehildenb, line 188: Formatting here is not matching. Use spaces for these lists.

kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -482,16 +512,75 @@ def _range_predicate(term, type_label: str):
...
+    if type_label == 'uint232':
+        return (True, KEVM.range_uint(232, term))
+    if type_label == 'uint224':
+        return (True, KEVM.range_uint(224, term))
+    if type_label == 'uint216':
+        return (True, KEVM.range_uint(216, term))
+    if type_label == 'uint208':
+        return (True, KEVM.range_uint(208, term))
+    if type_label == 'uint200':
+        return (True, KEVM.range_uint(200, term))
```
ehildenb, line 536: I think this can be replaced by logic that says

```
if type_label.startswith('uint'):
    width = int(type_label[4:])
    if not (0 < width and width <= 256 and width % 8 == 0):
        raise ValueError(f'Unsupported range predicate type: {label_type}')
    return (True, KEVM.range_uint(width, term))
```

You may be able to also make a similar simplification in `evm_base_sort` method.

ehildenb (approved):
One suggestion about making two python methods more compact, but otherwise LGTM!

kevm-pyk/src/kevm_pyk/solc_to_k.py
JuanCoRo: Included! If you think what I wrote looks good let's merge this :handshake:

Issue #1374, opened 2022-09-13
Directly construct KCFGs instead of parsing


ehildenb:
Currently (as of https://github.com/runtimeverification/evm-semantics/pull/1372), we are generating KCFGs as the proof obligations for each foundry proof. We do this by generating a giant K file with the proof obligations, kompiling it, the using `kprove --emit-json-spec ... --dry-run` on the same file to read in the specs and generate KCFGs from them.

Instead, we can just directly generate the KCFGs right after calling kompiling the definition, and never bother writing these claims to the giant K file. This skips an entire round trip of unparsing + parsing the claims.

See https://github.com/runtimeverification/evm-semantics/pull/1372#discussion_r969325536

PR #1377, opened 2022-09-14
Directly construct KCFGs in foundry-kompile

ehildenb:
Fixes: https://github.com/runtimeverification/evm-semantics/issues/1374

This makes a series of small improvements:

- We avoid including the spec-modules in the definition.
- We directly construct `kcfgs.json` from the Foundry output, instead of turning into K specs and then turning back into KCFGs after. As such, we can skip the `kprove --dry-run --emit-json-spec ...` step, and remove `--reparse` argument. Options `spec_module` and `parsed_spec`.
- We start the proofs with an arbitrary initial account set, instead of `.Bag`.
- Add option `--depth` to `kevm foundry-prove ...`, so you can inspect intermediate states.
- Add option `--[no-]minimize` to `kprove_args`, which means we by default minimize the proof output, but give the option to not minimize with `--no-minimize`.
- Adds `symbol` attribute to generated productions for `contract_to_k`, so that we can be sure the klabels will be exactly how we generate them.
- Adds options `--lemma` to `kprove_args`, so that users can pass in lemmas on the CLI to use for simplification.


tothtamas28 (approved):

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -150,26 +149,17 @@ def exec_foundry_to_k(
...
             if claims_module:
                 _LOGGER.info(f'Produced claim module: {claims_module.name}')
-                claims_modules.append(claims_module)
+                claims.extend((claims_module.name, claim) for claim in claims_module.claims)
     _main_module = KFlatModule(
         main_module if main_module else 'MAIN', [], [KImport(mname) for mname in [_m.name for _m in modules] + imports]
     )
-    _spec_module = KFlatModule(
-        spec_module if spec_module else 'SPEC', [], [KImport(mname) for mname in [_m.name for _m in claims_modules]]
-    )
```
RaoulSchaffranek, line 159: What's the reason behind removing the spec module? Is it to separate the auto-generated rules from the hand-written claims?

RaoulSchaffranek (approved):

kevm-pyk/src/kevm_pyk/__main__.py
ehildenb: Just to reduce the output size since we aren't using the claims anymore to make the input specifications. The KCFGs are stored structurally (instead of textual claims), and then when we want to run the proof, the KCFG is written as a claim to a file for calling KProve on.

PR #1388, opened 2022-09-21
Fixed typo on foundry.md

lucasmt:
`AssertTest.test_assert_true` to `AssumeTest.test_assume_true`

### DIFF
--- a/foundry.md
+++ b/foundry.md
@@ -59,7 +59,7 @@ WARNING 2022-09-11 15:36:00,449 kevm_pyk.solc_to_k - Using generic sort K for ty
 *And discharge some specific test as a proof obligation (inside virtual environment):*
 
 ```sh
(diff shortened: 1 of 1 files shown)


ehildenb:
Why this change? Both tests exist, so they shuold both be fine for the test.

lucasmt:
> Why this change? Both tests exist, so they shuold both be fine for the test.

Do they? I only see `AssumeTest` here: https://github.com/runtimeverification/evm-semantics/tree/master/tests/foundry/test

ehildenb:
The AssertTest are here: https://github.com/runtimeverification/evm-semantics/blob/master/tests/foundry/test/Simple.t.sol

The selector slug `--test ...` lets you specify ContractName.test_name, not FileName.test_name.

lucasmt:
You're right, my mistake

PR #1391, opened 2022-09-22
Separate generaion of main module and claims

tothtamas28:
Generating the state after `setUp()` requires that the main module is already kompiled.

### DIFF
--- a/kevm-pyk/src/kevm_pyk/__main__.py
+++ b/kevm-pyk/src/kevm_pyk/__main__.py
@@ -202,13 +202,7 @@ def _foundry_to_k(
         if json_file.endswith('.metadata.json'):
             continue
 
-        _LOGGER.info(f'Processing contract file: {json_file}')
-        with open(json_file, 'r') as cjson:
-            contract_json = json.loads(cjson.read())
-
-        contract_name = json_file.split('/')[-1]
-        contract_name = contract_name[0:-5] if contract_name.endswith('.json') else contract_name
-        contract = Contract(contract_name, contract_json, foundry=True)
+        contract = _contract_from_json(json_file)
 
         module = contract_to_main_module(contract, empty_config, imports=imports)
         _LOGGER.info(f'Produced contract module: {module.name}')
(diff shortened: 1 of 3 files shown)


kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -230,6 +193,73 @@ def exec_foundry_kompile(
...
+
+
+def _foundry_to_claims(
+    definition_dir: Path,
+    profile: bool,
+    contracts: Iterable[Contract],
+    main_module: Optional[str],
+) -> List[Tuple[str, KClaim]]:
+    kevm = KEVM(definition_dir, profile=profile)
+    empty_config = kevm.definition.empty_config(KSort('KevmCell'))
```
ehildenb, line 252: Maybe we should factor out `KEVM.Sort.KevmCell`? Or `KEVM.top_cell.sort`?

ehildenb: Just an idea.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -230,6 +193,73 @@ def exec_foundry_kompile(
...
+
+    claims: List[Tuple[str, KClaim]] = []
+    for contract in contracts:
+        claims_module = contract_to_spec_module(contract, empty_config, main_module=main_module)
+        _LOGGER.info(f'Produced claim module: {claims_module.name}')
+        claims.extend((claims_module.name, claim) for claim in claims_module.claims)
```
ehildenb, line 258: We should instead make a _LOGGER message for each additional claim we add, since we don't actually care about the modules anymore.

ehildenb: In fact, I guess that this logic can be moved to `contract_to_spec_module`, which should be renamed to `contract_to_claims` (avoid building a module altogether).

Then it could be renamed to `contract_to_claim_cfgs` or the like, and have it just directly construct KCFGs instead of even returning the KClaim.

ehildenb (approved):

kevm-pyk/src/kevm_pyk/__main__.py
tothtamas28: Done.

kevm-pyk/src/kevm_pyk/__main__.py
tothtamas28: Done. (Except I didn't move logic between `__main__` and `solc_to_k` for now, but eventually we should reorganize these modules.)

RaoulSchaffranek (approved):

PR #1393, opened 2022-09-23
Switch to using pyk kompile by default

ehildenb:
~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/1396~

This finishes out the last functionality to switch all calls to `kevm kompile` over to using pyk.

- Pass `--debug` through to `kevm_pyk kompile ...`.
- Add options `-ccopt`, `--backend`, and `--no-llvm-kompile` to `kevm kompile ...`.
- Factor out computation of `-ccopt` from `kompile_opts` in `kevm` runner script, because pyk's `kompile` must consume those arguments slightly differently.
- Enable kompiling with pyk by default in `Makefile`.
- Moves installation of `hub` CLI tool out of main Dockerfile into specific stage that needs it in Jenkinsfile.


kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -60,24 +60,42 @@ def exec_gst_to_kore(input_file: Path, schedule: str, mode: str, chainid: int, *
 
 def exec_kompile(
     definition_dir: Path,
+    backend: str,
```
tothtamas28, line 63: This could be of type `pyk.ktool.KompileBackend`.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -420,6 +448,7 @@ def parse(s: str) -> List[T]:
     )
 
     k_kompile_args = ArgumentParser(add_help=False)
+    k_kompile_args.add_argument('--backend', type=str, help='[llvm|haskell]')
```
tothtamas28, line 451: Consider using `pyk.ktool.KompileBackend`:

```suggestion
    k_kompile_args.add_argument('--backend', type=KompileBackend, help='[llvm|haskell]')
```

Then `exec_kompile`, `KEVM.kompile`, etc. can have `backend: KompileBackend` as argument.

tothtamas28 (approved):

kevm-pyk/src/kevm_pyk/__main__.py

ehildenb:
Blocked on: https://github.com/runtimeverification/evm-semantics/pull/1396

tothtamas28 (approved):

flake.nix
```diff
@@ -87,6 +87,8 @@
             patches = [ ./package/nix/kevm.patch ];
 
             postPatch = ''
+              substituteInPlace ./Makefile \
+                --replace 'PYK_ACTIVATE := . $(VENV_DIR)/bin/activate' 'PYK_ACTIVATE := true'
```
ehildenb, line 91: @goodlyrottenapple I think ti would be more natural to pass `PYK_ACTIVATE=true` as a CLI option than to do this substitution like this, but I could not figure out how to do that given this current flake. It seemed that I could possibly do something with `buildFlags`, but that is not used here, and maybe I could also do something with `buildPhase`, but that's also not used here. Any suggestions or is this the best method?

tothtamas28 (approved):

PR #1396, opened 2022-09-25
Single concrete rules file

ehildenb:
This PR consolidates our concrete rules strategy:

- All the concrete rules file contenst are sorted and uniqued into file `tests/specs/concrete-rules.txt`. Then several rules are removed which should just be simplified out.
- The other concrete rules files are removed.
- The Makefile is adjusted to use only this single new file.
- Some small generalizations to the Bihu proofs, in addition to adding some lemmas needed to keep them passing with the updated concrete-rules.txt file (it includes rules about `Csstore` that these proofs used to let execute through, but are now marked `concrete`).


ehildenb:
I am pretty sure this should pass CI now, but won't stay up to watch it. No need to review if it doesn't.

tothtamas28 (approved):

tests/specs/bihu/verification.k
```diff
@@ -81,3 +83,49 @@ module VERIFICATION
...
+      requires Ghasdirtysstore << SCHED >>
+      [simplification]
+
+    rule Csstore(SCHED, NEW, CURR, _ORIG)
+      => #if CURR ==Int 0 andBool NEW =/=Int 0
+           #then Gsstoreset < SCHED >
+           #else Gsstorereset < SCHED >
+         #fi
+      requires notBool Ghasdirtysstore << SCHED >>
+      [simplification]
```
anvacaru, line 129: These rules are already present in evm.md. Is there a reason for duplicating them?

ehildenb: Yes, it's because they are marked as `concrete` now, so they won't be simplified anymore in `evm.md`. But the Java backend needs these rules to apply, so we duplicate them here for the java backend.

anvacaru (approved):

PR #1399, opened 2022-09-26
Pass `KPrint` instance to `contract_to_claims`

tothtamas28:


kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -70,6 +70,7 @@ def kompile(
     @staticmethod
     def _patch_symbol_table(symbol_table: Dict[str, Any]) -> None:
         # fmt: off
+        symbol_table['binRuntime']                                    = applied_label_str('#binRuntime')
```
ehildenb, line 73: You should not need to have to do this. If so, it's because the incorrect KEVM is being passed somewhere. Somewhere, we are using the haskell-backend default build of KEVM instead of the foundry-spec generated build of KEVM.

ehildenb (approved):

Issue #1405, opened 2022-09-28
Refactors to `foundry.md` file


ehildenb:
Currently, `#finishCheatCode` does similar stuff to `#return` rule (in `evm.md`):

```
    syntax CallOp ::= "CALL"
 // ------------------------
    rule <k> CALL _GCAP ACCTTO VALUE ARGSTART ARGWIDTH RETSTART RETWIDTH
          => #checkCall ACCTFROM VALUE
          ~> #call ACCTFROM ACCTTO ACCTTO VALUE VALUE #range(LM, ARGSTART, ARGWIDTH) false
          ~> #return RETSTART RETWIDTH
         ...
         </k>
         <id> ACCTFROM </id>
         <localMem> LM </localMem>

...

    rule <statusCode> EVMC_SUCCESS </statusCode>
         <k> #halt ~> #return RETSTART RETWIDTH
          => #popCallStack ~> #dropWorldState
          ~> 1 ~> #push ~> #refund GAVAIL ~> #setLocalMem RETSTART RETWIDTH OUT
         ...
         </k>
         <output> OUT </output>
         <gas> GAVAIL </gas>
```

I think we could make the rules nicer by doing emulating more of the KEVM call signature (sketch of how to do it with `Vm.assume(....)` cheatcode):

```
syntax KItem ::= "#return_foundry" Int Int [klabel(foundry_return)]
         | "#call_foundry" Int ByteArray [klabel(foundry_call)]

    rule [foundry.call]:
         <k> CALL _ CHEAT_ADDR _VALUE ARGSTART ARGWIDTH RETSTART RETWIDTH
            => #call_foundry #asWord(#range(LM, ARGSTART, 4)) #range(LM, ARGSTART +Int 4, ARGWIDTH -Int 4)
           ~> #return_foundry RETSTART RETWIDTH
          ...
           </k> 
    requires CHEAT_ADDR ==Int #address(FoundryCheat)
    [priority(40)]

    rule [foundry.return]:
       <k> #return_foundry RETSTART RETWIDTH
         => #setLocalMem RETSTART RETWIDTH OUT
        ~> #refund GCALL
       ~> 1 ~> #push
          ... </k>
          <output> OUT </output>
        <callGas> GCALL </callGas>

    rule [foundry.call.assume]:
        <k> #call_foundry selector("assume(bool)") ARGS => #assume(ARGS ==K #bufStrict(32, 1))) ... </k>
         <output> _ => .ByteArray </output>
```

[...]

PR #1406, opened 2022-09-28
foundry.md: refactor the CALL rules

anvacaru:
Closes: https://github.com/runtimeverification/evm-semantics/issues/1405

This PR refactors the way the `CALL` rules are handled for the Foundry cheat codes.
- Now, a single `CALL` rule with a higher priority will match every call to the Foundry address.
- The logic of each cheat code is defined using `#call_foundry` rules.
- `#return_foundry` groups together all the steps that need to be taken after a `CALL` opcode has finished executing.
- Moved all helper rules under `Utils` section.
- The higher priority `CALL` rule and the `#call_foundry` rules are disabled for the java backend due to compilation errors ( the `#range` rules for the java backend expect a `Map` and in the current implementation we're passing a `WordStack`) and also due to failing proofs.
- Added new lemmas to simplify `#range(CONCRETE_BYTES ++ #buf(_, SYMBOLIC_VALUE), START, LEN)` which would not rewrite further.


foundry.md
```diff
@@ -334,18 +329,14 @@ This rule then takes the `uint256` value using `#asWord(#range(LM, ARGSTART +Int
...
-This rule then takes the `uint256` value using `#asWord(#range(LM, ARGSTART +Int 4, 32)` and updates the `<chainID>` cell.
+`foundry.call.chainId` will match when the `chainId` cheat code function is called.
+This rule then takes the `uint256` value from the function call data which is represented as `ARGS` and updates the `<chainID>` cell.
 
-```k
-    rule [call.chainId]:
-         <k> CALL _ CHEAT_ADDR _VALUE ARGSTART _ARGWIDTH _RETSTART _RETWIDTH => #finishCheatCode ... </k>
+```{.k .bytes}
+    rule [foundry.call.chainId]:
+         <k> #call_foundry 1078582738 ARGS => . ... </k>
```
ehildenb, line 337: We can't use `selector("chainId(uint256"))` here?

What if instead we do:

```
rule <k> #call_foundry FID ARGS => . ... </k>
      REST OF RULE
  requires FID ==Int selector("chainId(uint256)")
```

It just seems more readable if we can use the function selector.

But at the very least, we should have a comment saying that this Int is equal to `selector("chainId(uint256)"`

anvacaru: I left a comment here regarding the use of the `selector` function https://github.com/runtimeverification/evm-semantics/issues/1405#issuecomment-1260769951 . I'll give your suggestion a try.

foundry.md
```diff
@@ -498,28 +451,22 @@ This rule then loads the storage slot identified by `#asWord(#range(LM, ARGSTART
...
-         </account>
+      requires SELECTOR ==Int selector ( "store(address,bytes32,bytes32)" )
 ```
 
+Otherwise, ignore any other call to the Foundry contract.
+
+```{.k .bytes}
+    rule [foundry.call.owise]:
+         <k> #call_foundry _ _ => . ... </k> <output> _ => .ByteArray </output> [owise]
+```
```
ehildenb, line 469: We shouldn't ignore, we should just halt. So we can either just remove this rule, or have the `owise` rule go to a special `#error_foundry String` production, which holds the function selector that could not match and says "not implemented".


If we add such an error state, we can go through and add explicit error rules for all the unimplemented cheatcodes:

```
rule <k> #call_foundry NAME _ => #unimplemented_foundry STRING ... </k>
    requires NAME ==Int selector(STRING)
```

That way we get better error messages for the methods that are proper cheat codes, and we just have not implemented yet, and we have a record directly in this file of all the unimplemented cheat codes.

Then we can have one rule for unknown cheatcodes:

```
rule <k> #call_foundry NAME _ => #unimplemented_foundry ("UNKNOWN: " +String Int2String(NAME)) ... </k>
```

foundry.md
```diff
@@ -537,15 +484,106 @@ Utils
...
+           <acctID> ACCTID </acctID>
+           <storage> STORAGE => STORAGE [ LOC <- VALUE ] </storage>
+             ...
+         </account>
+```
+
+ - selectors for cheat code functions.
+
+```k
+    syntax Int ::= selector ( String ) [alias, klabel(abi_selector), symbol, function, no-evaluators]
```
ehildenb, line 570: This should be moved back to where it was defined, because there are other modules which use it.

foundry.md
anvacaru: For now I just defined the `#error_foundry` production, will open a new issue for the handling of unimplemented cheat codes.

foundry.md
anvacaru: Moved to `abi.md` instead.

foundry.md
```diff
@@ -185,24 +187,60 @@ First we have some helpers in K which can:
...
+    rule [foundry.call]:
+         <k> CALL _ CHEAT_ADDR _VALUE ARGSTART ARGWIDTH RETSTART RETWIDTH
+          => #call_foundry #asWord(#range(LM, ARGSTART, 4)) #range(LM, ARGSTART +Int 4, ARGWIDTH -Int 4)
+          ~> #return_foundry RETSTART RETWIDTH
+         ...
+         </k>
+         <localMem> LM </localMem>
+    requires CHEAT_ADDR ==Int #address(FoundryCheat)
+    [priority(40)]
+```
```
ehildenb, line 207: I think here we can set the `<output>` to `.ByteArray`, so that calls which don't need the `<output>` don't have to make any changes.

foundry.md
```diff
@@ -185,24 +187,60 @@ First we have some helpers in K which can:
...
-         <k> CALL _ CHEAT_ADDR _VALUE ARGSTART _ARGWIDTH _RETSTART _RETWIDTH => #assume(#range(LM, ARGSTART +Int 4, 32) ==K #bufStrict(32, 1)) ~> #finishCheatCode ... </k>
-         <output> _ => .ByteArray </output>
-         <localMem> LM </localMem>
-      requires CHEAT_ADDR ==Int #address(FoundryCheat)
-       andBool #asWord(#range(LM, ARGSTART, 4)) ==Int 1281615202 // selector ( "assume(bool)" )
-      [priority(40)]
+```{.k .bytes}
+    rule [foundry.call.assume]:
+         <k> #call_foundry SELECTOR ARGS => #assume(ARGS ==K #bufStrict(32, 1)) ... </k>
+      requires SELECTOR ==Int selector ( "assume(bool)" )
```
ehildenb, line 243: This doesn't put the same width restriction on `ARGS` (by doing `#range(ARGS, 0, 32)` basically) as the original rule.

I wonder if we should do that? In the below rule for `deal` you _do_ include that extra restriction.

ehildenb: See this issue: https://github.com/runtimeverification/evm-semantics/issues/1410

tests/specs/lemmas.k
```diff
@@ -415,4 +415,12 @@ module LEMMAS-HASKELL [symbolic, kore]
 
     rule B ==Bool false => notBool B [simplification]
 
+    rule #range(A ++ _B, START, LENGTH) => #range(A, START, LENGTH)
+      requires START +Int LENGTH <=Int #sizeByteArray(A) [simplification]
+
+    rule #range(A ++ B, START, LENGTH) => #range(B, START -Int #sizeByteArray(A), LENGTH)
+      requires #sizeByteArray(A) <=Int START [simplification]
+
+    rule #range ( #buf ( LEN , VALUE ) , 0 , LEN ) => #buf (LEN, VALUE) [simplification]
```
ehildenb, line 424:
```suggestion
    rule #range ( BA:ByteArray , 0 , LEN ) => BA requires #sizeByteArray(BA) ==Int LEN [simplification]
```

Does this work? It seems more general.

foundry.md
anvacaru: In the `rule` for `deal`, I do that because `ARGS` is supposed to contain two arguments. Here I assume that the restriction is made on line 200 with `#range(LM, ARGSTART +Int 4, ARGWIDTH -Int 4)`.

tests/specs/lemmas.k
anvacaru: Checking it locally now, will add your suggestion if it works.

PR #1408, opened 2022-09-30
added rules for ecrecoverloop00-sig1-invalid-spec

jinxinglim:
## Possible solution for failed proof for ecrecoverloop00-sig1-invalid-spec

**Problem:** Failed proof for ecrecoverloop00-sig1-invalid-spec
**Possible solution:** Added the following two rules in `tests/specs/benchmarks/verification.k` to debug ecrecoverloop00-sig1-invalid-spec:
```k
// for debugging ecrecoverloop00-sig1-invalid-spec
    rule A  <Int B => #Not ( { true #Equals B <=Int A } ) [simplification]
    rule A <=Int B => #Not ( { true #Equals B  <Int A } ) [simplification]
```
> **Note:** I am not sure if the above rules should be added in `tests/specs/benchmarks/verification.k` or `tests/specs/int-simplification.k`.

## Details
As pointed out in https://github.com/runtimeverification/haskell-backend/issues/2997, the proof of ecrecoverloop00-sig1-invalid-spec will fail due execution gets stuck with the following content of k cell:
```k
JUMPI 559 bool2Word ( 0 <Int maxUInt160 &Int #asWord ( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) ++ #range ( b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01`" [ 128 := #buf ( 32 , SIGV0:Int ) ++ #buf ( 32 , SIGV1:Int ) ] [ 192 := #buf ( 32 , SIGR0:Int ) ++ #buf ( 32 , SIGR1:Int ) ] [ 256 := #buf ( 32 , SIGS0:Int ) ++ #buf ( 32 , SIGS1:Int ) ] [ 320 := b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00" ] , #sizeByteArray ( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) ) +Int 320 , 32 -Int #sizeByteArray ( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) ) ) ) ) ~> #pc [ JUMPI ] ~> #execute ~> _Gen0:K
```
 One can find this cause by running `kevm prove` for `ecrecoverloop00-sig1-invalid-spec.k` through `Kore Repl`, i.e.,
```
cd evm-semantics
kevm prove tests/specs/benchmarks/ecrecoverloop00-sig1-invalid-spec.k --backend haskell --definition tests/specs/benchmarks/verification/haskell --debugger
```
and produce the graph through `graph` after stepping forward for say 3000 steps, i.e., `stepf 3000`.

The part of the graph that we concern about is the following:
Screenshot from 2022-09-30 12-42-32

---

### `konfig` of the crucial nodes

**Node 1760:** This is the node where the `JUMPI` command is waiting to be executed:
```k
<k>
      JUMPI 559 bool2Word ( 0 <Int maxUInt160 &Int #asWord ( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) ++ #range ( b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01`" [ 128 := #buf ( 32 , SIGV0:Int ) ++ #buf ( 32 , SIGV1:Int ) ] [ 192 := #buf ( 32 , SIGR0:Int ) ++ #buf ( 32 , SIGR1:Int ) ] [ 256 := #buf ( 32 , SIGS0:Int ) ++ #buf ( 32 , SIGS1:Int ) ] [ 320 := b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00" ] , #sizeByteArray ( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) ) +Int 320 , 32 -Int #sizeByteArray ( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) ) ) ) ) ~> #pc [ JUMPI ] ~> #execute ~> _Gen0:K
</k>
```

**Node 1761:** This is the node where `bool2Word ( 0 <Int maxUInt160 &Int #asWord ( #ecrec ( ... ) ++ #range ( ... ) ) )` is evaluated to 0, i.e., `0 <Int maxUInt160 &Int #asWord ( #ecrec ( ... ) ++ #range ( ... ) )` is false, which will be rewritten to `0 <Int maxUInt160 &Int #asWord ( #ecrec ( ... ) ++ #range ( ... ) ) <= 0` via the following rule found in `tests/specs/int-simplification.k`: 
```k
rule notBool (A  <Int B) => B <=Int A [simplification]
```
as shown in the `konfig` of node 1761:
```k
...
#And
  {
    true
  #Equals
    maxUInt160 &Int #asWord ( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) ++ #range ( b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01`" [ 128 := #buf ( 32 , SIGV0:Int ) ++ #buf ( 32 , SIGV1:Int ) ] [ 192 := #buf ( 32 , SIGR0:Int ) ++ #buf ( 32 , SIGR1:Int ) ] [ 256 := #buf ( 32 , SIGS0:Int ) ++ #buf ( 32 , SIGS1:Int ) ] [ 320 := b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00" ] , #sizeByteArray ( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) ) +Int 320 , 32 -Int #sizeByteArray ( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) ) ) ) <=Int 0
  }
...
```

**Node 1762:** This is the node where `bool2Word ( 0 <Int maxUInt160 &Int #asWord ( #ecrec ( ... ) ++ #range ( ... ) ) )` is evaluated to 1, i.e., `0 <Int maxUInt160 &Int #asWord ( #ecrec ( ... ) ++ #range ( ... ) )` is true, which is shown in the `konfig` of node 1762:
```k
...
#And
  {
    true
  #Equals
    0 <Int maxUInt160 &Int #asWord ( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) ++ #range ( b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01`" [ 128 := #buf ( 32 , SIGV0:Int ) ++ #buf ( 32 , SIGV1:Int ) ] [ 192 := #buf ( 32 , SIGR0:Int ) ++ #buf ( 32 , SIGR1:Int ) ] [ 256 := #buf ( 32 , SIGS0:Int ) ++ #buf ( 32 , SIGS1:Int ) ] [ 320 := b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00" ] , #sizeByteArray ( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) ) +Int 320 , 32 -Int #sizeByteArray ( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) ) ) )
  }
...
```

**Node 1763:** This is the (unexpected) node where `0 <Int maxUInt160 &Int #asWord ( #ecrec ( ... ) ++ #range ( ... ) )` is false and true at the same time, which is shown in the `konfig` of node 1763:
```k
#And
  #Not ( {
    true
  #Equals
    0 <Int maxUInt160 &Int #asWord ( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) ++ #range ( b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01`" [ 128 := #buf ( 32 , SIGV0:Int ) ++ #buf ( 32 , SIGV1:Int ) ] [ 192 := #buf ( 32 , SIGR0:Int ) ++ #buf ( 32 , SIGR1:Int ) ] [ 256 := #buf ( 32 , SIGS0:Int ) ++ #buf ( 32 , SIGS1:Int ) ] [ 320 := b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00" ] , #sizeByteArray ( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) ) +Int 320 , 32 -Int #sizeByteArray ( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) ) ) )
  } )
#And
  #Not ( {
    true
  #Equals
    maxUInt160 &Int #asWord ( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) ++ #range ( b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01`" [ 128 := #buf ( 32 , SIGV0:Int ) ++ #buf ( 32 , SIGV1:Int ) ] [ 192 := #buf ( 32 , SIGR0:Int ) ++ #buf ( 32 , SIGR1:Int ) ] [ 256 := #buf ( 32 , SIGS0:Int ) ++ #buf ( 32 , SIGS1:Int ) ] [ 320 := b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00" ] , #sizeByteArray ( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) ) +Int 320 , 32 -Int #sizeByteArray ( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) ) ) ) <=Int 0
  } )
```

**Possible reasons for node 1763:**
1) Simplification of `0 <Int maxUInt160 &Int #asWord (#ecrec ( ... ) ++ #range ( ... ) )` to `0 <Int #asWord (#ecrec ( ... ) ++ #range ( ... ) )` did not occur prior to node 1760. If it occurs, the prover should be able to figure it out.
2) The prover does not know that `<=Int` is total ordering, i.e., if `A <Int B`, then `B <=Int A` cannot hold.

---

### Attempt/s to tackle reason 1

To tackle reason A, I added the following rules to "force" the simplification of `0 <Int maxUInt160 &Int #asWord (#ecrec ( ... ) ++ #range ( ... ) )` to `0 <Int #asWord (#ecrec ( ... ) ++ #range ( ... ) )`:
```k
rule 0 <=Int #asWord( _ ) => true [simplification]
rule #asWord(#ecrec(_   , _   , _   , _   ) ++ _ ) <=Int maxUInt160 => true [simplification]
```
But the same problem still persists, i.e., `0 <Int #asWord ( #ecrec ( ... ) ++ #range ( ... ) )` is false and true at the same time, as shown in the following `konfig`:
```k
...
#And
  #Not ( {
    true
  #Equals
    0 <Int #asWord ( #ecrec ( #buf ( 32 , HASH ) , #buf ( 32 , SIGV0 ) , #buf ( 32 , SIGR0 ) , #buf ( 32 , SIGS0 ) ) ++ #range ( b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01`" [ 128 := #buf ( 32 , SIGV0 ) ++ #buf ( 32 , SIGV1 ) ] [ 192 := #buf ( 32 , SIGR0 ) ++ #buf ( 32 , SIGR1 ) ] [ 256 := #buf ( 32 , SIGS0 ) ++ #buf ( 32 , SIGS1 ) ] [ 320 := b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00" ] , #sizeByteArray ( #ecrec ( #buf ( 32 , HASH ) , #buf ( 32 , SIGV0 ) , #buf ( 32 , SIGR0 ) , #buf ( 32 , SIGS0 ) ) ) +Int 320 , 32 -Int #sizeByteArray ( #ecrec ( #buf ( 32 , HASH ) , #buf ( 32 , SIGV0 ) , #buf ( 32 , SIGR0 ) , #buf ( 32 , SIGS0 ) ) ) ) )
  } )
#And
  #Not ( {
    true
  #Equals
    #asWord ( #ecrec ( #buf ( 32 , HASH ) , #buf ( 32 , SIGV0 ) , #buf ( 32 , SIGR0 ) , #buf ( 32 , SIGS0 ) ) ++ #range ( b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01`" [ 128 := #buf ( 32 , SIGV0 ) ++ #buf ( 32 , SIGV1 ) ] [ 192 := #buf ( 32 , SIGR0 ) ++ #buf ( 32 , SIGR1 ) ] [ 256 := #buf ( 32 , SIGS0 ) ++ #buf ( 32 , SIGS1 ) ] [ 320 := b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00" ] , #sizeByteArray ( #ecrec ( #buf ( 32 , HASH ) , #buf ( 32 , SIGV0 ) , #buf ( 32 , SIGR0 ) , #buf ( 32 , SIGS0 ) ) ) +Int 320 , 32 -Int #sizeByteArray ( #ecrec ( #buf ( 32 , HASH ) , #buf ( 32 , SIGV0 ) , #buf ( 32 , SIGR0 ) , #buf ( 32 , SIGS0 ) ) ) ) ) <=Int 0
  } )
#And
  <kevm>
    <k>
      JUMPI 559 bool2Word ( 0 <Int #asWord ( #ecrec ( #buf ( 32 , HASH ) , #buf ( 32 , SIGV0 ) , #buf ( 32 , SIGR0 ) , #buf ( 32 , SIGS0 ) ) ++ #range ( b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01`" [ 128 := #buf ( 32 , SIGV0 ) ++ #buf ( 32 , SIGV1 ) ] [ 192 := #buf ( 32 , SIGR0 ) ++ #buf ( 32 , SIGR1 ) ] [ 256 := #buf ( 32 , SIGS0 ) ++ #buf ( 32 , SIGS1 ) ] [ 320 := b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00" ] , #sizeByteArray ( #ecrec ( #buf ( 32 , HASH ) , #buf ( 32 , SIGV0 ) , #buf ( 32 , SIGR0 ) , #buf ( 32 , SIGS0 ) ) ) +Int 320 , 32 -Int #sizeByteArray ( #ecrec ( #buf ( 32 , HASH ) , #buf ( 32 , SIGV0 ) , #buf ( 32 , SIGR0 ) , #buf ( 32 , SIGS0 ) ) ) ) ) ) ~> #pc [ JUMPI ] ~> #execute ~> _Gen0 ~> .
    </k>
...
...
```
---

### Attempt/s to tackle reason 2

The following are the attempts to introduce rules to the `tests/specs/benchmarks/verification.k` under the `VERIFICATION` module (the last one is the only one that succeeded):

- Attempt 1:
```k
rule #Not ( { true #Equals I1 <=Int I2 } ) => { true #Equals I2 <Int I1 } [simplification]
```

- Attempt 2:
```k
rule (A <Int B) andBool (B <=Int A) => false [simplification]
```

- Attempt 3:
```k
rule (notBool (A <Int B)) andBool (notBool (B <=Int A)) => false [simplification]
```

- Attempt 4:
```k
rule (A <Int B) orBool (B <=Int A) => true [simplification]
```

- Attempt 5:
```k
rule (A <Int B) andBool (B <=Int A) => false requires (A <Int B) orBool (B <=Int A) [simplification]
```

- Attempt 6:
```k
rule (notBool (A <Int B)) andBool (notBool (B <=Int A)) => false requires (A <Int B) orBool (B <=Int A) [simplification]
```

- Attempt 7 (the process went on for >30min, so I guess there is a loop of rewriting):
```k
rule A  <Int B => notBool (B <=Int A) [simplification]
rule A <=Int B => notBool (B  <Int A) [simplification]
```

- Attempt 8 (the process went on for >30min, so I guess there is a loop of rewriting):
```k
rule A  <Int B => false requires B <=Int A [simplification]
rule A <=Int B => false requires B  <Int A [simplification]
```

- Attempt 9 (the process went on for >30min, so I guess there is a loop of rewriting):
```k
rule A  <Int B => true requires notBool (B <=Int A) [simplification]
rule A <=Int B => true requires notBool (B  <Int A) [simplification]
```

- Attempt 10 (completed the proof in around 6 min):
```k
rule A  <Int B => #Not ( { true #Equals B <=Int A } ) [simplification]
rule A <=Int B => #Not ( { true #Equals B  <Int A } ) [simplification]
```
---

### Existing rules that might be helpful in the debugging process:

- In `tests/specs/int-simplification.k`:
```k
rule notBool (A  <Int B) => B <=Int A [simplification]
rule notBool (A <=Int B) => B  <Int A [simplification]
```

- In `tests/specs/benchmarks/verification.k`:
```k
rule 0 <=Int #asWord(#ecrec(_   , _   , _   , _   ))                  => true                                                      [simplification]
rule         #asWord(#ecrec(_   , _   , _   , _   )) <=Int maxUInt160 => true                                                      [simplification]

rule #asWord(WS) &Int maxUInt256 => #asWord(WS) [simplification]

rule maxUInt256 &Int X => X requires #rangeUInt(256, X) [simplification]

rule maxUInt160 &Int Y => Y requires 0 <=Int Y andBool Y <=Int maxUInt160 [simplification]
```
---

### DIFF
--- a/tests/specs/benchmarks/verification.k
+++ b/tests/specs/benchmarks/verification.k
@@ -153,9 +153,8 @@ module VERIFICATION-HASKELL [symbolic, kore]
 
     rule 0 <Int #asWord ( #ecrec( HASH, SIGV, SIGR, SIGS ) ) => notBool ( #ecrecEmpty ( HASH, SIGV, SIGR, SIGS ) )  [simplification]
     
-    // <Int <=/=> <=Int (to debug ecreoverloop00-sig1-invalid-spec)
-    rule A  <Int B => #Not ( { true #Equals B <=Int A } ) [simplification]
-    rule A <=Int B => #Not ( { true #Equals B  <Int A } ) [simplification]
+    // <Int => <=Int (to debug ecreoverloop00-sig1-invalid-spec)
+    rule A <Int B => A <=Int B [simplification] 
 
 endmodule
 
--- a/tests/specs/benchmarks/verification.k
+++ b/tests/specs/benchmarks/verification.k
@@ -152,9 +152,6 @@ module VERIFICATION-HASKELL [symbolic, kore]
     rule ( #sizeByteArray(#ecrec( _, _, _, _ )) <=Int K ) => true requires 32 <=Int K [simplification]
 
     rule 0 <Int #asWord ( #ecrec( HASH, SIGV, SIGR, SIGS ) ) => notBool ( #ecrecEmpty ( HASH, SIGV, SIGR, SIGS ) )  [simplification]
-    
-    // <Int => <=Int (to debug ecreoverloop00-sig1-invalid-spec)
-    rule A <Int B => A <=Int B [simplification] 
 
 endmodule
 
--- a/tests/specs/benchmarks/verification.k
+++ b/tests/specs/benchmarks/verification.k
@@ -132,6 +132,10 @@ module VERIFICATION
     rule X <=Int #ceil32(X)     => true   requires 0 <=Int X                       [simplification]
     rule #ceil32(X) <Int X      => false  requires 0 <=Int X                       [simplification]
     rule #ceil32(X) <=Int pow16 => true   requires 0 <=Int X andBool X <=Int pow16 [simplification]
+    
+    // for debugging ecrecoverloop00-sig1-invalid-spec
+    rule A  <Int B => #Not ( { true #Equals B <=Int A } ) [simplification]
+    rule A <=Int B => #Not ( { true #Equals B  <Int A } ) [simplification]
 
 endmodule
 
--- a/tests/specs/benchmarks/verification.k
+++ b/tests/specs/benchmarks/verification.k
@@ -132,10 +132,6 @@ module VERIFICATION
     rule X <=Int #ceil32(X)     => true   requires 0 <=Int X                       [simplification]
     rule #ceil32(X) <Int X      => false  requires 0 <=Int X                       [simplification]
     rule #ceil32(X) <=Int pow16 => true   requires 0 <=Int X andBool X <=Int pow16 [simplification]
-    
-    // for debugging ecrecoverloop00-sig1-invalid-spec
-    rule A  <Int B => #Not ( { true #Equals B <=Int A } ) [simplification]
-    rule A <=Int B => #Not ( { true #Equals B  <Int A } ) [simplification]
 
 endmodule
 
@@ -156,6 +152,10 @@ module VERIFICATION-HASKELL [symbolic, kore]
     rule ( #sizeByteArray(#ecrec( _, _, _, _ )) <=Int K ) => true requires 32 <=Int K [simplification]
 
     rule 0 <Int #asWord ( #ecrec( HASH, SIGV, SIGR, SIGS ) ) => notBool ( #ecrecEmpty ( HASH, SIGV, SIGR, SIGS ) )  [simplification]
+    
+    // <Int <=/=> <=Int (to debug ecreoverloop00-sig1-invalid-spec)
+    rule A  <Int B => #Not ( { true #Equals B <=Int A } ) [simplification]
+    rule A <=Int B => #Not ( { true #Equals B  <Int A } ) [simplification]
 
 endmodule
 
--- a/tests/failing-symbolic.haskell
+++ b/tests/failing-symbolic.haskell
@@ -1,4 +1,3 @@
-tests/specs/benchmarks/ecrecoverloop00-sig1-invalid-spec.k
 tests/specs/bihu/collectToken-spec.k
 tests/specs/mcd/flapper-tend-guy-diff-pass-rough-spec.k
 tests/specs/mcd/vat-frob-diff-zero-dart-pass-rough-spec.k
--- a/tests/failing-symbolic.haskell
+++ b/tests/failing-symbolic.haskell
@@ -3,3 +3,4 @@ tests/specs/mcd/flapper-tend-guy-diff-pass-rough-spec.k
 tests/specs/mcd/vat-frob-diff-zero-dart-pass-rough-spec.k
 tests/specs/mcd/vow-cage-deficit-pass-rough-spec.k
 tests/specs/mcd/vow-cage-surplus-pass-rough-spec.k
+tests/specs/opcodes/create-spec.k
--- a/tests/failing-symbolic.haskell
+++ b/tests/failing-symbolic.haskell
@@ -3,4 +3,3 @@ tests/specs/mcd/flapper-tend-guy-diff-pass-rough-spec.k
 tests/specs/mcd/vat-frob-diff-zero-dart-pass-rough-spec.k
 tests/specs/mcd/vow-cage-deficit-pass-rough-spec.k
 tests/specs/mcd/vow-cage-surplus-pass-rough-spec.k
-tests/specs/opcodes/create-spec.k
--- a/tests/specs/benchmarks/ecrecoverloop00-sig1-invalid-spec.k
+++ b/tests/specs/benchmarks/ecrecoverloop00-sig1-invalid-spec.k
@@ -137,7 +137,8 @@ _
      andBool #rangeBytes(32, SIGR1)
      andBool #rangeBytes(32, SIGS0)
      andBool #rangeBytes(32, SIGS1)
-     andBool #ecrecEmpty( #bufStrict(32, HASH), #bufStrict(32, SIGV1), #bufStrict(32, SIGR1), #bufStrict(32, SIGS1) )
-
+     andBool notBool #ecrecEmpty( #bufStrict(32, HASH), #bufStrict(32, SIGV0), #bufStrict(32, SIGR0), #bufStrict(32, SIGS0) )
+     andBool         #ecrecEmpty( #bufStrict(32, HASH), #bufStrict(32, SIGV1), #bufStrict(32, SIGR1), #bufStrict(32, SIGS1) )
+     
 endmodule
 


tests/specs/benchmarks/verification.k
```diff
@@ -132,6 +132,10 @@ module VERIFICATION
     rule X <=Int #ceil32(X)     => true   requires 0 <=Int X                       [simplification]
     rule #ceil32(X) <Int X      => false  requires 0 <=Int X                       [simplification]
     rule #ceil32(X) <=Int pow16 => true   requires 0 <=Int X andBool X <=Int pow16 [simplification]
+    
+    // for debugging ecrecoverloop00-sig1-invalid-spec
+    rule A  <Int B => #Not ( { true #Equals B <=Int A } ) [simplification]
+    rule A <=Int B => #Not ( { true #Equals B  <Int A } ) [simplification]
```
ehildenb, line 139: @jinxinglim you probably need to add these lemmsa to `VERIFICATION-HASKELL` module, not the `VERIFICATION` module, so that only the Haskell backend gets them.

The Java backend doesn't really use matching logic symbols like this, and it's not having a problem with this proof anyway.

jinxinglim: Okay, I have shifted the rules to `VERIFICATION-HASKELL` module.

ehildenb:
@jinxinglim you also need to remove this proof from `tests/failing-symbolic.haskell` as well, to enable it on CI.

jinxinglim:
Seems like adding the proposed rules solves `ecrecoverloop00-sig1-invalid-spec` but causes `requires01-a0gt0-spec` to fail.

I will try to find out the reason.

jinxinglim:
Found a much simpler rule to fix `ecrecoverloop00-sig1-invalid-spec` and at the same time not causing `requires01-a0gt0-spec` to fail.

Rule/Lemma added: `A <Int B => A <=Int B`    
Proof. `A <=Int B` implies `A ==Int B` or `A <Int B`. Since `A <Int B` by hypothesis, `A <=Int B` holds. Qed.

Possible reason why this will work:
With the added rule, `0 <Int maxUInt160 &Int #asWord (#ecrec ( ... ) ++ #range ( ... ) )` will be rewritten as `0 <=Int maxUInt160 &Int #asWord (#ecrec ( ... ) ++ #range ( ... ) )`. As `<=Int` is defined as **total** function (i.e., via the `functional` attribute) and we have `maxUInt160 &Int #asWord (#ecrec ( ... ) ++ #range ( ... ) ) <= 0`, the prover will be able to figure out that either one can hold true only.

Note:
We can add the converse rule `A <=Int B => A <Int B` instead and both the proofs for `ecrecoverloop00-sig1-invalid-spec` and `requires01-a0gt0-spec` will still succeed. But mathematically, it is wrong.

jinxinglim:
My latest rule:
```k
rule A <Int B => A <=Int B [simplification] 
```
can solve for `ecrecoverloop00-sig1-invalid-spec` and `requires01-a0gt0-spec` but causes `ecrecover00-sigvalid-spec` (and maybe others) to fail.

I will try to find a rule that can succeed for `ecrecoverloop00-sig1-invalid-spec` and at the same not not failing other proofs in `benchmarks`.

ehildenb:
We should not add a rule `A <Int B => A <=Int B`, this is not correct. Simplification rules are to be interpreted as _equalities_, meaning the two sides are equal, not as _implications_, where the LHS implies the RHS (as you have proved).

jinxinglim:
Thank you @ehildenb for pointing out that simplification rules are to be interpreted as equalities and not as implications. I have removed the previously added rule and will keep that in mind from now on.

Anyway, I believe that I have found a possible fix for `ecrecoverloop00-sig1-invalid-spec.k`, which is to add the following side condition in the spec file:
```k
notBool #ecrecEmpty( #bufStrict(32, HASH), #bufStrict(32, SIGV0), #bufStrict(32, SIGR0), #bufStrict(32, SIGS0) )
```

**Reason:**
As mentioned previously, the point where the proof fails is at node 1760 where the `k` cell is as follows:
```k
<k>
      JUMPI 559 bool2Word ( 0 <Int maxUInt160 &Int #asWord ( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) ++ #range ( b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01`" [ 128 := #buf ( 32 , SIGV0:Int ) ++ #buf ( 32 , SIGV1:Int ) ] [ 192 := #buf ( 32 , SIGR0:Int ) ++ #buf ( 32 , SIGR1:Int ) ] [ 256 := #buf ( 32 , SIGS0:Int ) ++ #buf ( 32 , SIGS1:Int ) ] [ 320 := b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00" ] , #sizeByteArray ( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) ) +Int 320 , 32 -Int #sizeByteArray ( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) ) ) ) ) ~> #pc [ JUMPI ] ~> #execute ~> _Gen0:K
</k>
```
The (possible) reason is that K cannot evaluate the value of `#sizeByteArray( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) )` as there is no condition on it. So I added the new side condition so that the `#sizeByteArray` will be rewritten as `32`, following this rule:
```k
rule #sizeByteArray(#ecrec(HASH, SIGV, SIGR, SIGS)) => 32 requires notBool #ecrecEmpty(HASH, SIGV, SIGR, SIGS) [simplification]
```
This is done similarly in the successful proof of `ecrecoverloop02-sig1-invalid-spec.k` (this is where I got my inspiration/idea of this new added side condition from). Although the claim says that "if the 0th signature is valid or invalid, and the 1th signature is invalid, then the program should always revert", we have `ecrecoverloop00-sig0-invalid-spec.k` to cover the case when 0th signature is invalid.

(Note: I only add the stated side condition and did not add any new simplification rule in `verification.k` or anywhere.)

ehildenb:
@jinxinglim that solution won't work.

The proof is for this Solidity code: https://github.com/runtimeverification/evm-semantics/blob/master/tests/specs/benchmarks/ecrecover00.sol

And this proof is specifically exercising the case: "what if we know the second signature is invalid, but th efirst we _do not know_". By adding that side-condition, you are saying that the first signature is invalid, so it's the same as proof `ecrecoverloop00-sig0-invalid-spec.k`. That's not what this proof is trying to exercise.

Vovanchik25071982 (approved):

jinxinglim:
Hi @ehildenb. Sorry, long comment ahead as I will try to explain the addition of the new side condition in details.

In `ecrecoverloop00-sig0-invalid-spec.k`, the side condition for saying the 0-th signature is invalid is: 
```k
             #ecrecEmpty( #bufStrict(32, HASH), #bufStrict(32, SIGV0), #bufStrict(32, SIGR0), #bufStrict(32, SIGS0) )
```
But for `ecrecoverloop00-sig1-invalid-spec.k`, in addition to the existing side condition that says the 1-th signature is invalid:
```k
            #ecrecEmpty( #bufStrict(32, HASH), #bufStrict(32, SIGV1), #bufStrict(32, SIGR1), #bufStrict(32, SIGS1) )
```
I added the new side condition that says the **0-th signature is valid**:
```k
notBool #ecrecEmpty( #bufStrict(32, HASH), #bufStrict(32, SIGV0), #bufStrict(32, SIGR0), #bufStrict(32, SIGS0) )
```

As you mentioned, the `ecrecoverloop00-sig1-invalid-spec.k` proof is specifically exercising the case: "what if we know the second (1-th) signature is invalid, but the first (0-th) we *do not know*?". So this specification can be broken down into two sub-cases:
1. What if we know the second (1-th) signature is invalid, but the first (0-th) is *invalid*?
2. What if we know the second (1-th) signature is invalid, but the first (0-th) is *valid*?

For sub-case 1, since the `for` loop in https://github.com/runtimeverification/evm-semantics/blob/master/tests/specs/benchmarks/ecrecoverloop00.sol will go through the 0-th signature before the 1-th, sub-case 1 will be covered by `ecrecoverloop00-sig0-invalid-spec.k`.

For sub-case 2, the specification is precisely stated in `ecrecoverloop00-sig1-invalid-spec.k` after adding the new condition I have mentioned above.

-------------------------------------------------

To add on, if you compare between https://github.com/runtimeverification/evm-semantics/blob/master/tests/specs/benchmarks/ecrecoverloop00.sol and https://github.com/runtimeverification/evm-semantics/blob/master/tests/specs/benchmarks/ecrecoverloop02.sol, the only difference in these two Solidity codes is the inclusion of the input `bytes memory data` for the latter, which is not in use for the function.

Now, if you compare between `ecrecoverloop00-sig0-invalid-spec.k` and `ecrecoverloop02-sig0-invalid-spec.k`, both have the side condition saying that the 0-th signature is invalid (and hence, both proofs passed):
```k
             #ecrecEmpty( #bufStrict(32, HASH), #bufStrict(32, SIGV0), #bufStrict(32, SIGR0), #bufStrict(32, SIGS0) )
```

Similarly, if you observe between `ecrecoverloop00-sigs-valid-spec.k` and `ecrecoverloop02-sigs-valid-spec.k`, both have the two side conditions that 0-th and 1-th signatures are valid (and hence, both proofs passed):
```k
notBool #ecrecEmpty( #bufStrict(32, HASH), #bufStrict(32, SIGV0), #bufStrict(32, SIGR0), #bufStrict(32, SIGS0) )
notBool #ecrecEmpty( #bufStrict(32, HASH), #bufStrict(32, SIGV1), #bufStrict(32, SIGR1), #bufStrict(32, SIGS1) )
```

However, if you compare between `ecrecoverloop00-sig1-invalid-spec.k` and `ecrecoverloop02-sig1-invalid-spec.k`, the latter has the side condition that the former does not have:
```k
notBool #ecrecEmpty( #bufStrict(32, HASH), #bufStrict(32, SIGV0), #bufStrict(32, SIGR0), #bufStrict(32, SIGS0) )
```

This is probably why the case for `ecrecoverloop00-sig1-invalid-spec.k` failed but the case for `ecrecoverloop02-sig1-invalid-spec.k` passed. I believe a side condition must be given for the value of `#sizeByteArray( #ecrec ( #buf ( 32 , HASH:Int ) , #buf ( 32 , SIGV0:Int ) , #buf ( 32 , SIGR0:Int ) , #buf ( 32 , SIGS0:Int ) ) )` to be evaluated. 
> **Note:** I tried to remove the condition `notBool #ecrecEmpty( #bufStrict(32, HASH), #bufStrict(32, SIGV0), #bufStrict(32, SIGR0), #bufStrict(32, SIGS0) )` from `ecrecoverloop02-sig1-invalid-spec.k` and it went from passing to failing the proof. The error that surfaced (stuck at `JUMPI`) is exactly the same one in `ecrecoverloop00-sig1-invalid-spec.k` without the new added condition.

ehildenb:
You are making the proof statement _less general_ than it is intended to be with this change. You are constraining the initial state more, so teh proof is not as strong.

The _point_ of this proof is to show that "we will _always_ fail if the second signature is invalid, regardless of whether the first signature is valid or not". You have changed the proof to say that "we will _always_ fail if the first signature is valid _and_ the second signature is invalid", which is a different and less general statement.

Instead, you must find a lemma that allows the haskell backend to not get stuck at that point, because it should be able to show _on both branches_ from there that the proof still passes. Either the first signature is invalid and we revert because of that, or it's valid and we revert because the second is invalid.

ehildenb:
@jinxinglim I think solution 10 should be fine, it keeps the proof as it was, but allows the backend to simplify the third branch to infeasible as it should.

jinxinglim:
> @jinxinglim I think solution 10 should be fine, it keeps the proof as it was, but allows the backend to simplify the third branch to infeasible as it should.

Hi @ehildenb. Sorry, may I know which "solution 10" are you referring to?

jinxinglim:
> You are making the proof statement _less general_ than it is intended to be with this change by constraining the initial state more.
> 
> The _point_ of this proof is to show that "we will _always_ fail if the second signature is invalid, regardless of whether the first signature is valid or not". You have changed the proof to say that "we will _always_ fail if the first signature is valid _and_ the second signature is invalid", which is a different and less general statement.
> 
> Instead, you must find a lemma that allows the haskell backend to not get stuck at that point, because it should be able to show _on both branches_ from there that the proof still passes. Either the first signature is invalid and we revert because of that, or it's valid and we revert because the second is invalid.

@ehildenb Okay. Does that mean that we should
1. remove the side condition `notBool #ecrecEmpty( #bufStrict(32, HASH), #bufStrict(32, SIGV0), #bufStrict(32, SIGR0), #bufStrict(32, SIGS0) )` from https://github.com/runtimeverification/evm-semantics/blob/master/tests/specs/benchmarks/ecrecoverloop02-sig1-invalid-spec.k, and
2. add a lemma such that **both** `ecreoverloop02-sig1-invalid-spec.k` and `ecreoverloop00-sig1-invalid-spec.k` proofs succeed without the above side condition?

ehildenb:
Yes @jinxinglim that's what we should do.

palinatolmach:
@jinxinglim @ehildenb do we still plan to do this, or should we close the issue?

ehildenb:
I've closed, likely we need to re-check if any more proofs can be enabled with Haskell backend.

PR #1412, opened 2022-10-05
Implement startPrank and stopPrank rules

anvacaru:
Implementing two `startPrank` and the `stopPrank` cheat codes.
Adding a new `<active>` cell to the `<prank/>` subconfiguration to signal if a prank is active.
Adding two new `CALL` rules with a higher priority to inject the pranked `msg.sender` and `tx.origin`.


anvacaru:
Currently blocked on #1414

foundry.md
```diff
@@ -561,7 +641,62 @@ Utils
...
+
+- `#setPrankWithOrigin NEWCALLER NEWORIGIN` will set the `<prank/>` subconfiguration for the given account.
+
+```k
+    syntax KItem ::= "#setPrankWithOrigin" Int Int [klabel(foundry_setPrankWithOrigin)]
+ // -----------------------------------------------------------------------------------
+    rule <k> #setPrankWithOrigin NEWCALLER NEWORIGIN => . ... </k>
+         <callDepth> CD </callDepth>
+         <caller> CL </caller>
+         <origin> OG </origin>
```
ehildenb, line 672: What happens if you start a prank with one already active? Maybe we can have `#setPrank ACCT => #setPrankWithOrigin ACCT .Account`, so that we only have one rule updating the prank sub-configuration.

Then we can also have a rule for `#setPrankWithOrigin ...` which handles the case where `<active> true </active>` (by going to a `#foundry_error "Prank already active"` state somehow?

foundry.md
```diff
@@ -561,7 +641,62 @@ Utils
...
+           <depth> _ => CD </depth>
+           <singleCall> _ => false </singleCall>
+         </prank>
+```
+
+- `#endPrank` will end the prank, restoring the previous caller and origin to the `<caller>` and `<origin>` cells in the configuration.
+
+```k
+    syntax KItem ::= "#endPrank" [klabel(foundry_endPrank)]
+ // -------------------------------------------------------
```
ehildenb, line 688: Similarly, what happens if you do `#endPrank` when `<active> false </active>`? Maybe we should have a rule which goes to `#foundry_error "No prank started"` instead.

Also, what about case where you do a `#startPrank ACCT`, then an `#endPrank`? It seems to me that you'll end up overriding the `<origin>` cell with `.Account`, because this rule doesn't check if `<prevOrigin>` is `.Account`. Maybe we need a rule to handle `<proveOrigin> .Account </prevOrigin>` case?

ehildenb: Ah wait! I misread the semantics! It appears that you already are handling this smoothly by having separate `<newOrigin>` and `<prevOrigin>` cell. Nice.

foundry.md
ehildenb: Similarly, if you end a prank with none active, we should go to an error state. Please also add tests of these cases, and add those tests to the `tests/foundry/exclude` list for now.

anvacaru: > What happens if you start a prank with one already active? Maybe we can have `#setPrank ACCT => #setPrankWithOrigin ACCT .Account`, so that we only have one rule updating the prank sub-configuration.
> 
> Then we can also have a rule for `#setPrankWithOrigin ...` which handles the case where `<active> true </active>` (by going to a `#foundry_error "Prank already active"` state somehow?

Starting a prank while another one is active will return a Fail error `[FAIL. Reason: You have an active prank already.]`. Marking the test with `testFail` will make the test pass.

ehildenb (approved):

Issue #1302, opened 2022-07-27
Update to BSD-3-clause license


ehildenb:
Same license used for K.

PR #1417, opened 2022-10-07
Update License

anvacaru:
Fixes: #1302 

Updating to the same license type as K uses.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 BSD 3-Clause License
 
-Copyright (c) 2022, Runtime Verification Inc.
+Copyright (c) 2016-2022, Runtime Verification Inc.
 All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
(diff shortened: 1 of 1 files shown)


LICENSE
```diff
@@ -1,60 +1,29 @@
-==============================================================================
-The EVM Semantics in K Release License
-==============================================================================
-University of Illinois/NCSA
-Open Source License
+BSD 3-Clause License
 
-Copyright (c) 2009-2019 University of Illinois at Urbana-Champaign.
+Copyright (c) 2022, Runtime Verification Inc.
```
ehildenb, line 3:
```suggestion
Copyright (c) 2106-2022, Runtime Verification Inc.
```

ehildenb (approved):

LICENSE
anvacaru: I assume it's 2016.

Issue #1385, opened 2022-09-21
Include solc in runtimeverificationinc/runtimeverification-evm-semantics docker image


enzoevers:
When trying to run `solc-to-k` in  [runtimeverificationinc/runtimeverification-evm-semantics:ubuntu-focal-master](https://hub.docker.com/r/runtimeverificationinc/runtimeverification-evm-semantics/tags) I get an error that `solc` cannot be found. And indeed, when going into the container solc is not installed.

If I'm not mistaking, [package/docker/Dockerfile.ubuntu](https://github.com/runtimeverification/evm-semantics/blob/5ec6128ca399dc770eb1ae32bb5dd3baeb6152e5/package/docker/Dockerfile.ubuntu) is the dockerfile used to create this docker image (looking at the [Jenkins](https://github.com/runtimeverification/evm-semantics/blob/96392cc2f4ec1ae593a0fa71dfdd0bcd1cac6fcc/Jenkinsfile#L133) file).

Could `solc` be installed in this image too so that `solc-to-k` works out of the box?

Or should `solc` be part of the `Depends` section in [control.focal](https://github.com/runtimeverification/evm-semantics/blob/5ec6128ca399dc770eb1ae32bb5dd3baeb6152e5/package/debian/control.focal#L13)?

anvacaru:
We should copy this line from the main Dockerfile https://github.com/runtimeverification/evm-semantics/blob/5ec6128ca399dc770eb1ae32bb5dd3baeb6152e5/Dockerfile#L11

omarzd:
Hi. I get this same error. I followed the README today, so my build should include the fix, right @anvacaru? Here is a dump from my terminal:
```
(venv) omarzd@PC:~/evm-semantics$ make build-prove
. .build/venv/bin/activate
. .build/venv/bin/activate && kevm solc-to-k tests/specs/examples/ERC20.sol ERC20 --pyk --verbose --profile --verbose --definition .build/usr/lib/kevm/haskell --main-module ERC20-VERIFICATION > tests/specs/examples/erc20-bin-runtime.k
INFO 2022-11-03 17:46:25,278 kevm_pyk.solc_to_k - Running: solc --standard-json
Traceback (most recent call last):
  File "/usr/lib/python3.8/runpy.py", line 194, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.8/runpy.py", line 87, in _run_code
    exec(code, run_globals)
[...]

PR #1418, opened 2022-10-11
Remove docker hub stage & file

anvacaru:
Fixes: https://github.com/runtimeverification/evm-semantics/issues/1385 by copying the imports from https://github.com/runtimeverification/evm-semantics/blob/master/Dockerfile#L17 .


package/docker/Dockerfile.ubuntu
```diff
@@ -5,13 +5,41 @@ ARG BASE_IMAGE=focal
...
-                        git             \
-                        python          \
-                        python3-pip
+RUN    apt-get update                                   \
+    && apt-get install --yes software-properties-common \
+    && add-apt-repository ppa:ethereum/ethereum
+
+RUN    apt-get update            \
+    && apt-get upgrade --yes     \
+    && apt-get install --yes     \
```
tothtamas28, line 14: If I understand correctly, the goal is to make `solc-to-k` work in the container. Is this the minimal list of dependencies to achieve that?

ehildenb:
Yes, we don't push it to dockerhub (or build it on CI) anymore. I would honestly advocate for the removal of this particular step in the build.

ehildenb:
And this is certainly a longer list of dependencies than needed to get solc-to-k working. KEVM package already pulls in most of these deps.

tothtamas28 (approved):

Issue #1416, opened 2022-10-07
Hash collision in KEVM


anvacaru:
Currently, KEVM has several lemmas which use clauses like this: [`0 <=Int X andBool X <Int 10`](https://github.com/runtimeverification/evm-semantics/pull/1270/files#diff-7b449c34f5bca7c7fd19162bbdf846d87463e4d75229bf0e37393ac732f5b45cR210).

What this means usually is "the number is small", and is used for "unlikely to find a hash collision at this offset".

- Could we generalize it to say "could not find a hash collision at any _concrete_ offset"?
- Or could we factor out a `#rangeSmall` predicate which sets that small integer `10` once for everywhere?

PR #1419, opened 2022-10-11
Add rangeSmall to range of types

anvacaru:
fixes: https://github.com/runtimeverification/evm-semantics/issues/1416

Created a new range to check if ` 0 <=Int X andBool X <Int 10`.

### DIFF
--- a/word.md
+++ b/word.md
@@ -260,6 +260,7 @@ Range of types
                   | #rangeAddress ( Int )             [alias]
                   | #rangeBytes   ( Int , Int )       [alias]
                   | #rangeNonce   ( Int )             [alias]
+                  | #rangeSmall   ( Int )             [alias]
  // ---------------------------------------------------------
     rule #rangeBool    (            X ) => X ==Int 0 orBool X ==Int 1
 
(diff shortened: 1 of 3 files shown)


word.md
```diff
@@ -304,6 +305,7 @@ Range of types
     rule #rangeAddress (            X ) => #range ( minUInt160      <= X <  pow160          )
     rule #rangeBytes   (   N ,      X ) => #range ( 0               <= X <  1 <<Byte N      )
     rule #rangeNonce   (   X          ) => #range ( 0               <= X < maxUInt64        )
+    rule #rangeSmall   (   X          ) => #range ( 0               <= X < 10               )
```
tothtamas28, line 308: Is this range only relevant for proof heuristics, or does it also have a significance for the semantics of EVM? (I'm thinking whether this is the right place to include it.)

anvacaru: I've only seen that range being used in proofs. I also considered `lemmas.k`, but that file only contains simplifications.

ehildenb: I think this is an appropriate place as well, it's next to the other range predicates.

ehildenb (approved):

PR #1424, opened 2022-10-18
Create individual definition for foundry

anvacaru:
Currently, the generated Haskell definition contains both the semantics for EVM and a new semantics for the Foundry property testing. 
The second one contains rules and productions which are not supposed to apply to the semantics of the EVM under normal circumstances, and which are causing our current proofs for EVM to fail ( see #1412 ).
The changes in this PR generate a new kompiled definition for the `FOUNDRY` module under a `foundry` folder, while the `haskell` will contain only the kompiled semantics which have `EDSL` as the main module.

### DIFF
--- a/tests/foundry/foundry.k.check.expected
+++ b/tests/foundry/foundry.k.check.expected
@@ -1,10 +1,8 @@
-requires "edsl.md"
 requires "foundry.md"
 requires "lemmas/lemmas.k"
 requires "lemmas/int-simplification.k"
 
 module ADDRTEST-BIN-RUNTIME
-    imports public EDSL
     imports public FOUNDRY
     imports public LEMMAS
     imports public INT-SIMPLIFICATION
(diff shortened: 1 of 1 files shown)


tests/foundry/foundry.k.check.expected
```diff
@@ -1,9 +1,11 @@
 requires "edsl.md"
+requires "foundry.md"
 requires "lemmas/lemmas.k"
 requires "lemmas/int-simplification.k"
 
 module ADDRTEST-BIN-RUNTIME
     imports public EDSL
+    imports public FOUNDRY
     imports public LEMMAS
     imports public INT-SIMPLIFICATION
```
ehildenb, line 10: I guess `FOUNDRY` already imports `EDSL`? So we can remove the `EDSL` import from these modules now?

anvacaru: That's correct. I'm updating that now.

anvacaru: @ehildenb  `FOUNDRY` currently imports only the `BIN-RUNTIME` module from `EDSL` (see line 83 from `foundry.md` above). I could just import the entire `EDSL` module in `FOUNDRY` and remove `EDSL` import here.

PR #1425, opened 2022-10-19
Minimal example of the issue with collectToken-spec.k

fiedlr:
This PR is not intended for a merge (at least not yet), but to describe the progress of fixing `collectToken-spec.k` by means of a minimal example.

# How to reproduce
If you have `evm-semantics` built together with `k-deps` (locally), the fastest way to reproduce this is to `cd` the main project directory, build tests
```
make build-prove
```
start the `kore-repl`:
```
KORE_REPL_OPTS='--smt-timeout 10000' kevm prove tests/specs/bihu/collectToken-spec.k --definition tests/specs/bihu/verification/haskell/ --debugger --backend haskell
```
and then run
```
load failing-branch-sesion
```
in the REPL.

*Note* If you have the newest `pyk` installed, you can use the `pykprint.sh` script to pinpoint the important cells like this:
```
config | bash pykprint.sh "tests/specs/bihu/verification/haskell/" "<k>,<pc>,<wordStack>,<statusCode>,<storage>"
```
You just need to change the path to your `pyk` in `pykprint.sh`.

# Problem (Updated)
The prover doesn't realize that the side condition is false, leading to a huge subtree that isn't pruned and should be. Eventually, the test fails on a false assert in this incorrect subtree.
At its root `977`, the side condition is false, in a similar way as the `<, >=` problem mentioned in https://github.com/runtimeverification/haskell-backend/issues/2997. See the [new post below for details](https://github.com/runtimeverification/evm-semantics/pull/1425#issuecomment-1323956926).

# My original thoughts (no longer relevant!)

## Description
The spec seems to fail at two places (see the entire proof tree in `proof_newest4.svg`), both places already have false conditions and seem to suffer from the same problem: the branch with a failing assert is not pruned and it is actually the same assert (didn't doublecheck if the side condition is the same). This is the code for the assert:
```
JUMPI 719 bool2Word ( chop ( COLLECTED:Int +Int BAL:Int *Int 10 /Int 100 *Int NOW:Int -Int START:Int /Int 31536000 -Int COLLECTED:Int ) <=Int COLLECTED:Int +Int BAL:Int *Int 10 /Int 100 *Int NOW:Int -Int START:Int /Int 31536000 )
```
When evaluated to true, it skips an invalid instruction (which is how the old Solidity translates failed asserts). The wrong branch that is not pruned evaluates this to false, gets to the invalid opcode and halts.

*Note* for debugging purposes I minimized it to this specific subtree (see `proof_branch_newest10.svg`) because it runs much faster. That is also what the `load` script does, it explores only this subtree. I also removed the passing claims and switched to virtual gas for now to avoid any issues with gas.

## Quirks
I added a test to `functional-spec.k` testing if the expression inside the assert evaluates to true. It passes without any changes:
```
kevm prove tests/specs/bihu/functional-spec.k  --backend haskell  --definition tests/specs/bihu/functional-spec/haskell
```
The side condition in `functional-spec.k` contains all the side conditions containing the used variables from the node `1973` (as identified in `proof_branch_newest10.svg`) This is what the functional test returns:
```
kore-exec: [4248467] Warning (WarnTrivialClaim):
    Claim proven during initialization: /home/ja/dev/evm-semantics/tests/specs/bihu/functional-spec.k:22:11-47:79
    The left-hand side of the claim may be undefined.
#Top
```
When I played around with this, it seems like the side condition already in `1973` is false because this passes:
```
claim <k> runLemma(true) => doneLemma(false) ...</k>
      requires
         #Not ( { 0 #Equals COLLECTED:Int +Int BAL:Int *Int 10 /Int 100 } )
         andBool #Not ( { 0 #Equals COLLECTED:Int +Int BAL:Int } )

         andBool #Not ( { true #Equals NOW:Int -Int START:Int <=Int 0 } )
         andBool #Not ( { true #Equals NOW:Int <=Int START:Int } )

         andBool #Not ( { true #Equals 115792089237316195423570985008687907853269984665640564039457584007913129639936 <=Int COLLECTED:Int +Int BAL:Int *Int 3153600 } )
         andBool #Not ( { true #Equals 115792089237316195423570985008687907853269984665640564039457584007913129639936 <=Int BAL:Int } )
         andBool #Not ( { true #Equals 115792089237316195423570985008687907853269984665640564039457584007913129639936 <=Int COLLECTED:Int } )
         andBool #Not ( { true #Equals 115792089237316195423570985008687907853269984665640564039457584007913129639936 <=Int NOW:Int } )
         andBool #Not ( { true #Equals 115792089237316195423570985008687907853269984665640564039457584007913129639936 <=Int NOW:Int -Int START:Int } )
         andBool #Not ( { true #Equals 115792089237316195423570985008687907853269984665640564039457584007913129639936 <=Int START:Int } )

         andBool { true #Equals 0 <=Int BAL:Int }

         andBool { true #Equals 0 <=Int COLLECTED:Int }
         andBool { true #Equals 0 <=Int COLLECTED:Int +Int BAL:Int }
         andBool { true #Equals 0 <=Int NOW:Int }

         andBool { true #Equals 0 <=Int START:Int }
         andBool { true #Equals #accumulatedReleasedTokens ( BAL:Int , COLLECTED:Int , START:Int , NOW:Int ) <Int BAL:Int +Int COLLECTED:Int +Int -10 }
         andBool { true #Equals #accumulatedReleasedTokens ( BAL:Int , COLLECTED:Int , START:Int , NOW:Int ) >Int COLLECTED:Int +Int 3 }
         andBool { true #Equals NOW:Int -Int START:Int /Int 31536000 <=Int 0 }
```
So it seems that the prover doesn't realize the side condition is false and still explores this branch even though it should have pruned it earlier. 

But why does the prover realize it in the test and not in the actual claim run?  Any ideas how to proceed?

When I tried to enforce that the assert evaluates to `true` by extending the side condition, it didn't help. The only thing that helped was making `chop` simplify by additional assumptions about its arguments (that they are less than `pow256`). But then a similar error occurred elsewhere.


ehildenb:
Stale.

PR #1427, opened 2022-10-20
Generate individual KCFGs for each method in contract

ehildenb:
This makes sure that every function in the contract has a KCFG generated for it, so that we can begin running proofs for any function and get summarized results. Two things:

- Directly constructs KCFGs, instead of constructing KClaim and then converting them to KCFGs for storage.
- Generates KCFGs for _every_ method in the contracts, test method or not. Makes sure that we add `foundry_success` predicate to the `test*` methods, and `notBool foundry_success` predicate to the `testFail` methods. The others it adds no post-condition to.
- Moves the generation of KCFGs to `foundry-prove` time, instead of at `foundry-kompile` time, and moves the `--reinit` argument to `foundry-prove` as well.
- Generates an individual file for each KCFG, instead of on file with all the KCFGs (should give faster CFG loading).
- Filters tests based on the `--test` and `--exclude-test` list _before_ loading any CFGs, to minimize how much needs to be loaded.
- Removes some now unused functions.
- Renames a test in `DealTest` to avoid a contract name clash.

This one may be easier to review commit-by-commit.


kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -431,15 +431,18 @@ def _init_term(
...
-    if not failing:
-        return CTerm(final_term).add_constraint(mlEqualsTrue(foundry_success))
-    else:
-        return CTerm(final_term).add_constraint(mlEqualsTrue(notBool(foundry_success)))
+    final_cterm = CTerm(final_term)
+    if is_test:
+        if not failing:
+            return final_cterm.add_constraint(mlEqualsTrue(foundry_success))
+        else:
+            return final_cterm.add_constraint(mlEqualsTrue(notBool(foundry_success)))
```
anvacaru, line 444: Why do we need to add a constraint on the `foundry_success` predicate here?

ehildenb: Here we are adding, to the `final_cterm`, the constraint `{ true #Equals foundry_success(...) }`. We're not adding a constarint to `foundry_success`, we're adding the `foundry_success` constraint to the CTerm.

ehildenb: You can find the implementation of `add_constraint` here: https://github.com/runtimeverification/pyk/blob/6bb1fcb27ddfa0d9c9bd45699951ad53ad9d678d/src/pyk/cterm.py#L95

ehildenb: That predicate is only added for `test*` functions, see the outer `if ...` statement.

ehildenb: The one where it says `if is_test: ...`

anvacaru (approved):
The `foundry.k.check.expected` file needs to be updated. Other than that, it looks good to me. I would also ask @tothtamas28 for a review on this one.

PR #1431, opened 2022-10-21
Handle auto-updates from pyk repo

ehildenb:
This makes sure that we are syncing the pyk dependency from file `deps/pyk_release`, and updating the poetry lock files and then nix files accordingly.


ehildenb:
Evidence this works: https://github.com/runtimeverification/evm-semantics/actions/runs/3299822013/jobs/5443629685 (and the subsequent update to the various poetry files pushed to this PR).

tothtamas28 (approved):

PR #1433, opened 2022-10-22
Organize repository, move sources into subdirectories

ehildenb:
This should help with some problems people have with files being auto-included, and with `kevm` being in the wrong place (running it from the root of the repo instead of from the installed location).

- Move K sources into `include/kframework/` subdirectory.
- Do not include non-generic lemmas into packaged product.
- Move generic lemmas `(tests/specs => include/kframework/lemmas)/(lemmas,int-simplification).k`.
- Make adjustments to the cmake builds.
- Remove now unneeded `-I` options that were there.
- Move `kevm` script into subdirectory `bin/`.
- Adjust flake.nix file.
- Make it so `LEMMAS` module depends on `BUF`, not on `EDSL`, and move some definitions around to accomodate that (`#ceil32` macro is moved).


anvacaru (approved):
Looks good to me 👍

PR #1436, opened 2022-10-25
Implement expectRevert cheat codes

anvacaru:
Fixes: https://github.com/runtimeverification/evm-semantics/issues/1345

Added a new subconfiguration named `<expected/>` to keep track of the fields required for the `expectRevert` cheat code.
Included new tests.


tests/foundry/test/ExpectRevertTest.sol
```diff
@@ -0,0 +1,50 @@
...
+    //     vm.expectRevert(bytes4("Revert"));
+    //     reverter.revertWithReason("But fail.");
+    // }
+
+    // function test_expectRevert_bytes4() public {
+    //     Reverter reverter = new Reverter();
+    //     vm.expectRevert(bytes4("Revert"));
+    //     reverter.revertWithReason("Revert");
+    // }
+}
```
ehildenb, line 50: Why are these tests commented? Better to uncomment them and add them to `tests/foundry/exclude`.

foundry.md
```diff
@@ -458,12 +466,75 @@ This rule then takes from the function call data the account using `#asWord(#ran
...
+```
+
+Change the status code from `EVMC_SUCCESS` to `EVMC_REVERT` if a revert is expected but the call succeded.
+
+```{.k .bytes}
+    rule <statusCode> EVMC_SUCCESS => EVMC_REVERT </statusCode>
+         <k> #halt ~> #return _RETSTART _RETWIDTH ... </k>
+         <callDepth> CD </callDepth>
+         <expected>
+           <expectedRevert> true => false </expectedRevert>
```
ehildenb, line 514: So, if I'm reading this correctly, `expectRevert` is only working for 1 top-level revert. It won't work for nested reverts, or for multiple reverts. Is that correct?

Can we make a test case for a nested revert and a multi-revert (two calls that both revert, and the expectRevert should absorb the first one, but the second one causes test failure with a different message). We should test this behavior.

anvacaru: I've added the requested tests.

ehildenb (approved):

PR #1438, opened 2022-10-27
Pass --md-selector to kprove too

ehildenb:
Due to https://github.com/runtimeverification/k/issues/3020, we have to make sure that `kprove` gets the appropriate tangle tags too.

We also turn of `--format-failures` by default on Java backend, an option that you probably only want sometimes.


jberthold (approved):
LGTM (from comparing to `run_kompile`)

PR #1446, opened 2022-11-02
Implement prank cheat code

anvacaru:
Implement the `prank(address)` and `prank(address, address)` cheat codes that set the `msg.sender` and `tx.origin` only for the next call.


include/kframework/foundry.md
```diff
@@ -511,6 +511,18 @@ Finally, the original sender of the transaction, `ACCTFROM` is changed to the ne
...
+We define a new rule for the `#halt ~> #return _ _` production that will trigger the `#endPrank` rules if the prank was set only for a single call.
+
+```{.k .bytes}
+    rule <k> (. => #endPrank) ~> #halt ~> #return _RETSTART _RETWIDTH ... </k>
+         <prank>
+           <singleCall> true </singleCall>
+           ...
+         </prank>
+      [priority(40)]
+
```
ehildenb, line 523: Sholud we also be checking that we're at the same calldepth as before? That way we're not stopping and ending the prank on some nested call?

anvacaru: I missed that. Thanks for pointing it out!

include/kframework/foundry.md
```diff
@@ -524,14 +539,31 @@ The `#loadAccount` production is used to load accounts into the state if they ar
...
+function prank(address) external;
+function prank(address sender, address origin) external;
+```
+
+```{.k .bytes}
+    rule [foundry.call.prank]:
+         <k> #call_foundry SELECTOR ARGS => #loadAccount #asWord(ARGS) ~> #setPrank #asWord(ARGS) .Account true ... </k>
+      requires SELECTOR ==Int selector ( "prank(address)" )
+
+    rule [foundry.call.rankWithOrigin]:
```
ehildenb, line 562:
```suggestion
    rule [foundry.call.prankWithOrigin]:
```

ehildenb (approved):

PR #1443, opened 2022-10-30
Incremental exploration based kprove using rpc endpoints

ehildenb:
~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/1447~
~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/1440~
~Blocked on: https://github.com/runtimeverification/pyk/pull/100~
~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/1458~
~Blocked on: https://github.com/runtimeverification/pyk/pull/112~
~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/1462~
~Blocked on: https://github.com/runtimeverification/pyk/pull/118~
~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/1476~
~Blocked on: https://github.com/runtimeverification/haskell-backend/pull/3386~
~Blocked on: https://github.com/runtimeverification/k/pull/3059~
~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/1484~
Blocked on: https://github.com/runtimeverification/evm-semantics/pull/1478
Blocked on: https://github.com/runtimeverification/evm-semantics/pull/1480

This brings in a decent amount of functionality needed for making progress incremental on Foundry proofs:

- Adds `KEVM.{is_terminal,extract_branches}`, which help with auto-exploration.
- Adds auto-exploration strategy based prover, which stores branch points for inspection (as opposed to using direct calls to `kprove`).
  - Use option `--max-iterations` to limit how many calls to prover to make while exploring.
  - Use option `--max-depth` to limit how long each exploration step will let the prover run.
  - Use option `--no-simplify-init` to turn of simplification of the initial state (up to 2 minutes at proof startup, does not need to be repeated on subsequent runs).

Doing something like `kevm foundry-prove --test MyContract.MyTest --max-depth 100` will make sure that every 100'th step is saved in the KCFG. That way you can save progress in case the prover crashes, and you can resume exploration from a later state with more lemmas in context.

PR #1447, opened 2022-11-03
CFG visualization and progress saving

ehildenb:
This PR is the simpler components pull out of https://github.com/runtimeverification/evm-semantics/pull/1443.

In particular, it has:

This brings in a decent amount of functionality needed for making progress incremental on Foundry proofs:

- Adds `KEVM.add_invariant`, and makes sure that the initial LHS states used for proofs are in the language invariant.
- The `kevm foundry-prove` command will now save the final counterexample state in the CFG on disk, so progress can be resumed from there.
- Adds `kevm foundry-show OUT_DIR Contract.Test` for displaying current CFG. Has options:
  - `--[no-]minimize`: minimize the configuration output.
  - `--node NODE_ID`: also display the selected node's configuration and constraints (respects `--[no-]minimize`, repetition allowed).
  - `--node-delta NODE_ID_1,NODE_ID_2`: display the configuration delta between the given nodes (respects `--[no-]minimie`, repetition allowed).
- Adds `kevm foundry-list OUT_DIR` for listing the current CFGs, and their progress towards being proved. Has options:
  - `--[no-]details`: show the details about progress to being proved for each CFG, instead of just their passed/failed state.

If you do something ilke `kevm foundry-prove --depth 100 ...`, the proof will fail at depth 100, and that state will be saved in the KCFG. Then you can continue progress from there with `kevm foundry-prove ...` again, and it will start from the 100th state the next time (you can add the `--lemma` argument the second call, for example). You can add the `--reinit` option to `foundry-prove` to start the proof over.


ehildenb:
@lucasmt and @qian-hu if you could test that the new functionality (`list`, `show`, and proof resuming) are working as expected, taht would be good.

ehildenb:
@qian-hu if everything else is working you should not press "Request Changes", because then I need to actually take some action to address your changes. See our [GitHub Guidelines](https://docs.google.com/document/d/1brNoaHuTDMtXkSvF17E4eOGOiS-4cN8JJCOXsKFZz7c/edit). As it is, you haven't actually given me anything actionable to do, I need to see the state that is not working for parsing, and ideally to see the entire input you are trying to run.

Also, it's a known issue that it cannot handle proof resuming from inside calls. I will be able to fix that too, but it will take me a couple weeks, it's not as straightforward as it sounds.

lucasmt:
I've pushed a commit adding parenthesis when printing the `up/Int` operator, since it was causing a parser error when running `kevm foundry-prove`. Some constraints written to `spec.k` for the test I was trying to run contained expressions of the form `#sizeByteArray(X) up/Int 32 *Int 32`, and the parser considers them ambiguous without parentheses.

lucasmt:
Now I am also getting the same error that @qian-hu mentioned when I try to run `kevm foundry-prove` the second time. For me the error looks like this:
```
[Error] Inner Parser: Parse error: unexpected token '_EXITCODE_CELL' following
token '_'.
	Source(/home/lucasmt/lsp-smart-contracts-verification/out/specs/erc725ytest-testsetdata1-spec.k)
	Location(40,19,40,33)
	40 |	               ( __EXITCODE_CELL => ?_EXIT_CODE_CELL_c89e25d9 )
	   .	                  ^~~~~~~~~~~~~~
```
It looks like there is an extra underscore being added to the variable names.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -325,38 +327,137 @@ def exec_foundry_prove(
...
+    node_deltas: Iterable[Tuple[str, str]] = (),
+    minimize: bool = True,
+    **kwargs: Any,
+) -> None:
+    definition_dir = foundry_out / 'kompiled'
+    use_directory = foundry_out / 'specs'
+    use_directory.mkdir(parents=True, exist_ok=True)
+    kcfgs_dir = foundry_out / 'kcfgs'
+    foundry = Foundry(definition_dir, profile=profile, use_directory=use_directory)
+    kcfg_file = kcfgs_dir / f'{test}.json'
```
tothtamas28, line 416: Consider extracting `definition_dir`, `use_dir`, `kcfg_dir`, `kcfg_file` as functions of a new module `config.py`, e.g.

```python3
def kcfg_file(out_dir: Path, test_name: str) -> Path:
    return kcfg_dir(out_dir) / f'{test_name}.json'
```

kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -49,3 +103,35 @@ def abstract_cell_vars(cterm: KInner, keep_vars: Collection[KVariable] = ()) ->
         if type(subst[s]) is KVariable and not is_anon_var(subst[s]) and subst[s] not in keep_vars:
             subst[s] = abstract_term_safely(KVariable('_'), base_name=s)
     return substitute(config, subst)
+
+
+def sanitize_config(defn: KDefinition, init_term: KInner) -> KInner:
```
tothtamas28, line 108: Let's create an issue to mark that there is duplicated functionality between `kevm_pyk` and `ksummarize` that eventually should be factored out.

tothtamas28 (approved):

ehildenb:
@qian-hu thanks for the more detail. I've added a test that this does not happen, and fixed this bug you found in the state saving.

kevm-pyk/src/kevm_pyk/utils.py
ehildenb: I don't think this duplicated functionality will be long-lived. Plan to remove calls to sanitize_config once we have `execute` endpoint working (in my local branch, but not pushed up yet).

kevm-pyk/src/kevm_pyk/__main__.py
ehildenb: You mean as helpers so we're not re-computing them over and again? Maybe they could be fields in the `KEVM` class so they're jsut set at function init?

ehildenb: I took a quick look at this, and seems like more trouble than it's worth at the moment. Will consider it in my local tree, see if I can find a solution.

tothtamas28: > You mean as helpers so we're not re-computing them over and again

Re-computing the values is fine. The point is to have a single definition of these expressions so that e.g. the folder layout can be easily reconfigured.

flake.nix
```diff
@@ -35,6 +35,7 @@
           autoconf
           cmake
           clang
+          curl
```
ehildenb, line 38: I'm trying to make it so that `curl` command is available to the command `nix build '.#kevm-test'`, but this doesn't seem to have done the trick. @Baltoli any quick suggestions?

ehildenb: I seem to have made it past teh `curl` issue.

ehildenb: But now I get this one: 
image

Are nix derivations not allowed to reach the network?

PR #1448, opened 2022-11-07
Support expectStaticCall cheat code

anvacaru:
Added cheat codes to expect opcodes like `STATICCODE` `DELEGATECALL` `CALL` `CREATE` and `CREATE2`
Added new subconfiguration for expected opcodes.
Added a new sort to differentiate between opcodes.


include/kframework/foundry.md
```diff
@@ -539,6 +552,136 @@ If the `expectRevert()` selector is matched, call the `#setExpectRevert` product
...
+If the `expect*OPCODE*` selector is matched, the rule will load the account into the state and add the `#setExpectOpcode` production to the K cell to initialize the `<expectedOpcode/>` subconfiguration with the given parameters.
+
+```{.k .bytes}
+    rule [foundry.call.expectStaticCall]:
+         <k> #call_foundry SELECTOR ARGS
+          => #loadAccount #asWord(#range(ARGS, 0, 32))
+          ~> #let CODE_START = 96 #in
+          #let DATA_LENGTH = #asWord(#range(ARGS, 64, 32)) #in
+          #setExpectOpcode #asWord(#range(ARGS, 0, 32)) #range(ARGS, CODE_START, DATA_LENGTH) 0 Static
+          ...
```
ehildenb, line 573:
```suggestion
          ~> #let CODE_START = 96 #in
             #let DATA_LENGTH = #asWord(#range(ARGS, 64, 32)) #in
             #setExpectOpcode #asWord(#range(ARGS, 0, 32)) #range(ARGS, CODE_START, DATA_LENGTH) 0 Static
          ...
```

ehildenb: I think it reads a little better formatted like this, up ot you. But do the same throughout the file either way.

ehildenb: Also, not sure why we're using the `#let` bindings here. It seems that we're only using the variables defined in them a single time in the `#in` body, and really the point of using a let binding is if the defined variable is used more than once.

I guess you could argue that this reads better with the let-binding. Also up to you, seems correct either way.

include/kframework/foundry.md
```diff
@@ -158,6 +158,12 @@ The configuration of the Foundry Cheat Codes is defined as follwing:
...
     - `<expectedDepth>` records the depth at which the call is expected to revert.
     - `<expectedBytes>` keeps the expected revert message as a ByteArray.
+3. The `<expectOpcode>` subconfiguration stores values used for `expect*OPCODE*` cheat codes.
+    - `<isOpcodeExpected>` flags if a call opcode is expected.
+    - `<expectedAddress>` keeps the expected caller.
+    - `<expectedValue>` keeps expected `msg.value`.
+    - `<expectedData>` keeps expected `calldata`.
+    - `<opcodeType>` keeps track of what `CALL*` Opcode is expected.
 
 ```k
```
ehildenb, line 168: I think we can probably implement `expectRevert` in terms of this more general machinery, and then remove the `expectRevert` subconfiguration and any other special handling related to that.

I would say if you don't want to implement it in this PR, make an issue for it.

ehildenb: That would also enable us to remove the third argument to `foundry_success` (where `EXPECTEDREVERT_FINAL` is stored). But it would mean that you can't do an `expectRevert` and an `expectCall` at the same time. Maybe that is good though, if you can only do one `expect*` at a time?

tests/foundry/src/Vm2.sol
```diff
@@ -0,0 +1,25 @@
...
+    // Expects a delegate call to an address with the specified calldata.
+    function expectDelegateCall(address,bytes calldata) external;
+    // Expects that no contract calls are made after invoking the cheatcode.
+    function expectNoCall() external;
+    // Expects the given address to deploy a new contract, using the CREATE opcode, with the specified value and bytecode.
+    function expectCreate(address,uint256,bytes calldata) external;
+    // Expects the given address to deploy a new contract, using the CREATE2 opcode, with the specified value and bytecode (appended with a bytes32 salt).
+    function expectCreate2(address,uint256,bytes calldata) external;
+    // Makes the storage of the given address completely symbolic.
+    function symbolicStorage(address) external;
```
ehildenb, line 24: `symbolicStorage` is added, but not implemented. Should it be added to the unimplimented list?

include/kframework/foundry.md
```diff
@@ -832,25 +1006,30 @@ If the production is matched when no prank is active, it will be ignored.
...
+    rule ( selector ( "startPrank(address,address)" )              => 1169514616 )
+    rule ( selector ( "stopPrank()" )                              => 2428830011 )
+    rule ( selector ( "expectStaticCall(address,bytes)" )          => 2232945516 )
+    rule ( selector ( "expectDelegateCall(address,bytes)" )        => 1030406631 )
+    rule ( selector ( "expectRegularCall(address,uint256,bytes)" ) => 1973496647 )
+    rule ( selector ( "expectCreate(address,uint256,bytes)" )      => 658968394  )
+    rule ( selector ( "expectCreate2(address,uint256,bytes)" )     => 3854582462 )
 ```
 
 - selectors for unimplemented cheat code functions.
```
ehildenb, line 1035: The unimplemented list here.

ehildenb (approved):

include/kframework/foundry.md
anvacaru: In the case of `expectRevert`, they specifically mention that `After calling expectRevert, calls to other cheat codes before the reverting call are ignored.`. This is not mentioned for `expectCall` or `expectEmit` cheat codes. My plan was to first merge this PR and the one implementing `expectEmit` using a similar subconfiguration. Next I'm going to open a new PR to refactor the `expect*` subconfigurations.

tests/foundry/src/Vm2.sol
anvacaru: Done. I've added them.

PR #1455, opened 2022-11-14
Sign cheatcode

### LINKED ISSUE #1464: sign cheat code

lisandrasilva:
closes #1464


include/kframework/foundry.md
```diff
@@ -657,6 +657,22 @@ Otherwise, throw an error for any other call to the Foundry contract.
...
 
+#### `sign` - Signs a digest with private key
+
+```
+function sign(uint256 privateKey, bytes32 digest) external returns (uint8 v, bytes32 r, bytes32 s);
+```
+
+`foundry.call.sign` will match when the `sign` cheat code function is called.
+This rule then takes from the `privateKey` to sign using `#range(ARGS,0,32)` and the `digest` to be signed using `#range(ARGS, 32, 32)`. To perform the signature we use the `ECDSASign ( String, String )` function (from KEVM). This function receives as arguments 2 strings: the data to be signed and the private key, therefore we use `#unparseByteStack` to convert the bytearrays with the `privateKey` and `digest` into strings. The `ECDSASign` function returns the signed data in [r,s,v] form, which we convert to a bytearray using `#parseByteStack`.
+
```
ehildenb, line 668:
```suggestion
This rule then takes from the `privateKey` to sign using `#range(ARGS,0,32)` and the `digest` to be signed using `#range(ARGS, 32, 32)`.
To perform the signature we use the `ECDSASign ( String, String )` function (from KEVM).
This function receives as arguments 2 strings: the data to be signed and the private key, therefore we use `#unparseByteStack` to convert the bytearrays with the `privateKey` and `digest` into strings.
The `ECDSASign` function returns the signed data in [r,s,v] form, which we convert to a bytearray using `#parseByteStack`.

```

One sentence per line please.

include/kframework/foundry.md
```diff
@@ -657,6 +657,22 @@ Otherwise, throw an error for any other call to the Foundry contract.
...
+```{.k .bytes}
+    rule [foundry.call.sign]:
+         <k> #call_foundry SELECTOR ARGS => . ... </k>
+         <output> _ => #parseByteStack(ECDSASign(#unparseByteStack(#range(ARGS, 32, 32)),#unparseByteStack(#range(ARGS,0,32)))) </output>
```
ehildenb, line 672: Hmmmm, this direct call of `ECDSASign` probably is not a good idea, because it will try to evaluate as a hooked symbol, which there may not be implementations for.

Maybe better to make a wrapper like this:

```
syntax ByteArray ::= #sign ( ByteArray , ByteArray ) [function]
rule #sign(BA1, BA2) => #parseByteStack(ECDSASign(#unparseByteStack(BA1), #unparseByteStack(BA2))) [concrete]
```

Then use `#sign` in this rule.

Currently the only test of this only feeds concrete values into `ECDSASign`, which is likely why you haven't see the backend blow up on a symbolic inputs to this yet. So this proposed modification will be fine for your test-suite (because it only exercises the concrete case), but will keep the backend from blowing up in the symbolic case (which you maybe should consider adding a test for).

lisandrasilva: I added a fuzzing test and indeed the rule doesn't work anymore. I will address your suggestions and see if it fixes.

include/kframework/foundry.md
```diff
@@ -1168,9 +1192,6 @@ If the production is matched when no prank is active, it will be ignored.
 - selectors for unimplemented cheat code functions.
 
 ```k
-    rule selector ( "expectRegularCall(address,bytes)" )        => 3178868520
-    rule selector ( "expectNoCall()" )                          => 3861374088
-    rule selector ( "symbolicStorage(address)" )                => 769677742
```
anvacaru, line 1173: Is there a reason for this remove?

lisandrasilva: That was the result of running `make tests/foundry/foundry.k.check CHECK=cp`.

include/kframework/foundry.md
```diff
@@ -1089,6 +1108,11 @@ If the production is matched when no prank is active, it will be ignored.
           <singleCall> _ => false </singleCall>
         </prank>
 ```
+```k
+    syntax ByteArray ::= #sign ( ByteArray , ByteArray ) [function,klabel(foundry_sign)]
+ // -----------------------------------------------------------
```
anvacaru, line 1113:
```suggestion
    syntax ByteArray ::= #sign ( ByteArray , ByteArray ) [function,klabel(foundry_sign)]
 // ----------------------------------------------------------------------------------
```

tests/foundry/exclude
```diff
@@ -4,6 +4,7 @@ AccountParamsTest.testFail_GetNonce_false
 AddrTest.test_addr_false
 AddrTest.testFail_addr_false
 AddrTest.testFail_addr_true
+AddrTest.test_addr_symbolic
```
anvacaru, line 7: Let's also either add a wrapper rule around the `addr` cheatcode with the  `[concrete]` tag, in a similar way to how `#sign` is defined and used. Either in this PR or in a new one. Just left this comment to keep track of the issue.

lisandrasilva: Actually I already tried that, but Everett suggested to put the `concrete` tag on `#addrFromPrivateKey` directly. But the problem for symbolic inputs remained. Let's open another issue to fix this.

anvacaru (approved):

PR #1459, opened 2022-11-21
expectEmit cheat code

anvacaru:
closes: #1463 
- Adds a new subconfiguration `<expectEmit/>`
    - `<recordEvent>` flags if the next emitted Event should be recorded and held for future matches.
    - `<isEventExpected>` flags if an Event is expected to match the one recorded previously.
    - `<checkedTopics>` will store a list of `bool` values that flag if a topic should be checked or not.
    - `<checkedData>` flags if the data field should be checked or not.
    - `<expectedEventAddress>` stores the emitter of an expected Event.

- Adds new rules for the `LOG(N)` opcode which : 
   - record the next event and store it as the `expectedEvent`.
   - check the next emitted event to match the previously recorded one.
   
- Enables tests for expectEmit.


kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -367,12 +367,12 @@ def _patch_symbol_table(symbol_table: Dict[str, Any]) -> None:
...
-    def success(s: KInner, dst: KInner, r: KInner, c: KInner) -> KApply:
-        return KApply('foundry_success', [s, dst, r, c])
+    def success(s: KInner, dst: KInner, r: KInner, c: KInner, e1: KInner, e2: KInner) -> KApply:
+        return KApply('foundry_success', [s, dst, r, c, e1, e2])
 
     @staticmethod
-    def fail(s: KInner, dst: KInner, r: KInner, c: KInner) -> KApply:
-        return notBool(Foundry.success(s, dst, r, c))
+    def fail(s: KInner, dst: KInner, r: KInner, c: KInner, e1: KInner, e2: KInner) -> KApply:
+        return notBool(Foundry.success(s, dst, r, c, e1, e2))
```
tothtamas28, line 375: These parameters should have a more descriptive name.  (It's OK if you do not address this in this PR.)

tothtamas28 (approved):

lisandrasilva (approved):

PR #1461, opened 2022-11-22
Smaller target for Haskell profiling job

ehildenb:
This makes two changes:

- Pulls out `test-prove-smoke` test-suite for KEVM, which is all the proofs that ran betwen 100s and 350s on my machine on the Haskell backend.
- Adjust Nix job for profiling to build the kompiled definitions ahead of item (which eats a lot of CPU power).
- Adjust the Nix job to run the smoke test, not the original test-suite.


package/nix/profile.nix
```diff
@@ -20,7 +20,8 @@ in stdenv.mkDerivation {
     mkdir -p .build/usr
     cp -r ${kevm}/* .build/usr/
     ${kore-exec}/bin/kore-exec --version &> log
-    make test-prove TEST_SYMBOLIC_BACKEND=haskell KEVM='${profile}/bin/profile log timeout 3000 kevm' -j6 -k
+    make build-prove-haskell -j4
+    make test-prove-smoke TEST_SYMBOLIC_BACKEND=haskell KEVM='${profile}/bin/profile log timeout 3000 kevm' -j6 -k
```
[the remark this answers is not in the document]
ehildenb: I think the parallelism can stay higher with the `build-prove-haskell` job running with lower parallelism ahead of time. We're also not running the long-running proofs, nad the backend tends to consume more memory over time.

ehildenb, line 24:
```suggestion
    make test-prove-smoke TEST_SYMBOLIC_BACKEND=haskell KEVM='${profile}/bin/profile log timeout 600 kevm' -j6 -k
```

jberthold:
While we are at it, may I suggest to modify this: 
https://github.com/runtimeverification/evm-semantics/blob/master/package/nix/profile.py#L84
Replacing `|` by  `+` breaks github table rendering, if we leave the `|` we can directly post the table as a comment.

A bit more involved: https://github.com/runtimeverification/evm-semantics/blob/master/package/nix/profile.py#L19 reads the full git commit hash , maybe we could shorten it to 7 characters?

jberthold:
> A bit more involved: https://github.com/runtimeverification/evm-semantics/blob/master/package/nix/profile.py#L19 reads the full git commit hash , maybe we could shorten it to 7 characters?

Actually, this gets written in https://github.com/runtimeverification/evm-semantics/blob/master/package/nix/profile.nix#L22 so we might change it there. Not pressing, though

package/nix/profile.py
```diff
@@ -81,6 +81,6 @@ def mkColumn(c, w):
     return ' ' + c.ljust(w) + ' '
 
 columns = ['|'.join((mkColumn(c0, column_width0), mkColumn(c1, column_width1), mkColumn(c2, column_width2), mkColumn(c3, column_width3))) for (c0, c1, c2, c3) in final_table]
-columns[1] = columns[1].replace('|', '+').replace(' ', '-')
```
gtrepta, line 84: Does this script run on one of our runners somewhere, or is it just here to generate manual reports?

gtrepta (approved):
LGTM

package/nix/profile.py
jberthold: This script is used for performance tests on a dedicated runner of ours. We have [a mechanism in `haskell-backend` to trigger the performance tests](https://github.com/runtimeverification/haskell-backend/blob/master/.github/workflows/kevm-performance-test.yaml#L81) via git PR comments .

PR #1477, opened 2022-11-30
Request for Comment: Limit size of calldata arrays to 2^128 bytes

hjorthjort:
It is helpful when slicing calldata to assume that the size of the byte array is within the addressable range, 2^256 bytes. It's even more helpful to know that the calldata is much smaller than that, to allow some wiggle room for arithmetic:

Motivating example in a side condition:

```
#And ( ( { true #Equals ( ( #sizeByteArray ( #bufStrict ( 32 , #getValue ( #uint256 ( #sizeByteArray ( VV3__data_3c5818c8 ) ) ) ) ) +Int ( ( ( ( #sizeByteArray ( VV3__data_3c5818c8 ) up/Int 32 ) ) *Int 32 ) ) ) ) <Int 115792089237316195423570985008687907853269984665640564039457584007913129639804 }
```

Since calldata needs to be actually allocated and initialized, we can give it realistic constraints.
(Unlike storage, which can just be declared and is implicitly initialized, initializing a storage array as `new bytes[](type(uint256).max)` compiles fine.)
2^128 is a crypotgraphy-strength number. 2^128 bytes would be far more than 10^21 times the amount of data on the deep web. (https://www.wolframalpha.com/input?i=2%5E128+bytes). If we could allocate 64 GB of data per picosecond, it would take 100 million years to generate. So arguably the assumption is safe.

Note that we want to avoid saying that `#sizeByteArray(BS) <Int 128 => true` since this would be unsound, something a theorem prover or SMT solver might discover and decide it has reached bottom.

Instead, we want to say that: the length of any calldata array in EVM is less that 2^128 bytes.

I'm not sure if this is the best way to insert that assumption everywhere, so I'm looking for comments on the approach.

### DIFF
--- a/kevm-pyk/src/kevm_pyk/kevm.py
+++ b/kevm-pyk/src/kevm_pyk/kevm.py
@@ -234,6 +234,15 @@ class KEVM(KProve, KRun):
     def range_bytes(width: KInner, ba: KInner) -> KApply:
         return KApply('#rangeBytes(_,_)_WORD_Bool_Int_Int', [width, ba])
 
+    # We can realistically say that, when using memory or calldata, the size of
+    # the array is less than 2**128 bytes, which is still 300 quintillion
+    # exabytes. This will not realistically fit in a transaction, within any gas
+    # limit, or be possible to allocate on any machine, so this limit is a good
+    # working assumption.
+    @staticmethod
+    def range_array_calldata(ba: KInner) -> KApply:
+        return KApply('_<Int_', [KApply('#sizeByteArray(_)_EVM-TYPES_Int_ByteArray', [ba]), intToken(2**128)])
+
     @staticmethod
     def bool_2_word(cond: KInner) -> KApply:
         return KApply('bool2Word(_)_EVM-TYPES_Int_Bool', [cond])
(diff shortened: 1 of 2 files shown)


hjorthjort:
We also leave out strings from this update. It would be sensible to do the same for strings, since they are just another form of byte arrays.

kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -234,6 +234,15 @@ def range_bool(i: KInner) -> KApply:
...
         return KApply('#rangeBytes(_,_)_WORD_Bool_Int_Int', [width, ba])
 
+    # We can realistically say that, when using memory or calldata, the size of
+    # the array is less than 2**128 bytes, which is still 300 quintillion
+    # exabytes. This will not realistically fit in a transaction, within any gas
+    # limit, or be possible to allocate on any machine, so this limit is a good
+    # working assumption.
+    @staticmethod
+    def range_array_calldata(ba: KInner) -> KApply:
+        return KApply('_<Int_', [KApply('#sizeByteArray(_)_EVM-TYPES_Int_ByteArray', [ba]), intToken(2**128)])
```
anvacaru, line 244:
```suggestion
        return KApply('_<Int_', [KEVM.size_bytearray(ba), intToken(2**128)])
```

anvacaru:
I've tried to generate a KCFG using this branch and I couldn't see any `#sizeByteArray(_) <Int 2**128` constraint generated.  The only one added was
```
andBool ( #sizeByteArray ( ExpectCallTest . testExpectStaticCall ( ) ) <Int pow256
```
which is generated by this line here: https://github.com/runtimeverification/evm-semantics/blob/master/kevm-pyk/src/kevm_pyk/kevm.py#L185

So, if I understood correctly, you want to change the constraint and say that the size of the calldata is less than `pow128`?

ehildenb:
This has been subsumed.

PR #1478, opened 2022-11-30
Several fixes to KEVM prover

ehildenb:
This is a handful of fixes/improvements to the KEVM prover.

- `--debug` and `--[no-]-minimize` arguments are improved slightly.
- We use `#gas(_)` infinite gas in initial state for Solidity-generate proofs, which makes the branching that happens in a given proof significantly less.
- Remove a few lemmas about `Cmem` which are unsound, and are somewhat redundant.
- Avoids questionable lemmas about `Cgascap` interacting with `#gas`, removing several tests.
- Adds lemma about `#allBut64th`.


ehildenb:
Mostly subsumed.

Issue #1481, opened 2022-12-01
Constrain `ACCOUNTS_INIT` variable for foundry tests


RaoulSchaffranek:
When we generate a specification from a foundry test, the `<accounts>`-rewrite-rule looks similar to the following:

~~~xml
<accounts>
  ( <account>
    <acctID>
      1032069922050249630382865877677304880282300743300
    </acctID>
    ...
  </account>
  ...
  ( _ACCOUNTS_INIT => ?_ACCOUNTS_FINAL ) )
</accounts>
~~~

When we run a proof, the prover will add the following side condition to the final configuration:

~~~k
#And
  {
    false
  #Equals
    <acctID>
      1032069922050249630382865877677304880282300743300
    </acctID> in_keys ( _ACCOUNTS_INIT )
  }
~~~

Notice that `_ACCOUNTS_INIT` is unconstrained in the specification but constrained in the final configuration. Hence, we get the following error message:

~~~
The configuration's term unifies with the destination's term, but the implication check between the conditions has failed.
~~~

Can we automatically generate the necessary constraints to the `_ACCOUNTS_INIT`-variable to not include any of the addresses mentioned in the surrounding `<accounts>`-context?

As a workaround, I replaced `_ACCOUNTS_INIT` with `.Bag` in my specification, but it would be nice if we could be more general.

anvacaru:
I'm closing this because of the comments from #1482.

PR #1482, opened 2022-12-01
Add constraints for ACCOUNTS_INIT

anvacaru:
fixes: #1481


ehildenb:
I don't think this should be necessary. We should confirm with @RaoulSchaffranek, but I think we fonud it was due to another revert path issue the proof was failing (the final `<statusCode>` cell was `EVMC_REVERT` which means the foundry test failed).

It's normal to have these constraints added automatically at startup by th ebackend, it's the way the bcakend basically says "I'm assuming the initial configuration you gave me is well-defined, which requires these extra conditions". So manually adding the extra conditions is redundant.

Issue #1485, opened 2022-12-02
`expectRevert` cheat code behaves different from foundry on internal calls


RaoulSchaffranek:
Our `expectRevert` cheat code behaves different from foundry when the `revert` happens from an internal call:

The following test passes on Foundry but fails on KEVM:

~~~Solidity
contract Demo is Test {

    function doRevert() internal {
        require(false, "");
    }

    function testExpectRevert() public {
        vm.expectRevert();
        doRevert();
    }

}
~~~

One problem is that internal calls compile down to `jump` instructions. As implication the following rule does not match on internal reverts because we don't see the `#return` symbol in the K-sequence:

~~~
    rule <statusCode> EVMC_REVERT => EVMC_SUCCESS </statusCode>
         <k> #halt ~> #return _RETSTART _RETWIDTH ... </k>
         <output> OUT </output>
         <callDepth> CD </callDepth>
         <expected>
           <expectedRevert> true => false </expectedRevert>
           <expectedDepth> CD </expectedDepth>
           <expectedBytes> EXPECTED </expectedBytes>
         </expected>
      requires (EXPECTED =/=K .ByteArray andBool EXPECTED ==K #range(OUT, 4, #sizeByteArray(OUT) -Int 4))
        orBool EXPECTED ==K .ByteArray
      [priority(40)]
~~~

Another problem is that we increase the expected call depth when we see the cheat code. However, because the internal call is compiled to `jump` we never increase the call depth during execution.

~~~
    rule <k> #setExpectRevert EXPECTED => . ... </k>
         <callDepth> CD </callDepth>
         <expected>
           <expectedRevert> false => true </expectedRevert>
           <expectedDepth> _ => CD +Int 1 </expectedDepth>
           <expectedBytes> _ => EXPECTED </expectedBytes>
         </expected>
~~~

PR #1487, opened 2022-12-05
refactor expectRevert

anvacaru:
Fixes: #1485 
Change the expectRevert rules, to support internal calls.
Previously, the `expectRevert` rules would apply only when the top of the K cell would contain the `#halt ~> #return _ _` productions. Since `#return _ _` would only be present for external calls, ignoring internal calls and contract creation calls.


include/kframework/foundry.md
```diff
@@ -213,20 +213,24 @@ module FOUNDRY-CHEAT-CODES
...
  // ------------------------------------------------------------------
     rule <k> #assume(B) => . ... </k> ensures B
 
-    rule <k> #assert(false) => . ... </k>
-         <account>
-           <acctID> #address(FoundryCheat) </acctID>
-           <storage> STORAGE => STORAGE [ #loc(FoundryCheat . Failed) <- 1 ] </storage>
-           ...
-         </account>
+    rule <k> #assert(false) => #markAsFailed ... </k>
```
anvacaru, line 224: This rule is not used anywhere and confuses. It should be removed.

include/kframework/foundry.md
```diff
@@ -532,46 +532,42 @@ Ignore all cheat code calls which take place while `expectRevert` is active.
...
-         </expected>
-      requires (EXPECTED =/=K .ByteArray andBool EXPECTED ==K #range(OUT, 4, #sizeByteArray(OUT) -Int 4))
-        orBool EXPECTED ==K .ByteArray
+         <expectedRevert>
+           <isRevertExpected> true </isRevertExpected>
+           <expectedDepth> ED </expectedDepth> 
+           ...
+         </expectedRevert>
+      requires CD <=Int ED
+       andBool SC =/=K EVMC_SUCCESS
```
ehildenb, line 553: Sholud it be `SC =/=K EVMC_SUCCESS` or `SC ==K EVMC_REVERT`? If it ends with out of gas instead, for example, should we count that as a revert?

include/kframework/foundry.md
```diff
@@ -212,21 +212,21 @@ module FOUNDRY-CHEAT-CODES
...
-           ...
-         </account>
+     syntax KItem ::= "#markAsFailed" [klabel(foundry_markAsFailed)]
+  // ---------------------------------------------------------------
+     rule <k> #markAsFailed => . ... </k>
+          <account>
+             <acctID> #address(FoundryCheat) </acctID>
+             <storage> STORAGE => STORAGE [ #loc(FoundryCheat . Failed) <- 1 ] </storage>
+             ...
+           </account>
```
ehildenb, line 229: Is this how Foundry works? It marks as failed for when an expectRevert fails? Or does it have a separate way of storing that?

Also, if we are marking as failed like this, do we need the `ISREVERTEXPECTED` to be passed into `foundry_success` anymore?

include/kframework/foundry.md
anvacaru: The only instance of the `Revert` status could that I could find in their rust implementation is here https://github.com/foundry-rs/foundry/blob/master/evm/src/executor/inspector/cheatcodes/mod.rs#L489-L505. I also looked at their `handle_expect_revert` function, but I didn't see any match for the actual status code. But my Rust knowledge is limited.

include/kframework/foundry.md
anvacaru: Well, `expectRevert` looks at the next call made after the invocation of the cheat code. If I'm not setting the storage location, then any upcoming revert call could clear the `ISREVERTEXPECTED`, at least that was my understanding.

>Also, if we are marking as failed like this, do we need the ISREVERTEXPECTED to be passed into foundry_success anymore?

The `ISREVERTEXPECTED` is still passed to the `foundry_success` predicate in the case in which the test ends with a pending `expectRevert` (i.e. the cheat code is invoked but no further call is made).

PR #1490, opened 2022-12-06
refactor #execute rules

anvacaru:
- add `total` tag to  `#dasmOpCode`
- refactor `#execute` rules to use a lookup function which is totally defined.


ana-pantilie (approved):
Thank you. I believe this will make the RHS of those rules I pointed out total.

PR #1492, opened 2022-12-06
Anvacaru/investigate assume

ehildenb:
Blocked on: https://github.com/runtimeverification/pyk/pull/131

Small changes, adds test which identified issue in how we're printing out bytes token.

I honestly have no idea how this worked before.


ehildenb:
This should be resolved now.

PR #1493, opened 2022-12-07
Better restriction on dynamic memory inputs

ehildenb:
This adds the restriction on dynamic type `bytes` input that it's size is between `0` and `pow128` (an absurdly large number, but much smaller than `pow256`, the actual max).

It seems the Solidity compiler is actually restricting the size even further to be less than `(2 ^ 64) - 1 == 18446744073709551615`.

Not sure this test is passing, or if it needs to be added to `exclude` list.


PetarMax:
In the work on Optimism, I observed the following constraint on a proof branch, I assume the compiler generated it

```
18446744073709551487 <Int ( ( ( ( ( ( 115792089237316195423570985008687907853269984665640564039457584007913129639904 &Int ( ( #sizeByteArray ( VV3__data_3c5818c8 ) +Int 31 ) ) ) ) +Int 63 ) ) &Int 115792089237316195423570985008687907853269984665640564039457584007913129639904 ) )
```

which is really

```
(pow64 - 129) <Int ( ( ( ( ( ( (pow256 - 32) &Int ( ( #sizeByteArray ( VV3__data_3c5818c8 ) +Int 31 ) ) ) ) +Int 63 ) ) &Int (pow256 - 32) ) )
```

and I believe that this constraint should be discharged. However, I've found an example that witnesses it, when `#sizeByteArray ( VV3__data_3c5818c8 )  = (2^64 - 64)`. If this constraint is generated by the compiler, this means that actually the data is somehow restricted to `2^64 - x`, for some small `x` that I can discover.

ehildenb:
I would still say this is worth merging @PetarMax , because it's (i) still general to KEVM (no way you can supply bytecode this long anyway), so we're not pruning potentially valid paths, and (ii) it gets rid of spurious branches due to `#Ceil` conditions on the input.

ehildenb:
Then I need an approval :wink:

anvacaru (approved):

PR #1494, opened 2022-12-07
kevm-pyk/: turn on strict=True for zip operations

ehildenb:


ehildenb:
Subsumed by: https://github.com/runtimeverification/evm-semantics/pull/1491

PR #1501, opened 2022-12-09
Add command `view-kcfg`

tothtamas28:
~Blocked on: https://github.com/runtimeverification/pyk/pull/139~
~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/1508~
~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/1517~

```
$ poetry run kevm-pyk view-kcfg ../tests/foundry/out kcfg.json
```

[kcfg-viewer-screencast.webm](https://user-images.githubusercontent.com/10085903/207008194-b2552d5d-d93a-4fb5-acc7-ceee97c3966f.webm)


ehildenb:
Currently, I have to select the new node I want to see, _then_ press _enter_, and only then it displays the new node on the right display. It would be nicer if it updated automatically as I scrolled through nodes.

Also, currently long chains of nodes are displays as how I would expect branches to  be displayed. By default, I think we should lean towards the [KLab way of displaying and navigating](https://github.com/dapphub/klab#key-bindings) (matching the same keystrokes doesn't matter, but the navigation concept itself). That is:

- The view on the left only shows the branching structure by default (only the nodes which are at beginning/end/branch points).
- Each branch displays they branch it's associated with inline (examples in the [KLab demo video](https://www.youtube.com/watch?v=z4mIo38x1u8&t=1s)).
- You can step through the chains of nodes at whatever granularity they are stored using "forward" and "backward" (KLab uses `n` (next) and `p` (previous)). Modifiers make the jumps bigger (`Shift-n` means jump all the way forward to next branch, `Shift-p` means jump backward all the way to previous branch, for example).
- Ability to filter by cells: both turning _only_ some cells on, and turning some off.

tothtamas28:
> * The view on the left only shows the branching structure by default (only the nodes which are at beginning/end/branch points).
> * Each branch displays they branch it's associated with inline (examples in the [KLab demo video](https://www.youtube.com/watch?v=z4mIo38x1u8&t=1s)).

So the original tree view displays `[Init]` and `[Branch]` nodes, and upon selection another view shows all direct children that are `[Step N]`?

tothtamas28:
> Currently at bootup it hangs for a bit. I guess because it's rendering all the nodes? Maybe instead it could render them as the user asks for each node the first time?

The right view is rendered only on tree node selection. Bootup time is spent mostly on loading the `KCFG` from file, but this can only be improved by changing the data model. But of course loading can be delayed, e.g. until the user expands the root of the tree view, which might make the wait a bit more tolerable.

tothtamas28:
> Currently, I have to select the new node I want to see, _then_ press _enter_, and only then it displays the new node on the right display. It would be nicer if it updated automatically as I scrolled through nodes.

https://github.com/runtimeverification/pyk/issues/147

tothtamas28:
> Ability to filter by cells: both turning _only_ some cells on, and turning some off.

https://github.com/runtimeverification/pyk/issues/149

ehildenb (approved):

Issue #1509, opened 2022-12-13
Overlapping rules for Foundry `#endPrank`


ehildenb:
The rules for `#endPrank` look like this:

```
    rule <k> #endPrank => . ... </k> 
        <prank> 
          <prevCaller> .Account </prevCaller> 
          <prevOrigin> .Account </prevOrigin> 
          <active> false </active> 
          ... 
        </prank> 
 
    rule <k> #endPrank => #clearPrank ... </k> 
        <caller> _ => CL </caller> 
        <origin> _ => OG </origin> 
        <prank> 
          <prevCaller> CL </prevCaller> 
          <prevOrigin> OG </prevOrigin> 
          ... 
        </prank> 
```

In the case the first rule can fire, the second rule can _always_ fire, which leads to a branch. Is this intentional? My guess is that the second rule should be behind an `owise`, but I'm not sure, and we should base our answer on a reading of the Foundry source code.

PR #1510, opened 2022-12-13
Tweak the rules for endPrank

anvacaru:
Fixes: #1509 

 - tweak the `#endPrank` rules to avoid overlapping.
 - added a new test for internal prank calls.

There are two differences between the way we inject the pranked address vs the way foundry does it ( https://github.com/foundry-rs/foundry/blob/master/evm/src/executor/inspector/cheatcodes/mod.rs#L380-L395)
     1. We only set the tx.origin at the targeted depth - foundry sets it whenever the `call depth  >= targeted depth`
     2. When setting the msg.caller, we do not check that the current caller is the same as the one who invoked the `startPrank` cheat code.


include/kframework/foundry.md
```diff
@@ -1145,20 +1147,19 @@ If the production is matched when no prank is active, it will be ignored.
...
+          <active> true </active>
+          ...
+        </prank>
+
+    rule <k> #endPrank => . ... </k>
+        <prank>
+          <active> false </active>
           ...
         </prank>
 ```
```
ehildenb, line 1165: Why would we be ending the prank for `<active> false </active>`? Is this actually a case we expect to be exercised? Do our tests exercise this case?

ehildenb: If not, then maybe the first rule is enough on its own?

anvacaru: That rule is for the case in which `vm.endPrank()` is invoked when there is no active prank.

ehildenb: Is this a case that Foundry allows?

ehildenb (approved):

include/kframework/foundry.md
anvacaru: Yes, we have tests for this scenario.

PR #1512, opened 2022-12-14
Eliminate volumes from `with-docker` action

### LINKED ISSUE #1457: Interference between CI runs
Consider the following run: https://github.com/runtimeverification/evm-semantics/actions/runs/3468304857/jobs/5794152325

The job fails because `make kevm-pyk` fails. The log indicates that during `poetry install`, several packages are already installed with another version, in particular:

```
  • Updating pyk (0.1.52 08ab1a0 -> 0.1.34 6d12754)
```

This downgrade probably leaves some garbage from `0.1.52`, thus `check-mypy` fails:

```
src/kevm_pyk/utils.py:5: error: Module "pyk.kast" has no attribute "KApply"
```

The root of the issue is probably in `./.github/actions/with-docker`: the container is set up by mounting both `/opt/workspace` and `${HOME}` as volumes. A possible solution is to use `docker cp` instead (an [example](https://github.com/runtimeverification/pyk/blob/2c0ccd010ad6489662794aed8917538b16863464/.github/workflows/test-pr.yml#L74) of how containers are set up in `pyk`).

tothtamas28:
Fixes #1457

### DIFF
--- a/.github/actions/with-docker/action.yml
+++ b/.github/actions/with-docker/action.yml
@@ -24,16 +24,16 @@ runs:
         --tag runtimeverification/${TAG_NAME}                      \
         --build-arg K_COMMIT=${K_COMMIT}                           \
         --build-arg Z3_VERSION=${Z3_VERSION}                       \
-        --build-arg USER_ID=$(id -u) --build-arg GROUP_ID=$(id -g)
 
       docker run                        \
         --name ${TAG_NAME}              \
-        --rm -it                        \
+        --rm                            \
+        --interactive                   \
+        --tty                           \
         --detach                        \
-        --workdir /opt/workspace        \
-        --user $(id -u):$(id -g)        \
-        -v ${HOME}:${HOME}              \
-        -v "$(pwd):/opt/workspace"      \
-        -v "/etc/passwd:/etc/passwd:ro" \
-        -v "/etc/group:/etc/group:ro"   \
+        --user root                     \
+        --workdir /home/user            \
         runtimeverification/${TAG_NAME}
+
+      docker cp . ${TAG_NAME}:/home/user
+      docker exec ${TAG_NAME} chown -R user:user /home/user
--- a/.github/workflows/test-pr.yml
+++ b/.github/workflows/test-pr.yml
@@ -23,21 +23,21 @@ jobs:
         with:
           tag: kevm-ci-concrete-${{ github.sha }}
       - name: 'Build kevm-pyk'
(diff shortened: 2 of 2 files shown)


.github/actions/with-docker/action.yml
```diff
@@ -37,3 +37,4 @@ runs:
 
       docker cp . ${TAG_NAME}:/home/user
       docker exec ${TAG_NAME} chown -R user:user /home/user
+      docker exec -u user ${TAG_NAME} cat /home/user/deps/plugin/deps/libff/build/CMakeCache.txt
```
tothtamas28, line 40: This demonstrates why the following error occurs ([build](https://github.com/runtimeverification/evm-semantics/actions/runs/3693720518/jobs/6254042903#step:5:40)):

```
CMake Error: The current CMakeCache.txt directory /home/user/deps/plugin/deps/libff/build/CMakeCache.txt is different than the directory /opt/workspace/deps/plugin/deps/libff/build where CMakeCache.txt was created. This may result in binaries being created in the wrong place. If you are not sure, reedit the CMakeCache.txt
```

Output ([build](https://github.com/runtimeverification/evm-semantics/actions/runs/3693720518/jobs/6254042903#step:3:171)):
```
docker exec -u user kevm-ci-concrete-7a392c544137dd6fb49dd8e3b1bb08d02d9f6e74 cat /home/user/deps/plugin/deps/libff/build/CMakeCache.txt
# This is the CMakeCache file.
# For build in directory: /opt/workspace/deps/plugin/deps/libff/build
...
```

So currently `/opt/workspace` contains checked out code that has already been built (at least in part). The `checkout` action should ensure the repo is cleaned, but the `clean` parameter only seems to run `git clean -ffdx` without recursing into submodules.

The PR can thus be unblocked by ensuring that the repo copied into the docker container is in a clean state, probably by doing some cleanup on the runner.

F-WRunTime: Clean is not acting as it should. According to docs from git-clean -d should enable recursive cleaning 
```
-d
           Normally, when no <path> is specified, git clean will not recurse into untracked directories to avoid removing too
           much. Specify -d to have it recurse into such directories as well. If any paths are specified, -d is irrelevant; all
           untracked files matching the specified paths (with exceptions for nested git directories mentioned under --force) will
           be removed.
```

ehildenb (approved):

F-WRunTime (approved):

PR #1519, opened 2022-12-16
Refactor foundry.call rule

anvacaru:
Refactor the `foundry.call` rule to intercept any kind of `CALL*` opcode using `#call` instead of only intercepting the `CALL` opcode. The current drawback is that in `#call` there is no longer the `ARGWIDTH` value and I had to recompute it using `#sizeByteArray`.


include/kframework/lemmas/lemmas.k
```diff
@@ -438,4 +438,9 @@ module LEMMAS-HASKELL [symbolic, kore]
 
     rule #range ( BA:ByteArray , 0 , LEN ) => BA requires #sizeByteArray(BA) ==Int LEN [simplification]
 
+    rule (#buf ( 32 , X ) [ 0 .. 28 ]):ByteArray => #buf(28, 0) requires #rangeUInt(32, X) [simplification]
+
+    rule (#range (#range (BA:ByteArray, S0, L0), S1, L1)) => 
+      #range ( BA, S0 +Int S1, minInt(L1, L0 -Int S1) ) ++ #buf(maxInt(0, L1 -Int (L0 -Int S1)), 0) [simplification]
```
ehildenb, line 444: I think you should add side-conditions:

```
requires 0 <=Int S0 andBool 0 <=Int L0 andBool 0 <=Int S1 andBool 0 <=Int L1
```

Not sure this holds for negative values.

ehildenb (approved):

PR #1520, opened 2022-12-16
Use infinite gas by default

RaoulSchaffranek:
This PR turns on infinite gas for all foundry-proofs by default.


anvacaru (approved):

ehildenb:
I've begun work on fixing the issues here. The basic issue is that any test which uses the `gasLeft()` instruction (`gas()` in Yul/inline-assembly), cannot use infinite-gas to get sensible results. For example [this primitive](https://github.com/runtimeverification/evm-semantics/blob/a7e1c8b13026fa4bf09e9815b8de82632c2469e4/tests/foundry/test/StoreTest.t.sol#L95) uses this feature to measure how much gas a call uses.

To work around this [this PR](https://github.com/runtimeverification/evm-semantics/pull/1524) introduces a `setGas(...)` cheatcode, which allows you to set the gas to some concrete (or symbolic) amount. The way I see it, we have three possible solutions:

- Default to infinite-gas on Foundry tests (what this PR does), but allow `setGas` for Foundry tests that want to do explicit gas measurements (need more lemmas to make this work). This is the approach the [linked PR](https://github.com/runtimeverification/evm-semantics/pull/1524) takes.
- Default to finite gas (same amount as Foundry), but expose an `infiniteGas` cheatcode that sets the gas to infinite. Or make it an option to `kevm foundry-prove ... --infinite-gas` to make it use infinite gas in the specification.
- Change the way infinite gas works. Instead of having an [algebraically defined infinity `#gas`](https://github.com/runtimeverification/evm-semantics/blob/master/include/kframework/infinite-gas.md), instead just [override the rules for deducting gas](https://github.com/runtimeverification/evm-semantics/blob/master/include/kframework/evm.md#overall-gas) with higher-priority rules that do not do the out of gas checks, instead they just deduct the gas. The annoying thing about this approach is that it requires adjusting module imports to turn off gas checking (so in some cases may require rekompilation).

I'd like people's feedback on the approaches, including @PetarMax . Perhaps we can discuss it in a meeting this week.

hjorthjort:
> - Default to infinite-gas on Foundry tests (what this PR does), but allow `setGas` for Foundry tests that want to do explicit gas measurements (need more lemmas to make this work). This is the approach the [linked PR](https://github.com/runtimeverification/evm-semantics/pull/1524) takes.

I think this is sensible enough. But another drawback is that this would make us diverge from Foundry in behavior. Ideally we want that we behave exactly like Foundry on all concrete inputs. Granted, it would be for some silly test meant to exhaust all the gas that Foundry gives by default, but still.

> - Default to finite gas (same amount as Foundry), but expose an `infiniteGas` cheatcode that sets the gas to infinite.

I think this is better from a product/user perspective. Especially since it would makes specs more explicit and readable: we would usually be specifying infinite gas, which would be fine for most functions, but for some they would look and feel wrong when they are wrong. The specs are perhaps the most important artifact and making them expressive and capture as much of our assumptions as possible is a value add in my view.

> - The same as options (1) or (2), but do it via command-line options to `kevm foundry-prove ... --[no-]infinite-gas`, so that you are setting on the CLI whether you want infinite gas or not for the test. This may be easy to use initially, but annoying later because it does not store the information with the test itself that it requires infinite gas.

Agree here, cheat code is much better.

> - Store a shadow-cell `<foundryGas>`, which tracks exactly the `<gas>` cell (we can make it stay in sync with some simple higher priority rules), and use that cell when the `GAS` opcode is called, but the normal cell for checking deductions.

One issue with this is that it may cause issues when gas goes negative, which I suspect would explode the proofs as much as what we are trying to avoid.

> - Add a higher-priority rule for the `GAS` opcode which checks if there is an infinite `#gas(...)` there, and if so strips the outer infinite gas. This may have unexpected results, not sure. It's not clear to me that our algebraically defined infinite gas is sensible.

I think this would in practice give the same issues as above. Instead of dealing with out of gas errors you are dealing with arithmetic errors, because the `GAS` opcode is usually used to perform some arithmetic.

> - Change the way infinite gas works. Instead of having an [algebraically defined infinity `#gas`](https://github.com/runtimeverification/evm-semantics/blob/master/include/kframework/infinite-gas.md), instead just [override the rules for deducting gas](https://github.com/runtimeverification/evm-semantics/blob/master/include/kframework/evm.md#overall-gas) with higher-priority rules that do not do the out of gas checks, instead they just deduct the gas. The annoying thing about this approach is that it requires adjusting module imports to turn off gas checking (so in some cases may require rekompilation).

I think this would also cause the same problem as above.

anvacaru:
I'm closing this as the `infiniteGas` cheat code has been added with https://github.com/runtimeverification/evm-semantics/pull/1524.

Issue #1528, opened 2023-01-12
Cheatcodes for asserting that only some accounts/storage slots are modified


lucasmt:
We need cheatcodes for the LUKSO engagement that would let us check that only a limited set of accounts or storage slots are modified. Something like

- `onlyModify(address[] accounts)`: causes the test to fail if a non-static call is made to any account outside the list
- `onlyModify(StorageSlot[] slots)`: causes the test to fail if it writes to any storage slot outside the list

where `StorageSlot` is a `struct` like

```
struct StorageSlot {
    address account;
    bytes32 slot;
}
```

These don't have to be the exact cheatcodes if there's a better way to implement the same functionality. Another possibility would be to have a cheatcode that adds an `address` or `StorageSlot` to the allowed list, and an assertion cheatcode that reverts if any other account or slot has been modified.

PR #1529, opened 2023-01-13
Cheatcodes for asserting that only some accounts/storage slots are modified

anvacaru:
Fixes: https://github.com/runtimeverification/evm-semantics/issues/1528

Drafted a whitelist system for calls and storage slots based on the issue above.

Currently supports the following cheat codes:
```solidity
    // Adds an address to the whitelist.
    function allowCallsToAddress(address) external;
    // Adds an address and a storage slot to the whitelist.
    function allowChangesToStorage(address,uint256) external;
```

Added a subconfiguration `whitelist` which contains two flags and two `Set`s considered as whitelists.
```
<whitelist>
  <isCallWhitelistActive> false </isCallWhitelistActive>
  <isStorageWhitelistActive> false </isStorageWhitelistActive>
  <addressSet> .Set </addressSet>
  <storageSlotSet> .Set </storageSlotSet>
</whitelist>
```

The `storageSlotSet` will contain elements of sort `StorageSlot`, while `addressSet` will contain `Int` values.

```
    syntax StorageSlot ::= "{" Int "|" Int "}"
```

The flags `isCallWhitelistActive` and `isStorageWhitelistActive` will be set to `true` as soon as a cheat code is used to add elements to the corresponding whitelists.

If the `isCallWhitelistActive` is true, the following rule will look for called addresses that are not in the whitelist. 
If an address which is not in the whitelist has been called, the `statusCode` gets updated to point a specific state and the `<k>` cell will contain the `#error_foundry` production with the address which has been called.
```
rule [foundry.catchNonWhitelistedCalls]:
    <k> #call _ ACCTTO _ _ _ _ false => #error_foundry ACCTTO ... </k>
    <statusCode> _ => FOUNDRY_WHITELISTCALL </statusCode>
    <whitelist>
      <isCallWhitelistActive> true </isCallWhitelistActive>
      <addressSet> WLIST </addressSet>
     ...
    </whitelist>
    requires notBool ACCTTO in WLIST
    [priority(40)]
```

A similar process goes for the `isStorageWhitelistActive`, but the `SSTORE` rule is overwritten instead of `#call`.


lucasmt (approved):
LGTM. Like you pointed out, `testAllow` does fail for me initially on symbolic values, but it passes if I use infinite gas.

PR #1530, opened 2023-01-13
Include sourcemap in KCFG output

ehildenb:
~Blocked on: https://github.com/runtimeverification/pyk/pull/227~
~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/1628~

This PR makes an attempt to take advantage of the KCFG custom view to display a solidity sourcemap. In particular:

- `solc` call now includes the `ast` output, which contains some needed sources information.
- Method `bytes_offset_to_lines` is added to utils to snip sections out of Solidity files based on file byte offsets.
- `Foundry` class now is able to compute sourcemaps in more detail, including snipping out chunks of Solidity files.
- When `KCFGViewer` is constructed with a `Foundry` in context, it is given the `custom_view` and improved `short_info` view for better display of the sourcemap.


kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -68,6 +76,61 @@ def contracts(self) -> Dict[str, Contract]:
...
+        contract = self.contracts[contract_name]
+        if pc not in contract.srcmap:
+            _LOGGER.info(f'pc not found in srcmap for contract {contract_name}: {pc}')
+            return None
+        s, l, f, _, _ = contract.srcmap[pc]
+        if f not in self.contract_ids:
+            _LOGGER.info(f'Contract id not found in sourcemap data: {f}')
+            return None
+        src_contract = self.contracts[self.contract_ids[f]]
+        # TODO: Hardcoding that out directory is just one deep in project
```
ehildenb, line 98: This is an indication that we should be taking the root of the foundry directory as input instead (or assuming `.`), and reading the appropriate `out` directory from the `foundry.toml`.

Then the directory to run in can become an optional argument to all our `foundry` commands.

ehildenb: Issue created: https://github.com/runtimeverification/evm-semantics/issues/1630

ehildenb:
I've confirmed locally that this gives roughly sensible sourcemaps that seem to work.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -420,11 +427,19 @@ def exec_run(
...
     kcfg = KCFGExplore.read_cfg(test, kcfgs_dir)
     if kcfg is None:
         raise ValueError(f'Could not load CFG {test} from {kcfgs_dir}')
-    viewer = KCFGViewer(kcfg, foundry.kevm, node_printer=foundry.kevm.short_info)
+
+    def _short_info(cterm: CTerm) -> Iterable[str]:
+        return foundry.short_info_for_contract(contract_name, cterm)
+
+    def _custom_view(elem: KCFGElem) -> Iterable[str]:
+        return foundry.custom_view(contract_name, elem)
```
tothtamas28, line 440: This is fine as it is, but if you want to make this a bit more concise you can use `functools.partial`.

kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -123,6 +123,26 @@ def parse(s: str) -> Tuple[T1, T2]:
     return parse
 
 
+def byte_offset_to_lines(lines: Iterable[str], byte_start: int, byte_width: int) -> Tuple[List[str], int, int]:
+    text_lines = []
+    line_start = 0
+    for l in lines:
```
tothtamas28, line 129:
```suggestion
    for line in lines:
```

tothtamas28 (approved):

kevm-pyk/src/kevm_pyk/__main__.py
ehildenb: I think I like the explicit look here.

kevm-pyk/src/kevm_pyk/utils.py
ehildenb: Done.

Issue #1531, opened 2023-01-20
README Instructions Issue


sskeirik:
When building KEVM, I found that the instruction order given didn't fully work for me. Quoting from the README:

>   ### Building
>   
>   Finally, you can build the semantics.
>   
>   ```sh
>   make build
>   ```
>   
>   And you need to set up the virtual environment:
>   
>   ```sh
>   make venv
>   ```
>   
>   Which should output (towards the end), a line like this: `. .build/venv/bin/activate`.
>   You should run this line in your shell, then you are in the virtual environment:
>   
>   ```sh
>   . .build/venv/bin/activate
>   ```

However, on a clean checkout, I found that I needed to perform:

```
make venv
make build
```

or else the build crashed as follows:

```
make[1]: Leaving directory '/home/stephen/work/evm-semantics/deps/plugin/deps/libff/build'
. .build/venv/bin/activate && kevm kompile --pyk --backend node                        \
    .build/usr/lib/kevm/include/kframework/evm-node.md \
    --main-module EVM-NODE            \
    --syntax-module EVM-NODE        \
    -I /usr/lib/kevm/include/kframework
/bin/sh: line 1: .build/venv/bin/activate: No such file or directory
make: *** [Makefile:270: .build/usr/lib/kevm/node/definition.kore] Error 1
```

It also makes sense the `make venv` is needed first, since the build command directly activates a virtual env.

sskeirik:
This issue appears to be resolved in master due to a different commit.

PR #1532, opened 2023-01-20
Update README instructions

sskeirik:
Closes: #1531

This PR attempts to re-order the instructions so the referenced issue is fixed.
I'm not 100% sure of where/how the virtual env is used, so I updated the text as best I could.


sskeirik:
The documentation issue that led to my PR appears to be resolved now. Either I was an old commit locally and missed this update or else someone else fixed the issue in the meantime.

Issue #1533, opened 2023-01-23
`expectRevert` cheatcode saves incorrect revert message for custom error types


lucasmt:
When calling the `expectRevert` cheatcode, we can pass as an argument the expected revert message, which can be the ABI encoding of a custom error type:
```
vm.expectRevert(
    abi.encodeWithSelector(CustomError.selector, 1, 2)
);
```
This error message is ABI encoded twice: first by `abi.encodeWithSelector` and then to encode the arguments of the `expectRevert` call. However, in the KEVM implementation of `expectRevert`, the twice-encoded cheatcode arguments are saved directly in the `expectedReason` cell:
```
rule [foundry.call.expectRevert]:
         <k> #call_foundry SELECTOR ARGS => #setExpectRevert ARGS ... </k>
      requires SELECTOR ==Int selector ( "expectRevert()" )
        orBool SELECTOR ==Int selector ( "expectRevert(bytes)" )

...

rule <k> #setExpectRevert EXPECTED => . ... </k>
         <callDepth> CD </callDepth>
         <expectedRevert>
           <isRevertExpected> false => true </isRevertExpected>
           <expectedDepth> _ => CD +Int 1 </expectedDepth>
           <expectedReason> _ => EXPECTED </expectedReason>
         </expectedRevert>
```
This means that `EXPECTED` in the second rule is not `abi.encodeWithSelector(CustomError.selector, 1, 2)`, but `abi.encode(abi.encodeWithSelector(CustomError.selector, 1, 2))`, which will not match the actual message once the revert happens. I think there are two options to fix this: either decode the message before saving it to the `expectedReason` cell, or encode the actual revert message before matching with the `expectedReason` cell.

PR #1535, opened 2023-01-25
expectRevert - encode output to match the expected bytes

anvacaru:
fixes: #1533 

The test `ExpectRevertTest.test_expectRevert_encodedSymbolic` is passing with infinite gas and additional lemmas which are not currently in the repo.

Suppose the `expectRevert` cheat code uses a custom error. In that case, the expected bytes are encoded as `abi.encode(abi.encodeWithSelector(CustomError.selector, 1, 2))`, while the `<output>` cell contains only `abi.encodeWithSelector(CustomError.selector, 1, 2)`. 
As a fix, if the output has an error selector different than the built-in `Error(string)`, then the output is encoded one more time to match the expected bytes.


include/kframework/foundry.md
```diff
@@ -558,9 +559,12 @@ Ignore all cheat code calls which take place while `expectRevert` is active.
...
+The `#halt` production is used to examine the end of each call in KEVM.
+If the call depth of the current call is lower than the call depth of the `expectRevert` cheat code and the `<statusCode>` is not `EVMC_SUCCESS`, the `#checkRevertReason` will be used to compare the output of the call with the expect reason provided.
+
 ```{.k .bytes}
     rule [foundry.handleExpectRevert]:
-         <k> (. => #checkRevertReason ~> #clearExpectRevert) ~> #halt ... </k>
+         <k> (. => #encodeOutput ~> #checkRevertReason ~> #clearExpectRevert) ~> #halt ... </k>
          <statusCode> SC </statusCode>
          <callDepth> CD </callDepth>
          <expectedRevert>
```
ehildenb, line 570: I think instead of having a whole `#encodeOutput` thing, since it's _only_ used here, let's jsut make it a function and inline it into this rule?

ehildenb: I'd also rather not modify the contents of the `<output>` cell directly like `<encodeOutput>` does, because anything that comes _after_ this step which expects the `<output>` cell to be unencoded will break. So I think it's best we just call the function to do the encoding at the time we're comparing the outptu against the expected value.

anvacaru: Done!

ehildenb (approved):

tests/foundry/exclude
```diff
@@ -40,6 +40,7 @@ EnvTest.testEnvUInt
 EnvTest.testEnvUIntArray
 ExpectRevertTest.testFail_expectRevert_bytes4
 ExpectRevertTest.test_expectRevert_bytes4
+ExpectRevertTest.test_expectRevert_encodedSymbolic
```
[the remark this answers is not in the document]
anvacaru: Because it requires infinite gas and additional lemmas that are not yet in the KEVM repo.

PR #1554, opened 2023-02-09
Remove java backend

ehildenb:
This removes the java backend entirely from the repository


Jenkinsfile
```diff
@@ -39,7 +39,6 @@ pipeline {
           options { timeout(time: 120, unit: 'MINUTES') }
           parallel {
             stage('Conformance (LLVM)') { steps {                                         sh 'make test-conformance -j4 TEST_CONCRETE_BACKEND=llvm'      } }
-            stage('Proofs (Java)')      { steps { lock("kevm-java-${env.NODE_NAME}")    { sh 'make test-prove       -j2 TEST_SYMBOLIC_BACKEND=java'    } } }
```
anvacaru, line 42: Do you know if the Jenkinsfile is currently being used? Could we remove it altogether?

README.md
```diff
@@ -44,7 +44,7 @@ Installing/Building
 
 ### K Backends
 
-There are three backends of K available: LLVM (default) for concrete execution and Haskell (default) and Java for symbolic execution.
+There are two backends of K available: LLVM for concrete execution and Haskell for symbolic execution.
```
anvacaru, line 47: Perhaps we should remove OpenJDK from the dependencies list as well; I can do that in #1558 .

anvacaru (approved):
There are two more modules, `INT-SIMPLIFICATION-JAVA` in `include/kframework/lemmas/int-simplification.k` and `LEMMAS-JAVA` in `include/kframework/lemmas/lemmas.k` . Is there a reason to exclude them?

README.md
ehildenb: OpenJDK is needed needed for K itself.

Jenkinsfile
ehildenb: It's used for the master release job.

PR #1563, opened 2023-02-14
Implement `KEVM.kompile` in terms of `kompile`

tothtamas28:
~Blocked on: https://github.com/runtimeverification/pyk/pull/204~
~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/1565~

### DIFF
--- a/kevm-pyk/src/kevm_pyk/kevm.py
+++ b/kevm-pyk/src/kevm_pyk/kevm.py
@@ -1,15 +1,13 @@
 import logging
 import sys
 from pathlib import Path
-from subprocess import CalledProcessError
 from typing import Final, Iterable, List, Optional
 
-from pyk.cli_utils import run_process
 from pyk.cterm import CTerm
 from pyk.kast.inner import KApply, KInner, KLabel, KSort, KToken, KVariable, build_assoc
 from pyk.kast.manip import flatten_label, get_cell
 from pyk.kast.outer import KFlatModule
-from pyk.ktool.kompile import KompileBackend
+from pyk.ktool.kompile import KompileBackend, kompile
 from pyk.ktool.kprint import SymbolTable, paren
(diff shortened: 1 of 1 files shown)


kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -75,33 +72,28 @@ def kompile(
...
-        except CalledProcessError as err:
-            sys.stderr.write(f'\nkompile stdout:\n{err.stdout}\n')
-            sys.stderr.write(f'\nkompile stderr:\n{err.stderr}\n')
-            sys.stderr.write(f'\nkompile returncode:\n{err.returncode}\n')
+            kompile(
+                main_file=main_file,
+                output_dir=definition_dir,
+                backend=backend,
+                emit_json=emit_json,
+                include_dirs=[include for include in includes if Path(include).exists()],
```
tothtamas28, line 81: This filtering is necessary as due to

```
KOMPILE_OPTS = -I $(INSTALL_INCLUDE)/kframework
```

the `make` targets add an include

```
-I /usr/lib/kevm/include/kframework
```

which does not necessarily exists.

kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -20,8 +18,6 @@
 from pyk.prelude.string import stringToken
 from pyk.utils import unique
 
-from .utils import add_include_arg
```
ehildenb, line 23: I believe this was the only use of this function, can we remove ti from `utils.py` too?

ehildenb (approved):

kevm-pyk/src/kevm_pyk/kevm.py
tothtamas28: There's one more reference to it in `__main__.exec_prove` unfortunately.

PR #1564, opened 2023-02-14
clearing the output after expectRevert

PetarMax:
If `expectRevert` suceeds, the `output` needs to be adjusted so that it can pass the correctness checks for a successful return. As setting it to `.ByteArray` did not work, Lucas and I believe that it has to be of the format `[32 bytes of return value length][that many bytes of actual return value]`. The simplest approach in that setting is to set it to 32 bytes denoting zero.

One example for which this is reproducible is the `LSP6KeyManagerTest.testCannotAddExtensions` in Lukso. We can try to create a more minimal test instead, if required.


anvacaru (approved):

ehildenb:
@anvacaru or @PetarMax is it possible to add a test for this? We should be able to minimize one from the example that was failing on Lukso.

PetarMax:
This should be possible, yes, perhaps a function that reverts with whatever
the relevant Lukso function was reverting with. I will synchronise with
Andrei.

lucasmt:
@PetarMax @anvacaru @ehildenb I've added a test confirming the behavior. Before the change implemented in this PR, the check succeeds in Foundry but fails in KEVM (in fact, it fails even without the `assertEq0`, because the format of the return value doesn't pass the compiler checks). With the change, it passes also in KEVM.

PR #1582, opened 2023-02-24
Create module `kevm_pyk.foundry`

tothtamas28:


ehildenb (approved):

ehildenb:
You may also consider moving the class `Foundry` from `kevm.py` to `foundry.py` now.

PR #1584, opened 2023-02-27
Factor out `print` from module `foundry`

tothtamas28:


kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -444,11 +471,12 @@ def foundry_simplify_node(
...
     with KCFGExplore(foundry, port=port, bug_report=br) as kcfg_explore:
         new_term = kcfg_explore.cterm_simplify(cterm)
-    new_term_minimized = new_term if not minimize else minimize_term(new_term)
-    print(f'Simplified:\n{foundry.pretty_print(new_term_minimized)}')
+    if minimize:
+        new_term = minimize_term(new_term)
     if replace:
         kcfg.replace_node(node, CTerm(new_term))
         KCFGExplore.write_cfg(test, kcfgs_dir, kcfg)
+    return foundry.pretty_print(new_term)
```
ehildenb, line 479: This is probably a bug that existed before, but maybe you can fix it now. But replacement functionality should be happening _before_ minimization, so we are storing the unminimized variant of the term, not the minimized variant.

anvacaru (approved):

PR #1594, opened 2023-02-28
EVM type refactoring

### LINKED ISSUE #1590: Remove distinction between #range and _[_.._]

anvacaru:
- [x] remove `_ [_ .. _]` operator (fixes #1590)
- [x] remove `Memory` sort
- [x] remove `ByteArray` sort
   - [x] remove `#sizeByteArray(_)` and use `lengthBytes(_)` instead
   - [x] remove `_ ++ _` and use `+Bytes` instead


include/kframework/lemmas/lemmas.k
```diff
@@ -147,13 +146,13 @@ module LEMMAS [symbolic]
...
 
-    rule #asWord(WS) >>Int M => #asWord(WS [ 0 .. #sizeByteArray(WS) -Int (M /Int 8) ])  requires 0 <=Int M andBool M modInt 8 ==Int 0 [simplification]
+    rule #asWord(WS) >>Int M => #asWord( #range(WS, 0, #sizeByteArray(WS) -Int (M /Int 8) ))  requires 0 <=Int M andBool M modInt 8 ==Int 0 [simplification]
 
     rule #asWord(#padRightToWidth(32, BUF)) &Int notMaxUInt224 =>
          #asWord(#padRightToWidth(32, BUF))
       requires #sizeByteArray(BUF) <=Int 4 [simplification]
 
-    rule #padRightToWidth(_, BUF)[ 0 .. WIDTH ] => BUF requires #sizeByteArray(BUF) ==Int WIDTH [simplification]
+    rule #range(#padRightToWidth(_, BUF), 0, WIDTH) => BUF requires #sizeByteArray(BUF) ==Int WIDTH [simplification]
```
ehildenb, line 156: I would try just removing simplification rules like this instead of adapting them.

ehildenb:
Some thoughts:

- Just remove `simplification` rules about `_[_.._]`, don't try to adapt them. When we see what fails we can bring back some of the lemmas needed.
- The ranged write operation `_[_:=_]` needs to be either ported to `Bytes`, or remade into something like `#writeRange(...)` (maybe @PetarMax has a preference on how to name this).
- We should be able to get rid of sorts `Memory` and `ByteArray` now, as they should just be aliases of the `Bytes` sort now (the Jvaa backend had different implementation of these sorts).

include/kframework/lemmas/lemmas.k
anvacaru: Done.

anvacaru:
Removed the `Memory` and `ByteArray` sort. Conformance tests are passing. I want also to remove the `++` operator since it is a duplicate of `+Bytes` currently.

include/kframework/evm-types.md
```diff
@@ -257,10 +257,10 @@ A cons-list is used for the EVM wordstack.
...
-    rule #take(N,  BS:Bytes) => #padRightToWidth(N, .Bytes)                 requires notBool lengthBytes(BS) >Int 0 andBool         N >Int 0
-    rule #take(N,  BS:Bytes) => BS ++ #take(N -Int lengthBytes(BS), .Bytes) requires         lengthBytes(BS) >Int 0 andBool notBool N >Int lengthBytes(BS)
-    rule #take(N,  BS:Bytes) => BS [ 0 .. N ]                               requires         lengthBytes(BS) >Int 0 andBool         N >Int lengthBytes(BS)
+    rule #take(N, _BS:Bytes) => .Bytes                                          requires                                        notBool N >Int 0
+    rule #take(N,  BS:Bytes) => #padRightToWidth(N, .Bytes)                     requires notBool lengthBytes(BS) >Int 0 andBool         N >Int 0
+    rule #take(N,  BS:Bytes) => BS +Bytes #take(N -Int lengthBytes(BS), .Bytes) requires         lengthBytes(BS) >Int 0 andBool notBool N >Int lengthBytes(BS)
+    rule #take(N,  BS:Bytes) => #range(BS, 0, N)                                requires         lengthBytes(BS) >Int 0 andBool         N >Int lengthBytes(BS)
```
ehildenb, line 263: I believe that this `#take` is actually no longer used. There is another `#take` for WordStack that is still used, but this one is not I believe. I'm pretty sure the same is true for `#drop` over `Bytes` type, no longer used. There is another `#drop` over wordstack that is still used, that one should stay.

include/kframework/evm.md
```diff
@@ -2191,21 +2192,21 @@ There are several helpers for calculating gas (most of them also specified in th
...
+                 | Caddraccess    ( Schedule , Bool )                    [function, total, smtlib(gas_Caddraccess)   ]
+                 | Cstorageaccess ( Schedule , Bool )                    [function, total, smtlib(gas_Cstorageaccess)]
+                 | Csload         ( Schedule , Bool )                    [function, total, smtlib(gas_Csload)        ]
+                 | Cextcodesize   ( Schedule )                           [function, total, smtlib(gas_Cextcodesize)  ]
+                 | Cextcodecopy   ( Schedule , Int )                     [function, total, smtlib(gas_Cextcodecopy)  ]
+                 | Cextcodehash   ( Schedule )                           [function, total, smtlib(gas_Cextcodehash)  ]
+                 | Cbalance       ( Schedule )                           [function, total, smtlib(gas_Cbalance)      ]
+                 | Cmodexp        ( Schedule , Bytes , Int , Int , Int ) [function, total, smtlib(gas_Cmodexp)       ]
  // ----------------------------------------------------------------------------------------------------------------------
```
ehildenb, line 2210: Small formatting issue.

include/kframework/foundry.md
```diff
@@ -1076,12 +1076,12 @@ Utils
...
+    syntax KItem ::= "#setCode" Int Bytes [klabel(foundry_setCode)]
+ // ---------------------------------------------------------------
     rule <k> #setCode ACCTID CODE => . ... </k>
          <account>
            <acctID> ACCTID </acctID>
-           <code> _ => #if #asWord(CODE) ==Int 0 #then .ByteArray:AccountCode #else {CODE}:>AccountCode #fi </code>
+           <code> _ => #if #asWord(CODE) ==Int 0 #then .Bytes:AccountCode #else {CODE}:>AccountCode #fi </code>
            ...
          </account>
 ```
```
ehildenb, line 1087: Do we actually need this semantic kast anymore with simplified types? Can we remove the `:AccountCode` and the `{ ... }:>AccountCode` and it still builds OK?

kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -177,7 +172,6 @@ def concrete_rules() -> List[str]:
             'EVM-TYPES.#asByteStack',
             'EVM-TYPES.#asByteStackAux.recursive',
             'EVM-TYPES.#asWord.recursive',
-            'EVM-TYPES.ByteArray.range',
             'EVM-TYPES.bytesRange',
```
ehildenb, line 175: Instead of deleting, should it be updated in the same way that `concrete-rules.txt` is updated?

ehildenb (approved):

kevm-pyk/src/kevm_pyk/kevm.py
anvacaru: @ehildenb This list already contains an `'EVM-TYPES.#range'`. I assumed I would duplicate it by renaming it.

include/kframework/foundry.md
anvacaru: It builds without an error now.

include/kframework/evm.md
anvacaru: Done.

include/kframework/evm-types.md
anvacaru: Done.

tests/specs/concrete-rules.txt
```diff
@@ -25,7 +25,7 @@ EVM.Rsstore.old
 EVM-TYPES.#asByteStack
 EVM-TYPES.#asByteStackAux.recursive
 EVM-TYPES.#asWord.recursive
-EVM-TYPES.ByteArray.range
+EVM-TYPES.Bytes.range
 EVM-TYPES.bytesRange
 EVM-TYPES.mapWriteBytes.recursive
```
ehildenb, line 30: Here in concrete-rules.txt we are making a change, that I don't see the same change happening in kevm.py? Are you sure we don't need to make the same change there?

kevm-pyk/src/kevm_pyk/kevm.py
ehildenb: I called out the other list I'm talking about with a comment, have a look. Basically, the concrete_rules list in `kevm.py` and the `concrete-rules.txt` file should agree with each other.

PR #1550, opened 2023-02-07
First step towards making KEVM storage transactional

ChristianoBraga:
- Add `Storage` as supersort of `Map` and function `write`.
  + Write simply behaves as Map in this commit.
- Changed semantics of `SLOAD` and `SSTORE` to cope with `Storage`.
- Changed signature of `#lookup` to cope with `Storage`.

PR #1598, opened 2023-03-01
Transactional storage

nwatson22:
This PR is based on #1550 and suggestions from @ehildenb. It aims to makes storage transactional within the execution of a contract.  The `#write()` symbol is used to track writes to the storage. At the end of execution the `#write`s are flattened into a map.


nwatson22:
What I am trying to do here requires calling `#resolveWrites` on all accounts that have `#write` symbols in their `<storage>`. I initially tried performing this based on the accounts listed in `<activeAccounts>`, but it makes more sense to do it using the keys of `<accessedStorage>`. The problem with this was that on schedules before Berlin, this map is never populated because, I guess, there is no need for it since those older types of transactions can't include an accessList.

For this reason I removed the check [here](https://github.com/runtimeverification/evm-semantics/blob/29fe23fade9f83ecb87cfa5471f48438552408e3/include/kframework/evm.md?plain=1#L1863), so now the storage locations are tracked regardless of the schedule.

This causes another problem, however. There are some tests, such as `tests/ethereum-tests/BlockchainTests/GeneralStateTests/stEIP2930/addressOpcodes.json`, which have a transaction with an `accessList` field referring to an account that doesn't have a corresponding account in the `pre` list. It seems like only the `pre` transactions get loaded into `<accounts>`, so when we try to resolve accounts, there is no account being pointed to by a map key in `<accessedStorage>`.

ehildenb:
We can reopen if we need.

Issue #1588, opened 2023-02-27
Upstream lemmas from previous engagements


anvacaru:
- [ ]  We should upstream lemmas from previous engagements to the KEVM repo.

- include/kframework/lemmas/lemmas.k
- include/kframework/lemmas/int-simplification.k
- include/kframework/infinite-gas.md

With `run-lemma`, `done-lemma` style tests in : tests/specs/functional/...

- [ ] Move Byte-array simplification lemmas to a new module include/kframework/lemmas/bytes-simplification.k

- [ ]  Add a module to int-simplification.k called BITWISE-ARITHMETIC-SIMPLIFICATION (edited)

PR #1609, opened 2023-03-06
Move byte-array simplification lemmas to a separate module

lucasmt:
Create new lemmas file `bytes-simplification.k` with `BYTES-SIMPLIFICATION` module, and move lemmas related to byte arrays from `lemmas.k` to the new module.

Part of Issue #1588.


include/kframework/lemmas/bytes-simplification.k
```diff
@@ -0,0 +1,108 @@
...
+module BYTES-SIMPLIFICATION-HASKELL [symbolic, kore]
+    imports BYTES-SIMPLIFICATION-COMMON
+    imports INT-SYMBOLIC
+    imports EVM
+    imports BUF
+
+  // ########################
+  // Buffer Reasoning
+  // ########################
+
```
ehildenb, line 14: We can just combine the modules into one `BYTES-SIMPLIFICATION [symbolic]` instead of having `COMMON` and `HASKELL` module now.

lucasmt: Makes sense, I've made that change.

ehildenb (approved):

PR #1616, opened 2023-03-07
Move more logic of kevm kompile into python

ehildenb:
This moves logic of `bin/kevm` into Python script for kompile.

- Logic which is finding paths on the usres system is still done in bash.
- Arguments are added to `kevm kompile ...` in python to allow passing in those discovered paths as arguments.
- Most of the logic of building the actual arguments to the backend is moved to python.
- Script is simplified by enforcing that pyk must be used for `kevm kompile ...`.
- When a proof crashes in `parallel_kcfg_prove`, we try to print back the stacktrace as well.


kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -73,26 +73,73 @@ def exec_kompile(
     backend: KompileBackend,
     main_file: Path,
     emit_json: bool,
+    kompile_mode: str,
```
tothtamas28, line 76:
```suggestion
    kompile_mode: Union[str, KompileMode],
```

where

```python3
class KompileMode(Enum):
    NODE = 'node'
    STANDALONE = 'standalone'
```

Then you can process it like this:

```python3
kompile_mode = KompileMode(kompile_mode)
```

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -73,26 +73,73 @@ def exec_kompile(
...
     includes: List[str],
     main_module: Optional[str],
     syntax_module: Optional[str],
-    md_selector: Optional[str],
     ccopts: Iterable[str] = (),
     llvm_kompile: bool = True,
+    target: Optional[str] = None,
```
tothtamas28, line 82: Same for target:

```suggestion
    target: Optional[Union[str, KompileTarget]] = None,
```

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -647,6 +693,19 @@ def parse(s: str) -> List[T]:
         'kompile', help='Kompile KEVM specification.', parents=[shared_args, k_args, k_kompile_args]
     )
     kompile_args.add_argument('main_file', type=file_path, help='Path to file with main module.')
+    kompile_args.add_argument(
+        '--kompile-mode', type=str, default='standalone', help='KEVM kompile mode, [standalone|node].'
```
tothtamas28, line 697:
```suggestion
        '--kompile-mode', choice=('standalone', 'node'), default='standalone', help='KEVM kompile mode, [standalone|node].'
```

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -647,6 +693,19 @@ def parse(s: str) -> List[T]:
...
+    )
+    kompile_args.add_argument('--plugin-include', type=dir_path, help='Path to plugin include directory.')
+    kompile_args.add_argument('--libff-dir', type=dir_path, help='Path to libff include directory.')
+    kompile_args.add_argument('--target', type=str, default='linux', help='Compilation target, [linux|darwin].')
```
tothtamas28, line 701:
```suggestion
    kompile_args.add_argument('--target', choice=('linux', 'darwin'), default='linux', help='Compilation target, [linux|darwin].')
```

kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -92,7 +93,7 @@ def _call_rpc(packed_args: Tuple[str, KCFG, int]) -> bool:
                     implication_every_block=implication_every_block,
                 )
             except Exception as e:
-                _LOGGER.error(f'Proof crashed: {_cfgid}\n{e}')
+                _LOGGER.error(f'Proof crashed: {_cfgid}\n{e}\n{traceback.format_exc()}')
```
tothtamas28, line 96: Adding this might not be necessary if you specify `exc_info=True` ([docs](https://docs.python.org/3/library/logging.html#logging.debug)). I haven't tried this though, so don't take my word for it.

tothtamas28 (approved):

ehildenb:
From @tothtamas28 , you can use `__file__` to get path to the current file (to determine where the libraries are installed).

kevm-pyk/src/kevm_pyk/utils.py
ehildenb: Done.

kevm-pyk/src/kevm_pyk/__main__.py
ehildenb: I ended up using an Enum directly, have a look.

kevm-pyk/src/kevm_pyk/__main__.py
ehildenb: I just used an Enum directly.

kevm-pyk/src/kevm_pyk/__main__.py
ehildenb: Used an Enum directly.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -644,6 +692,24 @@ def parse(s: str) -> List[T]:
...
+    kompile_args.add_argument(
+        '--kompile-mode',
+        type=KEVMKompileMode,
+        default=KEVMKompileMode.STANDALONE,
+        help='KEVM kompile mode, [standalone|node].',
+    )
+    kompile_args.add_argument('--plugin-include', type=dir_path, help='Path to plugin include directory.')
+    kompile_args.add_argument('--libff-dir', type=dir_path, help='Path to libff include directory.')
+    kompile_args.add_argument(
+        '--target', type=KompileTarget, default=KompileTarget.LINUX, help='Compilation target, [linux|darwin].'
```
tothtamas28, line 704: This solution is fine as well. Why I think `choices` is slightly better though is it prints valid options on bad input.

PR #1617, opened 2023-03-08
Allow custom RPC server command, optionally produce a dy.lib using llvm

jberthold:
This implements support for custom RPC servers when running foundry proofs
* in `kevm foundry-prove`, the RPC server command can be customised using `--with-custom-rpc "command line to start rpc server". Definition and `--main-module` option are however still appendedto the given custom command
* in `kevm foundry-kompile`, the option `--with-llvm-library` adds an llvm backend compilation with `--llvm-kompile-type c`. This will generate a dynamic library in target directory `<foundry-out>/kompiled-llvm/`.

For the LLVM compile option to succeed, the EVM semantics must not contain existentials. An [additional commit on this branch](https://github.com/runtimeverification/evm-semantics/commit/198eaf522a7a02942dd090eece9483cbfcdc93bd) removes the current existentials (`?_VGAS`, `?_VCALLGAS`, `?STORAGE`) but has been reverted here to keep downstream dependencies in working state.

Part of https://github.com/runtimeverification/hs-backend-booster/issues/128


kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -588,6 +599,13 @@ def parse(s: str) -> List[T]:
         action='store_false',
         help='Do not run llvm-kompile process.',
     )
+    k_kompile_args.add_argument(
+        '--with-llvm-library',
+        dest='llvm_library',
+        default=True,
```
jberthold, line 605:
```suggestion
        default=False,
```

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -280,6 +282,7 @@ def exec_foundry_prove(
     implication_every_block: bool = True,
     rpc_base_port: Optional[int] = None,
     bug_report: bool = False,
+    rpc_command: Optional[str] = None,
```
ehildenb, line 285: We can probably set the default here, and simplify the handling logic? `rpc_command: str = 'kore-rpc',`?

ehildenb: I want to see @tothtamas28 opinion on how to do this though.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -535,6 +539,13 @@ def parse(s: str) -> List[T]:
         help='Store every Nth state in the CFG for inspection.',
     )
 
+    explore_args.add_argument(
+        '--with-custom-rpc',
+        dest='rpc_command',
+        type=str,
+        help='Custom command to start RPC server',
+    )
```
ehildenb, line 547:
```suggestion
    explore_args.add_argument(
        '--with-custom-rpc',
        dest='rpc_command',
        type=str,
        default='kore-rpc',
        help='Custom command to start RPC server',
    )
```

Similarly, we _could_ set the default here. Does it make sense to do so?

kevm-pyk/src/kevm_pyk/__main__.py
jberthold: I was trying to keep `Foundry.foundry_prove` behaviour unchanged by default... maybe not necessary, though.
Happy to follow any other suggestions. What we need to keep is the second default in `utils.py:parallel_config_explore`, since that function is being called from `exec_prove`, too.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -171,6 +173,21 @@ def foundry_kompile(
...
+                foundry_main_file,
+                emit_json=True,
+                includes=includes,
+                main_module_name=main_module,
+                syntax_module_name=syntax_module,
+                md_selector=md_selector,
+                debug=debug,
+                ccopts=ccopts,
+                llvm_kompile_type=LLVMKompileType.C,
+            )
```
tothtamas28, line 190: Consider extracting a nested function that captures arguments shared between the two calls, and exposes the rest as parameters.

kevm-pyk/src/kevm_pyk/__main__.py
tothtamas28: For parameters that inherit the default value of a called function I like to go with `Optional`:

I.e. in that case I prefer this:

```python3
def foo(x=None):
    return bar(x)

def bar(x=None):
    x = 0 if x is None else x
    return x
```

Over this:

```python3
def foo(x=0):
    return bar(x)

def bar(x=0):
    return x
```

(It gets worse as nesting increases.)

(Also, there are cases when exposing the default value is not  an option / is not advised, e.g. if the value is mutable.)

jberthold: OK, so IIUC the code can stay as it is now (setting defaults later rather than earlier, using `Optional` while going through the call stack)

kevm-pyk/src/kevm_pyk/__main__.py
jberthold: According to Tamas' preference, it is better to use an optional parameter

ehildenb (approved):

PR #1619, opened 2023-03-08
Set empty initial storage for Foundry Contract

### LINKED ISSUE #1618: Initialize contract storage with the empty storage instead of the symbolic one

anvacaru:
fixes #1618


kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -622,7 +622,7 @@ def _init_term(
         Foundry.address_TEST_CONTRACT(),
         intToken(0),
         program,
-        KVariable('ACCT_STORAGE'),
+        KApply('.Map'),
         KVariable('ACCT_ORIGSTORAGE'),
         intToken(0),
     )
```
ehildenb, line 628: Should we also be setting `origStorage` to empty as well?

ehildenb:
It looks good to me, but @anvacaru did you confirm that this model makes sense given how Foundry works?

anvacaru:
I did not examine the kcfgs, but without the infiniteGas, the proofs were running for a long time (killed them after an hour of running).
If we chose to have initial concrete values for storage, I'd say to make them concrete for the other fields as well.

ehildenb:
Yes, I agree, I think taht makes most sense. Let me know when to review again.

ehildenb (approved):

tothtamas28 (approved):

PR #1627, opened 2023-03-11
Updates to Foundry integration

ehildenb:
This accumulates several small fixes to the Foundry integration that have happened over time:

- Adds names for the `JUMPI` rules, and option `--break-on-jumpi` which allows creating a new KCFG node each `JUMPI` instruction.
- Sets the correct max-depth to 1000 for explorations.
- Downgrades several annoying warning messages to info messages.
- Adds some foundry tests for `sum_to_N` for various `N`.
- Uses the alias `notMaxUInt160` in various places where the integer had been hardcoded.
- Adds another test of address reprojection, and includes a claim to test it, which came up from some foundry tests, and adds the lemma to make it pass.
- Changes the initialization of foundry tests to _not_ double up on `add_invariant`, which was causing us to send terms with `macro` symbols in them to the backend (issue here: https://github.com/runtimeverification/pyk/issues/248).


kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -334,16 +335,14 @@ def foundry_prove(
             method = [m for m in contract.methods if m.name == method_name][0]
             empty_config = foundry.kevm.definition.empty_config(GENERATED_TOP_CELL)
             kcfg = _method_to_cfg(empty_config, contract, method)
-            init_term = kcfg.get_unique_init().cterm.kast
-            target_term = kcfg.get_unique_target().cterm.kast
             _LOGGER.info(f'Expanding macros in initial state for test: {test}')
+            init_term = kcfg.get_unique_init().cterm.kast
             init_term = KDefinition__expand_macros(foundry.kevm.definition, init_term)
-            init_cterm = KEVM.add_invariant(CTerm(init_term))
```
ehildenb, line 341: This call to `add_invariant` was happening _after_ macro expansion, and was adding terms like `#rangeUInt(...)` to the constraint. This call is removed, because the earlier call to `_method_to_cfg` already calls `add_invariant` for us anyway.

kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -41,10 +41,11 @@ def parallel_kcfg_explore(
     kprove: KProve,
     proof_problems: Dict[str, KCFG],
     save_directory: Optional[Path] = None,
-    max_depth: int = 100,
+    max_depth: int = 1000,
```
tothtamas28, line 44: Consider

```suggestion
    max_depth: Optional[int] = None,
```

and setting the default `int` value in the function body. Then callers don't need to be aware of the actual default value.

tothtamas28 (commented):
Python part looks good.

kevm-pyk/src/kevm_pyk/utils.py
ehildenb: I feel like the default value gives people an idea of what a "sane" value here is though. Otherwise someone might think "oh, maybe 10 is a good value", or they may think "maybe 1000000 is a good value"?

tests/foundry/test/Loops.t.sol
```diff
@@ -48,6 +48,27 @@ contract LoopsTest is Test {
...
+    function test_sum_10() public returns (uint) {
+        return sum_N(10);
+    }
+
+    function test_sum_100() public returns (uint) {
+        return sum_N(100);
+    }
+
+    function test_sum_1000() public returns (uint) {
+        return sum_N(1000);
```
lucasmt, line 70: Is the purpose of these tests to just check that KEVM is able to execute the loop to the end? We could also add an `assert` that the result is correct.

lucasmt (approved):
K changes look good

tothtamas28 (approved):

PR #1628, opened 2023-03-12
Move initialization of directories into Foundry class

ehildenb:
Currently, all the places which construct a `Foundry` object do the same initialization. Also, we assume that everywhere that needs a `Foundry` needs a `KEVM`, and these objects really are different. This moves that common logic into the `Foundry` constructor instead, just passing in the `foundry_out` directory to that constructor, and makes it so that it can construct a default `KEVM`, but not _is-a_ `KEVM`.

- `Foundry` class no longer inherits from `KEVM` class, rather it has a method `kevm()` which provides the "default" way to get a KEVM from a foundry instantiation. This separation makes sense (Foundry is its own thing, not a KEVM), but also makes `foundry_kompile` logic simpler: we need to have a `Foundry` around _before_ there is a compiled definition for it (so we can read in the smart contracts and generate the K needed for it), but we also need a KEVM around (the default foundry build shipped with KEVM) to do the compilation after the K is generated.
- Pull logic of initializing directories for kompiled definition into `Foundry.kevm` method.
- Remove (currently useless) sourcemap information from contract calculation. Was not actually calculating a useful sourcemap, and does not fit in cleaned up architecture. Will bring back in: https://github.com/runtimeverification/evm-semantics/pull/1530.
- Construct `Foundry.contracts` map directly in `Foundry` class, not managed separately (helps with calculating sourcemap in later PR).

Overall this should be slightly better for performance too, as we're caching the loaded contracts in the `Foundry` object, and only constructing the `KEVM` object once when it's needed with `Foundry.kevm`.


anvacaru (approved):
LGTM, but I would like input from @tothtamas28  as well 👍

tothtamas28 (approved):
Great refactoring!

PR #1631, opened 2023-03-14
Use KORE structures in `gst_to_kore`

### LINKED ISSUE #1615: Migrate `gst_to_kore.py` over to constructing proper Kore data-structures.
Currently, `gst_to_kore.py` is directly building the Kore term as output for the input state to KEVM.

It should be better integrated with the rest of the tool-suite, instead directly constructing actual Kore (for sanity checks), and then using the `text` method to serialize to text.

Note that this functionality is used in `make test-conformance -j12`, which must be _fast_. The current overhead of runner scripts is around 5-15ms, and it needs to remain around that time, because there are ~40k conformance tests to run in parallel, most of which complete in less then 0.01s.

tothtamas28:
Fixes #1615


tothtamas28:
On `test-conformance` this is 12% slower than writing to the output stream directly.

```
git:(master) % /usr/bin/time poetry -C kevm-pyk run make test-conformance KEVM_OPTS+=--pyk -j6
10606.25user 486.00system 30:49.21elapsed 599%CPU (0avgtext+0avgdata 3829128maxresident)k
1998664inputs+14222792outputs (0major+340342967minor)pagefaults 0swaps
```

```
git:(gst-to-kore) % /usr/bin/time poetry -C kevm-pyk run make test-conformance KEVM_OPTS+=--pyk -j6
11875.21user 897.30system 35:29.36elapsed 599%CPU (0avgtext+0avgdata 3828924maxresident)k
2010608inputs+17259368outputs (1major+611031240minor)pagefaults 0swaps
```

ehildenb (approved):

PR #1635, opened 2023-03-17
Add definedness constraint to `KCFG` init state

tothtamas28:
Fixes https://github.com/runtimeverification/pyk/issues/263

~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/1636~


kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -338,7 +338,14 @@ def foundry_prove(
...
             init_term = KDefinition__expand_macros(foundry.kevm.definition, init_term)
-            kcfg.replace_node(kcfg.get_unique_init().id, CTerm(init_term))
+            init_cterm = CTerm(init_term)
+
+            _LOGGER.info(f'Computing definedness constraint for initial state for test: {test}')
+            with KCFGExplore(foundry.kevm, port=find_free_port(), bug_report=br) as kcfg_explore:
+                init_cterm = kcfg_explore.cterm_assume_defined(init_cterm)
+
+            kcfg.replace_node(kcfg.get_unique_init().id, init_cterm)
+
```
ehildenb, line 348: It would be good to just have a single `with KCFGExplore ...` if we need to do both definedness conditions and simplification, because starting up the RPC servers takes time.

tothtamas28: Good point, fixed.

ehildenb:
What's confusing me right now is why does KEVM work on master currently? We removed the assumption of definedness already in the execute endpoint, so how come the proofs on KEVM are currently passing without this change already?

Did the `execute` endpoint get changed to assume definedness of its input @jberthold ?

jberthold:
> Did the `execute` endpoint get changed to assume definedness of its input @jberthold ?

There were no changes to the `execute` endpoint in the past month AFAICT.

tothtamas28:
> We removed the assumption of definedness already in the execute endpoint, so how come the proofs on KEVM are currently passing without this change already?

@ehildenb, is it possible some change in the initial configuration, in particular https://github.com/runtimeverification/evm-semantics/pull/1619, made the definedness constraint `\top`?

ehildenb:
Hmmmm, I don't think that change would have done it, because the main definedness condition we needed was not on the storage of that one account, but on the contents of the rest of the account cell map I believe (see here: https://github.com/runtimeverification/evm-semantics/blob/master/kevm-pyk/src/kevm_pyk/foundry.py#L667). 

Are you able to run a few foundry tests, and have it log teh Ceil condition that it ends up adding to the debug output and observe it to post here. If simplifying `#Ceil(init_term)` is causing definedness conditions to show up, then we must be feeding potentially unndefined terms into the backend, and should look at why it can handle them.

nwatson22:
This change eliminated the unexpected branching I was getting in foundry proofs.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -329,23 +329,34 @@ def foundry_prove(
...
+            _LOGGER.info(f'Starting KCFGExplore for test: {test}')
+            with KCFGExplore(
+                foundry.kevm,
+                bug_report=br,
+                kore_rpc_command=kore_rpc_command,
+                smt_timeout=smt_timeout,
+                smt_retry_limit=smt_retry_limit,
+            ) as kcfg_explore:
+                _LOGGER.info(f'Computing definedness constraint for test: {test}')
```
ehildenb, line 352: See: https://github.com/runtimeverification/pyk/pull/271/files#r1145638379

No change needed here, but it reminded me that we should be logging in the cterm_assume_defined method.

ehildenb (approved):

ehildenb:
For completeness, I'm pasting a discussion that we had on Slack here: https://runtimeverification.slack.com/archives/C03NP1UP2KZ/p1679508683801689

Basically, when https://github.com/runtimeverification/pyk/pull/271 was merged into KEVM, we started getting an immediate branch on all initial states using teh RPC prover. The branch where the initial configuration was _not_ defined, led to immediate subsumption into the target state, because `#Bottom` implies anything. Then execution was continuing directly on the other branch (where the backend assumed definedness), and that was the actual execution we wanted.

Merging this change will just get rid of that spurious (and annoying) branch.

PR #1659, opened 2023-03-22
Remove `kast-json.py`

tothtamas28:
Closes https://github.com/runtimeverification/evm-semantics/pull/1612


bin/kevm
```diff
@@ -124,18 +124,22 @@ run_krun() {
...
                          execute kore-json.py "${run_file}" "${cSCHEDULE_kore}" "${cMODE_kore}" "${cCHAINID_kore}"
                      else
                          check_k_install
                          gst-to-kore "${run_file}" --schedule "${schedule}" --mode "${mode}" --chainid "${chainid}" "$@"
                      fi
                      ;;
+        *.json-kast) fatal "kast output is not supported for .json input"
+                     ;;
         *)           check_k_install ; kast --definition "${backend_dir}" "${run_file}" --output "${output_mode}" "$@"
                      ;;
```
ehildenb, line 144: Do we need it to be unsupported? Or could we just fallback to `kast`?

```suggestion
        *.json-kast) check_k_install ; kast --definition "${backend_dir}" "${run_file}" --output "${output_mode}" --sort JSON "$@"
                     ;;
        *)           check_k_install ; kast --definition "${backend_dir}" "${run_file}" --output "${output_mode}" "$@"
                     ;;
```

ehildenb: Also, now that I'm looking at it, gst-to-kore (and what `kore-json.py` and `kast-json.py` did before) are fundamentally different than this call to `kast`. This call to `kast` just parses a term, `gst-to-kore` builds an initial configuration. They probably should be pulled into their own functions.

ehildenb (approved):

bin/kevm
ehildenb: Going ot merge this because my point above doesn't really matter, no one uses this  functionality anyway.

PR #1662, opened 2023-03-23
Upstream lemmas from Lukso engagement

lucasmt:
Part of: https://github.com/runtimeverification/evm-semantics/issues/1588
~Blocked: https://github.com/runtimeverification/evm-semantics/pull/1675~

TODOS:

- [ ] Create a separate module for Set and Bool lemmas (probably better to make a separate PR)
- [ ] Simplify claims
- [ ] Remove redundant claims
- [ ] Split claims into different files


include/kframework/lemmas/bytes-simplification.k
```diff
@@ -3,6 +3,19 @@ module BYTES-SIMPLIFICATION [symbolic]
     imports EVM
     imports BUF
 
+  //
+  // concatHead/concatTail
+  //
+
+    syntax Bytes ::= concatHead(Bytes) [function, total] // Leftmost operand in a concat
+                   | concatTail(Bytes) [function, total] // Rest of concat
```
lucasmt, line 11: Is this the right place to put this definition?

ehildenb: Ah, I see it's to extract the left-most position now..... Hmmmmm, it seems to me that the same can be accomplished by just re-associating `_+Bytes_` with a rule like:

```
rule (BA1 +Bytes BA2) +Bytes BA3 => BA1 +Bytes (BA2 +Bytes BA3) [simplification]
```

Then have all the rules which use `concatHead` and `concatTail` just do direct pattern matching. Basically state in that we prefer right-associated byte-arrays.

PetarMax: It is more general than pattern-matching, it is able to traverse arbitrarily associated concats. 

It seems to me as well that that re-association would isolate the head and the tail, but I am not sure what it would do to other simplifications (e.g., for combining ranges) that we use or how it would change the state in the contracts that we analysed.

I do welcome the idea of having a preferred associativity in principle.

include/kframework/lemmas/bitwise-simplification.k
```diff
@@ -20,85 +22,108 @@ module BITWISE-SIMPLIFICATION [symbolic]
...
-    rule N &Int maxUInt8   => N  requires #rangeUInt(8,   N) [simplification]
-    rule maxUInt8 &Int N   => N  requires #rangeUInt(8,   N) [simplification]
+    // PETAR: Could the 64 subsequent simplifications be conflated into one?
+    //        Or two, if commutativity can't be made to work? Maybe:
+    // rule [bitwise-and-identity]:
+    //   X &Int Y => Y
+    //   requires 0 <=Int X
+    //    andBool X +Int 1 ==Int 2 ^Int log2Int(X +Int 1)
+    //    andBool #rangeUInt(log2Int(X +Int 1), Y)
+    //   [concrete(X), simplification, comm]
```
ehildenb, line 41: Did we try this? If it works it would be quite nice to remove all the other ones. I think it should work. `&Int` has the `comm` attribute, so it may work: 
image

ehildenb: I don't know how well `concrete(X)` will interact with `comm` though. It may be worth inspecting the generated `definition.kore` to make sure it's behaving as expected.

ehildenb: Added an issue, so it can be addressed in a later PR: https://github.com/runtimeverification/evm-semantics/issues/1709

include/kframework/lemmas/bitwise-simplification.k
```diff
@@ -20,85 +22,108 @@ module BITWISE-SIMPLIFICATION [symbolic]
...
     rule maxUInt160 &Int (X |Int (notMaxUInt160 &Int Y:Int)) => X requires #rangeUInt(160, X) andBool 0 <=Int Y [simplification]
 
+    rule [bitwise-and-asWord-tail]:
+      X &Int #asWord ( _ +Bytes B ) => X &Int #asWord ( B )
+      requires 0 <=Int X andBool X <Int 2 ^Int (8 *Int lengthBytes(B))
+      [simplification]
+
+    rule [bitwise-and-asWord-tail-shift]:
+      X &Int ( #asWord ( _ +Bytes B ) >>Int Z ) => X &Int ( #asWord ( B ) >>Int Z )
+      requires 0 <=Int X andBool X <Int 2 ^Int (8 *Int lengthBytes(B) -Int Z)
```
ehildenb, line 126:
```suggestion
      requires 0 <=Int X andBool X <Int 2 ^Int ((8 *Int lengthBytes(B)) -Int Z)
```

Clarity never hurts! Or maybe it does. But I prefer explicit parens.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -7,88 +7,240 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+    rule [bytes-concat-right-assoc]: (B1:Bytes +Bytes B2:Bytes) +Bytes B3:Bytes => B1 +Bytes (B2 +Bytes B3) [simplification]
 
-    rule #range(A +Bytes B, START, LENGTH) => #range(B, START -Int lengthBytes(A), LENGTH)
-      requires lengthBytes(A) <=Int START [simplification]
+    // #buf
 
-    rule #range(BA:Bytes, 0, LEN) => BA requires lengthBytes(BA) ==Int LEN [simplification]
+    rule [buf-ceil]:
+      #Ceil(#buf ( X , _:Int )) => 0 <=Int X
+      [simplification]
```
ehildenb, line 29: This rule is not allowed by new rules coming down, and likely is not actually meaning what you think it is: https://github.com/runtimeverification/k/issues/3284 and https://github.com/runtimeverification/k/issues/3286 for reference.

You may be able to get away with `#Ceil(#buf(W, _:Int)) => #Bottom requires W <Int 0 [simpilification]`, and `#Ceil(#buf(W, I)) => #Bottom requires 2 ^Int (8 *Int W) <=Int I`, for example, while still adhering to the rules linked above.

But recall that `#Ceil(...)` returns `#Top` or `#Bottom`, which is either the set `{true, false} == #Top`, or `{} = #Bottom`. So saying that `#Top => {true}` or `#Top => {false}` or `#Bottom => {true}` or `#Bottom => {false}` is a bit non-sensical in matching-logic, and is equivalent to adding an axiom which says "sort `Bool` only has one element".

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -7,88 +7,240 @@ module BYTES-SIMPLIFICATION [symbolic]
...
 
-    rule (#range(#range( BA:Bytes, S0, L0), S1, L1) ) =>#range(BA, S0 +Int S1, minInt(L1, L0 -Int S1)) +Bytes #buf(maxInt(0, L1 -Int (L0 -Int S1)), 0)
-      requires 0 <=Int S0 andBool 0 <=Int L0 andBool 0 <=Int S1 andBool 0 <=Int L1 [simplification]
+    rule [buf-inject-ml]:
+      { #buf(W:Int, X:Int) #Equals #buf(W:Int, Y:Int) } => { X #Equals Y }
+      requires 0 <=Int W
+       andBool #rangeUInt (8 *Int W, X)
+       andBool #rangeUInt (8 *Int W, Y)
+      [simplification]
```
ehildenb, line 43: Issue created:

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -7,88 +7,240 @@ module BYTES-SIMPLIFICATION [symbolic]
...
 
-    rule WS         +Bytes .Bytes => WS  [simplification]
-    rule .Bytes +Bytes WS         => WS  [simplification]
+    rule [buf-as-int-left]:  B:Bytes   ==K #buf(32, X:Int) => X ==Int #asInteger(B) requires lengthBytes(B) <=Int 32 [simplification, concrete(B)]
+    rule [buf-as-int-right]: #buf(32, X:Int) ==K B:Bytes   => X ==Int #asInteger(B) requires lengthBytes(B) <=Int 32 [simplification, concrete(B)]
```
ehildenb, line 47: Issue created: https://github.com/runtimeverification/k/issues/3317

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -7,88 +7,240 @@ module BYTES-SIMPLIFICATION [symbolic]
...
-    rule (BA1 +Bytes BA2) +Bytes BA3 => BA1 +Bytes (BA2 +Bytes BA3) [simplification]
+    rule [buf-asWord-invert]:
+      #buf (W:Int , #asWord(B:Bytes)) => #buf(W -Int lengthBytes(B), 0) +Bytes B
+      requires lengthBytes(B) <=Int W
+      [simplification]
 
-    // #range
+    rule [buf-zero-concat-base]:
+      #buf(W1:Int, 0) +Bytes #buf(W2:Int, 0) => #buf(W1 +Int W2, 0)
+      [simplification]
```
ehildenb, line 55:
```suggestion
    rule [buf-zero-concat-base]:
      #buf(W1:Int, 0) +Bytes #buf(W2:Int, 0) => #buf(W1 +Int W2, 0)
      requires 0 <=Int W1 andBool 0 <=Int W2
      [simplification]
```

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -7,88 +7,240 @@ module BYTES-SIMPLIFICATION [symbolic]
...
 
-    rule #range(_, _, K) => .Bytes requires K <=Int 0 [simplification]
+    rule [buf-zero-concat-conc]:
+      #buf(W1:Int, 0) +Bytes #buf(W2:Int, 0) +Bytes B => #buf(W1 +Int W2, 0) +Bytes B
+      [simplification]
```
ehildenb, line 59:
```suggestion
    rule [buf-zero-concat-conc]:
      #buf(W1:Int, 0) +Bytes (#buf(W2:Int, 0) +Bytes B) => #buf(W1 +Int W2, 0) +Bytes B
      requires 0 <=Int W1 andBool 0 <=Int W2
      [simplification]
```

ehildenb: With both of these where we have 0-bytes, we can probably do something like this:

```
rule #buf(W, 0) => padRightBytes(.Bytes, W, 0) requires 0 <=Int W [simplification]
rule BA1 +Bytes (BA2 +Bytes BA3) => (BA1 +Bytes BA2) +Bytes BA3 [concrete(BA1, BA2), symbolic(BA3), simplification]
```

This will allow us to collapsed zero-bytes together as a single concrete-byte-array.

The (potential?) downside is that now it's actually a 0-byte array like `b"\x00\x00...\x00"`, which the backend can treat as concrete (benefit), but which could be quite large and not amiable to manipulation with our body of `#buf(...)` lemmas (drawback).

May be worth a try?

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -7,88 +7,240 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+
+    rule [range-overfull]:
+      #range(B:Bytes, 0, W:Int) => B +Bytes #buf(W -Int lengthBytes(B), 0)
+      requires lengthBytes(B) <Int W
+      [simplification]
+
+    rule [range-params-equal-ml]:
+      { #range (B1:Bytes, S1:Int, W1:Int) #Equals #range (B2:Bytes, S2:Int, W2:Int) } => #Top
+      requires B1 ==K B2 andBool S1 ==Int S2 andBool W1 ==Int W2
+      [simplification]
```
ehildenb, line 81: I wonder why this rule is needed at all..... Are we sure it is? The backend should be able to do this unification problem. Is it used in any tests? (can try it by just removing the lemma nad seeing if anything breaks).

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -7,88 +7,240 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+
+    rule [range-included-in-cHead]:
+      #range(B1:Bytes +Bytes _:Bytes, S:Int, W:Int) => #range(B1, S, W)
+      requires S +Int W <=Int lengthBytes(B1)
+      [simplification]
```
ehildenb, line 86:
```suggestion
    rule [range-included-in-cHead]:
      #range(B1:Bytes +Bytes _:Bytes, S:Int, W:Int) => #range(B1, S, W)
      requires 0 <=Int S andBool 0 <=Int W andBool S +Int W <=Int lengthBytes(B1)
      [simplification]
```

Simplification priorities do not work like normal priorities, they are _suggestions_, not _requirements_ for the backend. So we must write `simplification` rules that are sound regardless of the higher-priority simplification rules that are present: https://kframework.org/docs/user_manual/#simplification-attribute

ehildenb: It looks like many of the `#range` rules may have this problem, make sure to check them all.

In cases where you have something ilke `lengthBytes(B) <=Int W` or such, then it's not needed because `lengthBytes` should always be positive.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -7,88 +7,240 @@ module BYTES-SIMPLIFICATION [symbolic]
...
 
-    rule #range(#writeRange(M, N, BUF:Bytes ), L, K) => #range(M, L, K)
-      requires 0 <Int K andBool N +Int lengthBytes(BUF) <=Int L [simplification]
+    // lengthBytes
 
-    rule #range(#writeRange(#writeRange(M, N1, BUF1:Bytes), N2, BUF2:Bytes), L, K ) => #range(#writeRange(M, N2, BUF2), L, K)
-        requires 0 <Int K andBool N1 +Int lengthBytes(BUF1) <=Int L [simplification]
+    rule [lengthBytes-geq-zero]    : 0 <=Int lengthBytes(_:Bytes)                       => true                                           [simplification, smt-lemma]
+    rule [lengthBytes-concat]      : lengthBytes(B1:Bytes +Bytes B2:Bytes)              => lengthBytes(B1) +Int lengthBytes(B2)           [simplification]
+    rule [lengthBytes-buf]         : lengthBytes(#buf(W:Int, _:Int))                    => W requires 0 <=Int W                           [simplification]
```
ehildenb, line 198: This could write a potentially undefined term to a defined one. We should have the extra condition `#rangeUInt(8 *Int W, I)`

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -7,88 +7,240 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+    rule [lengthBytes-range]       : lengthBytes(#range(_:Bytes, _:Int, W:Int))         => maxInt(0, W)                                   [simplification]
```
ehildenb, line 199: What if `_:Int` is negative, and `W` is positive?

include/kframework/lemmas/bytes-simplification.k
PetarMax: Yes yes, that rule should not be allowed, what I meant to write was `{ true #Equals 0 <=Int X }`, is that the correct way?

include/kframework/lemmas/bitwise-simplification.k
PetarMax: Yes, done.

include/kframework/lemmas/bytes-simplification.k
PetarMax: I've thought about adding this and chose not to because I believe it follows from the `#Ceil` of the LHS?

include/kframework/lemmas/bytes-simplification.k
PetarMax: My intuition is that bringing concrete byte-arrays together is better for reasoning. I did try something similar previously, but I think my `concrete` and `symbolic` attributes were wrong since I ran into an infinite simplification loop. I'm not running into it with these rules, so let's try them out. Since I'm not sure whether or not this will happen again in future, I've explicitly put the `symbolic` attribute on the right-associative rule(s).

include/kframework/lemmas/bytes-simplification.k
PetarMax: I believe the backend doesn't (and shouldn't?) work like that because `#range` is not injective. There were Lukso proofs that needed this, specifically to prune some impossible branches.

PetarMax: I would definitely like something like this do be done automatically by the backend.

include/kframework/lemmas/bytes-simplification.k
PetarMax: Actually, I'm not sure what's happening, `#buf` is declared as `total` in `buf.md`, so there should be no `#Ceil` issues, but there are some constraints in place on how it should be used.

include/kframework/lemmas/bytes-simplification.k
PetarMax: Hmm, but `#buf` is total?

include/kframework/lemmas/bytes-simplification.k
PetarMax: Then problems. Added `requires 0 <=Int S`.

include/kframework/lemmas/bytes-simplification.k
PetarMax: So, from what I see:

- `[range-empty]` is ok since it specifically targets the less-than-zero cases.
- `[range-full]`, `[range-overfull]`, and `[range-includes-cHead]` are ok because `S` is explicitly zero and `W` is bounded by `lengthBytes(B)`.
- `[range-params-equal-ml]` is ok because all parameters are equal on the LHS and RHS.
- `[range-included-in-cHead]` (the one you mentioned) is ok because if `S <Int 0` or `W <Int 0` on the LHS then `S <Int 0` or `W <Int 0` also on the RHS.
- `[range-outside-cHead]` is ok because `S` is bounded by `lengthBytes(B1)` and if `W <Int 0` on the LHS then also `W <Int 0` on the RHS.
- `[range-inside-cHead]` is ok because `0 <Int S` and if `W <Int 0` on the LHS then also `W <Int 0` on the RHS.
- `[range-join-base]`, `[range-join-conc]`, `[range-reform-base]`, `[range-reform-conc]`, and `[range-buf-zero-1]` are protected directly.
- `[range-writeRange-before]` and `[range-writeRange-in-between]`: added `0 <Int W2` as per the previous versions.
- `[range-writeRange-outside]` is ok because `S1` and `W2` are protected directly and `S2` is bounded by the (non-negative) expression in the `requires` clause.

`[range-buf-zero-2]`, however, is a bit of mess. It is now:
```
  rule [range-buf-zero-2]:
    #range(#buf(_:Int, 0), S:Int, W:Int) => #buf(W, 0)
    requires 0 <=Int S andBool 0 <=Int W
    [simplification]
```
but I'm not sure about the first parameter of `#buf`, given the (non-)totality of `#buf`.

include/kframework/lemmas/bytes-simplification.k
PetarMax: I've removed this simplification in the latest commit, as `#buf` is declared as total

include/kframework/lemmas/bytes-simplification.k
PetarMax: I've added a constraint on the first parameter, just to be safe.

include/kframework/lemmas/bytes-simplification.k
PetarMax: Actually, `#buf(W, D)` is defined for any `D`, it calculates a remainder with respect to `2 ^ (8 *Int W)`, so it should be fine.

ehildenb:
Subsumed by a sequence of other PRs endingi here: https://github.com/runtimeverification/evm-semantics/pull/1773

PR #1664, opened 2023-03-27
Added --foundry-project-root command line argument to read foundry.toml

### LINKED ISSUE #1630: Foundry integration should read from `foundry.toml` file to determine output directory
Currently, we have the user pass the foundry output directory to use as a mandatory argument to all the foundry commands.

Instead, we should:

- Allow the user to pass an _optional_ parameter `--foundry-project-root ...` which specifies the directory where the `foundry.toml` file lives.
- Read the output directory to use from the `foundry.toml` file.
- Read in solidity sources relative to that root, instead of assuming they are one directory up: https://github.com/runtimeverification/evm-semantics/pull/1530/files#r1134907926

Scott-Guest:
Added an optional command line argument `--foundry-project-root` to all foundry commands. This allows us to read the output directory from `foundry.toml` rather than specifying it manually. Closes #1630 

To parse the `foundry.toml` file, I added a dependency on [TOML Kit](https://pypi.org/project/tomlkit/). There is actually a TOML parser in the Python standard library, but only as of Python 3.11.


Makefile
```diff
@@ -506,13 +506,13 @@ tests/foundry/foundry.k.check: tests/foundry/out/kompiled/foundry.k
...
 tests/foundry/out/kompiled/foundry.k.prove: tests/foundry/out/kompiled/timestamp
-	$(KEVM) foundry-prove tests/foundry/out                              \
+	$(KEVM) foundry-prove --foundry-project-root tests/foundry           \
 	    -j$(FOUNDRY_PAR) --no-simplify-init --max-depth 1000             \
 	    $(KEVM_OPTS) $(KPROVE_OPTS)                                      \
 	    $(addprefix --exclude-test , $(shell cat tests/foundry/exclude))
 
 tests/foundry/out/kompiled/timestamp: $(foundry_out) $(KEVM_LIB)/$(foundry_kompiled) $(lemma_includes) poetry
-	$(KEVM) foundry-kompile $< $(KEVM_OPTS) --verbose
+	$(KEVM) foundry-kompile --foundry-project-root $(foundry_dir) $(KEVM_OPTS) --verbose
```
ehildenb, line 515: Can we use `$(foundry_dir)` in the line 509 as well?

kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -18,6 +20,21 @@
...
+    profile = os.getenv('FOUNDRY_PROFILE', default='default')
+    return all_profiles[profile].unwrap()
+
+
+def get_foundry_out_from_root(foundry_root: Path) -> Path:
+    """Get the Foundry out directory."""
+    foundry_toml = get_foundry_toml_profile(foundry_root)
+    return foundry_root / Path(foundry_toml['out'])
+
+
```
ehildenb, line 37: These should both be properties of the `Foundry` class instead. The first one can be called:

```
@property
def profile(self) -> Dict[str, Any]:
    ...

@property
def out(self) -> Path:
    out_rel_path = self.profile['out'] if 'out' in self.profile else '.'
    return foundry_out / out_rel_path
```

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -467,14 +479,14 @@ def pretty(self, *, details: bool = True) -> str:
         return '\n'.join(lines)
 
 
-def foundry_list(foundry_out: Path) -> List[CfgStat]:
-    kcfgs_dir = foundry_out / 'kcfgs'
+def foundry_list(foundry_root: Path) -> List[CfgStat]:
+    kcfgs_dir = get_foundry_out_from_root(foundry_root) / 'kcfgs'
     paths = kcfgs_dir.glob('*.json')
     return [CfgStat.from_file(path) for path in paths]
```
ehildenb, line 485:
```suggestion
def foundry_list(foundry_root: Path) -> List[CfgStat]:
    foundry = Foundry(foundry_root)
    kcfgs_dir = foundry.out / 'kcfgs'
    paths = kcfgs_dir.glob('*.json')
    return [CfgStat.from_file(path) for path in paths]
```

With `out` declared as a `property` of `Foundry` class, we can use it like this.

kevm-pyk/src/kevm_pyk/utils.py
Scott-Guest: Agreed, that's much cleaner.

However, there's a minor annoyance with approach. In a few places, we construct something like
```
br = BugReport(foundry_out / 'bug_reports)
```
and thus need to have `foundry_out` available before constructing the `Foundry` object.

A few possible solutions:
- Keep the `get_foundry_out_from_root` utility function just for this case
- Provide a setter `set_bug_report` for the bug_report rather than passing it to the `Foundry` constructor
- Create a separate class `FoundryProject` (or a better name) just to hold the logic for the project location, `foundry.toml`, and any other config details:
```
class FoundryProject:
    root: Path
    toml: Dict[str, Any]

    def __init__(foundry_root: Path) -> None:
        ...

    @property
    def profile(self) -> Dict[str, Any]:
        ...

    @property
    def out(self) -> Path:
        ...

class Foundry:
    _project: FoundryProject
    _bug_report: Optional[BugReport]

    def __init__(
        self,
        project: FoundryProject,
        bug_report: Optional[BugReport] = None,
    ) -> None:
        self._project = project
        self._bug_report = bug_report
```
then do
```
foundry_proj = FoundryProject(foundry_root)
br = BugReport(foundry_proj.out / 'bug_reports')
foundry = Foundry(foundry_proj, bug_report=br)
```

@ehildenb Any thoughts?

ehildenb: Let's just put the bug report in the current directory instead of in the `foundry_out` directory.

tothtamas28 (approved):

ehildenb (approved):

Issue #1371, opened 2022-09-13
solc-to-k: Alias generation for overloaded method


RaoulSchaffranek:
I used `solc-to-k` on a contract with an overloaded method `getData`. It supports the following two signatures:

~~~solidity
function getData(bytes32 dataKey)
function getData(bytes32[] memory dataKeys)
~~~

The generated k-file contained aliases for both signatures, but we cannot disambiguate them:

~~~k
    rule  ( selector ( "getData" ) => 1425412735 )
    
    rule  ( selector ( "getData" ) => 1312714396 )
~~~

I think our aliases should follow full argument form, e.g.:

~~~k
    rule  ( selector ( "getData(bytes32)" ) => 1425412735 )
    
    rule  ( selector ( "getData(bytes32[])" ) => 1312714396 )
~~~

The relevant entry point seems to be here:

https://github.com/runtimeverification/evm-semantics/blob/96392cc2f4ec1ae593a0fa71dfdd0bcd1cac6fcc/kevm-pyk/src/kevm_pyk/solc_to_k.py#L62-L64

PR #1643, opened 2023-03-18
Adjust default SMT settings for RPC based proofs

ehildenb:
Blocked on: https://github.com/runtimeverification/evm-semantics/pull/1646

This attempts to up the SMT timeout to find the smallest viable one for getting things merged faster.

This PR:

- Adds a new argument parser for `smt_args`, which contains:
  - `--smt-timeout`: control how long smt queries take.
  - `--smt-retry-limit`: control how many times to retry the SMT solver.
- Threads through these CLI arguments to all places that invoke the backend via `KCFGExplore`.
- Removes `find_free_port()` and relies on upstream port finding, which does not have the race condition present here.

PR #1667, opened 2023-03-27
Use full method signature in selector rules

nwatson22:
Fixes: #1371
This PR adds the full signature to the generated method selector rules. It also fixes a bug where the wrong productions were getting generated for overloaded methods, caused by looking up the methods (arg names and types) by their name only, meaning that the first method with the overloaded name would be selected every time. Also, in order to support overloaded methods, now that more than one production is generated for each method name, we also needed to change the label generation to use a different label for each overloaded method production.
To add an additional complication, we can have sets of overloaded methods which are indistinguishable when using base K sorts, so in order to make the method rules apply only to the correct overloaded call, I also had to add solidity type annotations to the method arguments as part of the productions.


tests/foundry/foundry.k.check.expected
```diff
@@ -3008,64 +3008,157 @@ module LOOPSTEST-BIN-RUNTIME
...
+    rule  ( selector ( "IS_TEST()" ) => 4202047188 )
+      
+    
+    rule  ( selector ( "failed()" ) => 3124842406 )
+      
+    
+    rule  ( selector ( "getNumber(uint32)" ) => 4233508440 )
+      
+    
+    rule  ( selector ( "getNumber(uint32)" ) => 144150998 )
```
ehildenb, line 3155: Hmmmm, are we extracting the signature correctly? One of these should have `uint256`.

nwatson22:
The reason the selector rules for `getNumber` (and other methods) is getting the wrong argument types is because `_get_method_abi` is looking up a method from a json list by its name only. I changed it to look up by name and argument types. This revealed another issue: there is a parsing ambiguity error for overridden methods because currently they get the same klabel for every version. This issue wasn't visible before because the same version of the method was actually being loaded every time, so while multiple rules with the same klabel were produced, it didn't matter because they were identical. I solved this parsing ambiguity by changing the method production klabels to include the (base sorts of) the arguments.

tests/foundry/foundry.k.check.expected
nwatson22: It turns out it was being caused by another spot in the code that assumed methods have unique names. It should be fixed now.

kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -42,20 +42,26 @@ def __init__(self, name: str, id: int, abi: Dict, contract_name: str, sort: KSor
...
             input_nonterminals = (KNonTerminal(_evm_base_sort(input_type)) for input_type in self.arg_types)
             items_before: List[KProductionItem] = [KTerminal(self.name), KTerminal('(')]
             items_args: List[KProductionItem] = list(intersperse(input_nonterminals, KTerminal(',')))
             items_after: List[KProductionItem] = [KTerminal(')')]
+            args_list = '_'.join([_evm_base_sort(input_type).name for input_type in self.arg_types])
             return KProduction(
                 self.sort,
                 items_before + items_args + items_after,
-                klabel=KLabel(f'method_{self.contract_name}_{self.name}'),
+                klabel=KLabel(f'method_{self.contract_name}_{self.name}_{args_list}'),
```
ehildenb, line 64: Let's factor out the `klabel` as a property of `Method`, and use that property here, and in `rule`, and in `application`.

kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -110,10 +116,13 @@ def application(self) -> KInner:
...
     def __init__(self, contract_name: str, contract_json: Dict, foundry: bool = False) -> None:
-        def _get_method_abi(_mname: str) -> Dict:
+        def _get_method_abi(_mname: str, _margs: List[str]) -> Dict:
             for _method in contract_json['abi']:
                 if _method['type'] == 'function' and _method['name'] == _mname:
-                    return _method
+                    if len(_margs) == len(_method['inputs']):
+                        inputs_match = [(_method['inputs'][i]['type'] == _margs[i]) for i in range(len(_margs))]
+                        if all(inputs_match):
+                            return _method
```
ehildenb, line 125:
```suggestion
                    _marg_types = (_marg['type'] for _marg in _method['inputs'])
                    if len(_margs) == len(_marg_types) and all(a1 == a2 for a1, a2 in zip(_margs, _marg_types, strict=True)):
                        return _method
```

kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -132,8 +141,10 @@ def _get_method_abi(_mname: str) -> Dict:
         _methods = []
         for msig in method_ids:
             mname = msig.split('(')[0]
+            margs_cs = msig.split('(')[1][:-1]
+            margs = [] if margs_cs == '' else margs_cs.split(',')
             mid = int(method_ids[msig], 16)
-            _m = Contract.Method(mname, mid, _get_method_abi(mname), contract_name, self.sort_method)
+            _m = Contract.Method(mname, mid, _get_method_abi(mname, margs), contract_name, self.sort_method)
             _methods.append(_m)
         self.methods = tuple(_methods)
```
ehildenb, line 149: `_get_method_abi` is only used for this purpose, so I would suggest the following change:

- Just pass in `msig` directly to `Contract.Method.__init__` as teh third argument.
- Move the processing of the signature into the `Contract.Method.__init__` constructor directly. Either by moving the method `_get_method_abi` into the `Contract.Method` class and calling it, or just by inlining its logic directly into the `Method` constructor.

kevm-pyk/src/kevm_pyk/solc_to_k.py
nwatson22: Thanks, this looks like a better way to write it. I incorporated this in the latest commit.

kevm-pyk/src/kevm_pyk/solc_to_k.py
nwatson22: I moved the signature processing and method lookup logic into the Method  constructor.

ehildenb:
@nwatson22 don't forget to put a complete description of changes in the PR main body, because that is what ends up in the commit log on squash merge. See for example https://github.com/runtimeverification/evm-semantics/pull/1643 or https://github.com/runtimeverification/evm-semantics/pull/1656.

kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -42,9 +57,19 @@ def __init__(self, name: str, id: int, abi: Dict, contract_name: str, sort: KSor
...
             self.payable = abi['stateMutability'] == 'payable'
 
+        @property
+        def signature(self) -> str:
+            arg_list = ','.join(self.arg_types)
+            return f'{self.name}({arg_list})'
+
+        @property
+        def klabel(self) -> KLabel:
+            args_list = '_'.join([_evm_base_sort(input_type).name for input_type in self.arg_types])
```
ehildenb, line 67:
```suggestion
            args_list = '_'.join(_evm_base_sort(input_type).name for input_type in self.arg_types)
```

tests/foundry/foundry.k.check.expected
```diff
@@ -3016,64 +3015,157 @@ module LOOPSTEST-BIN-RUNTIME
...
       
     
-    rule  ( selector ( "testMaxBroken" ) => 1372438930 )
+    rule  ( MethodDisambiguateTest . getNumber ( V0_x ) => #abiCallData ( "getNumber" , #uint256 ( V0_x ) , .TypedArgs ) )
+       ensures #rangeUInt ( 256 , V0_x )
       
     
-    rule  ( selector ( "testNthPrime" ) => 2702762242 )
+    rule  ( MethodDisambiguateTest . getNumber ( V0_x ) => #abiCallData ( "getNumber" , #uint32 ( V0_x ) , .TypedArgs ) )
+       ensures #rangeUInt ( 32 , V0_x )
```
ehildenb, line 3141: Hmmmm, how will K be able to tell which `getNumber` should apply here? It seems not really possible?

tests/foundry/foundry.k.check.expected
```diff
@@ -3016,64 +3015,157 @@ module LOOPSTEST-BIN-RUNTIME
...
+    syntax MethodDisambiguateTestMethod ::= "IS_SCRIPT" "(" ")" [symbol(), klabel(method_MethodDisambiguateTest_IS_SCRIPT_)]
+    
+    syntax MethodDisambiguateTestMethod ::= "IS_TEST" "(" ")" [symbol(), klabel(method_MethodDisambiguateTest_IS_TEST_)]
     
-    rule  ( selector ( "testIsPrimeOpt" ) => 404719852 )
+    syntax MethodDisambiguateTestMethod ::= "failed" "(" ")" [symbol(), klabel(method_MethodDisambiguateTest_failed_)]
+    
+    syntax MethodDisambiguateTestMethod ::= "getNumber" "(" Int ")" [symbol(), klabel(method_MethodDisambiguateTest_getNumber_Int)]
+    
+    syntax MethodDisambiguateTestMethod ::= "getNumber" "(" Int ")" [symbol(), klabel(method_MethodDisambiguateTest_getNumber_Int)]
```
ehildenb, line 3121: It looks like they're getting the same klabels.....

And they have the same production. So we need to use the solidity type in the KLabel, not the K sort.

Also, for methods that are overloaded like this, I think we will just have to include the types directly in the production. We could instead generate, for example:

```
syntax MethodDisambiguateTestMethod ::= "getNumber" "(" Int ":uint32" ")" [symbol(), klabel(method_MethodDisambiguateTest_getNumber_uint32]
```

ehildenb: So that it looks kinda like solidity, but we have the `:uint32` to disambiguate the types present in teh production.

tests/foundry/foundry.k.check.expected
ehildenb: Note that getting this wrong is pretty dangerous, if we leave it as is, K may take the wrong rule, and put the wrong restrictions on the input types.

kevm-pyk/src/kevm_pyk/solc_to_k.py
ehildenb: So instead of using `_evm_base_sort(input_type).name`, I think just use `input_type` here.

tests/foundry/foundry.k.check.expected
nwatson22: I've added it so that the solidity types are included in the productions, as suggested.

tests/foundry/foundry.k.check.expected
nwatson22: The rules should be distinguishable now.

kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -50,26 +65,43 @@ def __init__(self, name: str, id: int, abi: Dict, contract_name: str, sort: KSor
...
             items_before: List[KProductionItem] = [KTerminal(self.name), KTerminal('(')]
-            items_args: List[KProductionItem] = list(intersperse(input_nonterminals, KTerminal(',')))
+
+            items_args: List[KProductionItem] = []
+            count = 0
+            for input_type in self.arg_types:
+                if count > 0:
+                    items_args += [KTerminal(',')]
+                items_args += [KNonTerminal(_evm_base_sort(input_type)), KTerminal(':'), KTerminal(input_type)]
+                count += 1
```
ehildenb, line 92:
```suggestion
            for i, input_type in enumerate(self.arg_types):
                if i > 0:
                    items_args += [KTerminal(',')]
                items_args += [KNonTerminal(_evm_base_sort(input_type)), KTerminal(':'), KTerminal(input_type)]
```

ehildenb (approved):
Looks good, just one review comment to address, and update to master.

kevm-pyk/src/kevm_pyk/solc_to_k.py
nwatson22: I applied this suggestion.

Issue #1668, opened 2023-03-28
Revise byte buffer abstraction functions


jberthold:
[The code for byte buffers](https://github.com/runtimeverification/evm-semantics/blob/master/include/kframework/buf.md) appears to have some problems:

1. Function `#bufStrict` has a rule that explicitly takes it to a literal `#Bottom` in cases where the arguments are not in the required range. The effect of that rule on the semantics as a whole and its proofs is unclear, and it does not seem to be a sensible rule to have.
2. Function `#buf` is marked `total` but is not defined for `SIZE < 0` (because `padLeftBytes` and `^Int` might not be defined for negative arguments)

jberthold:
#1669

PR #1669, opened 2023-03-28
1668 remove #bufStrict(..) => #Bottom, make #buf total

jberthold:
Part of #1668

* removes the rule that rewrites to `\bottom`
* provides definition for `#buf` on `SIZE <= 0` arguments (making it total)


include/kframework/buf.md
```diff
@@ -47,14 +46,8 @@ module BUF
...
     rule #buf(SIZE, DATA) => #padToWidth(SIZE, #asByteStack(DATA %Int (2 ^Int (SIZE *Int 8))))
-      [concrete]
-
-endmodule
-
-module BUF-KORE [kore, symbolic]
-    imports BUF-SYNTAX
-
-    rule #bufStrict(_, _) => #Bottom              [owise]
+      requires SIZE >Int 0
```
ehildenb, line 49: We definitely want to keep the `concrete` attribute here, because all our lemmas about byte arrays are written assuming the `#buf` will be present.

On the new `owise` rule, we don't need concrete attribute, because it goes to strictly simpler value.

ehildenb, line 49:
```suggestion
      requires SIZE >Int 0
      [concrete]
```

ehildenb:
```suggestion
      requires 0 <Int SIZE
```

Also, stylistically, KEVM orients all disequalities in number-line order. So `0 <Int SIZE` instead of `SIZE >Int 0`. This also avoids one simplification rewrite per application of the rules with disequalities in the side conditions, because all our lemmas are written in terms of `_<Int_` and `_<=Int_`.

jberthold: > We definitely want to keep the `concrete` attribute here, because all our lemmas about byte arrays are written assuming the `#buf` will be present.
> 
> On the new `owise` rule, we don't need concrete attribute, because it goes to strictly simpler value.

I tried that first but got a compiler error (which we are once more getting with the current state)
```
Error:  Compiler: Found concrete attribute without simplification attribute on
function with one or more non-concrete rules.
	Source(/home/user/.build/usr/lib/kevm/include/kframework/buf.md)
	Location(48,10,49,27)
	   .               v~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
	48 |	    rule #buf(SIZE, DATA) => #padToWidth(SIZE, #asByteStack(DATA %Int (2^Int (SIZE *Int 8))))
	49 |	      requires 0 <Int SIZE
	   .	      ~~~~~~~~~~~~~~~~~~~^
```
Maybe the second rule has to be `concrete`, too?

jberthold: Compilation and tests go through with `[concrete]` on both `#buf` rules. If this is OK to keep the PR is ready to go.

ehildenb (approved):

PR #1673, opened 2023-03-29
Update address of Foundry test contract

lucasmt:
* Updates hard-coded address of the Foundry test contract to new value (see https://github.com/foundry-rs/foundry/blob/7398b65e831f2339d1d0a0bb05ade799e4f9d01e/testdata/core/ContractEnvironment.t.sol#L16):
    * `1032069922050249630382865877677304880282300743300` to `728815563385977040452943777879061427756277306518`
    * `0xb4c79daB8f259C7Aee6E5b2Aa729821864227e84` to `0x7FA9385bE102ac3EAc297483Dd6233D62b3e1496`
* Adds a new test checking that the built-in Foundry addresses are as expected. Before the above update, this test passed in Foundry but failed in KEVM. After the update, it passes also in KEVM.


ehildenb:
@lucasmt not sure why this is triggering a formatting error, but you can do `make format` in the `kevm-pyk` directory to get it to reformat it for you.

anvacaru:
@ehildenb @lucasmt I think that we should also set the balance value of the test contract to what foundry is using.
Below is a test that asserts the balance of the `msg.sender` and the test contract. 
Since we use the symbolic `CALLER_ID` for the `msg.sender` then it makes sense to have symbolic balance as well.
For the foundry test contract I'd say that we should update the balance to the concrete value.

```solidity
    function test_initBalance() public {
       assertEq(address(msg.sender).balance, 79228162514264337593543950335);
       assertEq(address(this).balance, 79228162514264337593543950335);
    }
```

ehildenb (approved):

PR #1675, opened 2023-03-29
Enforce right-association of bytes concatenation

lucasmt:
* Add lemma rewriting `+Bytes` expressions to be right-associated
* Add claim testing the lemma


lucasmt:
From @PetarMax:
> Given the design of the simplification engine, will the following:
> `rule (A +Bytes B) +Bytes C => A +Bytes B +Bytes C [simplification]`
> always simplify any concatenation to the fully right-associative form `A1 +Bytes A2 +Bytes … +Bytes An`? For example, I remember there is some caching going on, etc.

From @ana-pantilie:
> I can't say for sure, there are bugs/limitations regarding one of the caching mechanisms which couldn't be fixed.

ehildenb (approved):

Issue #1634, opened 2023-03-16
Concrete rules ambiguity


anvacaru:
In both the [concrete-rules.txt](https://github.com/runtimeverification/evm-semantics/blob/d597e7a7730f46eaea357188c3dbe7bb3177fc36/tests/specs/concrete-rules.txt#L28) and [kevm.py](https://github.com/runtimeverification/evm-semantics/blob/d597e7a7730f46eaea357188c3dbe7bb3177fc36/kevm-pyk/src/kevm_pyk/kevm.py#L183) there is a label named `EVM-TYPES.Bytes.range` which I can't find in the source files which leads to the following questions:

1) Kompile should have warnings emitted for each lemma in the concrete rules which is not found. Are any warnings emitted?
2) Still, there are rules in our codebase that use the `[concrete]`. Is there a reason for this?
3) Could we remove the hardcoded list from `kevm.py` and read the list from `concrete-rules.txt`?

anvacaru:
For 3. is there a reason to have the `concrete-rules.txt` under `tests/specs`? Could we use something in python like `return open(filename).read().splitlines()` to avoid duplication?

ehildenb:
Are you able to confirm with a minimal example that the concrete-rule is not actually in the definition, but kmopile does not give a warning? If so, let's open an issue for that one against K repo, with the minimal example.

And actually, I want to move the other direction, remove the `concrete-rules.txt` instead, because I want to move all the logic into Python. We have it set so that by default you get the pyk kompile: https://github.com/runtimeverification/evm-semantics/pull/1616. So we're really just keeping around the Bash compile "just in case" at this point.

ehildenb:
Todo:

- Minimal example in the frontend for `kompile --backend haskell --concrete-rules MODULE-THAT-EXISTS.rule-that-doesnt-exist`, and add as a test and error to th efrontend.
- Once that is merged, remove the offending rules from `kevm.py` in this repo.
- Also remove the now unused `concrete-rules.txt` in this repo.

Scott-Guest:
The following labels do not have corresponding rules (as confirmed by both grep and the error check I added):

- EVM-TYPES.#asByteStack
- EVM-TYPES.#asByteStackAux.recursive
- EVM-TYPES.#asWord.recursive
- EVM-TYPES.#padRightToWidth
- EVM-TYPES.#padToWidth
- EVM-TYPES.#range
- EVM-TYPES.Bytes.range
[...]

PR #1677, opened 2023-03-30
Remove non-existent rules from KEVM.concrete_rules

Scott-Guest:
As noted in #1634, the following labels listed in `KEVM.concrete_rules` don't correspond to actual rules, and the check added in runtimeverification/k#3289 will report an error on them:

- EVM-TYPES.#asByteStack
- EVM-TYPES.#asByteStackAux.recursive
- EVM-TYPES.#asWord.recursive
- EVM-TYPES.#padRightToWidth
- EVM-TYPES.#padToWidth
- EVM-TYPES.#range
- EVM-TYPES.Bytes.range
- EVM-TYPES.mapWriteBytes.recursive

This PR removes these labels as well as the unused `concrete-rules.txt`


anvacaru:
`concrete-rules.txt` seems to be used [here](https://github.com/runtimeverification/evm-semantics/blob/38f3bf6ee75bbcf67363313ee01af3957a98b3d1/optimizer/optimize.sh#L69) and [here](https://github.com/runtimeverification/evm-semantics/blob/38f3bf6ee75bbcf67363313ee01af3957a98b3d1/package/test-package.sh#L34). 
AFAIK, `kompile` should use the concrete rules defined in `kevm.py`, so I would also suggest removing these references. Other than that, looks good to me 👍 !

anvacaru (approved):

PR #1682, opened 2023-04-04
Add definedness constraint to `KCFG` init state in `exec_prove`

dkcumming:
Mirroring changes made to `foundry_prove` [here](https://github.com/runtimeverification/evm-semantics/pull/1635)


kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -280,6 +280,9 @@ def exec_prove(
             smt_timeout=smt_timeout,
             smt_retry_limit=smt_retry_limit,
         ) as kcfg_explore:
+            for _claim, cfg in proof_problems.items():
+                init_node = cfg.get_unique_init()
+                cfg.replace_node(init_node.id, kcfg_explore.cterm_assume_defined(init_node.cterm))
             proof_problems = {claim: kcfg_explore.simplify(claim, cfg) for claim, cfg in proof_problems.items()}
```
ehildenb, line 286: Did you verify that this has the intended effect? I guess I'm not sure about Python's modify in place semantics, so I would do something more like:

```suggestion
        ) as kcfg_explore:
            _proof_problems = {}
            for _claim, cfg in proof_problems.items():
                init_node = cfg.get_unique_init()
                cfg.replace_node(init_node.id, kcfg_explore.cterm_assume_defined(init_node.cterm))
                _proof_problems[_claim] = cfg
            proof_problems = _proof_problems
```

@tothtamas28 will the original also modify the `proof_problems` in place?

Also, @dkcumming , we want to do this assuming of definedness _even if_ the user does pass `--no-simplify-init`, which currently it will not enter this block if the user has passed that flag. So we want to move the `if simplify_init: ...`` guard into here.

dkcumming: @ehildenb 
From what I could tell with the proofs I did it was working in place, but clearer/safer with the ammedments. 
@tothtamas28 
True, I will refactor.

dkcumming:
I implemented the changes discussed above. Aside I skipped the logging message of initializing KCFGExplore, but I can put it in there again if that is necessary. I am having a bit of trouble getting the used to the versioning. There may be extra commits until I figure out how to consistently get it in the right state,

ehildenb (approved):

ehildenb:
You can always safely resolve the merge conflicts on `package/version` and `kevm-pyk/pyproject.toml` and `kevm-pyk/poetry.lock` with `git checkout --theirs -- package/version kevm-pyk/pyproject.toml kevm-pyk/poetry.lock` after you do a `git merge origin/master` and it causes the conflicts.

PR #1685, opened 2023-04-04
update nixpkgs, update solc

h0nzZik:
After https://github.com/runtimeverification/pyk/pull/309, PyK works with a very recent version of nixpkgs. Until now, kevm and K worked with old version of nixpkgs. One reason for using the old version was that solc is broken in recent nixpkgs. Although there is a PR (https://github.com/NixOS/nixpkgs/pull/219240) which fixes solc in nixpkgs, it is not merged yet. Therefore what I propose is that we use the new version of nixpkgs together with the new version of solc.

Previously I created a PR to [_update-deps/runtimeverification/pyk](https://github.com/runtimeverification/evm-semantics/tree/_update-deps/runtimeverification/pyk), but the CI tests didn't run on that. So I created this new PR, because I do not want to write to [_update-deps/runtimeverification/pyk](https://github.com/runtimeverification/evm-semantics/tree/_update-deps/runtimeverification/pyk) directly.


ehildenb:
Subsumed: https://github.com/runtimeverification/evm-semantics/pull/1674

Issue #1683, opened 2023-04-04
Raise a warning instead of an error in _get_method_abi(mname)


anvacaru:
The function 
```python
        def _get_method_abi(_mname: str) -> Dict:
            for _method in contract_json['abi']:
                if _method['type'] == 'function' and _method['name'] == _mname:
                    return _method
            raise ValueError(f'Method not found in abi: {_mname}')
```
raises an error if the `_mname` function is not found in the `abi` of a contract.

We've discovered an edge case in which the compiled output does not have an abi for the function `config()`.

```json
  "abi": [],
  "bytecode": {
    "object": ["..."],
    "sourceMap": "...",
    "linkReferences": {}
  },
  "deployedBytecode": {
    "object": "...",
    "sourceMap": "...",
    "linkReferences": {}
  },
  "methodIdentifiers": {
    "config()": "79502c55"
  },
```

While investigating what's the cause of this, we should also modify the function to return a warning instead and skip the function.

anvacaru:
Hmm, it looks like #1667 overwritten #1687.

nwatson22:
> Hmm, it looks like #1667 overwritten #1687. This should include a test as well.

Apologies, I merged it poorly. I've added another PR which should hopefully get the old behavior back.

PR #1686, opened 2023-04-04
Replace error with warning in _get_method_abi function

anvacaru:
Fixes: #1683


kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -141,8 +142,10 @@ def _get_method_abi(_mname: str) -> Dict:
         for msig in method_ids:
             mname = msig.split('(')[0]
             mid = int(method_ids[msig], 16)
-            _m = Contract.Method(mname, mid, _get_method_abi(mname), contract_name, self.sort_method)
-            _methods.append(_m)
+            _sort = _get_method_abi(mname)
```
tothtamas28, line 145: How about `sort` or `msort`?

anvacaru: Instead of `_sort`?

tothtamas28: Yes.

anvacaru: Updated to `msort` and tweaked the function to return `{}` instead of `None` since the function is supposed to return a `Dict`.

PR #1688, opened 2023-04-04
Set python recursion limit higher

ehildenb:
This sets the recursion limit higher by default on KEVM definitions.

I've confirmed with Python3.11 that this fixes the various segfaults/crashes we had seen before. We still need to up the default Python version somehow, but that's another issue.

Related: https://github.com/runtimeverification/pyk/issues/286


kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -42,6 +42,8 @@
 _LOGGER: Final = logging.getLogger(__name__)
 _LOG_FORMAT: Final = '%(levelname)s %(asctime)s %(name)s - %(message)s'
 
+sys.setrecursionlimit(10**8)
```
tothtamas28, line 45: This should be set in `main`, i.e. replace the value there with the new, higher value.

tothtamas28:
> I've confirmed with Python3.11 that this fixes the various segfaults/crashes

Is this related to segfaults? My understanding is that changing Python version already fixes that.

ehildenb:
I believe that both are needed. Let me double-check.

ehildenb:
On Python 3.11 without upping the recursion limit, loading a Kore term that a client gets back from RPC server: 
image


So we need both.

ehildenb:
Ah wait though, I didn't see that `main` already was upping the limit. Let me try my example with the limit `main` is already setting.


The limit in `__main__.py` is enough for my example. So I'll close this.

tothtamas28:
> So we need both.

Not for fixing the segfault though. Probably you'd get the same `RecursionError` on any Python version, but the segfault AFAIK is only reproducible on 3.8.

Issue #1689, opened 2023-04-04
Stop creating new accounts for certain cheat codes


anvacaru:
Currently, we use `#loadAccount` in each cheat code that uses an address as an argument. `#loadAccount ADDR` will create a new account if there is none defined with `<acctID> ADDR </acctID>`.  For cheat codes like `prank` or `load`, `getNonce`, that do not actively write to the storage this only increases the size of the account state. We could avoid this by either returning `0` or a fresh symbolic variable for the case in which the account is not in the `<accounts>` cell.

ehildenb:
Are we trying to avoid this because then we have the definedness condition generated with whether that account is in `ACCOUNTS_INIT` or not?

If so, we may be able to adjust the `#loadAccount` rules to automatically make sure that we don't get branching on that definedness condition with `ACCOUNTS_INIT`:

```
    syntax KItem ::= "#loadAccount" Int [klabel(foundry_loadAccount)]
 // -----------------------------------------------------------------
    rule <k> #loadAccount ACCT => #accessAccounts ACCT ... </k>
         <activeAccounts> ACCTS:Set </activeAccounts>
         <accounts> <account> <acctID> ACCT </acctID> ... </account> ... </accounts>
      requires ACCT in ACCTS

    rule <k> #loadAccount ACCT => #newAccount ACCT ~> #accessAccounts ACCT ... </k>
         [owise]
```

This should make sure taht either: (i) the account is in both `<activeAccounts>` and `<accounts>`, and is not created, or (ii) it's in neither, and it's created. May help out with weird definedness conditions we're getting.

ehildenb:
In principle your proposal sounds fine though, as long as it doesn't change the semantics of anything. Though I'm not sure that it won't change the semantics of things ....

For example, if we don't do `#loadAccount`, then the `#accountNonexistent` check may change behavior, and maybe that means that the code after a `prank` behaves differently. What does Foundry do for pranking to an address that doesn't exist? Or loading storage/nonce from an address that doesn't exist?

anvacaru:
I will write some tests and investigate. Assigning this one to myself.

anvacaru:
Loading a storage/nonce of a contract that doesn't exist returns `0`. We have the following tests passing both in foundry and kevm.

```solidity
    function testLoadNonExistent() public {
        bytes32 testNumber = vm.load(address(100), bytes32(uint256(23)));
        assert(uint256(testNumber) == 0);
    }
```
```solidity
[...]

PR #1692, opened 2023-04-04
Include SET-SYMBOLIC module

anvacaru:
Branchings with `VV0_addr_123:Int in .Set` take place. Imported the `SET-SYMBOLIC` module, which has simplifications for the Set sort.


tests/foundry/test/AccountParamsTest.t.sol
```diff
@@ -38,6 +38,15 @@ contract AccountParamsTest is Test {
        assert(nonce == 1);
     }
 
+    function test_getNonce_unknownSymbolic(address addr) public {
+      vm.assume(addr != address(vm));
+      vm.assume(addr != address(this));
+      vm.assume(addr != address(0x3fAB184622Dc19b6109349B94811493BF2a45362));
+      vm.assume(addr != address(0x4e59b44847b379578588920cA78FbF26c0B4956C));
+      uint64 nonce = vm.getNonce(addr);
+      assert(nonce == 0);
```
PetarMax, line 47: @anvacaru I think we had a discussion about this and whether it should be symbolic or not - what is the current idea?

ehildenb:
My guess after 2 failures in a row we need to actually inspect the proofs that are taking longer. Maybe we've introduced a looping simplification.

anvacaru:
Locally, all the proofs passed in 75 minutes. I'll look at the simplifications.
image

tests/foundry/test/AccountParamsTest.t.sol
anvacaru: @PetarMax in this PR I just want to add a test that shows the need of the `SET-SYMBOLIC` module. I've created [this issue](https://github.com/runtimeverification/evm-semantics/issues/1689) to track wether it is a good ideea or not to return a symbolic value for unknown accounts.

PetarMax: Understood, but won't the test need changing if a symbolic value is returned?

PetarMax (approved):

PR #1698, opened 2023-04-05
Expose definitions from foundry.k

### LINKED ISSUE #1545: Expose generated foundry-helper modules
When using the KEVM+Foundry integration, we manually add additional lemmas to drive proofs forward. We conventionally keep these lemmas in a file named `lemmas.k`.

Inside the lemmas file, we cannot refer to helper definitions from the generated `foundry.k` file. In particular, we cannot use `#binRuntime( ... )` or `Contract . function ( PARAM )` syntax. Consequently, our lemmas contain hardcoded bytecode, jump-table, and ABI encodings. This harms readability and makes our lemmas fragile. For example, we need to replace the bytecode on every change manually.

It would be a massive step forward in terms of user experience if we exposed the helper definitions to the lemmas file.

### Implementation

A possible path could work like this:

1. The author of the lemmas file adds "requires `foundry.k` to his module.
2. We ensure that the `foundry.k` is linked correctly during kompilation.

### Relevant source files

Here is the block that generates `foundry.k`
https://github.com/runtimeverification/evm-semantics/blob/62b5215c281a251e5f1b8e2d6bcf2e0a03888f17/kevm-pyk/src/kevm_pyk/__main__.py#L207

And then the block (immediately after) that calls kompile:
https://github.com/runtimeverification/evm-semantics/blob/62b5215c281a251e5f1b8e2d6bcf2e0a03888f17/kevm-pyk/src/kevm_pyk/__main__.py#L215

Above that, in the call to `_foundry_to_bin_runtime`, is the block that actually generates the modules in `foundry.k` (see file `solc_to_k.py` for the actual implementation, but here is the wrapper in `__main__`):
https://github.com/runtimeverification/evm-semantics/blob/62b5215c281a251e5f1b8e2d6bcf2e0a03888f17/kevm-pyk/src/kevm_pyk/__main__.py#L251

Slack Discussion: https://runtimeverification.slack.com/archives/C015XKM0X19/p1675247830822969?thread_ts=1675184939.248819&cid=C015XKM0X19

nwatson22:
Closes #1545 
This PR aims to allow kevm foundry users to more easily integrate their modules of helper lemmas by allowing them to use generated `#binRuntime( ... )`, `Contract . function ( PARAM )`, and `#computeValidJumpDests( ... )` syntax in those modules.
Currently in this branch, the user can gain access to the generated syntax for a specific contract `contractName` by importing `CONTRACTNAME-BIN-RUNTIME` into their imported module. e.g.:

`helper-lemmas.k`:

```k
module HELPER-LEMMAS
  imports ASSERTTEST-BIN-RUNTIME

  syntax Bytes ::= testFunction() [function]
  syntax Set ::= testFunction2(Bytes) [function]

  rule testFunction() => AssertTest.test_assert_true() [simplification]
  rule testFunction() => #binRuntime(AssertTest) [simplification]
  rule testFunction2(PGM) => #computeValidJumpDests(PGM) [simplification]

endmodule
```

`test/Simple.t.sol`:

```sol
pragma solidity =0.8.13;

import "forge-std/Test.sol";

contract AssertTest is Test {

    function setUp() public {}

    function test_assert_true() public pure {
        assert(true);
    }
}
```
```
kevm foundry-kompile --require helper-lemmas.k  --module-import HELPER-LEMMAS
```
A different approach could also be to add a different module to the generated `foundry.k` file which imports all contract modules and then have the `FOUNDRY-MAIN` module import this and all `imports` modules. This might be better because the user then only needs to know one module name that they always import. I'm not sure which way would be better.

There didn't seem to be any issue with `requires` importing. In fact, this already seems to work in master except for the fact that there is no way to import the generated modules without creating circular imports.

Finally, this probably needs some  way of letting the user know the module(s) they can import into their custom modules and also what syntax they use, since right now the only way for them to deduce that is to inspect the auto-generated K definitions.


ehildenb:
Let's add a test by adding file `tests/foundry/lemmas.k` which is just for one of the contracts in the directory and makes some silly rules about `#binRuntime(ContractName)`. Then make sure that `kevm foundry-kompile ... --require "lemmas.k" --module-import "MY-SILLY-MODULE"` works with those rules in place.

nwatson22:
> Let's add a test by adding file `tests/foundry/lemmas.k` which is just for one of the contracts in the directory and makes some silly rules about `#binRuntime(ContractName)`. Then make sure that `kevm foundry-kompile ... --require "lemmas.k" --module-import "MY-SILLY-MODULE"` works with those rules in place.

In order to not interfere with the normal kompilation, for this test I added an option to `foundry-kompile` which kompiles into an alternate subdirectory and just do the compilation twice. I am a bit worried that it clutters things up too much but I didn't immediately see a better way. Is this alright or would you suggest something else?

ehildenb:
So the import structure doesn't make much sense to me. We have:

- File `lemmas.k` (no requires):
  - Module `TEST-HELPERS-LEMMAS`: imports `ASSSERT-TEST-BIN-RUNTIME`
- File `foundry.k`:
  - Module `ASSERT-TEST-BIN-RUNTIME`: imports `FOUNDRY`
  - Module `FOUNDRY-MAIN`: imports `ASSERT-TEST-BIN-RUNTIME`

So I think we're not actually getting `TEST-HELPERS-LEMMAS` into the final structure at all. The test you have (that I suggested above) I think would pass anyway, because the `binRuntime` macro is defined elsewhere (in the main semantics). So I think we need a better test anyway, likely the best test would be a loop invariant for `sum-to-n` for example (`tests/foundry/test/Loops.t.sol`). So we could do:

- File: `contracts.k` (generated): Defines all the `CONTRACT_NAME-CONTRACT` definitions (what we currently call `foundry.k`).
- File: `lemmas.k` (written by user, `requires "contracts.k"` at the top must be present): Defines `SUM-TO-N-INVARIANT`, which imports `LOOPSTEST-CONTRACT`, and adds the loop invariant for `SUM-TO-N` program, but gets to use `binRuntime(...)` macro in both the `<program>` and `<jumpDests>` cells.
- File `foundry.k` (generated): `requires "contracts.k"` and `requires "lemmas.k"` at the top: Defines modules `*-VERIFICATION` for each module `*-BIN-RUNTIME`. `*-VERIFICATION` imports `*-CONTRACT`, but in the case of `LOOPSTEST-CONTRACT`, it also imports `SUM-TO-N-INVARIANT`. Then also defines module `FOUNDRY-MAIN` which imports all the `*-VERIFICATION` modules.

File `contracts.k` (generated):

```
requires "foundry.md"

module *-CONTRACT
    imports FOUNDRY

    ...
endmodule

module LOOPSTEST-CONTRACT
    imports FOUNDRY

    ...
endmodule
```

File `lemmas.k` (written):

```
requires "contracts.k"  // <-- User is required to supply this, put in documentation.

module SUM-TO-N-INVARIANT
    imports LOOPSTEST-CONTRACT

   rule <k> .... </k> [priority(40)] // <--- loop invariant for `sum_N` method.
endmodule
```

File `foundry.k` (generated):

```
requires "contracts.k"
requires "lemmas.k"

module *-VERIFICATION
    imports *-CONTRACT
endmodule

module LOOPSTEST-VERIFICATION
    imports LOOPSTEST-CONTRACT
    imports SUM-TO-N-INVARIANT
endmodule

module FOUNDRY-MAIN
    imports *-VERIFICATION
endmodule
```


It may be that you have to put `imports SUM-TO-N-INVARIANT` in `FOUNDRY-MAIN` instead of in `LOOPSTEST-VERIFICATION` for the time being, because you won't be able to have the user specify which module it should go into. This may be something that having a declarative build system like `kbuild` could help with later (ping @tothtamas28 ).

ehildenb:
We can likely change the way we do imports so that we have `--module-import ContractName:NEW-MODULE-NAME` or `--module-import NEW-MODULE-NAME`. In the first case, we know to import the `NEW-MODULE-NAME` into `ContractName-VERIFICATION`, and in the second case, we import it into `FOUNDRY-MAIN` directly.

nwatson22:
Blocked on [pyk#398](https://github.com/runtimeverification/pyk/issues/398)

ehildenb:
Try out a higher priority than 40 (like 35) so that you can be higher priority than the rules in `include/kframework/optimizations.md`.

nwatson22:
@ehildenb Is there any way I can make it so the user must `requires "contracts.k"`? Currently, due to the copied requires files living in the `kompiled/requires` subdirectory, they need to do `requires "../contracts.k"`

ehildenb:
Can we generate `contracts.k` in the `requires/` subdirectory instead of in the superdirectory?

Or you can use the `-I` option to tell K where to look for files that are `required`, and do `-I SUPERDIRECTORY`

tests/foundry/lemmas.k
```diff
@@ -0,0 +1,113 @@
...
+    rule 0 <=Int (N *Int (N +Int 1)) /Int 2 => true
+    requires 0 <=Int N
+    [simplification]
+
+    rule N *Int 178 <=Int V => N <=Int V /Int 178
+    [simplification]
+
+//     rule N *Int 178 <=Int 9223372036854772639 => true
+//     requires N <=Int 1641760222560149529742
+//     [simplification]
```
ehildenb, line 55: @PetarMax can you review these rules?

ehildenb:
Add a chapter to the GitBook with the `sumN` example, which shows how to make this proof (in isolation) work: https://runtime-verification-tooling.gitbook.io/kevm-integration-for-foundry/kevm-foundry-integration-example/k-control-flow-graph-kcfg

tests/foundry/lemmas.k
```diff
@@ -0,0 +1,113 @@
...
+    rule chop(S +Int N) => S +Int N
+    requires #rangeUInt(256, S +Int ((N *Int (N +Int 1)) /Int 2))
+    andBool #rangeUInt(256, S)
+    andBool #rangeUInt(256, N)
+    [simplification]
+
+    rule chop(N +Int -1) => N +Int -1
+    requires 1 <=Int N andBool N <Int pow256
+    andBool #rangeUInt(256, N)
+    [simplification]
```
PetarMax, line 40: This, when turned into a claim, can be proven by KEVM master. Why is it needed as a lemma?
```
    claim [chop-simpl-1]: runLemma ( chop(N +Int -1) ) => doneLemma ( N +Int -1 )
      requires 1 <=Int N andBool N <Int pow256
```

tests/foundry/lemmas.k
```diff
@@ -0,0 +1,113 @@
...
+
+    rule (N *Int (N +Int 1)) /Int 2 <Int 115792089237316195423570985008687907853269984665640564039457584007913129639936 => true
+    requires N <Int 481231938336009023090067544955250113853
+    [simplification]
+
+    rule 0 <=Int (N *Int (N +Int 1)) /Int 2 => true
+    requires 0 <=Int N
+    [simplification]
```
PetarMax, line 48: This is also provable by KEVM master. Why is it needed?
```
    claim [sum-to-n-nonneg]: runLemma ( 0 <=Int (N *Int (N +Int 1)) /Int 2 ) => doneLemma ( true )
      requires 0 <=Int N
```

tests/foundry/lemmas.k
```diff
@@ -0,0 +1,113 @@
...
+
+    rule chop(N +Int -1) => N +Int -1
+    requires 1 <=Int N andBool N <Int pow256
+    andBool #rangeUInt(256, N)
+    [simplification]
+
+    rule (N *Int (N +Int 1)) /Int 2 <Int 115792089237316195423570985008687907853269984665640564039457584007913129639936 => true
+    requires N <Int 481231938336009023090067544955250113853
+    [simplification]
```
PetarMax, line 44: This doesn't hold for `N <Int -481231938336009023090067544955250113854`. Why is it needed? The following claim is provable by KEVM master:
```
    claim [sum-to-n-bound]: runLemma ( (N *Int (N +Int 1)) /Int 2 <Int pow256 ) => doneLemma ( true )
      requires -481231938336009023090067544955250113854 <=Int N andBool N <Int 481231938336009023090067544955250113853
```

tests/foundry/lemmas.k
```diff
@@ -0,0 +1,113 @@
...
+
+    rule 0 <=Int (N *Int (N +Int 1)) /Int 2 => true
+    requires 0 <=Int N
+    [simplification]
+
+    rule N *Int 178 <=Int V => N <=Int V /Int 178
+    [simplification]
```
PetarMax, line 51: Two questions:

1. Why would we prefer division over multiplication?
2. Can we not say something more general than 178?

tests/foundry/lemmas.k
PetarMax, line 40: This, when turned into a claim, can be proven by KEVM master. Why is it needed as a lemma?
```
    claim [chop-simpl-1]: runLemma ( chop(N +Int -1) ) => doneLemma ( N +Int -1 )
      requires 1 <=Int N andBool N <Int pow256
```

tests/foundry/lemmas.k
```diff
@@ -0,0 +1,113 @@
...
+    requires #rangeUInt(256, S +Int ((N *Int (N +Int 1)) /Int 2))
+    andBool #rangeUInt(256, S)
+    andBool #rangeUInt(256, N)
+    [simplification]
+
+    rule chop(S +Int N) => S +Int N
+    requires #rangeUInt(256, S +Int ((N *Int (N +Int 1)) /Int 2))
+    andBool #rangeUInt(256, S)
+    andBool #rangeUInt(256, N)
+    [simplification]
```
PetarMax, line 35: This, when turned into a claim, can be proven by KEVM master. Why is it needed as a lemma?
```
    claim [chop-simpl-1]: runLemma ( chop(S +Int N) ) => doneLemma ( S +Int N )
      requires #rangeUInt(256, S +Int ((N *Int (N +Int 1)) /Int 2))
       andBool #rangeUInt(256, S)
       andBool #rangeUInt(256, N)
```
Separately, just having `0 <=Int S andBool 0 <=Int N` should be sufficient, the upper bound on them is not needed as it is established by the other constraint.

tests/foundry/lemmas.k
```diff
@@ -0,0 +1,113 @@
...
+
+module SUM-TO-N-INVARIANT
+
+  imports LOOPSTEST-CONTRACT
+
+    rule N xorInt maxUInt256 <Int S => false
+    requires #rangeUInt(256, S +Int ((N *Int (N +Int 1)) /Int 2))
+    andBool #rangeUInt(256, S)
+    andBool #rangeUInt(256, N)
+    [simplification]
```
PetarMax, line 23: It should be sufficient to have 
```
andBool 0 <=Int S andBool 0 <=Int N
``` 
instead of the `#rangeUInt`s.

This is really `maxUInt256 <Int S +Int N => false`, isn't it? I wonder how the `xorInt` gets introduced...

How about having just
```
rule N xorInt maxUInt256 => maxUInt256 -Int N 
  requires #rangeUInt(256, N)
  [simplification]
```
then this lemma will become provable given that the lemma below with `chop(S +Int N)` is provable (I've tried it with KEVM master).

tests/foundry/lemmas.k
PetarMax: Some of the lemmas have been removed now as superfluous, but I would still like to know why they were there in the first place - was the execution failing somehow? @nwatson22

tests/foundry/lemmas.k
ehildenb: `X xorInt maxUInt256` is introduced as an overflow check for addition by the Solidity compiler. Observe (we're trying to compute `S +Word N`):

- `S +Word N` overflows if `pow256 <=Int S +Int N`.
- ... if `pow256 -Int N <=Int S`.
- `pow256 -Int N == notInt N == maxUInt256 xorInt N` if `#rangeUInt(256, N)`.
- `S +Word N` overflows if `N xorInt maxUInt256 <=Int S`.

So it's a 1-operator way to check "will the addition overflow?"

tests/foundry/lemmas.k
nwatson22: With some of the rules, I'm not sure what was going on. They looked to be needed at the time they were added, but you're right. I removed the ones you said were unnecessary and it still passes.

tests/foundry/lemmas.k
nwatson22: I removed it.

tests/foundry/lemmas.k
nwatson22: I removed this rule.

tests/foundry/lemmas.k
nwatson22: 1. I meant to have V be concrete only
2. I replaced it with a more general lemma:
```
    rule A *Int B <=Int C => A <=Int C /Int B
    requires B >Int 0
     andBool A >=Int 0
     andBool C >=Int 0
    [simplification, concrete(B, 
```

tests/foundry/lemmas.k
nwatson22: I removed it.

tests/foundry/lemmas.k
nwatson22: I removed it.

tests/foundry/lemmas.k
```diff
@@ -0,0 +1,113 @@
...
+
+    rule (S +Int N) +Int (((N -Int 1) *Int N) /Int 2) <Int pow256 => true
+    requires #rangeUInt(256, S +Int ((N *Int (N +Int 1)) /Int 2))
+    andBool #rangeUInt(256, S)
+    andBool #rangeUInt(256, N)
+    [simplification]
```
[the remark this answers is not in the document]
nwatson22: I removed it.

tests/foundry/lemmas.k
nwatson22: Both lemmas work, but I replaced the existing one with the one @PetarMax suggested because it is simpler.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -901,6 +903,13 @@ def parse(s: str) -> list[T]:
         action='store_true',
         help='Rekompile foundry.k even if kompiled definition already exists.',
     )
+    foundry_kompile.add_argument(
+        '--kompiled-dir',
+        dest='kompiled_dir',
+        type=str,
+        default=Path('kompiled'),
```
ehildenb, line 910: Here you say the `default` is a `Path` but the `type` is a `str`?

Instead, it should be `type=str`, and don't provide a default. Then in `exec_foundry_kompile`, have something that says `kompiled = foundry.out / ('kompiled' if kompiled_dir is not None else kompiled_dir)`

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -227,18 +227,20 @@ def foundry_kompile(
...
+    foundry_definition_dir = foundry.out / kompiled_dir
     foundry_requires_dir = foundry_definition_dir / 'requires'
     foundry_llvm_dir = foundry.out / 'kompiled-llvm'
+    foundry_contracts_file = foundry_definition_dir / 'contracts.k'
     foundry_main_file = foundry_definition_dir / 'foundry.k'
     kompiled_timestamp = foundry_definition_dir / 'timestamp'
     ensure_dir_path(foundry_definition_dir)
     ensure_dir_path(foundry_requires_dir)
     ensure_dir_path(foundry_llvm_dir)
+    main_module = 'FOUNDRY-MAIN'
```
ehildenb, line 243:
```suggestion
    kompiled_timestamp = foundry_definition_dir / 'timestamp'
    main_module = 'FOUNDRY-MAIN'
    ensure_dir_path(foundry_definition_dir)
    ensure_dir_path(foundry_requires_dir)
    ensure_dir_path(foundry_llvm_dir)
```

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -261,23 +263,46 @@ def foundry_kompile(
...
-        requires = ['foundry.md']
+    _imports: dict[str, list[str]] = {contract.name: [] for contract in foundry.contracts.values()}
+    for i in imports:
+        imp = i.split(':')
+        if not len(imp) == 2:
+            raise ValueError(f'module imports must be of the form "[ContractName]:[MODULE-NAME]". Got: {i}')
+        if imp[0] in _imports:
+            _imports[imp[0]].append(imp[1])
+        else:
+            raise ValueError(f"Contract {imp[0]!r} doesn\'t exist.")
```
ehildenb, line 274:
```suggestion
            raise ValueError(f'Could not find contract: {imp[0]}')
```

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -261,23 +263,46 @@ def foundry_kompile(
...
+            extra_unparsing_modules=(bin_runtime_definition.all_modules + contract_main_definition.all_modules),
         )
+        with open(foundry_contracts_file, 'w') as fcf:
+            _LOGGER.info(f'Writing file: {foundry_contracts_file}')
+            fcf.write(kevm.pretty_print(bin_runtime_definition) + '\n')
         with open(foundry_main_file, 'w') as fmf:
             _LOGGER.info(f'Writing file: {foundry_main_file}')
-            kevm = KEVM(definition_dir, extra_unparsing_modules=bin_runtime_definition.all_modules)
-            fmf.write(kevm.pretty_print(bin_runtime_definition) + '\n')
+            fmf.write(kevm.pretty_print(contract_main_definition) + '\n')
```
ehildenb, line 305: We can probably say something simpler now:

```
foundry_contracts_file.write_text(kevm.pretty_print(bin_runtime_definition) + '\n')
_LOGGER.info(f'Wrote file: {foundry_contracts_file}')
foundry_main_file.write_text(kevm.pretty_print(contract_main_definition) + '\n')
_LOGGER.info(f'Wrote file: {foundry_main_file}')
```

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -634,31 +659,49 @@ def _write_cfg(cfg: KCFG, path: Path) -> None:
...
     requires: Iterable[str],
-    imports: Iterable[str],
 ) -> KDefinition:
     modules = []
+    main_module = Contract.contract_to_module_name(list(contracts)[0].name_upper, spec=False)
     for contract in contracts:
-        module = contract_to_main_module(contract, empty_config, imports=imports)
+        module = contract_to_main_module(contract, empty_config, imports=['FOUNDRY'])
         _LOGGER.info(f'Produced contract module: {module.name}')
         modules.append(module)
```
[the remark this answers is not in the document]
ehildenb: No need for the individual logging methods.

tests/foundry/lemmas.k
```diff
@@ -0,0 +1,85 @@
...
+
+    rule N xorInt maxUInt256 => maxUInt256 -Int N 
+    requires #rangeUInt(256, N)
+    [simplification]
+
+    rule A *Int B <=Int C => A <=Int C /Int B
+    requires B >Int 0
+     andBool A >=Int 0
+     andBool C >=Int 0
+    [simplification, concrete(B, C)]
```
ehildenb, line 27:
```suggestion
    rule A *Int B <=Int C => A <=Int C /Int B
    requires 0 <Int B
     andBool 0 <=Int A
     andBool 0 <=Int C
    [simplification, concrete(B, C)]
```

ehildenb: Is this true? I am having a hard time proving it....

tests/specs/examples/sum-to-n-foundry-spec.k
```diff
@@ -0,0 +1,89 @@
...
+    rule chop(S +Int N) => S +Int N
+    requires #rangeUInt(256, S +Int ((N *Int (N +Int 1)) /Int 2))
+    andBool #rangeUInt(256, S)
+    andBool #rangeUInt(256, N)
+    [simplification]
+
+    rule chop(N +Int -1) => N +Int -1
+    requires 1 <=Int N andBool N <Int pow256
+    andBool #rangeUInt(256, N)
+    [simplification]
```
ehildenb, line 29: Are we able to replace these with the two simpler lemmas that were used in the foundry test suite?

nwatson22: Yes, this works. I also updated the claim to match the simplification rule.

tests/foundry/lemmas.k
nwatson22: Updated this.

kevm-pyk/src/kevm_pyk/foundry.py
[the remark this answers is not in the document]
nwatson22: I made this change.

kevm-pyk/src/kevm_pyk/foundry.py
nwatson22: I made this change.

kevm-pyk/src/kevm_pyk/foundry.py
nwatson22: I made this change.

kevm-pyk/src/kevm_pyk/foundry.py
nwatson22: I made this change.

kevm-pyk/src/kevm_pyk/__main__.py
nwatson22: I made this change.

tests/foundry/lemmas.k
```diff
@@ -0,0 +1,85 @@
...
+  imports ASSERTTEST-CONTRACT
+
+  syntax Bytes ::= testFunction() [function]
+  syntax Set ::= testFunction2(Bytes) [function]
+
+  rule testFunction() => AssertTest.test_assert_true() [simplification]
+  rule testFunction() => #binRuntime(AssertTest) [simplification]
+  rule testFunction2(PGM) => #computeValidJumpDests(PGM) [simplification]
+
+endmodule
```
ehildenb, line 13: Is this module relevant anymore?

tests/specs/examples/sum-to-n-foundry-spec.k
```diff
@@ -0,0 +1,76 @@
...
+
+    rule N xorInt maxUInt256 => maxUInt256 -Int N 
+    requires #rangeUInt(256, N)
+    [simplification]
+
+    rule A *Int B <=Int C => A <=Int C /Int B
+    requires 0 <Int B
+     andBool 0 <=Int A
+     andBool 0 <=Int C
+    [simplification, concrete(B, C)]
```
ehildenb, line 16: What about this claim breaks without this lemma? I'm curious why we'd want to switch from a multiplication to a division as well. I'm also still having a hard time proving this to be true. It's obvious in the `Real` domain, but in `Int` is it true?

nwatson22: I'm rerunning it to get a more detailed answer, but it was something to do with the gas calculation, where it was unable to show `N *Int 178` was less than the current amount of gas, and this was happening repeatedly when gas was deducted so I needed a more general lemma.

tests/foundry/lemmas.k
nwatson22: I think this line is still testing that we can access the `Contract.function()` syntax from a lemmas file, but I'm not sure how important that is. 
```
rule testFunction() => AssertTest.test_assert_true() [simplification]
```

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -901,6 +903,12 @@ def parse(s: str) -> list[T]:
         action='store_true',
         help='Rekompile foundry.k even if kompiled definition already exists.',
     )
+    foundry_kompile.add_argument(
+        '--kompiled-dir',
+        dest='kompiled_dir',
+        type=str,
+        help='Folder (under [foundry-project-root]/out/) to output as kompiled directory.',
+    )
```
ehildenb, line 911: Why is this option added? Was it something that seemed needed before, but now you don't?

nwatson22: It was used when I was testing the lemma inclusion in a separate kompiled directory. It's no longer needed because testing lemmas are included in the main kompilation now.

ehildenb:
I have a little helper script for updating all the expected output, which you can see here: https://github.com/runtimeverification/evm-semantics/issues/1760

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -261,23 +262,47 @@ def foundry_kompile(
...
+        imp = i.split(':')
+        if not len(imp) == 2:
+            raise ValueError(f'module imports must be of the form "[ContractName]:[MODULE-NAME]". Got: {i}')
+        if imp[0] in _imports:
+            _imports[imp[0]].append(imp[1])
+        else:
+            raise ValueError(f'Could not find contract: {imp[0]}')
+
+    _LOGGER.info(f'foundry_contracts_file: {foundry_contracts_file}')
+    _LOGGER.info(f'foundry_main_file: {foundry_main_file}')
```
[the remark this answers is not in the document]
ehildenb: In general, I follow the principle of:

- Log  that you are _about_ to read a given file (so that if it crashes at read time, you know what it was trying to do).
- Log that you _just did_ a file write (so that the log message only happens if the file got written).

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -634,31 +659,42 @@ def _write_cfg(cfg: KCFG, path: Path) -> None:
...
-        module = contract_to_main_module(contract, empty_config, imports=imports)
-        _LOGGER.info(f'Produced contract module: {module.name}')
-        modules.append(module)
-    _main_module = KFlatModule(
-        main_module if main_module else 'MAIN',
-        imports=(KImport(mname) for mname in [_m.name for _m in modules] + list(imports)),
-    )
-    modules.append(_main_module)
+    modules = [contract_to_main_module(contract, empty_config, imports=['FOUNDRY']) for contract in contracts]
+    main_module = Contract.contract_to_module_name(list(contracts)[0].name_upper, spec=False)
```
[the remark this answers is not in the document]
ehildenb: If it's teh case, then I would make a comment in the code here about that, so it's clear to us in the future.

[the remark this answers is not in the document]
nwatson22: Yes, the KDefinition constructor requires a main_module_name, and it has to be a real module. I just have it use the first module arbitrarily instead of producing a dummy module to be the main module.

kevm-pyk/src/kevm_pyk/foundry.py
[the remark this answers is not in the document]
nwatson22: I removed these messages.

kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -209,11 +209,15 @@ def srcmap(self) -> dict[int, tuple[int, int, int, str, int]]:
 
     @staticmethod
     def contract_to_module_name(c: str, spec: bool = True) -> str:
-        m = c.upper() + '-BIN-RUNTIME'
+        m = c.upper() + '-CONTRACT'
         if spec:
             m = m + '-SPEC'
         return m
```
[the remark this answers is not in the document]
nwatson22: I made this change.

kevm-pyk/src/kevm_pyk/foundry.py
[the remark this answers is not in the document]
nwatson22: I added a comment to the code clarifying this as well.

ehildenb (approved):

PR #1699, opened 2023-04-05
Construct Methods using `abi` array and then lookup in `methodIdentifiers` instead of the other way around

nwatson22:
I mistakenly removed a [feature](https://github.com/runtimeverification/evm-semantics/commit/a16feb2c8f8be497eced9f8fe93ead661624e78b) (#1683) when merging with a later [pull request](https://github.com/runtimeverification/evm-semantics/commit/295280c783bb445bef0a5232337998b092810c01). This changes it back to a warning.

Edit:
While writing a test for this I found out that the system I implemented to allow `_get_method_abi` to allow it to look up methods by arguments as well as name was a naive implementation which makes it miss methods that it was previously able to find when we use complex argument types involving `tuple`. This was causing the warning to occur in places where it wasn't previously. If we want to support overloaded methods, should we maybe scrap the `_get_method_abi` approach entirely and instead iterate through the `"abi"` array to generate our `Method`s? Then when we need the method IDs, we will need to unparse to get the method signature and look it up from the `"methodIdentifiers"` field.


kevm-pyk/src/tests/integration/test_integration.py
```diff
@@ -1,5 +1,2882 @@
...
+  },
+  "id": 17
+}
+"""
+
+
+def test_contract_creation() -> None:
+    contract = Contract('TestContract', json.loads(contract_json), foundry=True)
+    contract = contract
+    assert 1 == 2
```
anvacaru, line 2882: CI is currently failing this test.

anvacaru:
The proofs are passing, but the stage fails because the `foundry.k.check.expected` file has not been updated.
To update it, you can run `make tests/foundry/foundry.k.check CHECK=cp` in the root of the repo.

kevm-pyk/src/tests/integration/test_integration.py
```diff
@@ -1,5 +1,77 @@
...
+              "type": "address"
+            },
+            {
+              "internalType": "bytes",
+              "name": "callData",
+              "type": "bytes"
+            }
+          ],
+          "internalType": "struct IMulticall3.Call[]",
+          "name": "calls",
```
ehildenb, line 32: These seem like they can go as unit tests, not integration tests, because they don't require any external programs, they're just testing functionality of the python code.

nwatson22: Thanks, I wasn't sure how we defined integration tests in the context of this project. I've moved it over to the unit tests file.

ehildenb (approved):

PR #1705, opened 2023-04-06
Cleaned up raw KEVM.account_cell to Foundry.account_TEST_CONTRACT_ADDRESS

### LINKED ISSUE #1687: Reuse helper method for creating initial foundry account
https://github.com/runtimeverification/evm-semantics/blob/260e45afbfa9ddb073cae389dde2347efddc9d20/kevm-pyk/src/kevm_pyk/foundry.py#L657

Here, we are using the raw KEVM.account_cell instead of reusing the Foundry.account_TEST_CONTRACT_ADDRESS: https://github.com/runtimeverification/evm-semantics/blob/master/kevm-pyk/src/kevm_pyk/foundry.py#L177

Scott-Guest:
Closes #1687 

Tiny cleanup to use `Foundry.account_TEST_CONTRACT_ADDRESS` instead of creating a raw `KEVM.account_cell`


kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -662,14 +662,7 @@ def _init_term(
...
     program = KEVM.bin_runtime(KApply(f'contract_{contract_name}'))
-    account_cell = KEVM.account_cell(
-        Foundry.address_TEST_CONTRACT(),
-        intToken(0),
-        program,
-        KApply('.Map'),
-        KApply('.Map'),
-        intToken(1),
-    )
+    account_cell = Subst({'TEST_CODE': program})(Foundry.account_TEST_CONTRACT_ADDRESS())
```
anvacaru, line 665: Hmm, I would 
1. modify `Foundry.account_TEST_CONTRACT_ADDRESS()` to take `program` as a parameter, similar to how `def account_CHEATCODE_ADDRESS(store_var: KInner) -> KApply:` is defined;
2. remove `account_cell` completely;
3. change the `account_cell` from `ACCOUNTS_CELL` on [line 713 ](https://github.com/runtimeverification/evm-semantics/blob/master/kevm-pyk/src/kevm_pyk/foundry.py#L713)with `Foundry.account_TEST_CONTRACT_ADDRESS(program)`.

anvacaru (changes requested):

kevm-pyk/src/kevm_pyk/foundry.py
Scott-Guest: Ah yeah will do, didn't realize nothing else relied on 'TEST_CODE' being there

ehildenb (dismissed):

ehildenb:
@Scott-Guest should it be that `_final_term` also uses the new helper? I also see a call to `KEVM.account_cell(...)` directly in `_final_term` in `foundry.py`, and likely it's the same as the initial term variant that you modified.

Scott-Guest:
Per discussion with @ehildenb, the helper can't be re-used unless we parameterize it to take in all the same arguments as `KEVM.account_cell`. At that point, it isn't very helpful, so let's just remove it instead.

PR #1707, opened 2023-04-07
remove symbolic accounts_init from initial configuration

### LINKED ISSUE #1689: Stop creating new accounts for certain cheat codes

anvacaru:
Removing the symbolic `ACCOUNTS_INIT` from the `<accounts>` cell in the initial state which caused branching.


tests/foundry/test/ContractAssert.t.sol
```diff
@@ -0,0 +1,17 @@
...
+        assert(false);
+    }
+}
+
+contract ContractAssertTest is Test {
+    ContractAssert c;
+
+    function test_call_empty_account() public view{
+        c.f();
+    }
```
ehildenb, line 16: I don't think I understand this test. Is teh idea that because the contract was not deployed with `new ContractAssert()`, the account is actually starting as empty, and so it does not execute the code with `assert(false)`?

ehildenb:
It still seems in that case, we can deploy a contract `C` with symbolic bytecode (by adding a `symbolicEtch` cheatcode, fer example), and then make an explicit call into that account. Would that allow for the full generality @lucasmt ?

lucasmt:
My concern is not about C, but about Y. There can be an arbitrary number of contracts Y that C could potentially modify, so we can't just deploy all of them and make them symbolic.

anvacaru:
Perhaps we could have a cheat code that adds a symbolic `ACCOUNTS_` to the `<accounts>` cell?

lucasmt:
Yeah, that might be a good option. That way it's not turned on by default, but it is available for use cases where it might be relevant.

PetarMax:
> The test that I mentioned in the meeting today was one where we had a contract C that could potentially have write permission to several other contracts, and the property we wanted to verify was that if contract X called C, then C would only modify X, and not any other contract Y.

How did you state this property?

Separately, I am seeing branchings that I don't think should be happening while trying to reproduce Lukso proofs, on whether or not a fixed address meant for some of the contracts already exists in `ACCOUNTS_INIT`, so I echo the separate cheatcode approach.

Lastly, if these branchings are allowed to happen, I think that any (or almost any) call to `#newAccount` will have an accompanying error branch, given
```    
      rule <k> #newExistingAccount ACCT => #end EVMC_ACCOUNT_ALREADY_EXISTS ... </k>
         <account>
           <acctID> ACCT  </acctID>
           <code>   CODE  </code>
           <nonce>  NONCE </nonce>
           ...
         </account>
      requires CODE =/=K .Bytes orBool NONCE =/=Int 0
```

PetarMax:
Importantly, also, there seems to be an invariant of the semantics that needs to be maintained, which is that all of the keys in `<activeAccounts>` are precisely the ids of all of the accounts in `<accounts>`. This is currently broken since there is the `ACCOUNTS_INIT` symbolic variable which potentially holds keys that are not in `<activeAccounts>`. This needs to be accounted for when using the future cheatcode.

@anvacaru also told me that there is a PR that would remove `<activeAccounts>` altogether, which I think is a good idea since then this invariant would not have to be maintained.

PetarMax (approved):

PR #1716, opened 2023-04-10
Foundry test failure summary

nwatson22:
This PR adds an option `--failure-information` to `foundry-show` and `foundry-prove` which causes a summary to be printed of all stuck and frontier nodes. In the case of stuck nodes it prints the path condition (All constraints on that node that weren't there from the beginning) and failure reason (attempts to isolate the non-matching cell or if cells match, shows the simplified failing implication).


kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -208,6 +257,83 @@ def abstract_cell_vars(cterm: KInner, keep_vars: Collection[KVariable] = ()) ->
...
+    concrete_config, *concrete_constraints = concrete
+    abstract_config, *abstract_constraints = abstract
+    config_match = abstract_config.match(concrete_config)
+    if config_match is None:
+        _, concrete_subst = split_config_from(concrete_config)
+        cell_names = concrete_subst.keys()
+        failing_cells = []
+        for cell in cell_names:
+            concrete_cell = get_cell(concrete_config, cell)
+            abstract_cell = get_cell(abstract_config, cell)
```
ehildenb, line 288:
```suggestion
            concrete_cell = concrete.cell(cell)
            abstract_cell = abstract.cell(cell)
```

This variant where we use the `CTerm.cell` interace is automatically memoizing.

kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -208,6 +257,83 @@ def abstract_cell_vars(cterm: KInner, keep_vars: Collection[KVariable] = ()) ->
...
+            if type(csubst_arg) is KVariable and csubst_arg.name.endswith('_CELL'):
+                return True
+        return False
+
+    def _is_negative_cell_subst(constraint: KInner) -> bool:
+        constraint_bool = ml_pred_to_bool(constraint)
+        if type(constraint_bool) is KApply and constraint_bool.label.name == 'notBool_':
+            negative_constraint = constraint_bool.args[0]
+            if type(negative_constraint) is KApply and negative_constraint.label.name == '_andBool_':
+                constraints = flatten_label('_andBool_', negative_constraint)
```
ehildenb, line 273: This whole routine should be upstreamed into `KCFGExplore`, as `implication_failure_reason`.

nwatson22: This is now upstreamed so I removed it here.

kevm-pyk/src/kevm_pyk/utils.py
nwatson22: This method was moved to pyk and this issue was addressed therre.

ehildenb:
To make it so we can test this, can we add `--failure-information` option to `foundry-show-kcfg`, and turn that on for the `test-foundry-kcfg-diff` tests? Then we'll be asserting that we're getting the failure reasons we expect.

We should also print them by default on `foundry-prove`, but can include this output in `foundry-show-kcfg` to make it easy to test.

nwatson22:
> To make it so we can test this, can we add `--failure-information` option to `foundry-show-kcfg`, and turn that on for the `test-foundry-kcfg-diff` tests? Then we'll be asserting that we're getting the failure reasons we expect.
> 
> We should also print them by default on `foundry-prove`, but can include this output in `foundry-show-kcfg` to make it easy to test.

Added this.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -373,6 +373,7 @@ def exec_foundry_prove(
     kore_rpc_command: str | Iterable[str] = ('kore-rpc',),
     smt_timeout: int | None = None,
     smt_retry_limit: int | None = None,
+    failure_info: bool = False,
```
ehildenb, line 376: We should probably have this default match the CLI argument default, which is `True`.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -422,6 +424,7 @@ def exec_foundry_show(
     omit_unstable_output: bool = False,
     frontier: bool = False,
     stuck: bool = False,
+    failure_info: bool = False,
```
ehildenb, line 427: Similarly, this should match the CLI arg default.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -347,6 +353,7 @@ def foundry_prove(
     kore_rpc_command: str | Iterable[str] = ('kore-rpc',),
     smt_timeout: int | None = None,
     smt_retry_limit: int | None = None,
+    failure_info: bool = False,
```
ehildenb, line 356: Match the CLI argument default.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -458,6 +466,7 @@ def foundry_show(
     omit_unstable_output: bool = False,
     frontier: bool = False,
     stuck: bool = False,
+    failure_info: bool = False,
```
ehildenb, line 469: Match the CLI argument default.

kevm-pyk/src/kevm_pyk/__main__.py
nwatson22: Updated.

kevm-pyk/src/kevm_pyk/foundry.py
nwatson22: Updated.

nwatson22:
> Let's use record production syntax on the `foundry_success` predicate, to get better output about whic of the fields are failing. Let's also post links in the output on failing proofs:
> 
>     * Link to the `foundry_success` predicate on failure.
> 
>     * Link to our Discord for support.

Maybe I'm not doing it right, but adding the record production syntax doesn't seem to do anything with the final output. `compiled.txt` also doesn't include the labels. Is it maybe only used in the semantics itself? I know you mentioned it was parse-time only.

include/kframework/foundry.md
```diff
@@ -145,7 +145,15 @@ Hence, checking if a `DSTest.assert*` has failed amounts to reading as a boolean
...
+    syntax Bool ::= 
+      "foundry_success" "("
+        statusCode: StatusCode "," 
+        failed: Int ","
+        revertExpected: Bool ","
+        opcodeExpected: Bool ","
+        recordEventExpected: Bool ","
+        eventExpected: Bool
+      ")" [function, klabel(foundry_success), symbol]
  // --------------------------------------------------------------------------------------------------------------------------------------------
```
ehildenb, line 157: Formatting here.

```suggestion
    syntax Bool ::= 
      "foundry_success" "("
        statusCode: StatusCode "," 
        failed: Int ","
        revertExpected: Bool ","
        opcodeExpected: Bool ","
        recordEventExpected: Bool ","
        eventExpected: Bool
      ")" [function, klabel(foundry_success), symbol]
 // -------------------------------------------------
```

tests/foundry/contracts.k.check.stripped
```diff
@@ -0,0 +1,7360 @@
...
+    syntax AccountParamsTestContract ::= "AccountParamsTest" [symbol(), klabel(contract_AccountParamsTest)]
+    
+      
+    
+    syntax Field ::= AccountParamsTestField
+    
+    syntax AccountParamsTestField ::= "IS_TEST" [symbol(), klabel(field_AccountParamsTest_IS_TEST)]
+    
+    syntax AccountParamsTestField ::= "_failed" [symbol(), klabel(field_AccountParamsTest__failed)]
+
```
ehildenb, line 17: Is this file supposed to be added? I think it's just for comparing output with? My guess is it got added accidentally?

Maybe it should be added to `.gitignore` to make sure this doesn't happen again?

tests/foundry/golden/AssertTest.testFail_assert_true.expected
```diff
@@ -841,3 +807,21 @@ module SUMMARY-ASSERTTEST-TESTFAIL-ASSERT-TRUE
...
+    #And ( { true #Equals 0 <=Int ORIGIN_ID:Int }
+    #And ( { true #Equals 0 <=Int NUMBER_CELL:Int }
+    #And ( { true #Equals CALLER_ID:Int <Int pow160 }
+    #And ( { true #Equals ORIGIN_ID:Int <Int pow160 }
+    #And { true #Equals NUMBER_CELL:Int <=Int maxSInt256 } ) ) ) ) ) #Implies { false #Equals foundry_success ( ... statusCode: EVMC_SUCCESS , failed: #lookup ( .Map , 46308022326495007027972728677917914892729792999299745830475596687180801507328 ) , revertExpected: false , opcodeExpected: false , recordEventExpected: false , eventExpected: false ) } )
+See `foundry_success` predicate for more information:
+https://github.com/runtimeverification/evm-semantics/blob/master/include/kframework/foundry.md#foundry-success-predicate
+Join the Runtime Verification Discord server for support: https://discord.gg/GHvFbRDD
+  Path condition:
+    #Top
```
ehildenb, line 827:
```suggestion
Stuck nodes:

  Node id: 30b447..c22b50
  Failure reason:
    Implication check failed, the following is the remaining implication:
    ( ( { true #Equals 0 <=Int CALLER_ID:Int }
    #And ( { true #Equals 0 <=Int ORIGIN_ID:Int }
    #And ( { true #Equals 0 <=Int NUMBER_CELL:Int }
    #And ( { true #Equals CALLER_ID:Int <Int pow160 }
    #And ( { true #Equals ORIGIN_ID:Int <Int pow160 }
    #And { true #Equals NUMBER_CELL:Int <=Int maxSInt256 } ) ) ) ) ) #Implies { false #Equals foundry_success ( ... statusCode: EVMC_SUCCESS , failed: #lookup ( .Map , 46308022326495007027972728677917914892729792999299745830475596687180801507328 ) , revertExpected: false , opcodeExpected: false , recordEventExpected: false , eventExpected: false ) } )

  Path condition:
    #Top

See `foundry_success` predicate for more information:
https://github.com/runtimeverification/evm-semantics/blob/master/include/kframework/foundry.md#foundry-success-predicate

Join the Runtime Verification Discord server for support: https://discord.gg/GHvFbRDD
```

nwatson22: Done.

kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -138,9 +139,59 @@ def kevm_apr_prove(
...
+
+            res_lines.append('  Path condition:')
+            res_lines += [f'    {kcfg_explore.kprint.pretty_print(_cfg.path_constraints(node.id))}']
+
+            if 'foundry_success' in reason:
+                res_lines.append('')
+                res_lines.append('See `foundry_success` predicate for more information:')
+                res_lines.append(
+                    'https://github.com/runtimeverification/evm-semantics/blob/master/include/kframework/foundry.md#foundry-success-predicate'
+                )
```
ehildenb, line 188: This is the only part that is specific to foundry actually in the reason here. So I would actually just remove this chunk for now, and inline this logic into the places where `print_failure_info` is called.

Then you can actually take this function and upstream it to `APRProof.failure_info` or something instead. Or perhaps as `APRProof.pretty`.

ehildenb: Issue made: https://github.com/runtimeverification/pyk/issues/463

ehildenb:
Let's also link users to the foundry gitbook as well: https://docs.runtimeverification.com/kevm-integration-for-foundry/

And we should factor out all the foundry-specific messages from `print_failure_info`, so that it becomes generic (even semantics independent), and output the foundry-speciic stuff (link to `foundry_success` and to GitBook) afterwards, at the site where `print_failure_info` is called.

kevm-pyk/src/kevm_pyk/utils.py
nwatson22: For now I've moved the `foundry_success` outside of `print_failure_info()` and added the link to the gitbook.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -525,6 +532,7 @@ def _init_and_run_proof(_init_problem: tuple[str, str]) -> bool:
                 kore_rpc_command=kore_rpc_command,
                 smt_timeout=smt_timeout,
                 smt_retry_limit=smt_retry_limit,
+                failure_info=failure_info,
                 trace_rewrites=trace_rewrites,
             )
```
ehildenb, line 537: I guess we need to do something like:

```
result = kevm_apr_prove(...)
if result is failing:
    print extra foundry information also show in `show`
return result
```

So maybe I'd factor out a helper `Foundry.help_info`, and just print that in both places.

nwatson22: I wasn't sure I fully understood what you wanted here. I factored out the repeated code into a `help_info` method and call this in `init_and_run_proof` instead of in `kevm_apr_prove`. Does this address the issue?

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -232,6 +238,24 @@ def account_CHEATCODE_ADDRESS(store_var: KInner) -> KApply:  # noqa: N802
...
+            res_lines.append('')
+            res_lines.append('See `foundry_success` predicate for more information:')
+            res_lines.append(
+                'https://github.com/runtimeverification/evm-semantics/blob/master/include/kframework/foundry.md#foundry-success-predicate'
+            )
+        res_lines.append('')
+        res_lines.append(
+            'Access documentation for KEVM foundry integration at https://docs.runtimeverification.com/kevm-integration-for-foundry/'
+        )
+        return res_lines
```
ehildenb, line 257: Let's make this always return only the foundry help lines, not the failure info. The call site then decides whether to invoke `print_failure_info` and `Foundry.help_info`.

ehildenb:
@nwatson22 can you alos add the `--failure-information` flag to `kcfg-show` command as well? It shouldn't display the foundry speciic failure information, just the generic stuff that `kcfg-show` displays.

Possible also `prove` command should display failure information too!

nwatson22:
> @nwatson22 can you alos add the `--failure-information` flag to `kcfg-show` command as well? It shouldn't display the foundry speciic failure information, just the generic stuff that `kcfg-show` displays.
> 
> Possible also `prove` command should display failure information too!

This option is now added to `prove` and `kcfg-show`.

kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -130,15 +130,57 @@ def kevm_apr_prove(
...
     if len(failure_nodes) == 0:
         _LOGGER.info(f'Proof passed: {_cfgid}')
-        return True
+        return True, None
     else:
         _LOGGER.error(f'Proof failed: {_cfgid}')
-        return False
+        failure_log: list[str] | None = None
+        failure_log = print_failure_info(proof.kcfg, proof.id, kcfg_explore)
+        return False, failure_log
```
ehildenb, line 143: I would still move the printing of the failure info out of this method, so that it doesn't need to return a tuple. At the call site, you can just check the result and decide whether to call `print_failure_info`. The call site has access to the `proof` (because it's passed in by reference), and to the `kcfg_explore`, so it has everything needed to call `print_failure_info` function.

kevm-pyk/src/kevm_pyk/cli.py
```diff
@@ -339,6 +339,19 @@ def explore_args(self) -> ArgumentParser:
...
+            default=True,
+            action='store_true',
+            help='Show failure summary for all failing tests',
+        )
+        args.add_argument(
+            '--no-failure-information',
+            dest='failure_info',
+            action='store_false',
+            help='Do not show failure summary for failing tests',
+        )
```
anvacaru, line 354: Do you know if both arguments are required? Is there a reason not to have only `--no_failure-information` with `default=False`?

ehildenb (approved):

kevm-pyk/src/kevm_pyk/cli.py
ehildenb: If the default is `True`, you have to supply both options so that it's possible to default to `True` but still set to `False` as needed. Here we want the default to be `True`..

nwatson22: It is not strictly required. I was told we usually add both positive and negative flags when the default is is true, for consistency, but not when the default is false. Since there's two different versions of the failure info flag, one with default false and the other with default true, I just included the positive and negative options so the two flag variants behave consistently apart from the default value.

PR #1726, opened 2023-04-13
Constructor implementation of infinite gas.

dkcumming:
`Gas` sort is added to `evm.md` as a supersort of `Int`, this allows all rules in `evm.md` to remain unchanged as `Gas` can be inferred as an `Int` from the operations. `infinite-gas.md` extends `Gas` with infinite gas `#gas(Int)`. `#deductgas` and `#execGas` needed additional constructors as `#gas(_)` would not longer match on int. It appears with the additional rules that the arithmetic operations are not necessary to pass all the proofs in `smoke` except `transfer-success-2-spec.k` - I am unsure why this is not passing, the stopped state appears equivalent to the target state but with `(looped back)`. Adding this PR for comment on the changes.


include/kframework/infinite-gas.md
```diff
@@ -10,129 +10,65 @@ In particular, this means that `#gas(_) <Int #gas(_) => false`, and `#gas(_) <=I
...
+    
+    // syntax Gas ::=  Gas  "+" Gas [function, total] 
+    //              |  Gas  "-" Gas [function, total]
+    //              |  Gas  "*" Gas [function, total]
+    //             //  |  Gas  "/" Gas [function]
+    
+    syntax Bool ::= Gas  "<" Gas [function, total]
+                  | Gas "<=" Gas [function, total]
+                  | Gas  ">" Gas [function, total]
+                  | Gas ">=" Gas [function, total]
```
ehildenb, line 28:
```suggestion
    syntax Bool ::= Gas  "<Gas" Gas [function, total]
                  | Gas "<=Gas" Gas [function, total]
                  | Gas  ">Gas" Gas [function, total]
                  | Gas ">=Gas" Gas [function, total]
```

include/kframework/infinite-gas.md
```diff
@@ -10,129 +10,65 @@ In particular, this means that `#gas(_) <Int #gas(_) => false`, and `#gas(_) <=I
...
+    rule  _:Int   < #gas(_) => true  
+    rule  #gas(_) <= _:Int  => false 
+    rule  _:Int  <= #gas(_) => true  
+    rule  #gas(_)  > _:Int  => true  
+    rule  _:Int   > #gas(_) => false 
+    rule  #gas(_) >= _:Int  => true  
+    rule  _:Int  >= #gas(_) => false 
+
+    rule <k>  G:Int ~> #deductGas => . ... </k> <gas> #gas(GAVAIL) => #gas(GAVAIL -Int G) </gas>
+
```
ehildenb, line 60:
```suggestion

    rule <k>  G:Int ~> #deductGas => . ... </k> <gas> #gas(GAVAIL) => #gas(GAVAIL -Int G) </gas>

```

We don't want to deplicate this rule, better ot just modify the semantics to use `G -Gas G'` in the semantics directly.

include/kframework/infinite-gas.md
```diff
@@ -10,129 +10,65 @@ In particular, this means that `#gas(_) <Int #gas(_) => false`, and `#gas(_) <=I
...
+    rule <k> #gasExec(SCHED, SSTORE INDEX NEW) => Csstore(SCHED, NEW, #lookup(STORAGE, INDEX), #lookup(ORIGSTORAGE, INDEX)) ... </k>
+         <id> ACCT </id>
+         <gas> #gas(_) </gas>
+         <account>
+           <acctID> ACCT </acctID>
+           <storage> STORAGE </storage>
+           <origStorage> ORIGSTORAGE </origStorage>
+           ...
+         </account>
+         <refund> R => R +Int Rsstore(SCHED, NEW, #lookup(STORAGE, INDEX), #lookup(ORIGSTORAGE, INDEX)) </refund>
```
ehildenb, line 70:
```suggestion
    rule <k> #gasExec(SCHED, SSTORE INDEX NEW) => Csstore(SCHED, NEW, #lookup(STORAGE, INDEX), #lookup(ORIGSTORAGE, INDEX)) ... </k>
         <id> ACCT </id>
         <gas> #gas(_) </gas>
         <account>
           <acctID> ACCT </acctID>
           <storage> STORAGE </storage>
           <origStorage> ORIGSTORAGE </origStorage>
           ...
         </account>
         <refund> R => R +Int Rsstore(SCHED, NEW, #lookup(STORAGE, INDEX), #lookup(ORIGSTORAGE, INDEX)) </refund>
```

Same here, we just want to keep the original semantics.

include/kframework/evm.md
```diff
@@ -1859,8 +1886,8 @@ Overall Gas
          <memoryUsed> MU => MU' </memoryUsed> <schedule> SCHED </schedule>
 
     rule <k> _G:Int ~> (#deductMemoryGas => #deductGas)   ... </k> //Required for verification
-    rule <k>  G:Int ~> #deductGas => #end EVMC_OUT_OF_GAS ... </k> <gas> GAVAIL                  </gas> requires GAVAIL <Int G
-    rule <k>  G:Int ~> #deductGas => .                    ... </k> <gas> GAVAIL => GAVAIL -Int G </gas> requires GAVAIL >=Int G
+    rule <k>  G:Int ~> #deductGas => #end EVMC_OUT_OF_GAS ... </k> <gas> GAVAIL:Gas                  </gas> requires GAVAIL <Gas G
+    rule <k>  G:Int ~> #deductGas => .                    ... </k> <gas> GAVAIL:Gas => GAVAIL -Gas G </gas> requires GAVAIL >=Gas G
```
ehildenb, line 1890:
```suggestion
    rule <k>  G:Int ~> #deductGas => .                    ... </k> <gas> GAVAIL:Gas => GAVAIL -Gas G </gas> requires G <=Gas GAVAIL
```

We should try to always use the `<` and `<=` variants. We may find it means we need less of the `*Gas` comparison operators if we make this change here...

dkcumming: Noted, I will change this. I am unsure how it helps with `G*` though

Makefile
```diff
@@ -523,11 +524,11 @@ tests/foundry/out/kompiled/foundry.k.prove: tests/foundry/out/kompiled/timestamp
...
 tests/foundry/out/kompiled/foundry.k.bmc-prove: tests/foundry/out/kompiled/timestamp
-	$(KEVM) foundry-prove --foundry-project-root $(foundry_dir)          \
-	    -j$(FOUNDRY_PAR) --no-simplify-init --max-depth 1000             \
+        $(KEVM) foundry-prove --foundry-project-root $(foundry_dir)          \
+            -j$(FOUNDRY_PAR) --no-simplify-init --max-depth 1000             \
             --bmc-depth 3                                                    \
-	    $(KEVM_OPTS) $(KPROVE_OPTS)                                      \
-	    $(addprefix --test , $(shell cat tests/foundry/bmc-tests))
+            $(KEVM_OPTS) $(KPROVE_OPTS)                                      \
+            $(addprefix --test , $(shell cat tests/foundry/bmc-tests))
```
ehildenb, line 531: The way that spaces and tabs are used in a Makefile matter actually, so we should not change this.

include/kframework/evm.md
```diff
@@ -10,11 +10,13 @@ This file only defines the local execution operations, the file `driver.md` will
...
+requires "gas.md"
 
 module EVM
-    imports STRING
-    imports EVM-DATA
-    imports NETWORK
+   imports STRING
+   imports EVM-DATA
+   imports NETWORK
+   imports GAS
```
ehildenb, line 19:
```suggestion
    imports STRING
    imports EVM-DATA
    imports NETWORK
    imports GAS
```

include/kframework/evm.md
```diff
@@ -460,8 +462,8 @@ Here we load the correct number of arguments from the `wordStack` based on the s
 The `CallOp` opcodes all interperet their second argument as an address.
 
 ```k
-    syntax InternalOp ::= CallSixOp Int Int     Int Int Int Int
-                        | CallOp    Int Int Int Int Int Int Int
+    syntax InternalOp ::= CallSixOp Gas Int     Int Int Int Int
+                        | CallOp    Gas Int Int Int Int Int Int
```
ehildenb, line 466: These should probably remain `Int`? They are being pulled from the `<wordStack>`, which shuold only have `Int`.

include/kframework/gas.md
```diff
@@ -0,0 +1,55 @@
...
+    syntax Gas ::= Gas "*Gas" Gas [function, left, total]
+                 | Gas "/Gas" Gas [function, left]
+                 > left:
+                   Gas "+Gas" Gas [function, left, total]
+                 | Gas "-Gas" Gas [function, left, total]
+
+    syntax Bool ::= Gas  "<Gas" Gas [function, left, total]
+                  | Gas "<=Gas" Gas [function, left, total]
+                  | Gas  ">Gas" Gas [function, left, total]
+                  | Gas ">=Gas" Gas [function, left, total]
```
ehildenb, line 21: I don't see `>Gas` or `>=Gas` actually used anywhere in the semantics, we just are providing lemmas for it. Can we just remove these two variants altogether?

include/kframework/gas.md
```diff
@@ -0,0 +1,55 @@
...
+    
+    syntax Gas ::= "minGas"  "(" Gas "," Gas ")" [function, total]
+    syntax Int ::= "gas2Int" "(" Gas ")"         [function, total]
```
ehildenb, line 24: Group the `gas2Int` definition with the definition of the subsort of `Int` into `Gas` at the top of the module, and group the `minGas` declaration with the other gas-returning operators above.

tests/specs/mcd/end-subuu-pass-spec.k
```diff
@@ -5,7 +5,7 @@ module END-SUBUU-PASS-SPEC
 
     // End_subuu
     claim [End.subuu.pass]:
-      <k> #execute ~> CONTINUATION => #execute ~> CONTINUATION </k>
+      <k> #execute ~> CONTINUATION => JUMP JMPTO ~> #pc [ JUMP ] ~> #execute ~> CONTINUATION </k>
```
ehildenb, line 8: The fixes to this proof and to `vat-subui` can be pulled out into its own PR that you can open up really quick (use a `git worktree` to do it in a separate directory). You can also then remove them from the list `tests/failing-symbolic.pyk` I believe?

include/kframework/evm.md
dkcumming: Okay, I think that might have been legacy from when I had `Gas` permissible on the wordstack

include/kframework/gas.md
dkcumming: I thought for completeness into the future, should someone wish to use the operator it would be there. But I don't think it is necessary.

ehildenb: We always, by convention, use the `<` variants, because it reduces how many lemmas we have to write. So not including the other variants (for `>`), is a way to enforce that convention on future users.

include/kframework/evm.md
```diff
@@ -1858,9 +1860,9 @@ Overall Gas
     rule <k> MU':Int ~> #deductMemory => (Cmem(SCHED, MU') -Int Cmem(SCHED, MU)) ~> #deductMemoryGas ... </k>
          <memoryUsed> MU => MU' </memoryUsed> <schedule> SCHED </schedule>
 
-    rule <k> _G:Int ~> (#deductMemoryGas => #deductGas)   ... </k> //Required for verification
-    rule <k>  G:Int ~> #deductGas => #end EVMC_OUT_OF_GAS ... </k> <gas> GAVAIL                  </gas> requires GAVAIL <Int G
-    rule <k>  G:Int ~> #deductGas => .                    ... </k> <gas> GAVAIL => GAVAIL -Int G </gas> requires GAVAIL >=Int G
+    rule <k> _G:Gas ~> (#deductMemoryGas => #deductGas)   ... </k> //Required for verification
+    rule <k>  G:Gas ~> #deductGas => #end EVMC_OUT_OF_GAS ... </k> <gas> GAVAIL:Gas                  </gas> requires GAVAIL <Gas G
+    rule <k>  G:Gas ~> #deductGas => .                    ... </k> <gas> GAVAIL:Gas => GAVAIL -Gas G </gas> requires G <=Gas GAVAIL
```
ehildenb, line 1865: This change from `>=Int` to `<=Gas` should make it so we never nede the `>` variants.

include/kframework/gas.md
```diff
@@ -0,0 +1,50 @@
...
+    syntax Gas ::= Int
+    syntax Int ::= "gas2Int" "(" Gas ")" [function, total]
+    
+    syntax Gas ::= "minGas" "(" Gas "," Gas ")" [function, total]
+                 > left:
+                   Gas "*Gas" Gas [function, left, total]
+                 | Gas "/Gas" Gas [function, left]
+                 > left:
+                   Gas "+Gas" Gas [function, left, total]
+                 | Gas "-Gas" Gas [function, left, total]
```
dkcumming, line 19: Is this what you meant @ehildenb

include/kframework/infinite-gas.md
```diff
@@ -5,134 +5,55 @@ Here we use the construct `#gas` to represent positive infinity, while tracking
...
+    rule minGas(#gas(_), G':Int)   => G'
 
-    rule #gas(G) *Int G' => #gas(G *Int G') requires 0 <=Int G' [simplification]
-    rule G *Int #gas(G') => #gas(G *Int G') requires 0 <=Int G  [simplification]
+    rule gas2Int(#gas(G)) => G
 
-    rule #gas(G) /Int G' => #gas(G /Int G') requires 0  <Int G' andBool G' <Int #gas(G)  [simplification]
-    rule G /Int #gas(G') => 0               requires 0 <=Int G  andBool G  <Int #gas(G') [simplification]
+    rule _ <=Gas #gas(_)        => true  [simplification]
+    rule         #gas(_) <Gas _ => false [simplification]
```
dkcumming, line 50: @ehildenb Would the rules on 49 and 50 speed things up in the event that `_` is a function? Previously I was thinking to leave them out as we have total functions, but maybe its a good idea to keep that in.

tests/specs/functional/infinite-gas-spec.k
```diff
@@ -20,124 +20,111 @@ module INFINITE-GAS-SPEC
  // Gas simplifications
  // -------------------
 
-    claim <k> runLemma(#if B #then (VGas -Int Csstore(ISTANBUL, V1, V2, V3)) +Int -435 #else (VGas -Int Csstore(ISTANBUL, V1, V2, V3)) +Int -4342 #fi)
+    claim [gas.arith.0]: <k> runLemma(#if B #then (VGas -Int Csstore(ISTANBUL, V1, V2, V3)) +Int -435 #else (VGas -Int Csstore(ISTANBUL, V1, V2, V3)) +Int -4342 #fi)
```
dkcumming, line 23: Should I remove the names now they are passing? Or is it better to keep them around now they are there?

tests/specs/functional/infinite-gas-spec.k
```diff
@@ -1,15 +1,15 @@
 requires "infinite-gas.md"
-requires "lemmas/int-simplification.k"
+requires "lemmas/lemmas.k"
```
dkcumming, line 2: This hierarchy seemed more intuitive to me. Is there any drawback to bringing `lemmas.k` in instead of just `int-simplification.k`?

tests/specs/functional/infinite-gas-spec.k
```diff
@@ -20,124 +20,111 @@ module INFINITE-GAS-SPEC
...
 
-    claim <k> runLemma(#gas(G) -Int 3 *Int (DATA_LEN up/Int 32)) => doneLemma(#gas(G -Int 3 *Int (DATA_LEN up/Int 32))) ... </k> requires 0 <=Int DATA_LEN andBool DATA_LEN <Int pow16
+    claim [gas.inf.9]: <k> runLemma(#gas(G) -Gas 3 *Int (DATA_LEN up/Int 32)) => doneLemma(#gas(G -Int 3 *Int (DATA_LEN up/Int 32))) ... </k> requires 0 <=Int DATA_LEN andBool DATA_LEN <Int pow16
 
-    claim <k> runLemma (#gas(-271 -Int Cmem(ISTANBUL, #memoryUsageUpdate(5, 160, DATA_LEN))) -Int 3 *Int (DATA_LEN up/Int 32) +Int -3)
-           => doneLemma(#gas(-271 -Int Cmem(ISTANBUL, #memoryUsageUpdate(5, 160, DATA_LEN)) -Int 3 *Int (DATA_LEN up/Int 32) +Int -3)) ... </k>
-      requires 0 <=Int DATA_LEN andBool DATA_LEN <Int pow16
+    // claim <k> runLemma (#gas(-271 -Int Cmem(ISTANBUL, #memoryUsageUpdate(5, 160, DATA_LEN))) -Gas 3 *Int (DATA_LEN up/Int 32) +Int -3)
+    //        => doneLemma(#gas(-271 -Int Cmem(ISTANBUL, #memoryUsageUpdate(5, 160, DATA_LEN)) -Int 3 *Int (DATA_LEN up/Int 32) +Int -3)) ... </k>
+    //   requires 0 <=Int DATA_LEN andBool DATA_LEN <Int pow16
```
dkcumming, line 64: How does `K` resolve order of operations between `Int` and `Gas`? I think that is likely why this is still failing.

include/kframework/gas.md
ehildenb: Yeah that's fine.

include/kframework/infinite-gas.md
ehildenb: I think these rules make the rules on line 36 and 40 redundant, so I would remove the rules on lines 36 and 40, and move these to those blocks of rules.

tests/specs/functional/infinite-gas-spec.k
ehildenb: It just becomes a less focused testing harness, because `lemmas.k` introduces a bunch of extra lemmas about other thingns.

tests/specs/functional/infinite-gas-spec.k
```diff
@@ -1,15 +1,15 @@
...
-    imports INT-SIMPLIFICATION
+    imports LEMMAS
 
     syntax KItem ::= runLemma ( Step ) | doneLemma ( Step )
  // -------------------------------------------------------
     rule <k> runLemma(S) => doneLemma(S) ... </k>
 
-    syntax Step ::= Bool | Int
+    syntax Step ::= Bool | Int | Gas
  // --------------------------
```
ehildenb, line 13:
```suggestion
    syntax Step ::= Bool | Int | Gas
 // --------------------------------
```

tests/specs/functional/infinite-gas-spec.k
ehildenb: Keep them please.

tests/specs/functional/infinite-gas-spec.k
ehildenb: Not sure, but I guess we want parenthesis like `#gas(...) -Gas ((3 *Int (...)) +Int 3)`, because that's how it would have been parenthesized before.

Notice I switched the `-3` to a `3`, to prreserve the semantics of the original test.

ehildenb: It looks liek all the haskell and foundry tests are passing anyway though, so I think it's OK to just remove this one if you can't get it passing again easily anyway.

tests/specs/functional/infinite-gas-spec.k
dkcumming: So keep it like this?

include/kframework/evm.md
```diff
@@ -10,11 +10,13 @@ This file only defines the local execution operations, the file `driver.md` will
 ```k
 requires "data.md"
 requires "network.md"
+requires "gas.md"
```
dkcumming, line 13: Would `evm-types.md` be a more appropriate place to import this?

include/kframework/infinite-gas.md
```diff
@@ -5,134 +5,49 @@ Here we use the construct `#gas` to represent positive infinity, while tracking
...
-    rule (notBool (A andBool B)) andBool A => (notBool B) andBool A                      [simplification]
-    rule #if B #then C +Int C1 #else C +Int C2 #fi => C +Int #if B #then C1 #else C2 #fi [simplification]
-
- // Symbolic Gas
-
-    syntax Int ::= #gas ( Int ) [function, total, no-evaluators, klabel(infGas), symbol, smtlib(infGas)]
- // ----------------------------------------------------------------------------------------------------
-    rule #gas(G) +Int G' => #gas(G +Int G') requires 0 <=Int G' orBool 0 -Int G' <Int #gas(G)  [simplification]
-    rule G +Int #gas(G') => #gas(G +Int G') requires 0 <=Int G  orBool 0 -Int G  <Int #gas(G') [simplification]
+    syntax Gas ::= #gas(Int)
```
dkcumming, line 20: Are there any modifiers that should be used for this? I see that the last version used `[symbol]` and `[smtlib(infGas)]` etc

ehildenb: We should probably give it `smtlib(infGas), symbol, klabel(infGas)`. The `symbol, klabel(infGas)` is so that [here](https://github.com/runtimeverification/evm-semantics/blob/3a4aaf884ff2df3455bd8f78b4a579087ea74e5f/kevm-pyk/src/kevm_pyk/kevm.py#L255) we get the correct AST node where building them programatically.

The `smtlib` is so that Z3 knows that this behaves like a function. It will get slightly better reasoning out of it that way.

include/kframework/evm.md
ehildenb: I think `data.md` is actually probably the best place. Let's see if we can make this change in this PR.

Are we able to inline the `infinite-gas.md` into `gas.md`? Probably not I guess, because `infinite-gas.md` mentions things from `evm.md`.... I wonder if ti makes sense to move all the function declarations about gas calculations from `evm.md` to `gas.md` actually, that may be the most clean. I've made an issue about this: https://github.com/runtimeverification/evm-semantics/issues/1789. No need to address it in this PR.

dkcumming: I tried refactoring `infinite-gas.md` in `gas.md`, but I couldn't avoid a cycle. You mean all the `Csstore` etc.? I think that makes sense

include/kframework/lemmas/lemmas.k
```diff
@@ -92,6 +93,63 @@ module LEMMAS [symbolic]
...
+    rule 0 <=Int Caddraccess(_, _)              => true [simplification]
+    rule 0 <=Int Csstore(_, _, _, _)            => true [simplification]
+    rule 0 <=Int Cgascap(_, _:Int, _:Int, _)    => true [simplification]
+    rule 0 <=Gas Cgascap(_, _, _, _)            => true [simplification]
+    rule 0 <=Int #allBut64th(_:Int)             => true [simplification]
+    rule 0 <=Gas #allBut64th(_)                 => true [simplification]
+    rule 0 <=Int Cextra(_, _, _, _)             => true [simplification]
+    rule 0 <=Int _:ScheduleConst < _:Schedule > => true [simplification]
+    rule 0 <=Int G up/Int _                     => true requires 0 <=Int G [simplification]
+    rule Cextra(_, _, _, _) <Int pow256         => true [simplification]
```
ehildenb, line 127:
```suggestion
    rule 0 <=Int Csload(_, _)                   => true                                       [simplification]
    rule 0 <=Int Cmem(_, N)                     => true requires 0 <=Int N                    [simplification]
    rule 0 <=Int Caddraccess(_, _)              => true                                       [simplification]
    rule 0 <=Int Csstore(_, _, _, _)            => true                                       [simplification]
    rule 0 <=Int Cgascap(_, _:Int, _:Int, _)    => true                                       [simplification]
    rule 0 <=Gas Cgascap(_, _, _, _)            => true                                       [simplification]
    rule 0 <=Int #allBut64th(_:Int)             => true                                       [simplification]
    rule 0 <=Gas #allBut64th(_)                 => true                                       [simplification]
    rule 0 <=Int Cextra(_, _, _, _)             => true                                       [simplification]
    rule 0 <=Int _:ScheduleConst < _:Schedule > => true                                       [simplification]
    rule 0 <=Int G up/Int I                     => true requires 0 <=Int G andBool 0 <Int I [simplification]
    rule Cextra(_, _, _, _) <Int pow256         => true                                     [simplification]
```

tests/specs/functional/infinite-gas-spec.k
ehildenb: If we can switch it back, I would prefer that. If it only passes with this change, then we can keep it.

dkcumming: A lot of the lemmas that were in `infitinite-gas.md` I moved to lemmas.k`, because they were referring to finite gas. But I could move them to `gas.md` instead. I think it would work then.

include/kframework/evm.md
```diff
@@ -1423,10 +1425,10 @@ The various `CALL*` (and other inter-contract control flow) operations will be d
          <output> OUT </output>
          <gas> GAVAIL </gas>
 
-    syntax InternalOp ::= "#refund" Exp [strict]
+    syntax InternalOp ::= "#refund" Exp
```
dkcumming, line 1428: @ehildenb 
Not sure if the update to `K` caused this, but `#refund` was running into an error where `~> #freezer#refund__EVM_InternalOp_Exp0_ ( )` was left on the `<k>` cell when refunding a `#gas`. The crashing tests were passing locally, but unsure if there are other issues or efficiency costs caused by removing `[strict]` though.

ehildenb: I am surprised it works at all with removing `strict` here actually, as far as I know, we definitely need that there....

ehildenb: Actually, looking at it, it seems that we probably don't need taht to be `strict`. But if that's so, we should lower teh sort from `Exp` to `Gas` instead (of the argument).

```suggestion
    syntax InternalOp ::= "#refund" Gas
```

include/kframework/evm.md
```diff
@@ -2165,57 +2167,58 @@ The intrinsic gas calculation mirrors the style of the YellowPaper (appendix H).
...
+         => Cextra(SCHED, ISEMPTY, VALUE, ISWARM) +Gas Cgascap(SCHED, GCAP, GAVAIL, Cextra(SCHED, ISEMPTY, VALUE, ISWARM)) ... </k>
+
+   rule <k> Ccallgas(SCHED, ISEMPTY:Bool, GCAP, GAVAIL, VALUE, ISWARM)
+         => Cgascap(SCHED, GCAP, GAVAIL, Cextra(SCHED, ISEMPTY, VALUE, ISWARM)) +Gas #if VALUE ==Int 0 #then 0 #else Gcallstipend < SCHED > #fi ... </k>
+
+   rule <k> Cselfdestruct(SCHED, ISEMPTY:Bool, BAL)
+         => Gselfdestruct < SCHED > +Int Cnew(SCHED, ISEMPTY andBool Gselfdestructnewaccount << SCHED >>, BAL) ... </k>
+
+   syntax Gas ::= Cgascap        ( Schedule , Gas , Gas , Int )         [function, total, smtlib(gas_Cgascap)       ]
+   syntax Int ::= Cgascap        ( Schedule , Int , Int , Int )         [function, total, smtlib(gas_Cgascap)       ]
```
ehildenb, line 2200:
```suggestion
   syntax Gas ::= Cgascap        ( Schedule , Gas , Gas , Int )         [function, total, smtlib(gas_Cgascap_Gas)       ]
   syntax Int ::= Cgascap        ( Schedule , Int , Int , Int )         [function, total, smtlib(gas_Cgascap_Int)       ]
```

We shouldn't use the same SMT symbols for these productions, because their sorts don't match.

include/kframework/evm.md
```diff
@@ -2306,6 +2309,7 @@ There are several helpers for calculating gas (most of them also specified in th
  // --------------------------------------------------------------------------------------------------
     rule #accountEmpty(CODE, NONCE, BAL) => CODE ==K .Bytes andBool NONCE ==Int 0 andBool BAL ==Int 0
 
+    syntax Gas ::= #allBut64th ( Gas ) [function, total, smtlib(gas_allBut64th)]
     syntax Int ::= #allBut64th ( Int ) [function, total, smtlib(gas_allBut64th)]
```
ehildenb, line 2313:
```suggestion
    syntax Gas ::= #allBut64th ( Gas ) [function, total, smtlib(gas_allBut64th_Gas)]
    syntax Int ::= #allBut64th ( Int ) [function, total, smtlib(gas_allBut64th_Int)]
```

include/kframework/evm.md
```diff
@@ -2165,57 +2167,58 @@ The intrinsic gas calculation mirrors the style of the YellowPaper (appendix H).
...
+         => Cextra(SCHED, ISEMPTY, VALUE, ISWARM) +Gas Cgascap(SCHED, GCAP, GAVAIL, Cextra(SCHED, ISEMPTY, VALUE, ISWARM)) ... </k>
+
+   rule <k> Ccallgas(SCHED, ISEMPTY:Bool, GCAP, GAVAIL, VALUE, ISWARM)
+         => Cgascap(SCHED, GCAP, GAVAIL, Cextra(SCHED, ISEMPTY, VALUE, ISWARM)) +Gas #if VALUE ==Int 0 #then 0 #else Gcallstipend < SCHED > #fi ... </k>
+
+   rule <k> Cselfdestruct(SCHED, ISEMPTY:Bool, BAL)
+         => Gselfdestruct < SCHED > +Int Cnew(SCHED, ISEMPTY andBool Gselfdestructnewaccount << SCHED >>, BAL) ... </k>
+
+   syntax Gas ::= Cgascap        ( Schedule , Gas , Gas , Int )         [function, total, smtlib(gas_Cgascap_Gas)       ]
+   syntax Int ::= Cgascap        ( Schedule , Int , Int , Int )         [function, total, smtlib(gas_Cgascap_Int)       ]
```
ehildenb, line 2200:
```suggestion
   syntax Gas ::= Cgascap        ( Schedule , Gas , Gas , Int )         [function, total, smtlib(gas_Cgascap_Gas)   ]
   syntax Int ::= Cgascap        ( Schedule , Int , Int , Int )         [function, total, smtlib(gas_Cgascap_Int)   ]
```

include/kframework/gas.md
```diff
@@ -0,0 +1,50 @@
...
+
+    syntax Bool ::= Gas  "<Gas" Gas [function, left, total]
+                  | Gas "<=Gas" Gas [function, left, total]
```
ehildenb, line 22:
```suggestion
    syntax Gas ::= "minGas" "(" Gas "," Gas ")" [function, total]
                 > left:
                   Gas "*Gas" Gas [function, total]
                 | Gas "/Gas" Gas [function]
                 > left:
                   Gas "+Gas" Gas [function, total]
                 | Gas "-Gas" Gas [function, total]

    syntax Bool ::= Gas  "<Gas" Gas [function, total]
                  | Gas "<=Gas" Gas [function, total]
```

In the case of the arithmetic operators, these `left` annotations are redundant with the `left: ...` component.

In the case of the boolean operators, they don't make sense because they aren't associative operator.s

include/kframework/infinite-gas.md
```diff
@@ -5,134 +5,49 @@ Here we use the construct `#gas` to represent positive infinity, while tracking
...
+    rule      G  /Gas #gas(G') => #gas(G /Int G')  requires G' =/=Int 0
+    rule #gas(G) +Gas #gas(G') => #gas(G +Int G')
+    rule #gas(G) -Gas #gas(G') => #gas(G -Int G')
+    rule #gas(G) *Gas #gas(G') => #gas(G *Int G')
+    rule #gas(G) /Gas #gas(G') => #gas(G /Int G')  requires G' =/=Int 0
+
+    rule  _:Int    <Gas #gas(_) => true
+    rule  #gas(_)  <Gas _       => false [simplification]
+    rule  #gas(_) <=Gas _:Int   => false
+    rule _        <=Gas #gas(_) => true  [simplification]
```
ehildenb, line 38:
```suggestion
    rule  _:Int    <Gas #gas(_) => true
    rule  #gas(_)  <Gas _       => false
    rule  #gas(_) <=Gas _:Int   => false
    rule _        <=Gas #gas(_) => true
```

Not sure, but it feels weird that these two rules are marked as simplification, and the others are not. Any reason for it?

include/kframework/evm.md
```diff
@@ -2286,8 +2289,9 @@ There are several helpers for calculating gas (most of them also specified in th
  // --------------------------------------------------------------------------------------------------
     rule #accountEmpty(CODE, NONCE, BAL) => CODE ==K .Bytes andBool NONCE ==Int 0 andBool BAL ==Int 0
 
-    syntax Int ::= #allBut64th ( Int ) [function, total, smtlib(gas_allBut64th)]
- // ----------------------------------------------------------------------------
+    syntax Gas ::= #allBut64th ( Gas ) [function, total, smtlib(gas_allBut64th_Gas)]
+    syntax Int ::= #allBut64th ( Int ) [function, total, smtlib(gas_allBut64th_Int)]
+ // --------------------------------------------------------------------------------
     rule [allBut64th.pos]: #allBut64th(N) => N -Int (N /Int 64) requires 0 <=Int N
     rule [allBut64th.neg]: #allBut64th(N) => 0                  requires N  <Int 0
```
ehildenb, line 2296: Do we actually need 2 productions for this? Can we just have the `Gas -> Gas` variant? The rules below will still work directly for that, because `Int` is a subsort, and `N` in the rules is restricted to being an `Int` by the requires clauses. You can add the extra sort constraint just to be sure, but I think it should work without both productions.

```suggestion
    syntax Gas ::= #allBut64th ( Gas ) [function, total, smtlib(gas_allBut64th)]
 // ----------------------------------------------------------------------------
    rule [allBut64th.pos]: #allBut64th(N:Int) => N -Int (N /Int 64) requires 0 <=Int N
    rule [allBut64th.neg]: #allBut64th(N:Int) => 0                  requires N  <Int 0
```

include/kframework/evm.md
```diff
@@ -2145,57 +2147,58 @@ The intrinsic gas calculation mirrors the style of the YellowPaper (appendix H).
...
+                | Cmem           ( Schedule , Int )                     [function, total, smtlib(gas_Cmem), memo    ]
+                | Caddraccess    ( Schedule , Bool )                    [function, total, smtlib(gas_Caddraccess)   ]
+                | Cstorageaccess ( Schedule , Bool )                    [function, total, smtlib(gas_Cstorageaccess)]
+                | Csload         ( Schedule , Bool )                    [function, total, smtlib(gas_Csload)        ]
+                | Cextcodesize   ( Schedule )                           [function, total, smtlib(gas_Cextcodesize)  ]
+                | Cextcodecopy   ( Schedule , Int )                     [function, total, smtlib(gas_Cextcodecopy)  ]
+                | Cextcodehash   ( Schedule )                           [function, total, smtlib(gas_Cextcodehash)  ]
+                | Cbalance       ( Schedule )                           [function, total, smtlib(gas_Cbalance)      ]
+                | Cmodexp        ( Schedule , Bytes , Int , Int , Int ) [function, total, smtlib(gas_Cmodexp)       ]
  // ------------------------------------------------------------------------------------------------------------------
```
ehildenb, line 2195:
```suggestion
   syntax Gas ::= Cgascap        ( Schedule , Gas , Gas , Int )         [function, total, smtlib(gas_Cgascap)       ]
   syntax Int ::= Csstore        ( Schedule , Int , Int , Int )         [function, total, smtlib(gas_Csstore)       ]
                | Rsstore        ( Schedule , Int , Int , Int )         [function, total, smtlib(gas_Rsstore)       ]
                | Cextra         ( Schedule , Bool , Int , Bool )       [function, total, smtlib(gas_Cextra)        ]
                | Cnew           ( Schedule , Bool , Int )              [function, total, smtlib(gas_Cnew)          ]
                | Cxfer          ( Schedule , Int )                     [function, total, smtlib(gas_Cxfer)         ]
                | Cmem           ( Schedule , Int )                     [function, total, smtlib(gas_Cmem), memo    ]
                | Caddraccess    ( Schedule , Bool )                    [function, total, smtlib(gas_Caddraccess)   ]
                | Cstorageaccess ( Schedule , Bool )                    [function, total, smtlib(gas_Cstorageaccess)]
                | Csload         ( Schedule , Bool )                    [function, total, smtlib(gas_Csload)        ]
                | Cextcodesize   ( Schedule )                           [function, total, smtlib(gas_Cextcodesize)  ]
                | Cextcodecopy   ( Schedule , Int )                     [function, total, smtlib(gas_Cextcodecopy)  ]
                | Cextcodehash   ( Schedule )                           [function, total, smtlib(gas_Cextcodehash)  ]
                | Cbalance       ( Schedule )                           [function, total, smtlib(gas_Cbalance)      ]
                | Cmodexp        ( Schedule , Bytes , Int , Int , Int ) [function, total, smtlib(gas_Cmodexp)       ]
 // ------------------------------------------------------------------------------------------------------------------
```

Do we actually need both productiions forr `Cgascap`?

include/kframework/evm.md
```diff
@@ -2145,57 +2147,58 @@ The intrinsic gas calculation mirrors the style of the YellowPaper (appendix H).
...
+
+   rule <k> Ccallgas(SCHED, ISEMPTY:Bool, GCAP, GAVAIL, VALUE, ISWARM)
+         => Cgascap(SCHED, GCAP, GAVAIL, Cextra(SCHED, ISEMPTY, VALUE, ISWARM)) +Gas #if VALUE ==Int 0 #then 0 #else Gcallstipend < SCHED > #fi ... </k>
+
+   rule <k> Cselfdestruct(SCHED, ISEMPTY:Bool, BAL)
+         => Gselfdestruct < SCHED > +Int Cnew(SCHED, ISEMPTY andBool Gselfdestructnewaccount << SCHED >>, BAL) ... </k>
+
+   syntax Gas ::= Cgascap        ( Schedule , Gas , Gas , Int )         [function, total, smtlib(gas_Cgascap_Gas)   ]
+   syntax Int ::= Cgascap        ( Schedule , Int , Int , Int )         [function, total, smtlib(gas_Cgascap_Int)   ]
+                | Csstore        ( Schedule , Int , Int , Int )         [function, total, smtlib(gas_Csstore)       ]
```
ehildenb, line 2181:
```suggestion
   syntax Gas ::= Cgascap        ( Schedule , Gas , Gas , Int )         [function, total, smtlib(gas_Cgascap)       ]
   syntax Int ::= Csstore        ( Schedule , Int , Int , Int )         [function, total, smtlib(gas_Csstore)       ]
```

Do we actually need both productions?

include/kframework/lemmas/lemmas.k
```diff
@@ -92,6 +93,63 @@ module LEMMAS [symbolic]
...
+  // ########################
+
+    rule 0 <=Int Csload(_, _)                   => true                                       [simplification]
+    rule 0 <=Int Cmem(_, N)                     => true requires 0 <=Int N                    [simplification]
+    rule 0 <=Int Caddraccess(_, _)              => true                                       [simplification]
+    rule 0 <=Int Csstore(_, _, _, _)            => true                                       [simplification]
+    rule 0 <=Int Cgascap(_, _:Int, _:Int, _)    => true                                       [simplification]
+    rule 0 <=Gas Cgascap(_, _, _, _)            => true                                       [simplification]
+    rule 0 <=Int #allBut64th(_:Int)             => true                                       [simplification]
+    rule 0 <=Gas #allBut64th(_)                 => true                                       [simplification]
```
ehildenb, line 123:
```suggestion
    rule 0 <=Int Cgascap(_, _:Int, _:Int, _)    => true                                       [simplification]
    rule 0 <=Int #allBut64th(_:Int)             => true                                       [simplification]
```

Are these rules actually needed? It seems to me that instead what should happen is the `#gas(...)` on the right bubbles up to the top, then we have already have a rule that says:

```
rule _ <=Gas #gas(_) => true
```

include/kframework/infinite-gas.md
dkcumming: I recall now the reason I had these here was for short circuit evaluation. Maybe I am wrong, but I thought that it would mean that any rules that could apply to the interior of a `#gas(...)` would be skipped. Maybe this is better achieved with priorities?

include/kframework/evm.md
dkcumming: I run into Sort errors when I strip this back. 
`rule #allBut64th(#gas(G)) => #gas(#allBut64th(G:Int))` errors on RHS as `#allBut64th` returns a `Gas`, in reality this is an `Int` for this case. Is there a way to communicate that aside from duplicating the rule?

ehildenb:
Ah OK, thanks for investigating,good to have these reasons on record. Let's roll back the changes to the working version from before.

include/kframework/evm.md
```diff
@@ -2287,7 +2290,8 @@ There are several helpers for calculating gas (most of them also specified in th
  // --------------------------------------------------------------------------------------------------
     rule #accountEmpty(CODE, NONCE, BAL) => CODE ==K .Bytes andBool NONCE ==Int 0 andBool BAL ==Int 0
 
-    syntax Int ::= #allBut64th ( Int ) [function, total, smtlib(gas_allBut64th)]
+    syntax Gas ::= #allBut64th ( Gas ) [function, total, smtlib(gas_allBut64th_Gas)]
+    syntax Int ::= #allBut64th ( Int ) [function, total, smtlib(gas_allBut64th_Int)]
  // ----------------------------------------------------------------------------
```
ehildenb, line 2295:
```suggestion
    syntax Gas ::= #allBut64th ( Gas ) [function, total, smtlib(gas_allBut64th_Gas)]
    syntax Int ::= #allBut64th ( Int ) [function, total, smtlib(gas_allBut64th_Int)]
 // --------------------------------------------------------------------------------
```

include/kframework/gas.md
ehildenb, line 19:
```suggestion
    syntax Gas ::= "minGas" "(" Gas "," Gas ")" [function, total]
                 > left:
                   Gas "*Gas" Gas [function, total]
                 | Gas "/Gas" Gas [function]
                 > left:
                   Gas "+Gas" Gas [function, total]
                 | Gas "-Gas" Gas [function, total]
```

include/kframework/gas.md
ehildenb, line 22:
```suggestion
    syntax Bool ::= Gas  "<Gas" Gas [function, total]
                  | Gas "<=Gas" Gas [function, total]
```

include/kframework/gas.md
```diff
@@ -0,0 +1,50 @@
...
+========
+
+Here is a representation of `Gas`. This sort is used by `<gas>`, `<callGas>`, and `<gasUsed>` cells of the EVM configuration. As is in this file, `Gas` is a super-sort of `Int` with no differing behaviour. However the distinction at this level allows for an extension of infinite gas later, see `infinite-gas.md`.
+
+```k
+module GAS-SYNTAX
+    imports INT-SYNTAX
+    
+    syntax Gas ::= Int
+    syntax Int ::= "gas2Int" "(" Gas ")" [function, total]
```
ehildenb, line 11:
```suggestion
    syntax Int ::= "gas2Int" "(" Gas ")" [function]
```

I don't think this function is total, is it?

I guess for `gas2Int(#gas(X)) => X` we kinda make it total, but does that really make sense? That means that `#gas` doesn't really behave like a point, does it?

include/kframework/infinite-gas.md
ehildenb, line 38:
```suggestion
    rule _:Int    <Gas #gas(_) => true
    rule #gas(_)  <Gas _       => false [simplification]
    rule #gas(_) <=Gas _:Int   => false
    rule _       <=Gas #gas(_) => true  [simplification]
```

include/kframework/infinite-gas.md
```diff
@@ -5,134 +5,49 @@ Here we use the construct `#gas` to represent positive infinity, while tracking
...
+
+    rule minGas(#gas(G), #gas(G')) => #gas(minInt(G, G'))
+    rule minGas(G:Int  , #gas(_))  => G
+    rule minGas(#gas(_), G':Int)   => G'
```
ehildenb, line 42:
```suggestion
    rule minGas(#gas(G), #gas(G')) => #gas(minInt(G, G'))
    rule minGas(G:Int  , #gas(G')) => #gas(minInt(G, G'))
    rule minGas(#gas(G), G':Int)   => #gas(minIntG, G'))
```

Wouldn't this be the pointed set semantics?

include/kframework/lemmas/lemmas.k
ehildenb: It still seems to me that the rules here with `<=Gas` are not needed.

include/kframework/gas.md
dkcumming: I agree, I think this is the weakest part of the PR. `gas2Int` only exists to process the `GAS` opcode by placing the `<gas>` cell on the wordstack, so a `#gas(G)` needs to be stripped to `G` so the sort doesn't conflict. But I think this as a solution is flawed, as the information that `G` came from a `#gas()` is lost.

include/kframework/infinite-gas.md
dkcumming: Do you have a tool that helps show the whitespace? You have also noticed trailing whitespace in the past that I have missed. Or maybe there is a system you have?

include/kframework/infinite-gas.md
dkcumming: If we are happy to go with this that is fine, it will mean that previous tests will change. This test here will change:
[`claim <k> runLemma(3 <Int minInt(#gas(_VGas), 2)) => doneLemma(false) ... </k>`](https://github.com/runtimeverification/evm-semantics/blob/8a2ba8b772ac2c853df286c7b86fe112af1fd85a/tests/specs/functional/infinite-gas-spec.k#L105)

ehildenb (approved):

PR #1702, opened 2023-04-06
Refactor test-foundry make command

anvacaru:
- added a check for `kevm foundry-list` to catch bugs like #1701.
- split `test-foundry` make command in : 
     -  `test-foundry-kompile` -  checks `foundry.k` file.
     -  `test-foundry-prove` - executes the proofs.
     -  `test-foundry-list` - runs `kevm foundry-list` and compares the output to an expected file in order to test the parsing of the KCFGs.

PR #1731, opened 2023-04-13
Fix foundry-list proof id lookup

nwatson22:
`foundry-list` was broken by a change to the way kcfgs are stored introduced in [pyk#325](https://github.com/runtimeverification/pyk/commit/3e369fc88a2253709b0af1b4c6ea48319cd52b79) resulting in a KeyError for 'proofid' whenever it was run. This changes it to use the correct key.


ehildenb:
@nwatson22 following https://github.com/runtimeverification/evm-semantics/pull/1702 being merged, please update this PR in the way you're thinking. Basically:

- We should always be reading/writing entire `AGProof`, not just a `KCFG`
- We should be using `AGProof.summary` to do the `foundry-list` command.

anvacaru:
Perhaps we should also sort [the `stats` in `foundry_list`](https://github.com/runtimeverification/evm-semantics/blob/master/kevm-pyk/src/kevm_pyk/__main__.py#L476-L480) before displaying. I've seen the foundry-list check fail because of a different kcfg order.

nwatson22:
> Perhaps we should also sort [the `stats` in `foundry_list`](https://github.com/runtimeverification/evm-semantics/blob/master/kevm-pyk/src/kevm_pyk/__main__.py#L476-L480) before displaying. I've seen the foundry-list check fail because of a different kcfg order.

Good point, I'll add that to this branch.

ehildenb (approved):

PR #1740, opened 2023-04-19
Enable more proofs to run with pyk prover

ehildenb:
This PR enables more proofs to run with pyk prover. The key insight is that the backend expects the `<k>` cell to contain a `kseq` element (in Kore), but many of the proofs just have `<k> #execute => #halt </k>` at the top of the `<k>` cell. So this PR automatically inserts the `KSequence` if it detects that it's not present.

- Automatically insert a `KSequence` at the top of the `<k>` cell for initial/target terms of proofs that are converted to KCFGs.
- Consolidate the logic of setting up textual proofs as KCFGs.
- Remove many proofs from `tests/failing-symbolic.pyk` that are now passing with the pyk RPC prover.
- Enable modifying variable `KEVM` as a user, while still getting the modifications the makefile makes when doing foundry proofs.


ehildenb:
@radumereuta I've asked for your review because I was surprised that these proofs are working on normal frontend. I guess the frontend is injecting `KSequence` to the proofs at some point after parsing it? Because we're reading the Kast parsed from the frontend.

radumereuta (commented):
The kprove front-end applies an entire compilation [pipeline](https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/backend/kore/KoreBackend.java#L178)
```java
m -> resolveComm
.andThen(resolveAnonVars)
.andThen(numberSentences)
.andThen(resolveSemanticCasts)
.andThen(generateSortProjections)
.andThen(propagateMacroToRules)
.andThen(expandMacros)
.andThen(checkSimplificationRules)
.andThen(addImplicitComputationCell)
.andThen(resolveFreshConstants)
.andThen(concretizeCells)
.andThen(subsortKItem)
.andThen(restoreDefinitionModulesTransformer(def))
```
If you used `kast --intput rule` that's much [simpler](https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/kast/KastFrontEnd.java#L155):
```java
if (options.steps.contains(KompileSteps.anonVars))
    r = (Rule) new ResolveAnonVar().resolve(r);
r = (Rule) new ResolveSemanticCasts(false).resolve(r); // move casts to side condition predicates
r = (Rule) new ConstantFolding().fold(unparsingMod, r);
r = (Rule) new CloseCells(configInfo, sortInfo, labelInfo).close(r);
if (options.steps.contains(KompileSteps.concretizeCells)) {
    r = (Rule) new AddImplicitComputationCell(configInfo, labelInfo).apply(mod, r);
    r = (Rule) new ConcretizeCells(configInfo, labelInfo, sortInfo, mod).concretize(mod, r);
}
```
Not to mention that we're also doing some extra [structuralChecks](https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/kompile/Kompile.java#L420) and [proverChecks](https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/kompile/Kompile.java#L401).

ehildenb:
This is where the JSON spec that the frontend emits is done: https://github.com/runtimeverification/k/blob/master/kernel/src/main/java/org/kframework/kprove/KProve.java#L79

Does that happen before or after the compilation pipeline you mention? It look to me like the definition is emitted after the compilation pipeline, so I'd expect the ksequence to be in place. Either that, or it happens at ModuleToKORE time.

Either way, the backend will not rewrite the term without the KSequence in place, and seem to be happy to do it with that in place.

tothtamas28 (approved):

radumereuta:
I can't see anything in `ModuleToKore` that would add that KSequence.
I think it happens in the `ConfigurationConcretization` step.
I couldn't reproduce it with a simple example. I will try to reproduce it with EVM later today. So please don't merge until then. Maybe there is a compiler bug, and this workaround is not needed.

ehildenb:
This is the code that extracts claims: https://github.com/runtimeverification/pyk/blob/master/src/pyk/ktool/kprove.py#L296

It's just calling `kprove ...` directly to get the parsed output.

Issue #1604, opened 2023-03-03
Remove activeAccounts cell


anvacaru:
Now that the Java backend has been deprecated, we might be able to remove the `<activeAccounts>` cell.

Issue #1670, opened 2023-03-28
`kevm interpret --debugger` not working correctly


anvacaru:
Nothing is displayed when `--debugger` flag is passed to `kevm interpret`. 
I've tried inputting `step` and got the response `The program is not being run`. 
Seems that the debugger is running but all the output is redirected somewhere else.
<img width="1437" alt="image" src="https://user-images.githubusercontent.com/16517508/228190350-0af2f574-056a-4201-b46a-7b77dec02c93.png">

anvacaru:
After debugging with @radumereuta, we found the following:
   1. we need to specify a way to enable `--enable-llvm-debug` in `kevm-pyk`, maybe add a new option `--kopts ARGS` to pyk so we can manually add new options that are not supported by pyk yet. 
   2. the `kevm` script is redirecting the output to a temp file [here](https://github.com/runtimeverification/evm-semantics/blob/master/bin/kevm#L215). We should add support for `--debugger` flag to stop hiding the output.
  
krun should fail when --debugger is provided but the definition is not kompiled with debug symbols. (`--enable-llvm-debug` or `-g -O1`)

tothtamas28:
> we need to specify a way to enable `--enable-llvm-debug` in `kevm-pyk`, maybe add a new option `--kopts ARGS` to pyk

I'd rather not add such a blanket parameter, because it makes working around missing flags very convenient for downstream projects. Too many such flags and we lose the benefits of the additional layer. Instead, people should just open an issue or PR on `pyk` for the missing feature, helping the tool evolve in the right direction.

PR #1743, opened 2023-04-19
Remove activeAccounts cell from the KEVM configuration

anvacaru:
fixes: #1604 

The Java backend is outdated and not used anymore. In this PR I remove the `<activeAccounts>`cell and adapt rules to use the `<accounts>` cell map.

~blocked on #1670~
~blocked on https://github.com/runtimeverification/k/issues/3365~


anvacaru:
Currently there is an issue with the `CALLBlake2f_MaxRounds.json` test, causing the timeout.

ehildenb:
Try changing this to `-O1` instead, and see if we get the same hang: https://github.com/runtimeverification/evm-semantics/blob/30606a30f95c591e384cd3acfe336129e72ca183/Makefile#L228

anvacaru:
@ehildenb The test passed with `-O1`

include/kframework/evm.md
```diff
@@ -613,12 +608,14 @@ After executing a transaction, it's necessary to have the effect of the substate
            ...
          </message>
 
+    rule <k> (.K => #newAccount MINER) ~> #finalizeTx(_)... </k>
+         <coinbase> MINER </coinbase> [owise]
```
ehildenb, line 612: I think we can just get rid of this rule? Because above the rule for `#finalizeTx(true) => ...` already injects `MINER` into the set  of accounts to call `#finalizeStorage` on, and `#finalizeStorage` will already make sure the account exists when it does its work.

include/kframework/evm.md
```diff
@@ -818,18 +805,16 @@ These are just used by the other operators for shuffling local execution state a
          </account>
       requires VALUE >Int ORIGFROM
 
-    rule <k> (. => #newAccount ACCTTO) ~> #transferFunds ACCTFROM ACCTTO VALUE ... </k>
-         <activeAccounts> ACCTS </activeAccounts>
+    rule <k> #transferFunds ACCTFROM ACCTTO VALUE => #transferFundsToNonExistent ACCTFROM ACCTTO VALUE ... </k> [owise]
+
+    rule <k> #transferFundsToNonExistent ACCTFROM ACCTTO VALUE => #newAccount ACCTTO ~> #transferFunds ACCTFROM ACCTTO VALUE ... </k>
```
ehildenb, line 810:
```suggestion
    rule <k> (. => #newAccount ACCTTO) ~> #transferFunds ACCTFROM ACCTTO VALUE ... </k> [owise]
```

Instead of making the new `#transferFundsNonExistent`, doesn't using this pattern work?

include/kframework/evm.md
anvacaru: @ehildenb `#finalizeStorage` is adding the account, but after `#finalizeTx(true)`. The execution gets stuck because in the rule for `#finalizeTx(false => true)`, the coinbase account is not added in the state.

include/kframework/evm.md
anvacaru: Not really, `#transferFundsNonExistent` is used to differentiate the cases when `Gemptyisnonexistent <<SCHED >>` is true or not.

ehildenb: Ok, sounds good. But looking at the rules, it seems that the case where `ACCTTO ==K ACCTFROM`, _and_ the account does _not_ exist in the configuration, is not handled. I guess this case doesn't usually make sense, but maybe it should be handled explicitly? Right now, it will just get stuck with `#transferFundsNonExistent` on the `<k>` cell.

anvacaru: @ehildenb Well, the rules don't cover the case when the `ACCTFROM` is not in the configuration. A rule for this case would also include the scenario you mentioned.

tothtamas28 (approved):

Issue #1709, opened 2023-04-07
Simplify and generalize powers of 2 lemmas


ehildenb:
We have many lemmas calling out specific powers of 2 (see, for example: https://github.com/runtimeverification/evm-semantics/pull/1662/files#r1160865517 and https://github.com/runtimeverification/evm-semantics/pull/1662/files#r1160866673).

Because the one argument in all these lemmas is concrete, we can likely get away with:

- Reducing all these cases to a single case that uses `log2Int` instead of specific powers of 2.
- Taking advantage of `concrete` attribute.
- Taking advantage of `comm` attribute.

PR #1745, opened 2023-04-19
General simplification for power-of-two bitwise-and identity

PetarMax:
Fixes: https://github.com/runtimeverification/evm-semantics/issues/1709

- Added a general simplification to capture that bitwise-and is an identity if one of the operands is one less than a given power of two.
- Moved previous simplifications to be (provable) claims
- Uncommented several claims in `lemmas-spec.k` that pass


ehildenb:
BTW, if you do this: 
image

Github will know to close the issue when this PR is merged.

ehildenb (approved):

PR #1747, opened 2023-04-19
Publish dockerhub images on CI

ehildenb:
This enables publishing dockerhub images on CI for use in other projects that want to run KEVM on CI.

~Blocked on: https://github.com/runtimeverification/pyk/pull/375~
~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/1749~

- Option `--read-only-kompiled-directory` is added to the `kevm-pyk` kompilation logic, so that we can make installs that are workable when in directories that are not writeable.
- Ubuntu Jammy packaging is brought back. The package is really not suitable for standalone install.
  - The files `package/debian/...` are updated for Jammy.
  - It requires that the user call `poetry install ... kevm-pyk` before trying to use the installed KEVM.
  - It requires that `poetry` be called on invocations to the packaging, which a patch file for the `kevm` runner script is supplied for.
  - The package is not included in cut releases, because it's not suitable for standalone install.
  - The package only includes the Foundry/Haskell builds of KEVM, as the others were not working.
- A Docker image with the package pre-installed is made.
  - It builds the Ubuntu Jammy package, then
  - Installs it in a Docker image, then
  - Sets up Poetry in the appropriate way for the main Docker image users, then
  - Runs a single Foundry test, then
  - Uploads the image to Dockerhub.


ehildenb:
image

ehildenb:
Evidence this works: https://github.com/runtimeverification/evm-semantics/actions/runs/4885005748/jobs/8722144151
https://hub.docker.com/layers/runtimeverificationinc/kevm/ubuntu-jammy-1.0.142/images/sha256-21ecb515feaf1bf371372487ac3d4deeb66c8cbfb37da5a1ee64601e32e6750a?context=repo

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -82,6 +82,8 @@ def exec_kompile(
     includes: list[str],
     main_module: str | None,
     syntax_module: str | None,
+    kevm_lib: Path | None = None,
```
tothtamas28, line 85: Why is this passed to the command?

Makefile
```diff
@@ -170,11 +169,17 @@ plugin-deps: $(plugin_includes) $(plugin_c_includes)
...
+KEVM_PYK_DIR := $(KEVM_LIB)/kevm-pyk
 POETRY       := poetry -C $(KEVM_PYK_DIR)
 POETRY_RUN   := $(POETRY) run --
 
-poetry-env:
+kevm_pyk_files := $(shell find kevm-pyk -type f -name '*.py') \
+                  kevm-pyk/Makefile                           \
+                  kevm-pyk/pyproject.toml                     \
+                  kevm-pyk/poetry.lock                        \
+                  kevm-pyk/.flake8
```
tothtamas28, line 180: For development this is not necessary. For distributing in a Docker image, why not install `kevm-pyk` with `pip`? Then

* The `Makefile` does not become more complicated
* `poetry` does not become a runtime dependency
* This will also be unnecessary:

> It requires that the user call `poetry install ... kevm-pyk` before trying to use the installed KEVM.

kevm-pyk/src/kevm_pyk/__main__.py
ehildenb: Removed.

Makefile
ehildenb: I have switched to doing a user-level install of kevm-pyk in the docker image using pip. I tried to get a global debian install working, to have the standalone package work, but not very hard, because the only deliverable is a docker image, and I figure we can fix that later if needed.

ehildenb:
Evidence this works: https://github.com/runtimeverification/evm-semantics/actions/runs/4941930324/jobs/8835009348?pr=1747

And the deployed (and tested) docker image: https://hub.docker.com/layers/runtimeverificationinc/kevm/ubuntu-jammy-1.0.151/images/sha256-61f8bf2f46f59f581f0358a066f74e42ca18e54869add07d7473dd927541fb82?context=explore

package/docker/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+        git                  \
+        make                 \
+        python3              \
+        python3-pip          \
+        sudo                 \
+        wget
+
+RUN echo '%sudo ALL=(ALL) NOPASSWD:ALL' >> /etc/sudoers
+
+RUN curl -sSL https://install.python-poetry.org | POETRY_HOME=/usr python3 - --version 1.3.2
```
tothtamas28, line 25: This can probably be dropped.

package/docker/Dockerfile
```diff
@@ -0,0 +1,48 @@
...
+    && sudo apt-get install --yes ./kframework_${K_VERSION}_amd64_ubuntu_jammy.deb
+
+ARG KEVM_VERSION
+ADD --chown=user:user ./kevm_${KEVM_VERSION}_amd64.deb ./
+RUN    sudo apt-get update        \
+    && sudo apt-get upgrade --yes \
+    && sudo apt-get install --yes ./kevm_${KEVM_VERSION}_amd64.deb
+
+ADD --chown=user:user ./kevm-pyk /home/user/kevm-pyk
+RUN cd /home/user/kevm-pyk && pip3 install .
```
tothtamas28, line 48: If you call `poetry build` in the job before building the image, you can do something like

```suggestion
COPY --chown=user:user ./kevm-pyk/dist/kevm_pyk-${KEVM_VERSION}-py3-none-any.whl /home/user/
RUN pip3 install /home/user/kevm_pyk-${KEVM_VERSION}-py3-none-any.whl
```

(Not tested.)

tothtamas28 (approved):

package/docker/Dockerfile
ehildenb: Recorded: https://github.com/runtimeverification/evm-semantics/issues/1813 (I don't want to have to go through another round of testing for something that's working now)

package/docker/Dockerfile
ehildenb: Recorded: https://github.com/runtimeverification/evm-semantics/issues/1813 (I don't want to have to go through another round of testing for something that's working now)

PR #1755, opened 2023-04-20
Proof digests for better rebuilding

ehildenb:
This PR introduces digests for the Foundry integration, for the purposes of detecting when rebuilds are necessary.

- The `Contract` class now has a `digest`, which is a hash of the contract json information.
- The `Foundry` class now has a `digest`, which is a hash of all the contracts involved. This is stored on disk.
- When calling `foundry-kompile`, the Foundry hash is inspected, and if it changed then regeneration/rekompilation is automatically triggered.
- The `AGProof` identifier used for storing proofs on disk now includes as well the Contract hash of the proof. This is so that when the contract bytecode changes, the hash will change, and we'll automatically detect that we need to regenerate it. This should help with server-side caching of proofs to (ping @asavienko ).

I've tested locally that the caching works as expected, when the relevant Solidity files are changed, then rekompilation is triggered, and when a non-related contract is changed, we do not re-run the proof.


kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -92,6 +92,21 @@ def contracts(self) -> dict[str, Contract]:
             _contracts[contract_name] = Contract(contract_name, contract_json, foundry=True)
         return _contracts
 
+    @cached_property
+    def digest(self) -> str:
+        sorted_names = sorted(self.contracts.keys())
```
tothtamas28, line 97:
```suggestion
        sorted_names = sorted(self.contracts)
```

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -523,8 +523,9 @@ def exec_foundry_view_kcfg(foundry_root: Path, test: str, **kwargs: Any) -> None
     foundry = Foundry(foundry_root)
     ag_proofs_dir = foundry.out / 'ag_proofs'
     contract_name = test.split('.')[0]
+    proof_digest = f'{test}:{foundry.contracts[contract_name].digest}'
```
tothtamas28, line 526: Consider making this a method on `Foundry`.

tothtamas28: Even better, you can hide all these details in `Foundry` and expose `ag_proof(test: str)`.

kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -166,6 +166,10 @@ def __init__(self, contract_name: str, contract_json: dict, foundry: bool = Fals
                 _fields[_l] = _s
             self.fields = FrozenDict(_fields)
 
+    @cached_property
+    def digest(self) -> str:
+        return hash_str(f'{self.name} - {json.dumps(self.contract_json)}')
```
tothtamas28, line 171:
```suggestion
        return hash_str(f'{self.name} - {json.dumps(self.contract_json, sort_keys=True)}')
```

tothtamas28 (approved):

kevm-pyk/src/kevm_pyk/__main__.py
ehildenb: I've factored out `Foundry.proof_digest` for now, since that was easy and a pre-req. Factoring out `Foundry.proof` is harder because of how the `setUp` functions work, but I've made an issue: https://github.com/runtimeverification/evm-semantics/issues/1769

kevm-pyk/src/kevm_pyk/foundry.py
ehildenb: Done.

kevm-pyk/src/kevm_pyk/solc_to_k.py
ehildenb: Done.

PR #1756, opened 2023-04-21
Lemmas for `#ecrec` and `#writeRange`

lucasmt:
- Add lemmas bounding the size of the `#ecrec` function
- Add the `smtlib` attribute to `#ecrec` in order to make the lemmas `smt-lemmas`
- Add a lemma simplifying the case when the starting index of `#writeRange` is greater than the length of the buffer
- Add claims testing these lemmas


tests/specs/functional/lemmas-spec.k
```diff
@@ -256,4 +260,18 @@ module LEMMAS-SPEC
     claim <k> runLemma ( N &Int maxUInt8   ) => doneLemma ( N ) ... </k> requires #rangeUInt(8,   N)
     claim <k> runLemma ( maxUInt8 &Int N   ) => doneLemma ( N ) ... </k> requires #rangeUInt(8,   N)
 
+ // ecrecover
+ // ---------
+    claim [ecrec-length]: <k> runLemma  ( lengthBytes ( #ecrec ( _ , _ , _ , _ ) ) <=Int 32 )
+                           => doneLemma ( true ) ... </k>
+
+    claim [ecrec-split]: <k> runLemma  ( #range ( #buf(160, 0) +Bytes _BUF , lengthBytes ( #ecrec ( HASH , SIGV , SIGR , SIGS ) ) +Int 128 , 32 -Int lengthBytes ( #ecrec ( HASH , SIGV , SIGR , SIGS ) ) ) )
+                          => doneLemma ( #range ( #buf( 32, 0)             , lengthBytes ( #ecrec ( HASH , SIGV , SIGR , SIGS ) )          , 32 -Int lengthBytes ( #ecrec ( HASH , SIGV , SIGR , SIGS ) ) ) ) ... </k>
```
ehildenb, line 269: I don't quite understand this claim I don't think. It's basically saying, either:

- `#ecrec` is valid (returns an address), and so the overall read of `#range` will return an empty bytes, or
- `#ecrec` is _invalid_ (returns empty), and so the overall read of `#range` will return 32 zeros?

Am I reading this right? Why does such a term show up?

ehildenb:
It looks good, just have one question.

tests/specs/functional/lemmas-spec.k
lucasmt: This appears as part of a larger expression:
```
#ecrec(...) +Bytes #range(#buf(32, 0), lengthBytes(#ecrec(...)), 32 -Int lengthBytes(#ecrec(...)))
```
It seems that this expression is essentially doing `padRightBytes(#ecrec(...), 32, 0)`:
- If `#ecrec` returns 0 bytes, then it will be appended with 32 zeros.
- If `#ecrec` returns 32 bytes, then it will be appended with an empty bytes.

I can add the `#ecrec(...) +Bytes ...` part to the claim to make its context more clear, if you think that would be better.

ehildenb: No, it's fine, thanks for the explanation!

ehildenb (approved):

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -78,6 +78,9 @@ module BYTES-SIMPLIFICATION [symbolic]
     rule #writeRange(#writeRange(MEM, K1, BUF1), K2, BUF2) => #writeRange(MEM, K2, BUF2)
       requires K1 +Int lengthBytes(BUF1) <=Int K2 +Int lengthBytes(BUF2) andBool K2 <=Int K1 [simplification]
 
+    rule #writeRange(MEM, K, BUF) => MEM +Bytes #buf(K -Int lengthBytes(MEM), 0) +Bytes BUF
+      requires lengthBytes(MEM) <=Int K [simplification]
```
PetarMax, line 82: This lemma is in preparation in the Lukso upstreaming PR, perhaps it is better:
```
    rule [writeRange-as-concat-outside]:
      #writeRange(B1, S, B2:Bytes) => #range(B1, 0, S) +Bytes B2
      requires 0 <=Int S andBool lengthBytes(B1) <=Int S +Int lengthBytes(B2)
      [simplification(60)]
 ```

PR #1757, opened 2023-04-21
Integrate BMC prover into `kevm foundry-prove`

### LINKED ISSUE #1727: Apply BMC prover to KEVM
A BMC is introduced here: https://github.com/runtimeverification/pyk/pull/347

First, we need to write 

```
function myloop(uint2656 n) {
  ...
}

function test_myloop_concrete() {
   y = myloop(3);
   assert(p(y));
}

function test_myloop_symbolic(uint256 input) {
  vm.assume(input <= 3);
  y = myloop(input);
  assert(p(y));
}
```

Then run with different `--bmc-depth ...` and get different results.

Then try instantiating the BMC prover with this, using this predicate for detecting same loop head: https://github.com/runtimeverification/ksummarize/blob/master/ksummarize/src/ksummarize/evm_strategy.py#L156

geo2a:
Fixes #1727

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -490,6 +490,7 @@ test-foundry-%: KEVM_OPTS += --pyk --verbose
 test-foundry-%: KEVM := $(POETRY_RUN) kevm
 test-foundry-kompile: tests/foundry/foundry.k.check
 test-foundry-prove: tests/foundry/out/kompiled/foundry.k.prove
+test-foundry-bmc-prove: tests/foundry/out/kompiled/foundry.k.bmc-prove
 test-foundry-list: tests/foundry/foundry-list.check
 
 foundry-forge-build: $(foundry_out)
@@ -519,6 +520,13 @@ tests/foundry/out/kompiled/foundry.k.prove: tests/foundry/out/kompiled/timestamp
 	    -j$(FOUNDRY_PAR) --no-simplify-init --max-depth 1000             \
 	    $(KEVM_OPTS) $(KPROVE_OPTS)                                      \
 	    $(addprefix --exclude-test , $(shell cat tests/foundry/exclude))
+
+tests/foundry/out/kompiled/foundry.k.bmc-prove: tests/foundry/out/kompiled/timestamp
+	$(KEVM) foundry-prove --foundry-project-root $(foundry_dir)          \
+	    -j$(FOUNDRY_PAR) --no-simplify-init --max-depth 1000             \
+            --bmc-depth 3                                                    \
+	    $(KEVM_OPTS) $(KPROVE_OPTS)                                      \
+	    $(addprefix --test , $(shell cat tests/foundry/bmc-tests))
 
 foundry_golden := tests/foundry/golden
 foundry_diff_tests := $(shell cat tests/foundry/checkoutput)
--- a/kevm-pyk/src/kevm_pyk/foundry.py
+++ b/kevm-pyk/src/kevm_pyk/foundry.py
@@ -21,7 +21,7 @@ from pyk.prelude.k import GENERATED_TOP_CELL
 from pyk.prelude.kbool import FALSE, notBool
 from pyk.prelude.kint import INT, intToken
 from pyk.prelude.ml import mlEqualsTrue
-from pyk.proof import AGProof
+from pyk.proof import AGProof, AGBMCProof
 from pyk.utils import shorten_hashes, single, unique
 
 from .kevm import KEVM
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -490,6 +490,7 @@ test-foundry-%: KEVM_OPTS += --pyk --verbose
 test-foundry-%: KEVM := $(POETRY_RUN) kevm
 test-foundry-kompile: tests/foundry/foundry.k.check
 test-foundry-prove: tests/foundry/out/kompiled/foundry.k.prove
+test-foundry-bmc-prove: tests/foundry/out/kompiled/foundry.k.bmc-prove
```
ehildenb, line 493: This needs to be added to `.PHONY` list.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -355,7 +356,7 @@ def foundry_prove(
             setup_methods[contract.name] = f'{contract.name}.{method.name}'
 
     def run_cfg_group(tests: list[str]) -> dict[str, bool]:
-        ag_proofs: dict[str, AGProof] = {}
+        ag_proofs: dict[str, AGProof | AGBMCProof] = {}
         for test in tests:
             if AGProof.proof_exists(test, ag_proofs_dir) and not reinit:
                 ag_proof = AGProof.read_proof(test, ag_proofs_dir)
```
ehildenb, line 362: what if the BMC proof already exists on disk? We should probably be cehcking "does the file exist on disk?", and then reading in the json, and inspecting the `type` field to determine if we should read a `AGProof` or `AGBMCProof`.

kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -267,6 +267,17 @@ def is_terminal(cterm: CTerm) -> bool:
...
+    def same_loop(cterm1: CTerm, cterm2: CTerm) -> bool:
+        branch1 = list(KEVM.extract_branches(cterm1))
+        branch2 = list(KEVM.extract_branches(cterm2))
+        same_branch = len(branch1) > 0 and len(branch2) > 0 and branch1[0] == branch2[0]
+        same_cell_structure = all(
+            cterm1.cell(cn) == cterm2.cell(cn) for cn in ['PC_CELL', 'CALLDATA_CELL', 'PROGRAM_CELL', 'JUMPDESTS_CELL']
+        )
+        same_wordstack_structure = KEVM.wordstack_len(cterm1) == KEVM.wordstack_len(cterm2)
+        return same_branch and same_cell_structure and same_wordstack_structure
+
```
ehildenb, line 280: This probably isn't doing what you expect, and will fail for any loops that are not very simple.

Here: https://github.com/runtimeverification/ksummarize/blob/master/ksummarize/src/ksummarize/evm_strategy.py#L156, we are using method `_extract_branch`, which returns a tuple of `pcount, cond` (program counter and condition) for the _pre-branch_ node, but you are using `KEVM.extract_branches` (which is implemented in the above link as `extract_branch_condition`, which returns the actual branch conditions.

So in the original, we are comparing that the program counter being jumped to (the first argument of `JUMPI`) is the same in both, but here you are comparing that the branching condition itself is exactly the same (which is likely not to be the case for anything but trivial loops).

Likely, what you need to do, is something more like:

```suggestion
    @staticmethod
    def same_loop(cterm1: CTerm, cterm2: CTerm) -> bool:
        # In the same program, at the same calldepth, at the same program counter
        for cell in ['PC_CELL', 'CALLDEPTH_CELL', 'PROGRAM_CELL']:
            if cterm1.cell(cell) != cterm2.cell(cell):
                return False
        # duplicate from KEVM.extract_branches
        jumpi_pattern = KEVM.jumpi_applied(KVariable('###PCOUNT'), KVariable('###COND'))
        pc_next_pattern = KApply('#pc[_]_EVM_InternalOp_OpCode', [KEVM.jumpi()])
        branch_pattern = KSequence([jumpi_pattern, pc_next_pattern, KEVM.sharp_execute(), KVariable('###CONTINUATION')])
        subst1 = branch_pattern.match(cterm1.cell('K_CELL'))
        subst2 = branch_pattern.match(cterm2.cell('K_CELL'))
        # Jumping to the same program counter
        if subst1 is not None and subst2 is not None and subst1['###PCOUNT'] == subst2['###PCOUNT']:
            if KEVM.wordstack_len(cterm1.cell('WORDSTACK_CELL')) == KEVM.wordstack_len(cterm2.cell('WORDSTACK_CELL')):
                return True
        return False
```

Note that above I've assumed that we pass a `KInner` into `KEVM.wordstack_len`.

ehildenb:
@geo2a you'll also have to update the various expected output check files (since you modified the Solidity source):

```
make tests/foundry/foundry.k.check CHECK=cp
make test-foundry-prove -j4 FOUNDRY_PAR=12 # takes a long time, ~1 hour
make tests/foundry/foundry-list.check CHECK=cp
make test-foundry-kcfg-diff -j4 FOUNDRY_PAR=12 CHECK=cp # takes a long time, ~15 minutes
```

ehildenb (approved):

PR #1761, opened 2023-04-22
Uniform lemmas for minInt and maxInt

PetarMax:
This PR introduces a set of uniform lemmas for the `minInt` and `maxInt` functions.


include/kframework/lemmas/int-simplification.k
```diff
@@ -115,13 +115,27 @@ module INT-SIMPLIFICATION-COMMON
...
+    rule [minInt-factor-left]:  minInt ( A:Int +Int B:Int, A:Int +Int C:Int ) => A +Int minInt ( B, C ) [simplification]
+    rule [minInt-factor-right]: minInt ( A:Int +Int B:Int, C:Int +Int B:Int ) => minInt ( A, C ) +Int B [simplification]
+
+    rule [maxint-left]:  maxInt(A:Int, B:Int) => B requires A <=Int B [simplification]
+    rule [maxint-right]: maxInt(A:Int, B:Int) => A requires A >=Int B [simplification]
+
+    rule [maxint-lt]:  maxInt(A:Int, B:Int)  <Int C:Int => A  <Int C andBool B  <Int C [simplification]
+    rule [maxint-leq]: maxInt(A:Int, B:Int) <=Int C:Int => A <=Int C andBool B <=Int C [simplification]
+    rule [maxint-gt]:  A:Int  <Int maxInt(B:Int, C:Int) => A  <Int B  orBool A  <Int C [simplification]
+    rule [maxint-geq]: A:Int <=Int maxInt(B:Int, C:Int) => A <=Int B  orBool A <=Int C [simplification]
```
ehildenb, line 135: Do we need these lemmas, or can we do with ones that go straight to `true` and `false`? In general, would like to retain as much structure of the original problem as possible. As long as one of the sides of each disjunct/conjunct is decidable as `true/false`, then these lemmas should be simplifiable, and will lead to terms that are stricly smaller. The LHS here has 2 AST nodes, and the RHS has 3, so these are expanding, and they lose the information that the disequalities came froma maxInt/minInt.

So could we instead do the splitting for each, and have only lemmas that go to `true/false`?

include/kframework/lemmas/int-simplification.k
```diff
@@ -115,13 +115,27 @@ module INT-SIMPLIFICATION-COMMON
...
+    rule [minint-left]:  minInt(A, B) => A requires A <=Int B [simplification]
+    rule [minint-right]: minInt(A, B) => B requires B <=Int A [simplification]
 
-    rule [minint-lt]:  minInt(A, B) <Int C  => true requires A  <Int C  orBool B  <Int C [simplification]
-    rule [minint-gt]:  A  <Int minInt(B, C) => true requires A  <Int B andBool A  <Int C [simplification]
-    rule [minint-geq]: A <=Int minInt(B, C) => true requires A <=Int B andBool A <=Int C [simplification]
+    rule [minint-lt]:  minInt(A, B)  <Int C => A  <Int C  orBool B  <Int C [simplification]
+    rule [minint-leq]: minInt(A, B) <=Int C => A <=Int C  orBool B <=Int C [simplification]
+    rule [minint-gt]:  A  <Int minInt(B, C) => A  <Int B andBool A  <Int C [simplification]
+    rule [minint-geq]: A <=Int minInt(B, C) => A <=Int B andBool A <=Int C [simplification]
```
ehildenb, line 124: Same here, can we split each and have lemmas that only go to `true/false` (or directly to one of the branches)?

In the case that the resulting RHS is simplifiable, this expression will go directl yto `true/false`, or to one of the branches anyway. So having it go directly there happens in 1 rewrite instead of 2. In the case that the RHS here is not simplifiable further, now we have an expression with lower structure, and we don't know that these came from a maxInt/minInt.

ehildenb:
Do we have any tests for these? If not, can we write some tests for them? In a new file `tests/specs/functional/int-simplification.k`.

include/kframework/lemmas/int-simplification.k
PetarMax: Are you sure that information about where the inequalities came from would be lost? With `true/false`, we'd log the application of the simplification, but we'd do the same for these ones, so this information should ultimately be tracked, no?

When it comes to the AST structure, I do understand the point about expansion. My reasoning is that we are unconditionally eliminating a function symbol and moving to integer comparison, and so, for me, having the RHS is simpler. With `true/false`, my intuition is that we would have more conditional simplifications that would be attempted more times, which might slow the execution down.

include/kframework/lemmas/int-simplification.k
PetarMax: Same as above.

PetarMax:
> Do we have any tests for these? If not, can we write some tests for them? In a new file `tests/specs/functional/int-simplification.k`. We don't necessarily need to test them all, just a few tests and a testing harness for later tests to be added.

Will look into that.

include/kframework/lemmas/int-simplification.k
ehildenb: Sure, it's in the log, but later states don't have it, and so anything relying on `maxInt` being there won't have it, so it's no longer in the current path condition.

Same reasoning as below: either the RHS here will simplify directly down to `true/false` or one of the branches of the conjunct/disjunct _anyway_, or you'll be left with a boolean expression which gets pulled apart that came from something with higher structure (in which case, I don't really see the benefit of the boolean expression).

Using unconditional rules is good though, that does help the booster backend. But I suspect these particular conditional rules won't be too challenging for it.

ehildenb: Let's go with your solution for now, sounds fine. I guess either way time will tell!

PetarMax:
> > Do we have any tests for these? If not, can we write some tests for them? In a new file `tests/specs/functional/int-simplification.k`. We don't necessarily need to test them all, just a few tests and a testing harness for later tests to be added.
> 
> Will look into that.

Ah, I just pushed but forgot the new file bit - should we separate the other claims into their separate files, for example the 64 bitwise-and ones?

Also, I am not sure why the CI is not passing here but passed on the up/Int lemmas...

ehildenb:
I'd like these lemmas separated into their own file, so that we can start an `int-simplification-spec.k` file with the goal of eventually merging that into upstream K for use in all semantics.

You'll need to add your new file to this list: https://github.com/runtimeverification/evm-semantics/blob/79adbc19a1330680728f1a76593ab273aea3cc40/Makefile#L432, and probably to this list: https://github.com/runtimeverification/evm-semantics/blob/master/tests/failing-symbolic.pyk

PetarMax:
> I'd like these lemmas separated into their own file, so that we can start an `int-simplification-spec.k` file with the goal of eventually merging that into upstream K for use in all semantics.
> 
> You'll need to add your new file to this list:
> 
> https://github.com/runtimeverification/evm-semantics/blob/79adbc19a1330680728f1a76593ab273aea3cc40/Makefile#L432
> , and probably to this list:
> 
> https://github.com/runtimeverification/evm-semantics/blob/79adbc19a1330680728f1a76593ab273aea3cc40/tests/failing-symbolic.pyk#L1

Should be done with the latest commit.

ehildenb:
Tests look good. I see that you cancelled the test run, so I'll wait for the tests to go green to mark approved.

PetarMax:
> Tests look good. I see that you cancelled the test run, so I'll wait for the tests to go green to mark approved.

Yes, Nix (macos-12) failed randomly, so I'm manually re-running them...

ehildenb (approved):

include/kframework/lemmas/int-simplification.k
```diff
@@ -115,13 +115,27 @@ module INT-SIMPLIFICATION-COMMON
...
 
-    rule [maxint-geq-left]: A <=Int maxInt(B, C) => true requires A <=Int B orBool A <=Int C [simplification]
+    rule [minInt-factor-left]:  minInt ( A:Int +Int B:Int, A:Int +Int C:Int ) => A +Int minInt ( B, C ) [simplification]
+    rule [minInt-factor-right]: minInt ( A:Int +Int B:Int, C:Int +Int B:Int ) => minInt ( A, C ) +Int B [simplification]
+
+    rule [maxint-left]:  maxInt(A:Int, B:Int) => B requires A <=Int B [simplification]
+    rule [maxint-right]: maxInt(A:Int, B:Int) => A requires A >=Int B [simplification]
```
ehildenb, line 130:
```suggestion
    rule [maxint-right]: maxInt(A:Int, B:Int) => A requires B <=Int A [simplification]
```

ehildenb: Also note that these lemmas overlap on the case that `A` and `B` are equal. Probably not a problem since they are `simplification` lemmas, but for example ,if they were not simplification lemmas, then the backend would refuse to apply either if both applied.

ehildenb:
One final comment for style, feel free ta add automerge after addressing.

include/kframework/lemmas/int-simplification.k
PetarMax: Wouldn't the rules then be rewrite rules, and any applicable would apply?

PR #1762, opened 2023-04-22
up/Int lemmas

PetarMax:
This PR introduces several lemmas for `up/Int`.


ehildenb:
Can we add tests for these? They can go in `tests/specs/functional/lemmas-spec.k`. Just some handwritten tests for `up/Int` is fine.

PetarMax:
> Can we add tests for these? They can go in `tests/specs/functional/lemmas-spec.k`. Just some handwritten tests for `up/Int` is fine.

Yes, will do.

PetarMax:
> > Can we add tests for these? They can go in `tests/specs/functional/lemmas-spec.k`. Just some handwritten tests for `up/Int` is fine.
> 
> Yes, will do.

Added tests now.

tests/specs/functional/lemmas-spec.k
```diff
@@ -274,4 +274,19 @@ module LEMMAS-SPEC
...
+                           => doneLemma ( true ) ... </k>
+                          requires 0 <Int Y andBool Y <=Int Z
+
+    claim [upInt-02]: <k> runLemma ( lengthBytes(B) +Int X -Int Y <Int notMaxUInt5 &Int lengthBytes(B) +Int 31 )
+                           => doneLemma ( true ) ... </k>
+                          requires Y -Int X >Int 0
+
+    claim [upInt-03]: <k> runLemma ( lengthBytes(B) +Int X -Int Y >Int notMaxUInt5 &Int lengthBytes(B) +Int 31 )
+                           => doneLemma ( true ) ... </k>
+                          requires Y -Int X <Int -64
```
ehildenb, line 290: I don't see `up/Int` used in these claims.... Should they perhaps go somewhere else? More tests is always better, just not sure how these ones are relevant here.

include/kframework/lemmas/int-simplification.k
```diff
@@ -158,4 +161,26 @@ module INT-SIMPLIFICATION-COMMON
...
+      requires 0 <Int Y
+      [simplification]
+
+    rule [lengthBytes-upInt-32-lower-bound]:
+      lengthBytes(X) <=Int notMaxUInt5 &Int lengthBytes(X) +Int 31 => true
+      [simplification, smt-lemma]
+
+    rule [lengthBytes-upInt-32-upper-bound]:
+      lengthBytes(X) +Int 32 >Int notMaxUInt5 &Int lengthBytes(X) +Int 31 => true
+      [simplification, smt-lemma]
```
ehildenb, line 184: Not sure these lemmas belong here, and they seem pretty specific. Could we replace `lengthBytes(X)` with just a variable `Y #And Y >=Int 0`, and then these lemmas just become about comparisons and bitwise operations (so moved to `bitwise-simplification.k`)?

tests/specs/functional/lemmas-spec.k
PetarMax: `notMaxUInt5 &Int lengthBytes(X) +Int 31` is `lengthBytes(X) up/Int 32`, I think?

include/kframework/lemmas/int-simplification.k
PetarMax: So, `notMaxUInt5 &Int lengthBytes(X) +Int 31` is `lengthBytes(X) up/Int 32`, that's why I put them here instead of there, and these have appeared in that form in Optimism when copying some data in memory, so they are likely to reappear. 

We could also make the `up/Int` explicit with a simplification and then reason about it instead? Possibly more generally, also, if the second operand is a power of two...

PetarMax:
I've revisited the lemmas and claims. I'm a bit surprised I couldn't use 
```
    rule [upInt-lt-false]:
      ((X up/Int Y) *Int Y) <Int X => false
      requires 0 <Int Y
      [simplification]
```
to prove
```
    claim [upInt-02]: <k> runLemma ( notMaxUInt5 &Int lengthBytes(B) +Int 31 >=Int lengthBytes(B) )
                           => doneLemma ( true ) ... </k>
```
but had to explicitly write the negation.

Also, if any of the simplifications are marked as `smt-lemma`s, Z3 dies because it doesn't know that the overall set of smt-lemmas is consistent.

PetarMax:
Actually, I don't like removing the lemmas about `notMaxUInt5 &Int lengthBytes(X) +Int 31` and converting it to `up/Int`, because I was able to have two `smt-lemmas` on it, and I can't have them on the `up/Int` because z3 breaks. Optimism proofs are likely to break with this change, and it's not clear to me how well future proofs would work.

PetarMax:
I'm not sure how to write good tests for these lemmas, if they were SMT-lemmas I could just put some other operations around them, add something here and there, but like this, I can't. I could just instantiate them with concrete values, but those will obviously pass. Any advice? @ehildenb

include/kframework/lemmas/int-simplification.k
```diff
@@ -158,4 +161,48 @@ module INT-SIMPLIFICATION-COMMON
...
+
+    rule [upInt-lt-false]:
+      ((X up/Int Y) *Int Y) <Int Z => false
+      requires Z <=Int X andBool 0 <Int Y
+      [simplification]
+
+    rule [upInt-refl-lt]:
+      ((X up/Int Y) *Int Y) <Int X => false
+      requires 0 <Int Y
+      [simplification]
```
lucasmt, line 181: Isn't this lemma subsumed by the previous one?

include/kframework/lemmas/int-simplification.k
```diff
@@ -158,4 +161,48 @@ module INT-SIMPLIFICATION-COMMON
...
+
+    rule [upInt-refl-gt]:
+      X <Int ((X up/Int Y) *Int Y) => X modInt Y =/=Int 0
+      requires 0 <Int Y
+      [simplification]
+
+    rule [upInt-refl-geq]:
+      X <=Int ((X up/Int Y) *Int Y) => true
+      requires 0 <Int Y
+      [simplification]
```
lucasmt, line 196: I assume that this is one of the lemmas that breaks z3 if it's made an `smt-lemma`, right? Because if we had this as an `smt-lemma` and also another that said `((X up/Int Y) *Int Y) <Int X + Y => true`, then we wouldn't need `upInt-lt-true` and `upInt-lt-false`

include/kframework/lemmas/int-simplification.k
PetarMax: It is, I left it there for completeness. Shall I remove?

include/kframework/lemmas/int-simplification.k
PetarMax: Yes, I think that's correct, we wouldn't need those. 

Any lemma with `up/Int` will break Z3 if `smt-lemma`' is added. I think this is because Z3 has no idea what `up/Int` actually is and then when faced with constraints it just goes 'I don't know!'.

lucasmt: I'm looking at the definition of `up/Int` in [`evm-types.md`](https://github.com/runtimeverification/evm-semantics/blob/master/include/kframework/evm-types.md) and it's defined with an `smtlib` attribute, which should give it uninterpreted function semantics. I don't get why this only happens to `up/Int` and not any of the other functions with the `smtlib` attribute. Does @ehildenb know anything about this?

include/kframework/lemmas/int-simplification.k
lucasmt: Maybe turn it into a claim?

include/kframework/lemmas/int-simplification.k
PetarMax: It doesn't only happen to `up/Int`, I can also break it by requiring `smt-lemma` on the facts that `&Int` cannot be greater than its operands...

PetarMax: It might just be the limit of Z3 reasoning that we're hitting...

include/kframework/lemmas/int-simplification.k
PetarMax: Yes, that is better.

include/kframework/lemmas/int-simplification.k
ehildenb: We could try alternative Z3 theories. For example, add the `smt-lemma`, then get hte Z3 transcript, then at the top it specifies which Z3 theories it's reasoning in. We may be able to just change the theory support it has and suddenly get better results.

include/kframework/lemmas/int-simplification.k
```diff
@@ -158,4 +161,43 @@ module INT-SIMPLIFICATION-COMMON
...
+
+    rule [upInt-refl-leq]:
+      ((X up/Int Y) *Int Y) <=Int X => X modInt Y ==Int 0
+      requires 0 <Int Y
+      [simplification]
+
+    rule [upInt-refl-gt]:
+      X <Int ((X up/Int Y) *Int Y) => X modInt Y =/=Int 0
+      requires 0 <Int Y
```
ehildenb, line 185: We should really pull these out into their own module, `EVM-WORD-SIMPLIFICATION` perhaps.

That's because `up/Int` is not part of the prelude, and I intend for `INT-SIMPLIFICATION` to eventually be upstreamed to the K prelude so that all K definitions benefit from it.

include/kframework/lemmas/int-simplification.k
PetarMax: The transcript I got doesn't have the theories, it starts with:
```
(set-option :print-success true )
; success

(set-option :timeout 40 )
; success

(declare-sort |HB_#String| 0 )
; success

(declare-sort |HB_SortAccessedAccountsCellOpt| 0 )
; success

(declare-sort |HB_SortAccessedStorageCellOpt| 0 )
; success

...
```
and then just continues with the declarations and assertions...

ehildenb: Hmmmm, that probably means we're just using Z3's default theory, which is probably not very useful. I wonder if we could try some different Z3 theories on that transcript you have?

Here is an example of setting a different logic: https://microsoft.github.io/z3guide/docs/theories/Arithmetic/#difference-arithmetic, and here are a bunch of supported logics: https://stackoverflow.com/questions/23450572/which-logics-are-supported-by-z3

These logics should really be called "theories", but whatever.

Makefile
```diff
@@ -54,7 +54,7 @@ export PLUGIN_FULL_PATH
         test-prove test-failing-prove test-foundry-kcfg-diff                                                                                 \
         test-prove-benchmarks test-prove-functional test-prove-opcodes test-prove-erc20 test-prove-bihu test-prove-examples test-prove-smoke \
         test-prove-mcd test-klab-prove                                                                                                       \
-        test-parse test-failure test-foundry-kompile test-foundry-prove test-foundry-bmc-prove test-foundry-list                             \
+        test-parse test-failure test-foundry-kompile test-foundry-prove test-foundry-list                                                    \
         test-interactive test-interactive-help test-interactive-run test-interactive-prove test-interactive-search                           \
```
ehildenb, line 58: Why was this removed?

Makefile
```diff
@@ -491,7 +493,6 @@ test-foundry-%: KEVM_OPTS += --pyk --verbose
 test-foundry-%: KEVM := $(POETRY_RUN) kevm
 test-foundry-kompile: tests/foundry/foundry.k.check
 test-foundry-prove: tests/foundry/out/kompiled/foundry.k.prove
-test-foundry-bmc-prove: tests/foundry/out/kompiled/foundry.k.bmc-prove
 test-foundry-list: tests/foundry/foundry-list.check
```
ehildenb, line 496: Why deleted?

Makefile
```diff
@@ -522,13 +523,6 @@ tests/foundry/out/kompiled/foundry.k.prove: tests/foundry/out/kompiled/timestamp
...
 	    $(addprefix --exclude-test , $(shell cat tests/foundry/exclude))
 
-tests/foundry/out/kompiled/foundry.k.bmc-prove: tests/foundry/out/kompiled/timestamp
-	$(KEVM) foundry-prove --foundry-project-root $(foundry_dir)          \
-	    -j$(FOUNDRY_PAR) --no-simplify-init --max-depth 1000             \
-            --bmc-depth 3                                                    \
-	    $(KEVM_OPTS) $(KPROVE_OPTS)                                      \
-	    $(addprefix --test , $(shell cat tests/foundry/bmc-tests))
-
 foundry_golden := tests/foundry/golden
```
ehildenb, line 526: Why deleted?

include/kframework/lemmas/evm-int-simplification.k
```diff
@@ -0,0 +1,55 @@
...
+
+    rule [upInt-refl-leq]:
+      ((X up/Int Y) *Int Y) <=Int X => X modInt Y ==Int 0
+      requires 0 <Int Y
+      [simplification]
+
+    rule [upInt-refl-gt]:
+      X <Int ((X up/Int Y) *Int Y) => X modInt Y =/=Int 0
+      requires 0 <Int Y
+      [simplification]
```
ehildenb, line 38: I still don't really like these lemmas. Are we sure they're needed?

Could we also replace them with lemmas like this:

```
rule ((X up/Int Y) *Int Y) => X +Int (X modInt Y) requires 0 <Int Y andBool 0 <=Int X [simplification]
```

This I feel expresses the semantics of `_up/Int_` better, while being more general (so that we can have general lemmas about `_<=Int_` and `_<Int_` without `_up/Int_`).

PetarMax: I'm not sure that holds... take, for example: `Y ==Int 5` and `X ==Int 6, 7, 8, 9`. We have that `(X up/Int Y) *Int Y ==Int 10` for all of them, but `X +Int (X modInt Y) ==Int 7, 9, 11, 13`, right?

By definition, I think `X up/Int Y ==Int (X +Int Y -Int 1) /Int Y`?

The lemmas above aren't necessarily needed, I just have all six `<=, <, >, >=, ==, =/=` for completeness. I think @lucasmt ran into the `<Int`  version only?

ehildenb (approved):

include/kframework/lemmas/evm-int-simplification.k
ehildenb: Yes you're right, not sure what I was thinking. I would prefer lemma which directly represented the semantics of `up/Int` better, but I guess we will have to live with ones that define it relative to eqaulity instead.

PR #1763, opened 2023-04-22
lengthBytes lemmas

PetarMax:
This PR adds several lemmas concerning `#lengthBytes`.


PetarMax:
Will add tests as well as soon as the simplifications are assessed to be ok.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -83,11 +83,15 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+    rule [lengthBytes-geq-zero]: 0 <=Int lengthBytes ( _ )                          => true                                                [simplification, smt-lemma]
+    rule [lengthBytes-concat]:   lengthBytes(BUF1 +Bytes BUF2)                      => lengthBytes(BUF1) +Int lengthBytes(BUF2)            [simplification]
+    rule [lengthBytes-buf]:      lengthBytes(#buf(SIZE, _))                         => SIZE   requires 0 <=Int SIZE                        [simplification]
+    rule [lengthBytes-range]:    lengthBytes(#range(_, START, WIDTH))               => WIDTH  requires 0 <=Int START andBool 0 <=Int WIDTH [simplification]
+    rule [lengthBytes-prtw]:     lengthBytes(#padRightToWidth(W:Int, B:Bytes))      => maxInt(lengthBytes(B), W)                           [simplification]
+    rule [lengthBytes-padleft]:  lengthBytes(padLeftBytes(B:Bytes, W:Int, _:Int) )  => maxInt(lengthBytes(B), W)                           [simplification]
+    rule [lengthBytes-padright]: lengthBytes(padRightBytes(B:Bytes, W:Int, _:Int) ) => maxInt(lengthBytes(B), W)                           [simplification]
```
ehildenb, line 92: I don't think we actually have this guarantee of `padLeftbytes` and `padRightBytes` behaving this way on negative `W`, I'm pretty sure they're actually undefined. In `domains.md`, they're marked as partial, and implemented with hooks: https://kframework.org/k-distribution/include/kframework/builtin/domains/#bytes-padding

That's part of why `#padToWidth` and `#padRightToWidth` were introduced, so we could enforce totality on them. So we really should not be writing lemmas about these hooked functions ever. Either all the arguments to `#padToWidth` and `#padRightToWdith` are `concrete`, in which case they get lowered to the hooked functions and evaluated fully, or they are symbolic in which case we write lemmas directly about the higher-level constructs.

We may have made a mistake and let slip through lemmas which use `padRightBytes` and `padLeftBytes` through PR review before (on the RHS or requires clause of a lemma or claim), and that should be reversed. We should not be reasoning about these constructs directly, as it's not safe because their implementation is opaque (handled by backend) and not so much can be guaranteed about it.

ehildenb:
As a simple test, you can make a K definition which calls `padRightBytes` on some user input, and call `krun` on it. Then supply it with a negative number, and see what the different backends do. My guess is they both crash.

include/kframework/lemmas/bytes-simplification.k
PetarMax: You are correct, we don't have it:
```
rule #Ceil(padLeftBytes(_, LEN, VAL))  => {(0 <=Int LEN andBool 0 <=Int VAL andBool VAL <Int 256) #Equals true} [simplification]
```

But then, does that mean that if I write `padLeftBytes(B:Bytes, -4, 300)`, the backend first adds the `#Ceil` condition, which collapses into `#Bottom` and the claim vacuously holds?

Or if I have `padLeftBytes(B, W, BYTE)`, then the `#Ceil` condition automatically adds the constraints that we need?

ehildenb: I think the bottom line is we should not rely on definedness constraints at all, and should just avoid anything that is potentially problematic. I would treat the behavior of the backend on these cases as "undefined" besacially, but not in the matching-logic sense, but in the "implementations can do whatever they want" sense. Better to just avoid having to reason about them altogether.

ehildenb: The other reason to avoid touching definedness reasoning is because the new haskell backend basically throws away any part of the definition that does not obviously "preserve definedness", which means that the backend will throw away your rules for reasoning. This is to speed up reasoning (dramatically), because checking definedness every step is too expensive.

PetarMax: Ok, understood. That last bit might end being problematic for verification, because I had to have explicit inversion rules using `#Ceil` for KPlutus, and Lucas is likely to also need them right now in his engagement.

So then, there are two paths I see:

1. remove those simplifications altogether; or
2. add the definedness constraints explicitly to the `requires`.

I am leaning towards 2. What are your thoughts?

PetarMax: > I don't think we actually have this guarantee of `padLeftbytes` and `padRightBytes` behaving this way on negative `W`, I'm pretty sure they're actually undefined. In `domains.md`, they're marked as partial, and implemented with hooks: https://kframework.org/k-distribution/include/kframework/builtin/domains/#bytes-padding
> 
> That's part of why `#padToWidth` and `#padRightToWidth` were introduced, so we could enforce totality on them. So we really should not be writing lemmas about these hooked functions ever. Either all the arguments to `#padToWidth` and `#padRightToWdith` are `concrete`, in which case they get lowered to the hooked functions and evaluated fully, or they are symbolic in which case we write lemmas directly about the higher-level constructs.
> 
> We may have made a mistake and let slip through lemmas which use `padRightBytes` and `padLeftBytes` through PR review before (on the RHS or requires clause of a lemma or claim), and that should be reversed. We should not be reasoning about these constructs directly, as it's not safe because their implementation is opaque (handled by backend) and not so much can be guaranteed about it.

There are no lemmas that I see that use `padLeftBytes` or `padRightBytes` on `master`, at least.

ehildenb: I would lean towards 1 for now. What are examples where we need those simplifications? If we see those examples, we can understand where they are coming from, and avoid generating them altogether.

PetarMax: Ok, then they go away.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -83,11 +83,13 @@ module BYTES-SIMPLIFICATION [symbolic]
...
-    rule [lengthBytes-range]:        lengthBytes(#range(_, START, WIDTH))      => WIDTH  requires 0 <=Int START andBool 0 <=Int WIDTH [simplification]
-    rule [lengthBytes-prtw-leq-len]: lengthBytes(#padRightToWidth(WIDTH, BUF)) => WIDTH  requires lengthBytes(BUF) <=Int WIDTH        [simplification]
+    rule [lengthBytes-geq-zero]: 0 <=Int lengthBytes ( _ )                          => true                                                [simplification, smt-lemma]
+    rule [lengthBytes-concat]:   lengthBytes(BUF1 +Bytes BUF2)                      => lengthBytes(BUF1) +Int lengthBytes(BUF2)            [simplification]
+    rule [lengthBytes-buf]:      lengthBytes(#buf(SIZE, _))                         => SIZE   requires 0 <=Int SIZE                        [simplification]
+    rule [lengthBytes-range]:    lengthBytes(#range(_, START, WIDTH))               => WIDTH  requires 0 <=Int START andBool 0 <=Int WIDTH [simplification]
+    rule [lengthBytes-prtw]:     lengthBytes(#padRightToWidth(W:Int, B:Bytes))      => maxInt(lengthBytes(B), W)                           [simplification]
```
ehildenb, line 90:
```suggestion
    rule [lengthBytes-geq-zero]: 0 <=Int lengthBytes ( _ )                     => true                                                [simplification, smt-lemma]
    rule [lengthBytes-concat]:   lengthBytes(BUF1 +Bytes BUF2)                 => lengthBytes(BUF1) +Int lengthBytes(BUF2)            [simplification]
    rule [lengthBytes-buf]:      lengthBytes(#buf(SIZE, _))                    => SIZE   requires 0 <=Int SIZE                        [simplification]
    rule [lengthBytes-range]:    lengthBytes(#range(_, START, WIDTH))          => WIDTH  requires 0 <=Int START andBool 0 <=Int WIDTH [simplification]
    rule [lengthBytes-prtw]:     lengthBytes(#padRightToWidth(W:Int, B:Bytes)) => maxInt(lengthBytes(B), W)                           [simplification]
```

ehildenb (approved):

ehildenb:
One formatting change, then you can add `automerge`.

Issue #1765, opened 2023-04-22
Generating addrFromPrivateKey lemmas for missing ecdsaPubKey hook


ehildenb:
Because of ECDSAPubKey not being implemented on concrete inputs for the Haskell backend, we cannot rely on it being present when writing Foundry property tests (see: https://github.com/runtimeverification/haskell-backend/issues/3573).

But it's not uncommon to supply a random private key in a property test, then use `vm.addr(...)` cheatcode to turn said private key into an ethereum address that can be used. This pattern is especially used with contracts where signatures are checked onchain.

To get around this https://github.com/runtimeverification/evm-semantics/pull/1766 allows us to write lemmas directly for the addresses needed in the tests. Given a private key `KEY` (encoded in hex, for example: `0x2b8ba85ff7f2c7fa62cddb20ae40ef2553b9fdbff086f60d1647ecaab15af867`), and the associated Etherem address `ADDR` (encoded as a decimal integer, for example: `1204788665628904891128108629487072145087352593841`), you can add the following K lemma to your `lemmas.k` for a project to help the Haskell backend reason about that case:

```
rule #addrFromPrivateKey(KEY) => ADDR [priority(40)]
```

So, given `KEY`, how do you figure out the correct `ADDR`?

First, install `ganache-cli`:

```
npm install -g ganache-cli
```

Then run, in one terminal, `ganache-cli`:

```
ganache-cli
```

Wait for it to start up, then in a file called `ganache.in`, put the following contents:

```
{
  "jsonrpc": "2.0",
  "id": 1,
  "method": "personal_importRawKey",
  "params": ["KEY", "0x"]
}
```

Then, in another terminal, run the following command to send this request to ganache-cli:

```
curl -s -X POST 'http://127.0.0.1:8545' --data @ganache.in -H "Content-Type: application/json"
```

It will output a response that looks like this:

```
{"id":1,"jsonrpc":"2.0","result":"HEX_ADDR"}
```

Take the `HEX_ADDR`, and put it in this website to convert it to decimal: https://www.rapidtables.com/convert/number/hex-to-decimal.html, which gives you `ADDR`, so that you can make your lemma!

palinatolmach:
Merged this issue with the one that describes the missing `ECDSA hook` problem https://github.com/runtimeverification/haskell-backend/issues/3573.

palinatolmach:
[...]

PR #1766, opened 2023-04-22
Enable `test_addr_true` test

ehildenb:
This PR enables `AddrTest.test_addr_true` on our Foundry test-suite.

- The test is enabled.
- A bug in how we were calling `ECDSAPubKey` is fixed, which was causing an undefinedness crash (now that the undefinedness crash is fixed, instead we get a hook not implemented error).
- An example lemma is added which demonstrates how to get around the unimplemented hook error (and the process is documented here: https://github.com/runtimeverification/evm-semantics/issues/1765).


include/kframework/foundry.md
```diff
@@ -1023,9 +1023,9 @@ function sign(uint256 privateKey, bytes32 digest) external returns (uint8 v, byt
 ```
 
 `foundry.call.sign` will match when the `sign` cheat code function is called.
-This rule then takes from the `privateKey` to sign using `#range(ARGS,0,32)` and the `digest` to be signed using `#range(ARGS, 32, 32)`. 
-To perform the signature we use the `ECDSASign ( String, String )` function (from KEVM). 
-This function receives as arguments 2 strings: the data to be signed and the private key, therefore we use `#unparseByteStack` to convert the `Bytes` with the `privateKey` and `digest` into strings. 
+This rule then takes from the `privateKey` to sign using `#range(ARGS,0,32)` and the `digest` to be signed using `#range(ARGS, 32, 32)`.
+To perform the signature we use the `ECDSASign ( String, String )` function (from KEVM).
+This function receives as arguments 2 strings: the data to be signed and the private key, therefore we use `#unparseByteStack` to convert the `Bytes` with the `privateKey` and `digest` into strings.
```
lisandrasilva, line 1028: The comment should be updated because now we are using `#unparseDataBytes` instead of `#unparseByteStack`.

lisandrasilva (approved):

PR #1768, opened 2023-04-24
Base lemmas for buffer reasoning

PetarMax:
This PR introduces several base lemmas for buffer reasoning, including

- equality of concatenations
- associativity enforcement on `+Bytes`
- injectivity of `#buf`

### DIFF
--- a/include/kframework/lemmas/bytes-simplification.k
+++ b/include/kframework/lemmas/bytes-simplification.k
@@ -7,19 +7,63 @@ module BYTES-SIMPLIFICATION [symbolic]
   // Buffer Reasoning
   // ########################
 
-    rule BA:Bytes       ==K #buf(32, DATA) => #buf(32, DATA) ==K              BA                                    [simplification, concrete(BA)]
-    rule #buf(32, DATA) ==K BA:Bytes       =>          DATA  ==Int #asInteger(BA) requires lengthBytes(BA) <=Int 32 [simplification, concrete(BA)]
+    rule [bytes-not-equal-length]:
+      BA1:Bytes ==K BA2:Bytes => false
+      requires lengthBytes(BA1) =/=Int lengthBytes(BA2)
+      [simplification]
+
+    rule [bytes-equal-concat-split-k]:
+      A:Bytes +Bytes B:Bytes ==K C:Bytes +Bytes D:Bytes => A ==K C #And B ==K D
+      requires lengthBytes(A) ==Int lengthBytes(C)
+        orBool lengthBytes(B) ==Int lengthBytes(D)
+      [simplification]
+
+    rule [bytes-equal-concat-split-ml]:
+      { A:Bytes +Bytes B:Bytes #Equals C:Bytes +Bytes D:Bytes } => { A #Equals C } #And { B #Equals D }
+      requires lengthBytes(A) ==Int lengthBytes(C)
+        orBool lengthBytes(B) ==Int lengthBytes(D)
+      [simplification]
 
     rule [bytes-concat-empty-right]: B:Bytes +Bytes .Bytes  => B [simplification]
-    rule [bytes-concat-empty-left]:  .Bytes  +Bytes B:Bytes => B [simplification]
+    rule [bytes-concat-empty-left]:   .Bytes +Bytes B:Bytes => B [simplification]
 
-    rule [bytes-concat-assoc-right]: (BA1 +Bytes BA2) +Bytes BA3 => BA1 +Bytes (BA2 +Bytes BA3) [simplification]
+    rule [bytes-concat-right-assoc-symb-l]: (B1:Bytes +Bytes B2:Bytes) +Bytes B3:Bytes => B1 +Bytes (B2 +Bytes B3) [symbolic(B1), simplification]
+    rule [bytes-concat-right-assoc-symb-r]: (B1:Bytes +Bytes B2:Bytes) +Bytes B3:Bytes => B1 +Bytes (B2 +Bytes B3) [symbolic(B2), simplification]
+    rule [bytes-concat-left-assoc-conc]:    B1:Bytes +Bytes (B2:Bytes +Bytes B3:Bytes) => (B1 +Bytes B2) +Bytes B3 [concrete(B1, B2), symbolic(B3), simplification]
 
     // #buf
 
-    // TODO: custom ==K unification doesn't work in Haskell yet
-    // +Bytes unification
-    rule #buf(N, A) +Bytes BUF1 ==K #buf(N, B) +Bytes BUF2 => #buf(N, A) ==K #buf(N, B) andBool BUF1 ==K BUF2     [simplification]
+    rule [buf-inject-k]:
+      #buf(W:Int, X:Int) ==K #buf(W:Int, Y:Int) => X ==Int Y
+      requires 0 <=Int W
+       andBool #rangeUInt (8 *Int W, X)
+       andBool #rangeUInt (8 *Int W, Y)
+      [simplification]
+
+    rule [buf-inject-ml]:
+      { #buf(W:Int, X:Int) #Equals #buf(W:Int, Y:Int) } => { X #Equals Y }
+      requires 0 <=Int W
+       andBool #rangeUInt (8 *Int W, X)
+       andBool #rangeUInt (8 *Int W, Y)
+      [simplification]
+
+    rule [buf-as-int-left]:  B:Bytes   ==K #buf(32, X:Int) => X ==Int #asInteger(B) requires lengthBytes(B) <=Int 32 [simplification, concrete(B)]
+    rule [buf-as-int-right]: #buf(32, X:Int) ==K B:Bytes   => X ==Int #asInteger(B) requires lengthBytes(B) <=Int 32 [simplification, concrete(B)]
+
+    rule [buf-asWord-invert]:
+      #buf (W:Int , #asWord(B:Bytes)) => #buf(W -Int lengthBytes(B), 0) +Bytes B
+      requires lengthBytes(B) <=Int W
+      [simplification]
+
+    rule [buf-zero-concat-base]:
+      #buf(W1:Int, 0) +Bytes #buf(W2:Int, 0) => #buf(W1 +Int W2, 0)
+      requires 0 <=Int W1 andBool 0 <=Int W2
+      [simplification]
+
+    rule [buf-zero-concat-conc]:
+      #buf(W1:Int, 0) +Bytes #buf(W2:Int, 0) +Bytes B => #buf(W1 +Int W2, 0) +Bytes B
+      requires 0 <=Int W1 andBool 0 <=Int W2
+      [simplification]
 
     // #range
 
@@ -107,4 +151,5 @@ module BYTES-SIMPLIFICATION [symbolic]
 
     rule #asWord ( #ecrec ( _ , _ , _ , _ ) ) <Int pow160 => true
         [simplification, smt-lemma]
+
 endmodule
--- a/tests/specs/functional/lemmas-spec.k
+++ b/tests/specs/functional/lemmas-spec.k
@@ -13,6 +13,17 @@ module VERIFICATION
  // --------------------------------------
     rule runLemma( T ) => doneLemma( T )
 
+ // Shorthand for previous notations so that more complex
+ // claims do not have to be rewritten
+
+    syntax Bytes ::= Bytes "[" Int ".." Int "]" [function, total]
+ // -------------------------------------------------------------
+    rule B:Bytes [ S:Int .. W:Int ] => #range(B, S, W)
+
+    syntax Bytes ::= Bytes "[" Int ":=" Bytes "]" [function, total]
+ // ---------------------------------------------------------------
+    rule B:Bytes [ S:Int := B1:Bytes ] => #writeRange(B, S, B1)
+
 endmodule
 
 module LEMMAS-SPEC


tests/failing-symbolic.haskell
```diff
@@ -1,5 +1,7 @@
 tests/specs/benchmarks/ecrecoverloop00-sig1-invalid-spec.k
 tests/specs/bihu/collectToken-spec.k
+tests/specs/erc20/ds/transferFrom-failure-2-a-spec.k
+tests/specs/erc20/hkg/transferFrom-failure-2-spec.k
```
ehildenb, line 4: We don't want to add to this list. It's OK to add to `tests/failing-symbolic.pyk`.

This list says "this proof doesn't work at all", and the other list `tests/failing-symbolic.pyk` says "this proof doesn't work with the pyk RPC based prover, but does work with the old prover".

So if you add this to the other list `tests/failing-symbolic.pyk`, and it's then passing (which you can test by removing from this list and adding to that list and running `make tests/specs/erc20/ds/transferFrom-failure-2-a-spec.k.prove`, for example), then we're good. But making the test fail altogether, we should isolate the failure reason and make a run-lemma-donel-lemma style claim for it and add the needed lemmas to get it passing with your new lemmas.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -7,19 +7,63 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+    rule [bytes-not-equal-length]:
+      BA1:Bytes ==K BA2:Bytes => false
+      requires lengthBytes(BA1) =/=Int lengthBytes(BA2)
+      [simplification]
+
+    rule [bytes-equal-concat-split-k]:
+      A:Bytes +Bytes B:Bytes ==K C:Bytes +Bytes D:Bytes => A ==K C #And B ==K D
+      requires lengthBytes(A) ==Int lengthBytes(C)
+        orBool lengthBytes(B) ==Int lengthBytes(D)
+      [simplification]
```
ehildenb, line 19: This is not correct, I believe. Here, on the RHS, we are saying set-intersection between the truth/false values of the two. So, for example, if we had `{ A ==K C #Equals true }`, and `{ B ==K D #Equals false }`, the RHS of this expression would be `#Bottom`, not `false` as intended.

I think we can just use `andBool` on the RHS here instead of `#And`?

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -7,19 +7,63 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+
+    rule [bytes-equal-concat-split-ml]:
+      { A:Bytes +Bytes B:Bytes #Equals C:Bytes +Bytes D:Bytes } => { A #Equals C } #And { B #Equals D }
+      requires lengthBytes(A) ==Int lengthBytes(C)
+        orBool lengthBytes(B) ==Int lengthBytes(D)
+      [simplification]
```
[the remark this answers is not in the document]
ehildenb: If not, do you have an example from Lukso that does break, which can be turned into a test added here?

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -7,19 +7,63 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+       andBool 0 <=Int X andBool X <Int 2 ^Int (8 *Int W)
+       andBool 0 <=Int Y andBool Y <Int 2 ^Int (8 *Int W)
+      [simplification]
+
+    rule [buf-inject-ml]:
+      { #buf(W:Int, X:Int) #Equals #buf(W:Int, Y:Int) } => { X #Equals Y }
+      requires 0 <=Int W
+       andBool 0 <=Int X andBool X <Int 2 ^Int (8 *Int W)
+       andBool 0 <=Int Y andBool Y <Int 2 ^Int (8 *Int W)
+      [simplification]
```
ehildenb, line 48: Once again, do any of the added tests break if you remove this one for now? Similarly, if not, does anything from Lukso break? We'd like to not introduce more ML connective simplifications if we don't havea  direct need for them.

Even if we do have an example that beraks, perhaps we omit this anyway, and just keep copying that between projects while we try to resolve taht issue in the backend instead.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -7,19 +7,63 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+
+    rule [buf-as-int-left]:  B:Bytes   ==K #buf(32, X:Int) => X ==Int #asInteger(B) requires lengthBytes(B) <=Int 32 [simplification, concrete(B)]
+    rule [buf-as-int-right]: #buf(32, X:Int) ==K B:Bytes   => X ==Int #asInteger(B) requires lengthBytes(B) <=Int 32 [simplification, concrete(B)]
```
ehildenb, line 51: Can these be simplified by using `comm` intead? We should have `comm` on `_==K_` now https://github.com/runtimeverification/k/pull/3342

include/kframework/lemmas/bytes-simplification.k
PetarMax: Oh no no no no, that has to be `andBool`, my mistake...

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -7,19 +7,62 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+      [simplification]
+
+    rule [buf-inject-ml]:
+      { #buf(W:Int, X:Int) #Equals #buf(W:Int, Y:Int) } => { X #Equals Y }
+      requires 0 <=Int W
+       andBool 0 <=Int X andBool X <Int 2 ^Int (8 *Int W)
+       andBool 0 <=Int Y andBool Y <Int 2 ^Int (8 *Int W)
+      [simplification]
+
+    rule [buf-as-int]: B:Bytes ==K #buf(32, X:Int) => X ==Int #asInteger(B) requires lengthBytes(B) <=Int 32 [simplification, concrete(B), comm]
```
ehildenb, line 50: I don't think this is quite true is it?

This lemma means: `lengthBytes(B) <=Int 32      IMPLIES     (B ==K #buf(32, X)    EQUALS    X ==Int #asInteger(B))`

Let's choose:

- `X == (2 ^Int 256) +Int 4`
- `B == #buf(1, 4)`

Then we have:

- `lengthBytes(B) <=Int 32` is `true`
- `#buf(32, X)` is `#buf(31, 0) +Bytes #buf(1, 4)`
- `#asInteger(B)` is `4`

So we have that:

- `X ==Int #asInteger(B)` is FALSE.
- `B ==K #buf(32, X)` is TRUE.

ehildenb: A way to fix this is to also restrict `X <Int pow256`, or even better, `#rangeUInt(256, X)`

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -7,19 +7,62 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+
+    rule [buf-asWord-invert]:
+      #buf (W:Int , #asWord(B:Bytes)) => #buf(W -Int lengthBytes(B), 0) +Bytes B
+      requires lengthBytes(B) <=Int W
+      [simplification]
+
```
ehildenb, line 56: I don't think this one is quite true either. Let's choose:

- `W == 64`
- `B == #buf(64, pow256 +Int 4)`

Then we have:

- `#asWord(B)` is `4` (see definition of `#asWord` in `evm-types.md`).
- `lengthBytes(B)` is `64`.
- REQ: `lengthBytes(B) <=Int W` is TRUE.
- LHS: `#buf(W, #asWord(B))` is `#buf(32, 0) +Bytes #buf(32, 4)`
- RHS: `#buf(W -Int lengthBytes(B), 0)` is `#buf(0, 0)`

So we don't have that `REQ     IMPLIES     (LHS    EQUALS     RHS)` is true.

tests/specs/functional/lemmas-spec.k
```diff
@@ -13,6 +13,17 @@ module VERIFICATION
...
+ // claims do not have to be rewritten
+
+    syntax Bytes ::= Bytes "[" Int ".." Int "]" [function, total]
+ // -------------------------------------------------------------
+    rule B:Bytes [ S:Int .. W:Int ] => #range(B, S, W)
+
+    syntax Bytes ::= Bytes "[" Int ":=" Bytes "]" [function, total]
+ // ---------------------------------------------------------------
+    rule B:Bytes [ S:Int := B1:Bytes ] => #writeRange(B, S, B1)
+
```
ehildenb, line 26: I actually don't understand why these are introduced again. Is it because there are places in the Lukso or Optimism proofs where we are using these shorthands? If so, let's just define them there.

We had https://github.com/runtimeverification/evm-semantics/pull/1594 to switch away from having this dual representation, so we should pick one and stick to it.

include/kframework/lemmas/bytes-simplification.k
PetarMax: Oh I didn't sanity-check that one, that one was there from long ago. Yes, `#rangeUInt` will do.

tests/specs/functional/lemmas-spec.k
PetarMax: These are used only for the claim testing, because there are claims that use them a lot and it'd be a substantial pain to rewrite them. They shouldn't seep into the proofs, right?

include/kframework/lemmas/bytes-simplification.k
PetarMax: I didn't realise `#asWord` did the chop. I will add `W <=Int 32`.

tests/specs/functional/lemmas-spec.k
PetarMax: I've removed the notation here, but in the next PR there will be claims like
```
claim [range-34]:
      <k> runLemma(
            #range ( b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x80\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00E\xb5`x\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
                       [ 132 := #buf ( 32 , _VV5__txOrigin_3c5818c8:Int ) ]
                       [ 128 := b"I\x1c\xc7\xc2\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00" ]
                       [ 128 := b"" ]
                       [ 132 := b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01" ]
                       [ 128 := b"\xe9\xe0\\B\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00" ]
                       [ 164 := b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01" ]
                       [ 160 := b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00" ]
                       [ 132 := b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00" ]
                       [ 164 := #buf ( 32 , _VV1__value_3c5818c8:Int ) ]
                       [ 196 := #buf ( 32 , _VV2__gasLimit_3c5818c8:Int ) ]
                       [ 228 := b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x01" ]
                       [ 260 := b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\xa0" ]
                       [ 292 := #buf ( 32 , lengthBytes ( VV6__data_3c5818c8:Bytes ) ) ]
                       [ 324 := VV6__data_3c5818c8:Bytes ] , 324 , minInt ( ( notMaxUInt5 &Int lengthBytes ( VV6__data_3c5818c8:Bytes ) +Int 31 ) +Int 196 , lengthBytes ( VV6__data_3c5818c8:Bytes ) +Int 196 ) +Int -196 )
          )
          => doneLemma( VV6__data_3c5818c8 ) ...
      </k>
```
I guess I thought claims like that would also appear in this one, but in the end they didn't.

ehildenb: But this claim should be changed to use the modern notation with `writeRange` I guess? This is exactly what the `_[_:=_]` notation was developed for though, because it is easier to read in this setting. So we may want to consider revisitng the previous decision about choosing the `#writeRange` and `#range` over the misfix variants of the operators.

ehildenb (approved):

tests/specs/functional/lemmas-spec.k
PetarMax: So, for `#writeRange` the notation with `[ ... ]` is much more readable for me - for `#range`, the notation that we have is much more readable, does that make sense? We could bring back just one...

PR #1773, opened 2023-04-25
range and writeRange simplifications

PetarMax:
This PR introduces:

- a number of `range` and `writeRange` simplifications;
- simplifications on `up/Int` when it is expressed in terms of `&Int`;
- a more general rule for calculating the length of a range.

Should be merged after #1762 and #1768.


include/kframework/lemmas/bitwise-simplification.k
```diff
@@ -43,4 +43,12 @@ module BITWISE-SIMPLIFICATION [symbolic]
...
     rule maxUInt160 &Int (X |Int (notMaxUInt160 &Int Y:Int)) => X requires #rangeUInt(160, X) andBool 0 <=Int Y [simplification]
 
+    rule [lengthBytes-upInt-32-lower-bound]:
+      lengthBytes(X) <=Int notMaxUInt5 &Int lengthBytes(X) +Int 31 => true
+      [simplification, smt-lemma]
+
+    rule [lengthBytes-upInt-32-upper-bound]:
+      lengthBytes(X) +Int 32 >Int notMaxUInt5 &Int lengthBytes(X) +Int 31 => true
+      [simplification, smt-lemma]
+
```
ehildenb, line 53: Can we add parens to the RHS of the comparison operators on these? It's confusing to me which binds tighter, the `&Int` or the `+Int`, and it wastes a bit of time hhhaving to look it up each time.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,50 +68,155 @@ module BYTES-SIMPLIFICATION [symbolic]
 
     // #range
 
-    rule #range(_, _, K) => .Bytes requires K <=Int 0 [simplification]
+    rule [range-empty]:
+      #range(_:Bytes, S:Int, W:Int) => .Bytes
+      requires S <Int 0 orBool W <=Int 0
+      [simplification(1)]
```
ehildenb, line 74:
```suggestion
      [simplification(40)]
```

Any reason to choose 1 in particular? That leaves little room for inserting rules before it.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,50 +68,155 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+      #range(B1:Bytes +Bytes B2:Bytes, 0, W:Int) => B1 +Bytes #range(B2, 0, W -Int lengthBytes(B1))
+      requires lengthBytes(B1) <=Int W
+      [simplification]
 
-    rule #range(#writeRange(M, N, BUF:Bytes ), L, K) => #range(M, L, K)
-      requires 0 <Int K andBool N +Int lengthBytes(BUF) <=Int L [simplification]
+    rule [range-inside-cHead-base-conc-left]:
+      #range(B:Bytes, S1:Int +Int S2:Int, W:Int) => #range(#range(B, S1, lengthBytes(B) -Int S1), S2, W)
+      requires 0 <Int S1 +Int S2 andBool S1 +Int S2 <=Int lengthBytes(B)
+      [concrete(B), concrete(S1), simplification]
```
ehildenb, line 109: - `B:Bytes == #buf(6, 1)`
- `S1 == -3`
- `S2 == 7`
- `W == 2`

Then

- `#range(B, S1 +Int S2, W) => #range(B, 4, 2) => #buf(2, 1)`
- `#range(B, S1 +Int S2, W) => #range(#range(B, -3, lengthBytes(B) -Int -3), 7, 2) => #range(.Bytes, 7, 2) => #buf(2, 0)`

ehildenb: This same counterexample may apply to other range lemmas you have about. Probably easily solved by requiring that `S1` and `S2` are individually non-negative too.

include/kframework/lemmas/bytes-simplification.k
PetarMax: Good point, I'll change to something more flexible.

include/kframework/lemmas/bitwise-simplification.k
PetarMax: Yes, done.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -128,11 +233,11 @@ module BYTES-SIMPLIFICATION [symbolic]
...
-    rule [lengthBytes-concat]:   lengthBytes(BUF1 +Bytes BUF2)                 => lengthBytes(BUF1) +Int lengthBytes(BUF2)            [simplification]
-    rule [lengthBytes-buf]:      lengthBytes(#buf(SIZE, _))                    => SIZE   requires 0 <=Int SIZE                        [simplification]
-    rule [lengthBytes-range]:    lengthBytes(#range(_, START, WIDTH))          => WIDTH  requires 0 <=Int START andBool 0 <=Int WIDTH [simplification]
-    rule [lengthBytes-prtw]:     lengthBytes(#padRightToWidth(W:Int, B:Bytes)) => maxInt(lengthBytes(B), W)                           [simplification]
+    rule [lengthBytes-geq-zero]: 0 <=Int lengthBytes ( _ )                     => true                                     [simplification, smt-lemma]
+    rule [lengthBytes-concat]:   lengthBytes(BUF1 +Bytes BUF2)                 => lengthBytes(BUF1) +Int lengthBytes(BUF2) [simplification]
+    rule [lengthBytes-buf]:      lengthBytes(#buf(S, _))                       => S  requires 0 <=Int S                    [simplification]
+    rule [lengthBytes-range]:    lengthBytes(#range(_, S:Int, W:Int))          => maxInt(0, W) requires 0 <=Int S          [simplification]
```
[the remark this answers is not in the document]
PetarMax: Oops...

ehildenb:
@PetarMax it looks like changes to `k-web-theme` and `k` submodule snuck in.

You can revert those changes before this is merged.

tests/specs/functional/lemmas-spec.k
```diff
@@ -13,6 +13,13 @@ module VERIFICATION
...
     rule runLemma( T ) => doneLemma( T )
 
+ // Shorthand for previous notations so that more complex
+ // claims do not have to be rewritten
+
+    syntax Bytes ::= Bytes "[" Int ".." Int "]" [function, total]
+ // -------------------------------------------------------------
+    rule B:Bytes [ S:Int .. W:Int ] => #range(B, S, W)
+
 endmodule
```
ehildenb, line 23: Lookin ghrough, I only see a handful of new lemmas using the shorthand, can we just simplify to the `#range(...)` syntax those lemmas? No need to give future-selves the option, which will lead to more confusion for future new people.

include/kframework/lemmas/bitwise-simplification.k
```diff
@@ -43,4 +43,12 @@ module BITWISE-SIMPLIFICATION [symbolic]
...
     rule maxUInt160 &Int (X |Int (notMaxUInt160 &Int Y:Int)) => X requires #rangeUInt(160, X) andBool 0 <=Int Y [simplification]
 
+    rule [lengthBytes-upInt-32-lower-bound]:
+      lengthBytes(X) <=Int notMaxUInt5 &Int ( lengthBytes(X) +Int 31 ) => true
+      [simplification, smt-lemma]
+
+    rule [lengthBytes-upInt-32-upper-bound]:
+      lengthBytes(X) +Int 32 >Int notMaxUInt5 &Int ( lengthBytes(X) +Int 31 ) => true
+      [simplification, smt-lemma]
+
```
ehildenb, line 53: Is there any reason these have to say `lengthBytes(X)`?

```suggestion
    rule [lengthBytes-upInt-32-lower-bound]:
      X <=Int notMaxUInt5 &Int ( X +Int 31 ) => true requires 0 <=Int X
      [simplification, smt-lemma]

    rule [lengthBytes-upInt-32-upper-bound]:
      X +Int 32 >Int notMaxUInt5 &Int ( X +Int 31 ) => true requires 0 <=Int X
      [simplification, smt-lemma]

```

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,168 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+
+    rule [range-full]:
+      #range(B:Bytes, 0, W:Int) => B
+      requires lengthBytes(B) ==Int W
+      [simplification]
+
+    rule [range-overfull]:
+      #range(B:Bytes, 0, W:Int) => B +Bytes #buf(W -Int lengthBytes(B), 0)
+      requires lengthBytes(B) <Int W
+      [simplification]
```
ehildenb, line 84: I guess these two lemmas could be combined with one for `lengthBytes(B) <=Int W`. Any reason not to?

PetarMax: Hmm, yes, they should be fine together, given that if we know the equality, the tail zeros will simplify out.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,168 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+
+    rule [range-inside-cHead-base-conc-left]:
+      #range(B:Bytes, S1:Int +Int S2:Int, W:Int) => #range(#range(B, S1, lengthBytes(B) -Int S1), S2, W)
+      requires 0 <=Int S1 andBool 0 <=Int S2 andBool 0 <Int S1 +Int S2 andBool S1 +Int S2 <=Int lengthBytes(B)
+      [concrete(B), concrete(S1), simplification]
+
+    rule [range-inside-cHead-base-conc-right]:
+      #range(B:Bytes, S1:Int +Int S2:Int, W:Int) => #range(#range(B, S2, lengthBytes(B) -Int S2), S1, W)
+      requires 0 <=Int S1 andBool 0 <=Int S2 andBool 0 <Int S1 +Int S2 andBool S1 +Int S2 <=Int lengthBytes(B)
+      [concrete(B), concrete(S2), simplification]
```
ehildenb, line 114: We should already be moving all concrete terms in `+Int` to the right, so I don't think the first rule here is needed actually. Does it work if you remove it? https://github.com/runtimeverification/evm-semantics/blob/8a8dc22a2f56460339d6bc29fa90f6b7e58eb256/include/kframework/lemmas/int-simplification.k#L16

We could consider bumping the priority of the `+Int` and `+Bytes` re-associtation lemmas up to priority `40`.

include/kframework/lemmas/bitwise-simplification.k
PetarMax: I don't think that work for `X >=Int pow256`, I will bound it to `[0, pow256)`.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,168 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+       andBool 0 <=Int S1 andBool 0 <=Int W1 andBool 0 <=Int S2 andBool 0 <=Int W2
+      [simplification]
 
-    rule #range( M [ N := BUF:Bytes ], L, K ) => #range(M, L, K)
-      requires 0 <Int K andBool N +Int lengthBytes(BUF) <=Int L [simplification]
+    rule [range-join-concat]:
+      #range(B1, S1, W1) +Bytes #range(B1, S2, W2) +Bytes B2 => #range(B1, S1, W1 +Int W2) +Bytes B2
+      requires S2 ==Int S1 +Int W1
+       andBool 0 <=Int S1 andBool 0 <=Int W1 andBool 0 <=Int S2 andBool 0 <=Int W2
+      [simplification]
```
ehildenb, line 137: Explicit parens please.

```suggestion
      (#range(B1, S1, W1) +Bytes #range(B1, S2, W2)) +Bytes B2 => #range(B1, S1, W1 +Int W2) +Bytes B2
      requires S2 ==Int S1 +Int W1
       andBool 0 <=Int S1 andBool 0 <=Int W1 andBool 0 <=Int S2 andBool 0 <=Int W2
      [simplification]
```

ehildenb:
The submodule change at `web/k-web-theme` should also be undone.

PetarMax:
> The submodule change at `web/k-web-theme` should also be undone.

Done.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,168 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+       andBool W3 ==Int S +Int W1 -Int lengthBytes(B)
+       andBool 0 <=Int S andBool 0 <=Int W1 andBool 0 <=Int W2
+      [simplification]
+
+    rule [range-reform-concat]:
+      #range(B1:Bytes, S:Int, minInt(W1:Int, W2:Int)) +Bytes #buf(maxInt(0, W3:Int), 0) +Bytes B2:Bytes => #range (B1, S, W1) +Bytes B2
+      requires W2 ==Int lengthBytes(B1) -Int S
+       andBool W3 ==Int S +Int W1 -Int lengthBytes(B1)
+       andBool 0 <=Int S andBool 0 <=Int W1 andBool 0 <=Int W2
+      [simplification]
```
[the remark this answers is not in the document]
PetarMax: But this parenthetisation will be undone immediately by the right-associativity enforcement, will it not?

include/kframework/lemmas/bitwise-simplification.k
PetarMax: This breaks the backend with:
```
kore-exec: [3416348] Error (DecidePredicateUnknown):
    Failed to decide predicate:        
        /* Sfa */
        \equals{SortBool{}, _}(
            /* T Fn D Sfa Cl */ \dv{SortBool{}}("true"),
            /* T Fn D Sfa */
            Lbl'Unds'orBool'Unds'{}(
                /* T Fn D Sfa */
                Lbl'UndsEqlsEqls'Int'Unds'{}(
                    /* T Fn D Sfa */ VarUINT8:SortInt{},
                    /* T Fn D Sfa Cl */ \dv{SortInt{}}("0")
                ),
                /* T Fn D Sfa */
                Lbl'UndsEqlsEqls'Int'Unds'{}(
                    /* T Fn D Sfa */ VarUINT8:SortInt{},
                    /* T Fn D Sfa Cl */ \dv{SortInt{}}("1")
                )
            )
        )
```
on any execution. The `smt-lemma`s are not behaving well, I guess.

include/kframework/lemmas/bytes-simplification.k
PetarMax: This works without change of priority.

include/kframework/lemmas/bitwise-simplification.k
PetarMax: Keeping `lengthBytes` for now...

include/kframework/lemmas/bytes-simplification.k
ehildenb: Yes, I assume it would. But these lemmas are redundant. If we're ni the situation `S1 +Int S2` with `concrete(S1), symbolic(S2)`, then the second rule won't fire, but we don't need to consider that case because the re-association lemma will flip around the `_+Int_`. So we really only need one of these lemmas, the second one (with `concrete(S2)`).

Suggesting to increase the priority of the reassociation lemma is just a side-thought, maybe it would be good to make sure that veerything as associated properly before applying other lemmas. Should not change the overall outcome.

include/kframework/lemmas/bytes-simplification.k
ehildenb: In order to determine that, I have to search though `domains.md` for this location: https://kframework.org/k-distribution/include/kframework/builtin/domains/#bytes-concatenation

No need to put that extra mental burden on people I think. Better to just put the parens in explicitly.

ehildenb: Code needs to be read more frequently than written, and reading code to extract intent is harder than writing code given an intent. So better to write something a little less elegant if it expresses the intent more clearly.

include/kframework/lemmas/bytes-simplification.k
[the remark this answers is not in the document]
ehildenb: Your lemma, as written, will be parsed as `(B1 +Bytes B2) +Bytes B3`. That doesn't have anything to do with the rules that enforce that our normal form is right-associative. Better to put the parens in so that people reading the lemma have no doubt about which AST you mean explicitly.

include/kframework/lemmas/bytes-simplification.k
ehildenb: The case where this would matter is if you have a variable `W` that you know `0 <=Int W` but not whether `0 ==Int W` or `0 <Int W`. In this case, having one lemma cover `0 <=Int W` will allow it to apply, but the backend will not case split on the two separate lemmas.

include/kframework/lemmas/bitwise-simplification.k
ehildenb: Well, was worth a try. Then I guess these lemmas shoul be moved to `include/kframework/lemmas/lemmas.k` instead, since `notMaxUInt5` is a KEVM specific constant.

include/kframework/lemmas/bytes-simplification.k
PetarMax: Yes yes, I agree, I meant that your suggestion works, we do need only one and the other passes as a claim. 

I'm happy to increase priority of associativity also.

ehildenb:
I think it'd be good to also change in this PR the definition of `_[_:=_]` to use `<=Int` instead of `>=Int`:

image

It would be nice to always use the same direction of disequalities for easier brain-thinking-time.

include/kframework/lemmas/bytes-simplification.k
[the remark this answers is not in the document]
PetarMax: I can't think of how to break them into simpler ones right now, will think a bit more.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,168 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+      requires 0 <=Int S andBool lengthBytes(B1) <=Int S +Int lengthBytes(B2)
+      [simplification(60)]
 
-    // overwritten memory write
-    rule MEM [ K1 := BUF1 ] [ K2 := BUF2 ] => MEM [ K2 := BUF2 ]
-      requires K1 +Int lengthBytes(BUF1) <=Int K2 +Int lengthBytes(BUF2) andBool K2 <=Int K1 [simplification]
+    rule [memUpdate-reorder]:
+      B:Bytes [ S1:Int := B1:Bytes] [ S2:Int := B2:Bytes ] => B [ S2 := B2 ] [ S1 := B1 ]
+      requires S2 <Int S1 andBool S2 +Int lengthBytes(B2) <=Int S1
+      [simplification]
```
ehildenb, line 219: Counterexample:

- `B == #buf(2, 1)`
- `S1== 1`
- `B1 == #buf(1, 2)`
- `S2 == -1`
- `B2 == #buf(1, 3)`

Then:

- `B [ S1 := B1 ] [ S2 := B2 ] => (B [ S1 := B1 ]) [ -1 := #buf(1, 3) ] => .Bytes`
- `B [ S1 := B1 ] [ S2 := B2 ] => B [ S2 := B2 ] [ S1 := B1 ] => (B [ -1 := #buf(1, 3) ]) [ S1 := B1 ] => .Bytes [ S1 := B1 ] => #buf(1, 0) +Bytes #buf(2, 1)`

ehildenb: I think restricting `0 <=Int S2` should solve it.

ehildenb: Any counterexamples produced would be good to put in as new `runLemma => doneLemma` style claims.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,168 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+      requires 0 <=Int S andBool S +Int lengthBytes(B2) <Int lengthBytes(B1)
+      [simplification(60)]
 
-    // re-order memory write
-    rule MEM [ K1 := BUF1:Bytes ] [ K2 := BUF2:Bytes ] => MEM [ K2 := BUF2 ] [ K1 := BUF1 ]
-      requires lengthBytes(BUF2) <=Int K1 -Int K2 andBool K2  <Int K1 [simplification]
+    rule [memUpdate-as-concat-outside-1]:
+      B1:Bytes [ S:Int := B2:Bytes ] => #range(B1, 0, S) +Bytes B2
+      requires 0 <=Int S andBool lengthBytes(B1) <=Int S +Int lengthBytes(B2)
+      [simplification(60)]
```
ehildenb, line 214: Why are we restricting here `lengthBytes(B1) <=Int S +Int lengthBytes(B2)`? Would it be enough to say `lengthBytes(B1) <=Int S` on its own? Also, if we have `lengthBytes(B1) <=Int S`, then we already are guaranteed that `0 <=Int S` because `0 <= Int lengthBytes(_)`.

include/kframework/lemmas/bytes-simplification.k
PetarMax: Yes, the (now) usual `0 <=Int ...` - will be fixed with next push.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,168 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+      [simplification]
 
-    rule #range(BA:Bytes, 0, LEN) => BA requires lengthBytes(BA) ==Int LEN [simplification]
+    rule [range-memUpdate-in-between]:
+      #range(B1:Bytes [ S1:Int := B2:Bytes ], S2:Int, W2:Int) =>
+        #let S = S2 -Int S1 #in
+        #let W = minInt(W2, lengthBytes(B2) -Int S) #in
+          #range(B2, S, W) +Bytes #range(B1, S1 +Int lengthBytes(B2), W2 -Int W)
+      requires S1 <=Int S2 andBool S2 <=Int S1 +Int lengthBytes(B2)
+      [simplification]
```
ehildenb, line 173: Do we need here additionally that `0 <=Int S1`?

include/kframework/lemmas/bytes-simplification.k
PetarMax: What we have now is more expressive, isn't it? `lengthBytes(B1)` could be less than or equal to `S`, but also in between `S` and `S +Int lengthBytes(B2)`?

include/kframework/lemmas/bytes-simplification.k
PetarMax: Very probably, yes.

include/kframework/lemmas/bytes-simplification.k
PetarMax: > Any counterexamples produced would be good to put in as new `runLemma => doneLemma` style claims.

I'm not sure what that would entail here?

include/kframework/lemmas/bitwise-simplification.k
ehildenb: Ah, well I guess I'm just wrong then. I suppose that should be cleaned up eventually, but maybe that's a problem for another day.

include/kframework/lemmas/bytes-simplification.k
ehildenb: Something like `runLemma(#buf(2, 1) [ 1 := #buf(1, 2) ] [ -1 := #buf(1, 3) ]) => doneLemma(.Bytes)` perhaps?

PetarMax: Done and passing.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,158 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+    rule [range-memUpdate-outside]:
+      #range(B1:Bytes [ S1:Int := B2:Bytes ], S2:Int, W2:Int) => #range(B1, S2, W2)
+      requires 0 <=Int S1 andBool 0 <=Int W2 andBool S1 +Int lengthBytes(B2) <=Int S2 [simplification]
 
-    rule (#range(#range(BA:Bytes, S0, L0), S1, L1)) => #range(BA, S0 +Int S1, minInt(L1, L0 -Int S1)) +Bytes #buf(maxInt(0, L1 -Int (L0 -Int S1)), 0)
-      requires 0 <=Int S0 andBool 0 <=Int L0 andBool 0 <=Int S1 andBool 0 <=Int L1 [simplification]
+    rule [range-buf-zero]:
+      #range (#buf(W1:Int, X:Int), S2:Int, W2:Int) => #buf(W2, 0)
+      requires 0 <=Int X andBool X <Int 2 ^Int (8 *Int (W1 -Int (S2 +Int W2))) andBool S2 +Int W2 <=Int W1
+      [simplification]
```
ehildenb, line 182: Counterexample:

- `W1 == 8`
- `X == 1`
- `S2 == -3`
- `W2 == 1`

Then:

- `0 <=Int X => true`
- `X <Int 2 ^Int (8 *Int (W1 -Int (S2 +Int W2))) => 1 <Int 2 ^Int (8 *Int (8 -Int (-3 +Int 1))) => 1 <Int 2 ^Int (8 *Int 10) => true`
- `S2 +Int W2 <=Int W1 => -3 +Int 1 <=Int 8 => true`
- `#range(#buf(W1, X), S2, W2) => #range(#buf(W1, X), -3, 1) => .Bytes`
- `#range(#buf(W1, X), S2, W2) => #range(#buf(8, 1), -3, 1) => #buf(1, 0)`

I think conditions `0 <=Int S2 andBool 0 <=Int W2` should be good?

PetarMax: I think `0 <=Int S2` is sufficient because if `W2 <Int 0` then both sides collapse to `.Bytes`, but I'm happy to add both for sanity.

include/kframework/lemmas/bitwise-simplification.k
```diff
@@ -43,4 +43,12 @@ module BITWISE-SIMPLIFICATION [symbolic]
     rule notMaxUInt160 &Int X => 0 requires #rangeUInt(160, X) [simplification]
     rule maxUInt160 &Int (X |Int (notMaxUInt160 &Int Y:Int)) => X requires #rangeUInt(160, X) andBool 0 <=Int Y [simplification]
 
+    rule [lengthBytes-upInt-32-lower-bound]:
+      lengthBytes(X) <=Int notMaxUInt5 &Int ( lengthBytes(X) +Int 31 ) => true
+      [simplification, smt-lemma]
+
+    rule [lengthBytes-upInt-32-upper-bound]:
+      lengthBytes(X) +Int 32 >Int notMaxUInt5 &Int ( lengthBytes(X) +Int 31 ) => true
+      [simplification, smt-lemma]
```
ehildenb, line 52: Counterexample: `lengthBytes(X) >=Int 32`. This is because `notMaxUInt5` is `maxUInt256 xorInt maxUInt5`.

ehildenb: I think just adding `requires lengthBytes(X) <=Int 32` should be fine?

The counterexample only applies to the first lemma I suppose.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
 
     // #range
 
-    rule #range(_, _, K) => .Bytes requires K <=Int 0 [simplification]
+    rule [range-empty]:
+      #range(_:Bytes, S:Int, W:Int) => .Bytes
+      requires S <Int 0 orBool W <=Int 0
+      [simplification(30)]
```
ehildenb, line 74: Reviewed.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
 
-    rule #range( M [ N := BUF:Bytes ], L, K ) =>
-         #let W = minInt(K, N -Int L) #in #range(M, L, W) +Bytes #range(M [ N := BUF ], N, K -Int W)
-      requires 0 <Int K andBool L <Int N [simplification]
+    rule [range-full]:
+      #range(B:Bytes, 0, W:Int) => B +Bytes #buf(W -Int lengthBytes(B), 0)
+      requires lengthBytes(B) <=Int W
+      [simplification]
```
ehildenb, line 79: Reviewed.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+
+    rule [range-params-equal-ml]:
+      { #range (B1:Bytes, S1:Int, W1:Int) #Equals #range (B2:Bytes, S2:Int, W2:Int) } => #Top
+      requires B1 ==K B2 andBool S1 ==Int S2 andBool W1 ==Int W2
+      [simplification]
```
ehildenb, line 84: Reviewed.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+
+    rule [range-included-in-cHead]:
+      #range(B1:Bytes +Bytes _:Bytes, S:Int, W:Int) => #range(B1, S, W)
+      requires S +Int W <=Int lengthBytes(B1)
+      [simplification]
```
ehildenb, line 89: Reviewed.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+
+    rule [range-outside-cHead]:
+      #range(B1:Bytes +Bytes B2:Bytes, S:Int, W:Int) => #range(B2, S -Int lengthBytes(B1), W)
+      requires lengthBytes(B1) <=Int S
+      [simplification]
```
ehildenb, line 94: Reviewed.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+
+    rule [range-includes-cHead]:
+      #range(B1:Bytes +Bytes B2:Bytes, 0, W:Int) => B1 +Bytes #range(B2, 0, W -Int lengthBytes(B1))
+      requires lengthBytes(B1) <=Int W
+      [simplification]
```
ehildenb, line 99: Reviewed.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+
+    rule [range-inside-cHead-base-conc]:
+      #range(B:Bytes, S1:Int +Int S2:Int, W:Int) => #range(#range(B, S2, lengthBytes(B) -Int S2), S1, W)
+      requires 0 <=Int S1 andBool 0 <=Int S2 andBool 0 <Int S1 +Int S2 andBool S1 +Int S2 <=Int lengthBytes(B)
+      [concrete(B), concrete(S2), simplification]
```
ehildenb, line 104:
```suggestion
    rule [range-inside-cHead-base-conc]:
      #range(B:Bytes, S1:Int +Int S2:Int, W:Int) => #range(#range(B, S2, lengthBytes(B) -Int S2), S1, W)
      requires 0 <=Int S1 andBool 0 <=Int S2 andBool 0 <Int S1 +Int S2 andBool S1 +Int S2 <=Int lengthBytes(B)
      [concrete(B, S2), simplification]
```

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+    rule [range-inside-cHead-concat]:
+      #range(B1:Bytes +Bytes B2:Bytes, S:Int, W:Int) => #range(#range(B1, S, lengthBytes(B1) -Int S) +Bytes B2, 0, W)
+      requires 0 <Int S andBool S <=Int lengthBytes(B1)
+      [simplification]
+
+    rule [range-of-range]:
+      #range(#range(B:Bytes, S1:Int, W1:Int), S2:Int, W2:Int) =>
+        #range(B, S1 +Int S2, minInt(W2, W1 -Int S2)) +Bytes #buf(maxInt(0, W2 -Int (W1 -Int S2)), 0)
+      requires 0 <=Int S1 andBool 0 <=Int W1 andBool 0 <=Int S2 andBool 0 <=Int W2
+      [simplification]
```
ehildenb, line 115: Reviewed.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
-    rule #range( M [ N := BUF:Bytes ], L, K ) =>
-           #let S = L -Int N #in
-           #let W = minInt(K, lengthBytes(BUF) -Int S) #in
-           #range(BUF, S, W) +Bytes #range(M, N +Int lengthBytes(BUF), K -Int W)
-      requires 0 <Int K andBool N <=Int L andBool L <=Int N +Int lengthBytes(BUF) [simplification]
+    rule [range-join-base]:
+      #range(B, S1, W1) +Bytes #range(B, S2, W2) => #range(B, S1, W1 +Int W2)
+      requires S2 ==Int S1 +Int W1
+       andBool 0 <=Int S1 andBool 0 <=Int W1 andBool 0 <=Int S2 andBool 0 <=Int W2
+      [simplification]
```
ehildenb, line 121: Reviewed.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+       andBool 0 <=Int S andBool 0 <=Int W1 andBool 0 <=Int W2
+      [simplification]
 
-    rule #range(A +Bytes _B, START, LENGTH) => #range(A, START, LENGTH)
-      requires START +Int LENGTH <=Int lengthBytes(A) [simplification]
+    rule [range-buf-zero-1]:
+      #range(#buf(32, X:Int), 0, W:Int) => #buf(W, 0)
+      requires 0 <=Int W andBool W <=Int 32
+       andBool #rangeUInt (8 *Int (32 -Int W), X)
+      [simplification, concrete(W)]
```
ehildenb, line 147: Are we sure this rule is doing anything?

`#rangeUInt(8 *Int (32 -Int W), X)` probably actually doesn't get understood by the backend, because `#rangeUInt(...)` is an `alias` (which is a `macro`), so to th ebackend it's probably just an uninterpreted symbol (and likely leads to it being very confused). Macros (and aliases) are expanded at kompile time, and the backend should not see the rules for it. Does removing this lemma break anything?

If it is needed, I think it should be woreded instead as:

```suggestion
    rule [range-buf-zero-1]:
      #range(#buf(32, X:Int), 0, W:Int) => #buf(W, 0)
      requires 0 <=Int W andBool W <=Int 32
       andBool 0 <=Int X andBoool X <Int 8 *Int (32 -Int W)
      [simplification, concrete(W)]
```

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
 
-    rule #range(A +Bytes B, START, LENGTH) => #range(B, START -Int lengthBytes(A), LENGTH)
-      requires lengthBytes(A) <=Int START [simplification]
+    rule [range-buf-zero-2-symb]:
+      #range(#buf(W:Int, 0), S1:Int, W1:Int) => #buf(W1, 0)
+      requires 0 <=Int W andBool 0 <=Int S1 andBool 0 <=Int W1
+      [simplification]
```
ehildenb, line 152: Reviewed.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+      requires 0 <=Int S1 andBool 0 <=Int W1
+       andBool B ==K #buf(lengthBytes(B), 0)
+      [concrete(B), simplification]
+
+    rule [range-memUpdate-before]:
+      #range(B1:Bytes [ S1:Int := B2:Bytes ], S2:Int, W2:Int) =>
+        #let W = minInt(W2, S1 -Int S2) #in
+          #range(B1, S2, W) +Bytes #range(B1 [ S1 := B2 ], S1, W2 -Int W)
+      requires 0 <=Int S2 andBool S2 <Int S1
+      [simplification]
```
ehildenb, line 165: Reviewed.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+      #range(B1:Bytes [ S1:Int := B2:Bytes ], S2:Int, W2:Int) =>
+        #let S = S2 -Int S1 #in
+        #let W = minInt(W2, lengthBytes(B2) -Int S) #in
+          #range(B2, S, W) +Bytes #range(B1, S1 +Int lengthBytes(B2), W2 -Int W)
+      requires 0 <=Int S1 andBool S1 <=Int S2 andBool S2 <=Int S1 +Int lengthBytes(B2)
+      [simplification]
+
+    rule [range-memUpdate-outside]:
+      #range(B1:Bytes [ S1:Int := B2:Bytes ], S2:Int, W2:Int) => #range(B1, S2, W2)
+      requires 0 <=Int S1 andBool 0 <=Int W2 andBool S1 +Int lengthBytes(B2) <=Int S2 [simplification]
```
ehildenb, line 177: Reviewed.

include/kframework/lemmas/bytes-simplification.k
ehildenb: Is this actually subsumed completely by `range-buf-zero` further down?

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
 
-    rule (#range(#range(BA:Bytes, S0, L0), S1, L1)) => #range(BA, S0 +Int S1, minInt(L1, L0 -Int S1)) +Bytes #buf(maxInt(0, L1 -Int (L0 -Int S1)), 0)
-      requires 0 <=Int S0 andBool 0 <=Int L0 andBool 0 <=Int S1 andBool 0 <=Int L1 [simplification]
+    rule [range-buf-zero]:
+      #range (#buf(W1:Int, X:Int), S2:Int, W2:Int) => #buf(W2, 0)
+      requires 0 <=Int X andBool X <Int 2 ^Int (8 *Int (W1 -Int (S2 +Int W2)))
+       andBool 0 <=Int S2 andBool 0 <=Int W2 andBool S2 +Int W2 <=Int W1
+      [simplification]
```
ehildenb, line 183: Reviewed.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
 
-    // lookup range in memory
-    rule #range(MEM, 0, W) => MEM  requires W ==Int lengthBytes(MEM) [simplification]
+    rule [range-padRightToWidth]:
+      #range(#padRightToWidth(_, BUF), 0, WIDTH) => BUF
+      requires lengthBytes(BUF) ==Int WIDTH [simplification]
```
ehildenb, line 187: Reviewed.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
-    rule #range((BUF1 +Bytes  _   ), START, WIDTH) =>  #range(BUF1, START                       , WIDTH                        )                                                                    requires START +Int WIDTH <=Int lengthBytes(BUF1)                       andBool 0 <=Int START andBool 0 <=Int WIDTH    [simplification]
-    rule #range((BUF1 +Bytes  BUF2), START, WIDTH) => (#range(BUF1, START                       , lengthBytes(BUF1) -Int START)) +Bytes (#range(BUF2, 0, START +Int WIDTH -Int lengthBytes(BUF1)))  requires                        lengthBytes(BUF1) <Int START +Int WIDTH andBool #range(0 <= START < lengthBytes(BUF1)) [simplification]
+    // Memory update
 
-    rule #range(#padRightToWidth(_, BUF), 0, WIDTH) => BUF requires lengthBytes(BUF) ==Int WIDTH [simplification]
+    rule [memUpdate-is-empty]:
+      _:Bytes [ S:Int := _:Bytes ] => .Bytes
+      requires S <Int 0
+      [simplification]
```
ehildenb, line 194: Reviewed.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
 
-    // WS [ START := WS' ]
+    rule [memUpdate-as-concat-inside]:
+      B1:Bytes [ S:Int := B2:Bytes ] =>
+        #range(B1, 0, S) +Bytes B2 +Bytes #range (B1 , S +Int lengthBytes(B2), lengthBytes(B1) -Int (S +Int lengthBytes(B2)))
+      requires 0 <=Int S andBool S +Int lengthBytes(B2) <Int lengthBytes(B1)
+      [simplification(60)]
```
ehildenb, line 200: Reviewed.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
 
-    // re-order memory write
-    rule MEM [ K1 := BUF1:Bytes ] [ K2 := BUF2:Bytes ] => MEM [ K2 := BUF2 ] [ K1 := BUF1 ]
-      requires lengthBytes(BUF2) <=Int K1 -Int K2 andBool K2  <Int K1 [simplification]
+    rule [memUpdate-as-concat-outside-1]:
+      B1:Bytes [ S:Int := B2:Bytes ] => #range(B1, 0, S) +Bytes B2
+      requires 0 <=Int S andBool lengthBytes(B1) <=Int S +Int lengthBytes(B2)
+      [simplification(60)]
```
ehildenb, line 205: Reviewed.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
 
-    // overwritten memory write
-    rule MEM [ K1 := BUF1 ] [ K2 := BUF2 ] => MEM [ K2 := BUF2 ]
-      requires K1 +Int lengthBytes(BUF1) <=Int K2 +Int lengthBytes(BUF2) andBool K2 <=Int K1 [simplification]
+    rule [memUpdate-reorder]:
+      B:Bytes [ S1:Int := B1:Bytes] [ S2:Int := B2:Bytes ] => B [ S2 := B2 ] [ S1 := B1 ]
+      requires 0 <=Int S2 andBool S2 <Int S1 andBool S2 +Int lengthBytes(B2) <=Int S1
+      [simplification]
```
ehildenb, line 210: Reviewed.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
 
-    rule MEM [ K := BUF ] => MEM +Bytes #buf(K -Int lengthBytes(MEM), 0) +Bytes BUF
-      requires lengthBytes(MEM) <=Int K [simplification]
+    rule [memUpdate-subsume]:
+      B:Bytes [ S1:Int := B1:Bytes] [ S2:Int := B2:Bytes ] => B [ S2 := B2 ]
+      requires 0 <=Int S2 andBool S2 <=Int S1 andBool S1 +Int lengthBytes(B1) <=Int S2 +Int lengthBytes(B2)
+      [simplification]
```
ehildenb, line 215: Reviewed.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
-    rule [lengthBytes-buf]:      lengthBytes(#buf(SIZE, _))                    => SIZE   requires 0 <=Int SIZE                        [simplification]
-    rule [lengthBytes-range]:    lengthBytes(#range(_, START, WIDTH))          => WIDTH  requires 0 <=Int START andBool 0 <=Int WIDTH [simplification]
-    rule [lengthBytes-prtw]:     lengthBytes(#padRightToWidth(W:Int, B:Bytes)) => maxInt(lengthBytes(B), W)                           [simplification]
+    rule [lengthBytes-geq-zero]: 0 <=Int lengthBytes ( _ )                     => true                                     [simplification, smt-lemma]
+    rule [lengthBytes-concat]:   lengthBytes(BUF1 +Bytes BUF2)                 => lengthBytes(BUF1) +Int lengthBytes(BUF2) [simplification]
+    rule [lengthBytes-buf]:      lengthBytes(#buf(S, _))                       => S requires 0 <=Int S                     [simplification]
+    rule [lengthBytes-range]:    lengthBytes(#range(_, S:Int, W:Int))          => maxInt(0, W) requires 0 <=Int S          [simplification]
+    rule [lengthBytes-prtw]:     lengthBytes(#padRightToWidth(W:Int, B:Bytes)) => maxInt(lengthBytes(B), W)                [simplification]
```
ehildenb, line 223: Reviewed.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+      [simplification]
 
-    rule #range( M [ N1 := BUF1:Bytes ] [ N2 := BUF2:Bytes ], L, K ) => #range( M [ N2 := BUF2 ], L, K )
-        requires 0 <Int K andBool N1 +Int lengthBytes(BUF1) <=Int L [simplification]
+    rule [range-reform-base]:
+      #range(B:Bytes, S:Int, minInt(W1:Int, W2:Int)) +Bytes #buf(maxInt(0, W3:Int), 0) => #range (B, S, W1)
+      requires W2 ==Int lengthBytes(B) -Int S
+       andBool W3 ==Int S +Int W1 -Int lengthBytes(B)
+       andBool 0 <=Int S andBool 0 <=Int W1 andBool 0 <=Int W2
+      [simplification]
```
ehildenb, line 134: Some observations (all apply to the next lemma `range-reform-concat` below too):

```
W3 ==Int S +Int W1 -Int lengthBytes(B)
```

Can be re-arranged as:

```
W1 ==Int W3 +Int (lengthBytes(B) -Int S)
```

Which can then be re-written as:

```
W1 ==Int W2 +Int W3
```

So the side-condition can become:

```suggestion
    rule [range-reform-base]:
      #range(B:Bytes, S:Int, minInt(W1:Int, W2:Int)) +Bytes #buf(maxInt(0, W3:Int), 0) => #range (B, S, W1)
      requires W1 ==Int W2 +Int W3
       andBool W2 ==Int lengthBytes(B) -Int S
       andBool 0 <=Int S andBool 0 <=Int W1 andBool 0 <=Int W2
      [simplification]
```

which is perhaps clearer for understanding this lemma.

ehildenb: With the above presentation of the side condition, I was able to easily show that the lemma holds. I'd recommend making the adjustment.

include/kframework/lemmas/bitwise-simplification.k
PetarMax: Yes, that will work.

include/kframework/lemmas/bytes-simplification.k
PetarMax: Yes, good point - but `X <Int 2 ^Int (8 *Int (32 -Int W))`, right?

It is subsumed completely, yes, by `range-buf-zero`.

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -68,71 +68,159 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+      #range(#buf(W:Int, 0), S1:Int, W1:Int) => #buf(W1, 0)
+      requires 0 <=Int W andBool 0 <=Int S1 andBool 0 <=Int W1
+      [simplification]
 
-    rule #range(BA:Bytes, 0, LEN) => BA requires lengthBytes(BA) ==Int LEN [simplification]
+    rule [range-buf-zero-2-conc]:
+      #range(B, S1:Int, W1:Int) => #buf(W1, 0)
+      requires 0 <=Int S1 andBool 0 <=Int W1
+       andBool B ==K #buf(lengthBytes(B), 0)
+      [concrete(B), simplification]
```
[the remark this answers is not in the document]
PetarMax: Yes, better.

include/kframework/lemmas/bytes-simplification.k
PetarMax: Yes, that is clearer.

include/kframework/lemmas/bitwise-simplification.k
PetarMax: It is not clear to me that Optimism proofs will pass with this, because there is no built-in limit on the length of `Bytes`. There probably should be one, since `Bytes` of length `pow256` could never possibly appear?

include/kframework/lemmas/bytes-simplification.k
PetarMax: Ok, this is bizarre. When rewritten as you suggest, the claim
```
    claim [range-36]:
      <k> runLemma ( #range(B:Bytes, 5 +Int S2:Int, W:Int) ) => doneLemma ( #range(#range(B, 5, lengthBytes(B) -Int 5), S2, W) ) ... </k>
      requires 0 <=Int S2 andBool (5 +Int S2 <=Int lengthBytes(B))
```
does not pass because `B` is not concrete and the simplification is not triggered, and it shouldn't pass, actually. But originally, with the two `concrete`s separated, the simplification is somehow triggered. I'm not sure who should look at this, perhaps @radumereuta if frontend or @jberthold if backend?

radumereuta: If you duplicate attributes on a rule, only one of them is taken into consideration.
```
// test.k
  rule A + B => A +Int B [concrete(A), concrete(B)]
// parsed.txt
  rule `_+__TEST-SYNTAX_Exp_Exp_Exp`(#SemanticCastToInt(A),#SemanticCastToInt(B))=>`_+Int_`(#SemanticCastToInt(A),#SemanticCastToInt(B)) requires #token("true","Bool") ensures #token("true","Bool")
   [concrete(B),
    org.kframework.attributes.Location(Location(10,8,10,25)),
    org.kframework.attributes.Source(Source(/home/radu/work/test/test.k)),
    org.kframework.definition.Production(syntax #RuleContent ::= #RuleBody [klabel(#ruleNoConditions),
    symbol])]
```
The front end should throw an error in this case. I'll make an issue.

radumereuta: I've added a check in the front end: https://github.com/runtimeverification/k/pull/3438
You should get an error message from now on.

ehildenb (approved):

PR #1774, opened 2023-04-25
Port `kevm kompile` logic to Python

tothtamas28:
* Consolidates KEVM kompilation logic under module `kevm_pyk.kompile`.
* Decouples `foundry_kompile` from `kevm_kompile`.
* Introduces  flags `--target` and `--kevm-lib` for `kevm-pyk kompile` and uses their values to infer `ccopts` and other `kompile` arguments.


tothtamas28:
@goodlyrottenapple, can you look into the Nix failure? (Simply deleting `kevm.patch` and the corresponding attribute didn't solve the issue.)

ehildenb:
So is the idea that the configuration living in `kompile` module would basically be moved to `kbuild` declaritive build setup?

It looks good, want to wait for Nix things to be fixed for approval.

ehildenb:
@tothtamas28 I think you need to remove this line: https://github.com/runtimeverification/evm-semantics/blob/31b6d7a030635451cad86a9b281138ee4bebda27/flake.nix#L88

ehildenb:
Also, not sure if it's still necessary, but maybe you need to detect if it's a Nix build, and make sure you have the same effect as teh patch: https://github.com/runtimeverification/evm-semantics/blob/master/package/nix/kevm.patch

tothtamas28:
> looks like you already tried something similar. I used a build flag instead: #1777

Yes, I ended up with that solution as well. (Adding the file didn't work because it's not there for the installed script.)

PR #1783, opened 2023-04-27
Further simplify kompilation

tothtamas28:
* Factor out remaining kompilation logic from `bin/kevm` and `Makefile`
* Extract kompilation template from `make` targets
* Use `kevm-pyk` directly for kompilation in `make` targets


kevm-pyk/src/kevm_pyk/config.py
```diff
@@ -1,10 +1,31 @@
...
+    env = os.getenv(var)
+    if env is None:
+        raise ValueError(f'Environment variable not set: {var}')
+    return env
+
+
 NIX_BUILD: Final = bool(os.getenv('NIX_BUILD'))
+
+KEVM_LIB: Final = Path(_env_or_raise('KEVM_LIB')).resolve()
+check_dir_path(KEVM_LIB)
```
ehildenb, line 23: We definitely should not be requiring that `KEVM_LIB` is always set, we used to do that and it was horrible user experience for KEVM installed as a package.

We should do:

- If the user sets `KEVM_LIB` use that version.
- Otherwise, look at where `which kevm` returns (usually something like `/usr/bin/kevm` or `/path/to/evm-semantics/.build/usr/bin/kevm`), and go up one directory to to `lib/` (for example `/usr/lib/kevm` or `/path/to/evm-semantics/.build/usr/lib/kevm`).

Then, if neither of those produce a `KEVM_LIB` that exists, we can fail.

ehildenb: I guess here we're assuming that we're always calling this from `kevm` script anyway, which will make sure to set `KEVM_LIB`? Maybe it's OK then, but we could also just move that logic to this script I suppose.

ehildenb (approved):

PR #1785, opened 2023-04-28
Annotate some rules as definedness-preserving

jberthold:
The annotations are relevant to faster execution using an optimised backend (which does not support definedness checks).
Also contains a few simplifications to rules to reduce overhead.


include/kframework/buf.md
```diff
@@ -47,8 +47,8 @@ module BUF
 
     rule #buf(SIZE, DATA) => #padToWidth(SIZE, #asByteStack(DATA %Int (2 ^Int (SIZE *Int 8))))
       requires 0 <Int SIZE
-      [concrete]
-    rule #buf(_SIZE, _) => .Bytes [owise, concrete] // SIZE <= 0
+      [concrete, preserves-definedness]
+    rule #buf(_SIZE, _) => .Bytes [owise, concrete, preserves-definedness] // SIZE <= 0
```
ehildenb, line 51: Why does this need to be marked as preserving definedness? This is a total function, and `.Bytes` is as well: 
image

include/kframework/evm-types.md
```diff
@@ -281,7 +280,7 @@ A cons-list is used for the EVM wordstack.
 
 ```k
     syntax Int ::= #sizeWordStack ( WordStack )       [function, total, smtlib(sizeWordStack)]
-                 | #sizeWordStack ( WordStack , Int ) [function, total, klabel(sizeWordStackAux), smtlib(sizeWordStackAux)]
+                 | #sizeWordStack ( WordStack , Int ) [function, total, klabel(sizeWordStack), smtlib(sizeWordStack)]
  // -----------------------------------------------------------------------------------------------------------------------
```
ehildenb, line 284: This seems a strange modification.... Why would we want to give this the same label as the single-argument variant?

include/kframework/evm-types.md
```diff
@@ -360,16 +359,16 @@ Bytes helper functions
     syntax Bytes ::= #range ( Bytes , Int , Int ) [function, total]
  // ---------------------------------------------------------------
     rule                #range(_, START, WIDTH)  => .Bytes                                                                       requires notBool (WIDTH >=Int 0 andBool START >=Int 0)
-    rule [bytesRange] : #range(WS, START, WIDTH) => substrBytes(padRightBytes(WS, START +Int WIDTH, 0), START, START +Int WIDTH) requires WIDTH >=Int 0 andBool START >=Int 0 andBool START <Int lengthBytes(WS)
+    rule [bytesRange] : #range(WS, START, WIDTH) => substrBytes(padRightBytes(WS, START +Int WIDTH, 0), START, START +Int WIDTH) requires WIDTH >=Int 0 andBool START >=Int 0 andBool START <Int lengthBytes(WS) [preserves-definedness]
```
ehildenb, line 362: `#range` is total.... I guess since `#range` is total, it should be already preserving definedness? That should be derivable from the axioms of total functions I think. Basically, "for every single-element input, it will return a single value" (totality, functionality), and the variables here are all element variables (not set variables). So I think we can just make the definedness checker stronger. Currently we only inspect the RHS and REQ clause to see if they are all total functions and constructors, but another sufficient criteria could be: The LHS only contains total functions, constructors, domain values, and element variables. Is taht true?

ehildenb:
I think all of these cases can be handled by making the "preserves definedness" pass handle the case where the LHS is all total functions, constructors, domain values, and element variables. In that case, we shouldn't even need to inspect the RHS. Someone should check that, but it seems to be the case if you look at the axioms of `total` and `function`.

include/kframework/buf.md
jberthold: That's true, an oversight.
About totality see the other comment/response

include/kframework/evm-types.md
jberthold: The `total` attribute to a function is currently not considered to infer definedness-preservation. 
I agree we could modify the analysis to just assume a rule is preserving definedness  when it uses a total function on the LHS.
Currently the rule won't be considered definedness-preserving since `padRightBytes` , a partial function, is used on the RHS. So the `preserves-definedness` is re-stating the totality of `#range`, in a way.

jberthold:
Should be obsolete with https://github.com/runtimeverification/hs-backend-booster/issues/167

PR #1786, opened 2023-04-28
Increase size of dummy return data for `expectRevert` to 512 bytes

lucasmt:
Since the `expectRevert` cheatcode causes execution to continue after a function reverts, it needs to provide dummy return data, otherwise it will not pass the checks that the Solidity compiler performs on the return value. This issue was partially addressed in PR #1564 by returning 32 bytes of 0, but this is not enough if the compiler expects the return data to be greater than 32 bytes. Therefore, this PR makes the following changes:

* Increase the dummy return data from 32 to 512 bytes of 0, matching the behavior implemented in Forge (see [here](https://github.com/foundry-rs/foundry/blob/91f2a442119106a4b2b0cc639e5a01d542294fdb/evm/src/executor/inspector/cheatcodes/expect.rs#L22))
* Modifies the `test_expectRevert_returnValue` to test a return type that requires more than 32 bytes. This test did not pass before the change above, but passes after.


anvacaru (approved):
Looks good to me. You still need to update the expected files to pass CI:
- tests/foundry/foundry-list.check.expected
- tests/foundry/golden/*.expected

lucasmt:
> Looks good to me. You still need to update the expected files to pass CI:
> 
>     * tests/foundry/foundry-list.check.expected
> 
>     * tests/foundry/golden/*.expected

@anvacaru Ah, sorry, I didn't realize. Is there an automated way to do that or do I need to just do it manually?

anvacaru:
@lucasmt I updated the files for now, but there are a few ways:

For the golden test files you can do `make test-foundry-kcfg-diff -j2 FOUNDRY_PAR=12 CHECK=cp` - takes around 25 mins
For the foundry-list you can do
- `make test-foundry-list  -j2 FOUNDRY_PAR=12 CHECK=cp` - this runs all the foundry proofs, so that it might take a while. 
-  You can instead grab the diff from the failed stage and apply it locally. Save the diff in a file (i.e., `tests/foundry/diff`) and then run `git apply diff`. Make sure to format the beginning of the `diff` file to modify the same file.
```diff
diff --git b/tests/foundry/foundry-list.check.expected a/tests/foundry/foundry-list.check.expected
index e6a59b0..581cb2c 100644
--- b/tests/foundry/foundry-list.check.expected
+++ a/tests/foundry/foundry-list.check.expected
```

tests/foundry/foundry-list.check.expected
PetarMax, line 1: I'm not sure what this file does.

tests/foundry/test/ExpectRevertTest.t.sol
```diff
@@ -120,8 +134,14 @@ contract ExpectRevertTest is Test {
...
+        assertEq0(returnValueBytes, "");
+
         vm.expectRevert("Error");
-        bytes memory returnValue = reverter.revertUnless(false);
-        assertEq0(returnValue, "");
+        (uint256 fst, uint256 snd) = reverter.returnTupleUnless(true);
+        assertEq(fst, 0);
+        assertEq(snd, 0);
     }
 }
```
PetarMax, line 147: How does this connect with the fact that the dummy return data is 512 bytes? Is it that is passes now and would not pass in the previous version?

PetarMax:
Is there/can we craft an example in which more than 512 bytes is returned? What would happen in Forge then?

tests/foundry/test/ExpectRevertTest.t.sol
lucasmt: Yes, because previously it would only return 32 bytes, which is not enough for `(uint256, uint256)` (64 bytes)

lucasmt:
@PetarMax I'll merge this for now and then try to work on an example with more than 512 bytes. My guess is that Forge should also fail on that, but it's unlikely to have a function that would require the return data to be 512 bytes. This would correspond to 16 `uint256`s.

lucasmt:
@PetarMax For the record, if I try to create a function with a return type that needs 512 bytes, the compiler gives me a "stack too deep" error.

Issue #1759, opened 2023-04-21
Initialize KCFGs in parallel


ehildenb:
Even when specifying `-jN` to run a bunch of proofs, KCFG initialization still runs in serial, which is very expensive time-wise for large proof-suites.

It would be nice if that ran in parallel instead.

PR #1790, opened 2023-05-01
Parallelize initial proof generation process

ehildenb:
Fixes: https://github.com/runtimeverification/evm-semantics/issues/1759

Currently, when running the Foundry test-suite, it takes about 20 minutes just to initialize all teh proofs. This is because each initialization is running in serial, even if the user specifies higher parallelism.

This PR:

- Changes `Contract.method_by_name` to return a `dict[str, Method]` instead of the original `tuple[Method]`. Also fixes a bug where it only was able to return the `setUp` method of a contract, which did not matter before because we were only using it get the `setUp` method, but matters now.
- Factors out the creation of the initial Proof into it's own function `_method_to_apr_proof`.
- Runs this function in parallel for initializing each `APRProof | APRBMCProof`.


kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -355,9 +356,9 @@ def foundry_prove(
     br = BugReport(foundry_root / 'bug_report') if bug_report else None
     foundry = Foundry(foundry_root, bug_report=br)
 
-    ag_proofs_dir = foundry.out / 'ag_proofs'
-    if not ag_proofs_dir.exists():
-        ag_proofs_dir.mkdir()
+    save_directory = foundry.out / 'ag_proofs'
```
geo2a, line 359: Just a nitpick: rename `'ag_proofs'` to `'apr_proofs'`, or perhaps even to just `'proofs'`.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -391,79 +392,37 @@ def foundry_prove(
     setup_methods: dict[str, str] = {}
     contracts = unique({test.split('.')[0] for test in tests})
     for contract_name in contracts:
+        if 'setUp' in foundry.contracts[contract_name].method_by_name:
+            setup_methods[contract_name] = f'{contract_name}.setUp'
+
+    def _init_apr_proof(_init_problems: tuple[str, str]) -> APRProof | APRBMCProof:
```
geo2a, line 398: Consider renaming the argument `_init_problems` -> `_init_problem`, since it represents a single pair of contract name and contract method.

geo2a (approved):
I'm approving to unblock to PR, consider addressing two minor comments on naming.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -355,9 +356,9 @@ def foundry_prove(
     br = BugReport(foundry_root / 'bug_report') if bug_report else None
     foundry = Foundry(foundry_root, bug_report=br)
 
-    ag_proofs_dir = foundry.out / 'ag_proofs'
-    if not ag_proofs_dir.exists():
-        ag_proofs_dir.mkdir()
+    save_directory = foundry.out / 'ag_proofs'
+    if not save_directory.exists():
+        save_directory.mkdir()
```
tothtamas28, line 361:
```suggestion
    save_directory.mkdir(exist_ok=True)
```

tothtamas28 (approved):

kevm-pyk/src/kevm_pyk/foundry.py
ehildenb: Done.

kevm-pyk/src/kevm_pyk/foundry.py
ehildenb: Done.

kevm-pyk/src/kevm_pyk/foundry.py
ehildenb: Done.

Issue #1795, opened 2023-05-02
Reuse `setUp` output even if bytecode changes


ehildenb:
Currently, we are not reusing the output of the `setUp` method if the bytecode changes, because the bytecode hash is contained in the `setUp` method directly.

We should instead have the `setUp` method be a hash of the contents of:

- The Solidity code in the `setUp` method.
- The Storage layout itself.

That way, we can continue to re-use the post-state of the setup method even if the property test changes.
We should also print a warning when we detect a bytecode change (but not a setup method digest change), so that the user knows that we are reusing the setup method for speed, but they should reinit afterwards to verify that the proof works directly as well.

PR #1798, opened 2023-05-04
Reuse setup output

nwatson22:
Per #1795, this PR improves our ability to detect when a test and its associated setup method has already been run and when it has been invalidated and needs to be rerun. Each test method that is run and any needed setup methods all save a digest to a file, and will automatically reinit when that digest is invalidated. They also save their contract's digest to a file to provide the user with a warning when a method is not being rerun even though the contract changed (should only happen with setup methods).


Makefile
```diff
@@ -513,7 +513,7 @@ foundry-forge-test: foundry-forge-build
 	cd $(foundry_dir) && forge test --ffi
 
 $(foundry_out):
-	rm -rf $@
+#   rm -rf $@
```
ehildenb, line 516: Just remove this, don't comment.

kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -39,30 +40,87 @@ class Method:
...
+
+        def up_to_date(self, digest_file: Path) -> bool:
+            if not digest_file.exists():
+                return False
+            text = digest_file.read_text()
+            try:
+                digest_dict = json.loads(text)
+            except JSONDecodeError:
+                _LOGGER.warning('json text: ')
+                exit(1)
```
ehildenb, line 87:
```suggestion
            digest_dict = json.loads(digest_file.read_text())
```

No need to re-catch this error just to exit, Python will exit with the appropriate error code here.

kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -144,13 +202,28 @@ def __init__(self, contract_name: str, contract_json: dict, foundry: bool = Fals
...
+                            continue
+                        if node1['functionSelector'] == method_identifier:
+                            return node1
+            return None
+
         _methods = []
         for method in contract_json['abi']:
             if method['type'] != 'function':
                 continue
+            _LOGGER.info(f'Current method: {method["name"]}')
```
ehildenb, line 221: Is this debugging message supposed to stay?

kevm-pyk/src/tests/unit/test_unit.py
```diff
@@ -206,7 +206,8 @@ def test_method_sig_from_abi() -> None:
     "aggregate((address,bytes)[])": "252dba42"
   },
   "ast": {
-    "absolutePath": "lib/forge-std/src/interfaces/IMulticall3.sol"
+    "absolutePath": "lib/forge-std/src/interfaces/IMulticall3.sol",
+    "nodes": []
   },
```
ehildenb, line 211: Do we need to adjust `tests/foundry/foundry.toml` to include this extra output?

kevm-pyk/src/kevm_pyk/solc_to_k.py
nwatson22: I was intending for it to print out the file it was trying to parse when it crashed, but it was a typo.

Makefile
nwatson22: Removed.

kevm-pyk/src/kevm_pyk/solc_to_k.py
nwatson22: Removed.

kevm-pyk/src/tests/unit/test_unit.py
nwatson22: I think what's happening here is that we're already specifying the `"ast"` to be included, and the `"nodes"` is a subobject of that which should be generated whenever `"ast"` is. For the `foundry.toml` file, I haven't been able to find any documentation confirming this, but it seems the `"ast"` is generated by default when calling the solidity compiler through forge. The reason I had to change this test was that it was manually constructed and didn't have this json field that our constructors now expect.

kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -39,30 +40,92 @@ class Method:
...
+
+        def up_to_date(self, digest_file: Path) -> bool:
+            if not digest_file.exists():
+                return False
+            text = digest_file.read_text()
+            try:
+                digest_dict = json.loads(text)
+            except JSONDecodeError:
+                _LOGGER.warning(f'json text: {text}')
+                exit(1)
```
ehildenb, line 87:
```suggestion
            digest_dict = json.loads(digest_file.read_text())
```

No need to catch this exception.

ehildenb, line 87:
```suggestion
            digest_dict = json.loads(digest_file.read_text())
```

No need to catch this exception.

kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -39,30 +40,92 @@ class Method:
...
+
+        def contract_up_to_date(self, digest_file: Path) -> bool:
+            if not digest_file.exists():
+                return False
+            text = digest_file.read_text()
+            try:
+                digest_dict = json.loads(text)
+            except JSONDecodeError:
+                _LOGGER.warning(f'json text: {text}')
+                exit(1)
```
ehildenb, line 103:
```suggestion
            digest_dict = json.loads(digest_file.read_text())
```

kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -144,13 +207,27 @@ def __init__(self, contract_name: str, contract_json: dict, foundry: bool = Fals
...
+            for node in contract_ast['nodes']:
+                if node['nodeType'] == 'ContractDefinition':
+                    for node1 in node['nodes']:
+                        if not node1['nodeType'] == 'FunctionDefinition':
+                            continue
+                        if not 'functionSelector' in node1:
+                            continue
+                        if node1['functionSelector'] == method_identifier:
+                            return node1
+            return None
```
ehildenb, line 220:
```suggestion
            function_definitions = [
                node for contract_defn in contract_ast['nodes'] if contract_defn['nodeType'] == 'ContractDefinition'
                for node in contract_defn['nodes'] if node['nodeType'] == 'FunctionDefinition'
            ]
            matching_functions = [
                fdefn for fdefn in function_definitions if 'functionSelector' in fdefn and fdefn['functionSelector'] == method_identifier
            ]
            return single(matching_functions)
```

kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -39,30 +40,92 @@ class Method:
         sort: KSort
         arg_names: tuple[str, ...]
         arg_types: tuple[str, ...]
+        contract: Contract
```
ehildenb, line 43: Hmmm, it seems strange to me that we have to pass in the `contract` to the `Method`, but `Method` is a nested class in `Contract`. I guess it doesn't matter type-wise.

But if we're going to do this, I guess we shouldn't also store `contract_name` separately, instead just use `self.contract.name` everywhere.

kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -144,13 +207,27 @@ def __init__(self, contract_name: str, contract_json: dict, foundry: bool = Fals
...
+                if node['nodeType'] == 'ContractDefinition':
+                    for node1 in node['nodes']:
+                        if not node1['nodeType'] == 'FunctionDefinition':
+                            continue
+                        if not 'functionSelector' in node1:
+                            continue
+                        if node1['functionSelector'] == method_identifier:
+                            return node1
+            return None
+
```
ehildenb, line 221:
```suggestion
        contract_ast = single([node for node in self.contract_json['ast']['nodes'] if node['nodeType'] == 'ContractDefinition' and node['name]' == self.name])
        function_asts = {node['functionSelector']: node for node in contract_ast['nodes'] if node['nodeType'] == 'FunctionDefinition' and 'functionSelector' in node} 
```

kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -144,13 +207,27 @@ def __init__(self, contract_name: str, contract_json: dict, foundry: bool = Fals
...
         for method in contract_json['abi']:
             if method['type'] != 'function':
                 continue
             msig = method_sig_from_abi(method)
-            mid = int(evm['methodIdentifiers'][msig], 16)
-            _m = Contract.Method(msig, mid, method, contract_name, self.sort_method)
+            method_selector: str = str(evm['methodIdentifiers'][msig])
+            method_ast = method_ast_from_contract_ast(self.contract_json['ast'], method_selector)
+            mid = int(method_selector, 16)
+            _m = Contract.Method(msig, mid, method, method_ast, self, self.sort_method)
```
ehildenb, line 230:
```suggestion
            method_selector: str = str(evm['methodIdentifiers'][msig])
            mid = int(method_selector, 16)
            method_ast = function_asts[method_selector] if method_selector in function_asts else None
            _m = Contract.Method(msig, mid, method, method_ast, self, self.sort_method)
```

ehildenb, line 230:
```suggestion
            method_selector: str = str(evm['methodIdentifiers'][msig])
            mid = int(method_selector, 16)
            method_ast = function_asts[method_selector] if method_selector in function_asts else None
            _m = Contract.Method(msig, mid, method, method_ast, self, self.sort_method)
```

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -391,8 +402,29 @@ def foundry_prove(
     setup_methods: dict[str, str] = {}
     contracts = unique({test.split('.')[0] for test in tests})
     for contract_name in contracts:
-        if 'setUp' in foundry.contracts[contract_name].method_by_name:
-            setup_methods[contract_name] = f'{contract_name}.setUp'
+        contract = foundry.contracts[contract_name]
+        if 'setUp' in contract.method_by_name:
+            setup_method = contract.method_by_name['setUp']
+            setup_methods[contract_name] = f'{setup_method.contract_name}.{setup_method.name}'
```
ehildenb, line 408: Why was this change made? It doesn't seem to compute anything different...

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -403,7 +435,7 @@ def _init_apr_proof(_init_problem: tuple[str, str]) -> APRProof | APRBMCProof:
             contract,
             method,
             save_directory,
-            reinit=reinit,
+            reinit=(method.qualified_name in out_of_date_methods),
```
ehildenb, line 438:
```suggestion
            reinit=(method.qualified_name in out_of_date_methods or reinit),
```

ehildenb: If the user specifies to reinit, we should always do so.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -443,6 +475,11 @@ def run_cfg_group(tests: list[str]) -> dict[str, bool]:
     failed = [setup_cfg for setup_cfg, passed in results.items() if not passed]
     if failed:
         raise ValueError(f'Running setUp method failed for {len(failed)} contracts: {failed}')
+    contracts = unique({test.split('.')[0] for test in tests})
+    _LOGGER.info(f'contracts: {contracts}')
```
ehildenb, line 479: I guess this is for debugging? Can it be removed?

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -695,6 +732,7 @@ def _method_to_apr_proof(
         if method_name != 'setUp' and 'setUp' in contract.method_by_name:
             setup_digest = f'{contract_name}.setUp:{contract.digest}'
             _LOGGER.info(f'Using setUp method for test: {test}')
+            _LOGGER.info(f'method.digest: {method.digest}')
```
ehildenb, line 735: I'm guessing this is also for debugging and can be removed?

kevm-pyk/src/kevm_pyk/foundry.py
nwatson22: If the reinit flag set I mark all methods as out of date so we can print that they're out of date.

kevm-pyk/src/kevm_pyk/foundry.py
nwatson22: Removed

kevm-pyk/src/kevm_pyk/foundry.py
nwatson22: Removed

kevm-pyk/src/kevm_pyk/solc_to_k.py
nwatson22: I applied this change.

nwatson22: I applied this change.

kevm-pyk/src/kevm_pyk/solc_to_k.py
nwatson22: I applied this change.

kevm-pyk/src/kevm_pyk/solc_to_k.py
nwatson22: I applied this change.  I was trying to figure out why we were failing to parse the digest file, but this crash seems to have stopped happening. I wonder if the old digest format of just one hash was somehow being cached.

kevm-pyk/src/kevm_pyk/solc_to_k.py
nwatson22: Same as in the other JSONDecodeError case.

kevm-pyk/src/kevm_pyk/solc_to_k.py
nwatson22: I changed it so we pass in the contract name and digest instead of a reference to the whole contract.

kevm-pyk/src/kevm_pyk/foundry.py
nwatson22: I might have been trying to isolate a crash. It's changed back now.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -391,9 +402,32 @@ def foundry_prove(
     setup_methods: dict[str, str] = {}
     contracts = unique({test.split('.')[0] for test in tests})
     for contract_name in contracts:
+        #          contract = foundry.contracts[contract_name]
+        #          if 'setUp' in contract.method_by_name:
+        #              setup_method = contract.method_by_name['setUp']
+        #              setup_methods[contract_name] = f'{setup_method.contract_name}.{setup_method.name}'
```
nwatson22, line 408:
```suggestion
```

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -389,11 +400,32 @@ def foundry_prove(
...
+            _LOGGER.info(f'Method {method.qualified_name} is out of date')
+        else:
+            _LOGGER.info(f'Method {method.qualified_name} skipped because it is up to date')
+            if not method.contract_up_to_date(foundry.out / 'digest'):
+                _LOGGER.warning(
+                    f'Method {method.qualified_name} skipped because digest was up to date, but the contract it is a part of has changed.'
+                )
+
+    _LOGGER.info(f'out of date methods: {out_of_date_methods}')
+
```
nwatson22, line 428:
```suggestion
```

nwatson22:
I tested locally that:
- Changing the fields of the contract results in both the test method and the associated setUp method are reinitialized
- Changing the setUp method results in both being reinitialized
- Changing the test method results in only the test method being reinitialized, with a warning printed.
- Changing nothing results in neither method being reinitialized.

ehildenb:
Looking at CI, it appears that it thinks the contracts are out of date on second run, which causes it to timeout: 
image

These methods should probably not be considered out of date on that second run.

Also looking at the Foundry Prove stage, it looks like it says they are "out of date" when they have not been initialized the first time (probably because the digest file doesn't exist yet). So it may be nice to have a better message there, like "Method ... has not been initialized." instead of "Method ... is out of date": image

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -470,7 +500,12 @@ def run_cfg_group(tests: list[str]) -> dict[str, bool]:
         raise ValueError(f'Running setUp method failed for {len(failed)} contracts: {failed}')
 
     _LOGGER.info(f'Running test functions in parallel: {tests}')
-    return run_cfg_group(tests)
+    results = run_cfg_group(tests)
+
+    for method in test_methods:
+        method.update_digest(foundry.out / 'digest')
+
+    return results
```
ehildenb, line 508: Why are we updating the digests _after_ running the proofs? Could we do it much earlier, after we've initialized the KCFGs? For example, we could do it in `run_cfg_group` right after we do `write_proof()`?

That way, if we kill a proof partway through, we don't lose the up-to-date digest (which should be the same if the solidity code itself didn't change).

Does this make sense?

nwatson22: Yes, that makes sense. I was thinking of it in terms of the proof being up to/out of date, but yes, if it's stopped part of the way through it will currently be reinitialized.

nwatson22: Is there any reason we can't just update each digest right after we check it? Putting it after `write_proof()` would result in it being called a bunch of times unnecessarily, I think, since that is in `advance_proof()`.

ehildenb (approved):

kevm-pyk/src/kevm_pyk/foundry.py
ehildenb: Yep, makes sense!

Issue #1742, opened 2023-04-19
Load entire basic blocks at a time


ehildenb:
One of the problems of proof-reuse with KEVM is taht we have a flat byte-array representing the current program, and use a program-counter to point at an offset into that byte-array. This means that even if we do a proof about a chunk of a program that remains the same, another part of the program changing may cause that proof to become invalid. This is annoying in the CI setting, for example, where loop invariants need to mention specific chunks of the program.

One way to make this less of a problem is to load entire basic blocks of the program at a time. Essentially, we would:

- At program-load time, break the bytecode up into basic-blocks directly, and store the blocks in a map of `PROGRAM_COUNTER_BLOCK_START |-> bytecode_chunk`
- Each time we step into a basic block, we pull that entire bytecode chunk onto the `<basic-block>` cell, and execute that by consuming it from the beginning (until it is empty, then we load the next basic block).

This would enable us to write proofs about chunks of bytecode at more flexible offsets rather than having them all be fixed together. It may enable more proof-reuse.

--------

Tasks:

- [ ] Break KEVM execution into basic-blocks when loading programs (any time `<program>` cell changes). Use a new `#loadProgram` construct which:
  - Updates the `<program>` cell.
  - Loads the `<basic-blocks>` cell. As a first cut, just break the program into `JUMPDEST` labelled basic blocks.
  - Whenever we have an empty `<basic-block>` cell, load the basic-block at the current `<pc>` into this cell.
  - When we get to `#execute`, just take the first byte off `<basic-block>` cell as the next opcode, instead of reading at an offset into the `<program>`.
- [ ] Make an example looping proof with a single basic block.

Resources:

- Main execution driver: https://github.com/runtimeverification/evm-semantics/blob/master/include/kframework/evm.md#execution-macros
- Example hand-written EVM "interactive" test you can extend/use: https://github.com/runtimeverification/evm-semantics/blob/master/tests/interactive/sumTo10.evm. Example of running an "interactive" test: https://github.com/runtimeverification/evm-semantics/blob/0e7955ec833d363ad2aed1248fdcc87b84167d19/Makefile#L462
- Want to scale to target `make test-conformance -j12`.

lucasmt:
This makes sense to me. This should also make the proofs more composable, right? For example, we can symbolically execute a function while abstracting away the rest of the bytecode.

When writing the invariant, would we then identify the basic block by the bytecode in the `<basic-block>` cell?

PetarMax:
[...]

PR #1805, opened 2023-05-08
Rework program loading and opcode fetching

geo2a:
Work on #1742 

This PR changes the program loading and opcode fetching logic:
* The program is sliced by the `JUMPDEST` instructions, and the chunks are places into the `<basic-blocks>` cell of sort `Map` with the jump destination's PC as the key;
* The `<basic-block>` cell (singular) is populated with the current basic block, and the opcodes are bitten off it until it is either:
  - exhausted;
  - a `JUMP*` instruction is encountered;
* The new basic block is loaded onto the  `<basic-block>`
* The execution terminated as before, i.e. when the `step` rule passes `.NoOpCode` to `#next`.

NOTE: the basic blocks are not actually basic, because they can contain `JUMP*` instructions in the middle, and not only as the last instruction. However, this seems to be a good enough approximation for our purposes. If a jump is encountered in the middle of a block, we just abandon the current block and load the target one.

State of this PR today:
- [x] Passes `make test-conformance`, but I had to exclude three more tests, see the [diff](https://github.com/runtimeverification/evm-semantics/pull/1805/files#diff-19796954caf4ef545e3950aa659c4cf17566a82414d6d2c2936840e8d90158d2)
- [x] Passes `make foundry-prove`
- [ ] Fails 'make test-foundry-kcfg-diff`, the golden KFCGs need to updated to include new basic blocks configuration cells
- [ ] Fails the old non-Foundry `make test-prove`, the claims need to be updated to include new basic blocks configuration cells 

The last item, porting the archive of proofs to pass `make test-prove`, likely requires a lot of manual labor.


include/kframework/evm.md
```diff
@@ -284,6 +286,17 @@ OpCode Execution
...
+    syntax MaybeOpCode ::= "#unconsOpCode" "(" Bytes "," Schedule ")" [function, total]
+ // -------------------------------------------------------------------------------------------
+    rule #unconsOpCode(BA, SCHED) => #dasmOpCode(BA[0], SCHED) requires 0 <Int lengthBytes(BA)
+    rule #unconsOpCode(_, _)      => .NoOpCode                 [owise]
+
+    syntax Bytes ::= #dropOpCode(Bytes, MaybeOpCode)  [function, total]
+ // --------------------------------------------------------------
+    rule #dropOpCode(BA, _)         => .Bytes                                         requires lengthBytes(BA) ==Int 0
+    rule #dropOpCode(BA, .NoOpCode) => BA
+    rule #dropOpCode(BA, OP)        => substrBytes(BA, #widthOp(OP), lengthBytes(BA)) [owise]
```
ehildenb, line 299: You'll need to make a special case for `PUSH(N)` opcodes. The arguments for `PUSH(...)` are directly inline in the bytecode, so for example `PUSH(1) 80` is encoded in the bytecode as `0x6080`, and `PUSH(2) 80` is encoded in the bytecode as `0x610080`.

So you need to special-case the `PUSH(N)` opcodes in these methods, to select more than just the first byte (or drop more than just the first byte).

You may also consider `head(...)` and `tail(...)` as the names of these?

include/kframework/evm.md
```diff
@@ -1385,6 +1435,23 @@ The various `CALL*` (and other inter-contract control flow) operations will be d
...
+
+   rule #computeBasicBlocksAux(_PGM, .List, _, RESULT) => RESULT
+   rule #computeBasicBlocksAux(PGM,
+                            ListItem(NEXT_BB_START) JUMPDESTS,
+                            CURRENT_BB_START,
+                            RESULT) =>
+        #computeBasicBlocksAux(PGM,
+                            JUMPDESTS,
+                            NEXT_BB_START,
+                            RESULT (CURRENT_BB_START |-> #range(PGM, CURRENT_BB_START, NEXT_BB_START -Int CURRENT_BB_START)))
```
ehildenb, line 1454: Hmmm, it seems to me that relying on `qsort` makes this at least `n log(n)`, whereas just iterating over the original bytecode makes it `n`? Basically just consume the bytecode 1 byte at a time, and look for exactly the byte that is `JUMPDEST`, hwile keeping track of the offset, and then emit the map as you go?

geo2a: Yeah, makes sense, it's indeed easier to just traverse the bytecode the second time than trying to reuse the computed jumpdests. Done.

include/kframework/evm.md
```diff
@@ -1385,6 +1429,33 @@ The various `CALL*` (and other inter-contract control flow) operations will be d
...
+        #computeBasicBlocks(PGM, I +Int 1, I, RESULT)
+     requires PGM [ I ] ==Int 91
+      andBool I ==Int 0
+   rule #computeBasicBlocksWithinBound(PGM, I, J, RESULT) =>
+        #computeBasicBlocks(PGM, I +Int 1, I, RESULT (J |-> substrBytes(PGM, J, I)))
+     requires PGM [ I ] ==Int 91
+      andBool I >Int 0
+   rule #computeBasicBlocksWithinBound(PGM, I, J, RESULT) =>
+        #computeBasicBlocks(PGM, I +Int #widthOpCode(PGM [ I ]), J, RESULT)
+     requires notBool PGM [ I ] ==Int 91
```
ehildenb, line 1458: I guess `computeValidJumpDests` could just re-use `computeBasicBlocks` then, by inspecting each map element and keeping the key if the first item of the basic block is a `JUMPDEST`?

Or, do we really need `computeValidJumpDests` anymore? Right now it's used here: https://github.com/runtimeverification/evm-semantics/blob/master/include/kframework/evm.md#evm-control-flow

```
    rule <k> JUMP DEST => #endBasicBlock... </k>
         <pc> _ => DEST </pc>
         <jumpDests> DESTS </jumpDests>
      requires DEST in DESTS

         <jumpDests> DESTS </jumpDests>
      requires notBool DEST in DESTS
```

But these uses could be pretty easily replaced with:

```
    rule <k> JUMP DEST => #endBasicBlock... </k>
         <schedule> SCHED </schedule>
         <pc> _ => DEST </pc>
         <basic-blocks> ... DEST |-> BA ... </jumpDests>
      requires #unconsOpCode(SCHED, BA) ==K JUMPDEST

    rule <k> JUMP DEST => #end EVMC_BAD_JUMP_DESTINATION ... </k> [owise]
```

include/kframework/k-utils.md
```diff
@@ -0,0 +1,33 @@
...
+    rule getIntElementsSmallerThan (X, (ListItem(I:Int) L), RESULTS) => getIntElementsSmallerThan (X, L, RESULTS)             requires I >=Int X
+
+    rule getIntElementsGreaterThan (_, .List ,              RESULTS) => RESULTS
+    rule getIntElementsGreaterThan (X, (ListItem(I:Int) L), RESULTS) => getIntElementsGreaterThan (X, L, ListItem(I) RESULTS) requires I  >Int X
+    rule getIntElementsGreaterThan (X, (ListItem(I:Int) L), RESULTS) => getIntElementsGreaterThan (X, L, RESULTS)             requires I <=Int X
+
+```
+
+```k
+endmodule
```
ehildenb, line 32: I guess this can be dropped.

include/kframework/evm.md
geo2a: It indeed does not look like we need `<jumpDest>` anymore. I think we can use a even simpler side condition:

```
    rule <k> JUMP DEST => #endBasicBlock... </k>
         <schedule> SCHED </schedule>
         <pc> _ => DEST </pc>
         <basic-blocks> BLOCKS </jumpDests>
      requires DEST in_keys(BLOCKS)

    rule <k> JUMP DEST => #end EVMC_BAD_JUMP_DESTINATION ... </k> [owise]
```
since the the keys in the `BLOCKS` map are all JUMPDEST (and 0 if the first opcode is not JUMPDEST)

include/kframework/k-utils.md
geo2a: Done.

PR #1807, opened 2023-05-09
Include <activeAccounts> cell in output of setUp method for Foundry tests

ehildenb:
Currently, we are only taking the `<accounts>` cell content from the final state of the `setUp` method, but the `<activeAccounts>` cell must remain in sync with this account. This PR:

- Modifies the rules which access the `<accounts>` and `<activeAccounts>` cells to be a little simpler via `owise`. This amy make it easier to remove the `<activeAccounts>` cell altogether later: https://github.com/runtimeverification/evm-semantics/issues/1604
- Adds a test where the `<activeAccounts>` being out of sync with `<accounts>` following the `setUp` method causes a problem in verification of a property test.
- Fixes the Foundry integration to also pull the `<activeAccounts>` cell from the final state of the `setUp` method.


tests/foundry/golden/SetUpDeployTest.test_extcodesize.expected
```diff
@@ -0,0 +1,649 @@
...
+                       ( 0 |-> 0 )
+                     </storage>
+                     <origStorage>
+                       .Map
+                     </origStorage>
+                     <nonce>
+                       1
+                     </nonce>
+                     ...
+                   </account>
```
ehildenb, line 150: Here we can see that the extra deployed account is present in the configuration.

ehildenb:
The changes to `<accounts>` cells in the Golden files are just re-arranging of the output (reordering of how it's printed).

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -766,12 +766,11 @@ def _init_cterm(init_term: KInner) -> CTerm:
...
 
-def get_final_accounts_cell(proof_digest: str, proof_dir: Path) -> KInner:
+def get_final_accounts_cell(proof_digest: str, proof_dir: Path) -> tuple[KInner, KInner]:
     apr_proof = APRProof.read_proof(proof_digest, proof_dir)
     target = apr_proof.kcfg.get_unique_target()
-    cover = single(apr_proof.kcfg.covers(target_id=target.id))
-    accounts_cell = get_cell(cover.source.cterm.config, 'ACCOUNTS_CELL')
-    return accounts_cell
+    cterm = single(apr_proof.kcfg.covers(target_id=target.id)).source.cterm
+    return (cterm.cell('ACCOUNTS_CELL'), cterm.cell('ACTIVEACCOUNTS_CELL'))
```
lucasmt, line 773: I'm assuming that `cterm.cell(...)` is equivalent to `get_cell(cterm.config, ...)`, right?

tests/foundry/foundry-list.check.expected
```diff
@@ -1,484 +1,496 @@
-APRProof: AccountParamsTest.testDealConcrete:d50737a25b9af563feaa75d11ca1053406a57e4c2437645dfe1e3c9bb7059443
+APRProof: AccountParamsTest.testDealConcrete:aa6ebb29536834f3fc654cf9d98d91d29982b8ac015daeaf7efd6a747ba72776
```
lucasmt, line 1: I've been meaning to ask, what causes the change in this hash?

include/kframework/evm.md
```diff
@@ -1135,81 +1135,61 @@ Operators that require access to the rest of the Ethereum network world-state ca
...
-       requires #accountEmpty(CODE, NONCE, BAL)
-
-    rule <k> EXTCODEHASH ACCT => 0 ~> #push ... </k>
-         <activeAccounts> ACCTS </activeAccounts>
-      requires notBool ACCT in ACCTS
-```
+    rule <k> EXTCODEHASH ACCT => 0 ~> #push ... </k> [owise]
 
-TODO: What should happen in the case that the account doesn't exist with `EXTCODECOPY`?
-Should we pad zeros (for the copied "program")?
```
lucasmt, line 1197: Same as above, do we have confirmation that what we have is the correct behavior?

lucasmt (approved):
I'm approving, but see my questions and suggestions.

include/kframework/evm.md
```diff
@@ -1135,81 +1135,61 @@ Operators that require access to the rest of the Ethereum network world-state ca
 
 ### Account Queries
 
-TODO: It's unclear what to do in the case of an account not existing for these operators.
-`BALANCE` is specified to push 0 in this case, but the others are not specified.
-For now, I assume that they instantiate an empty account and use the empty data.
-
```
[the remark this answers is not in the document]
ehildenb: We pass the same conformance test-suite as the other implementations, and this behavior has been the same since 6 years ago when it was introduced: https://github.com/runtimeverification/evm-semantics/commit/29913b40a558f0066b8f277c850b6ea5a57f264e#diff-fbe2eecda4210ecd36a23352476c19f2ae691860a076a5e66a4886af8bcaf1d7R618

That comment was added when the semantics was very new and we were just beginning ot test it, before even RV took over the semantics.

kevm-pyk/src/kevm_pyk/foundry.py
ehildenb: Yep! But it's better because once you've called `cterm.cell(...)` once for any cell, every subsequent call will be faster (it's memoized at the `CTerm` level).

Here is the implementation of `CTerm.cell`: https://github.com/runtimeverification/pyk/blob/d9d6e8c101bc48570ebfdc62bf7b851af468aa23/src/pyk/cterm.py#L103
And the implementation of `get_cell`: https://github.com/runtimeverification/pyk/blob/d9d6e8c101bc48570ebfdc62bf7b851af468aa23/src/pyk/kast/manip.py#LL528C6-L528C6

Almost always using the `CTerm` interface means faster access and better behavior overall.

tests/foundry/foundry-list.check.expected
ehildenb: I actually don't know. Here is where the `proof_digest` is computed for a given proof in a Foundry property test suite: https://github.com/runtimeverification/evm-semantics/blob/18567e5fe70f8ff62d7204e112efe72bde4b1638/kevm-pyk/src/kevm_pyk/foundry.py#L102

And here is where the contract digest is computed (which is what's changing here): https://github.com/runtimeverification/evm-semantics/blob/18567e5fe70f8ff62d7204e112efe72bde4b1638/kevm-pyk/src/kevm_pyk/solc_to_k.py#LL170C13-L170C13

So it is basically a hash of all the Solidity compiler output. Could be many things changing. Likely we should be hashing less data as part of that digest to keep it more stable. @nwatson22 is working on something similar to that for the `setUp` methods specifically: https://github.com/runtimeverification/evm-semantics/issues/1795

include/kframework/evm.md
ehildenb: Yep, should be fine.

include/kframework/evm.md
```diff
@@ -1135,81 +1135,61 @@ Operators that require access to the rest of the Ethereum network world-state ca
...
            <acctID> ACCT </acctID>
            <balance> BAL </balance>
            ...
          </account>
+      requires ACCT in ACCTS
 
-    rule <k> BALANCE ACCT => 0 ~> #push ... </k>
-         <activeAccounts> ACCTS </activeAccounts>
-      requires notBool ACCT in ACCTS
+    rule <k> BALANCE ACCT => 0 ~> #push ... </k> [owise]
```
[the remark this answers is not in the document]
ehildenb: Done.

include/kframework/evm.md
```diff
@@ -1135,81 +1135,61 @@ Operators that require access to the rest of the Ethereum network world-state ca
...
-           <balance> BAL </balance>
-           ...
-         </account>
-       requires #accountEmpty(CODE, NONCE, BAL)
-
-    rule <k> EXTCODEHASH ACCT => 0 ~> #push ... </k>
-         <activeAccounts> ACCTS </activeAccounts>
-      requires notBool ACCT in ACCTS
-```
+    rule <k> EXTCODEHASH ACCT => 0 ~> #push ... </k> [owise]
```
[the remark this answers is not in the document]
ehildenb: Done.

include/kframework/evm.md
```diff
@@ -1135,81 +1135,61 @@ Operators that require access to the rest of the Ethereum network world-state ca
...
            <acctID> ACCT </acctID>
            <code> PGM </code>
            ...
          </account>
+      requires ACCT in ACCTS
 
-    rule <k> EXTCODECOPY ACCT _MEMSTART _PGMSTART _WIDTH => . ... </k>
-         <activeAccounts> ACCTS </activeAccounts>
-      requires notBool ACCT in ACCTS
+    rule <k> EXTCODECOPY ACCT _MEMSTART _PGMSTART _WIDTH => . ... </k> [owise]
```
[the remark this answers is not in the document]
ehildenb: Done.

PR #1817, opened 2023-05-12
Mark various functions as total

kcsongor:
This PR marks the `#ecrec` and the `#sign` functions as `total`, as they were triggering spurious definedness checks. I think (at least morally) it makes sense to mark them as total.

### DIFF
--- a/include/kframework/evm.md
+++ b/include/kframework/evm.md
@@ -1690,8 +1690,8 @@ Precompiled Contracts
          <callData> DATA </callData>
          <output> _ => #ecrec(#range(DATA, 0, 32), #range(DATA, 32, 32), #range(DATA, 64, 32), #range(DATA, 96, 32)) </output>
 
-    syntax Bytes ::= #ecrec ( Bytes , Bytes , Bytes , Bytes ) [function, smtlib(ecrec)]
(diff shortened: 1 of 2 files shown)


jberthold (commented):
I agree `#ecrec` should be marked `total`, and `#sign` appears to use `parseByteStack` only with arguments for which we can guarantee that that function is defined.
However, the `parse*`  family  of functions all rely on `String2Int` or `String2Base` which are undefined on strings that are not valid number literals.

kcsongor:
Thanks @jberthold, I removed `total` from those functions.

On that note, I think I found the concrete semantics of `String2Int` https://github.com/runtimeverification/haskell-backend/blob/7f51e8c9038f7fa20e90f7886503683456866d2a/kore/src/Kore/Builtin/String.hs#L355-L364, but didn't have that much luck looking for `ECDSASign`. Do you know where that function is defined? My understanding is that it's still total as far as K is concerned -- for an invalid private key, the call should revert at the EVM level.

ehildenb:
It's defined here: https://github.com/runtimeverification/blockchain-k-plugin/blob/master/plugin/krypto.md

Issue #1792, opened 2023-05-02
Improve the output of the KEVM Visualizer


anvacaru:

anvacaru:
In foundry.md there are a few subconfigurations only used by specific cheat codes such as expectRevert, expectEmit or prank. Setting those cells as `optional` perhaps could improve performance for the tests that do not use these cheat codes.

ehildenb:
If we want to omit some cells in the output (and improve the output), I think a better way is:

- In the `KEVM` class (https://github.com/runtimeverification/evm-semantics/blob/c6fb207374940b2b12f509d58a1451d76768cf0b/kevm-pyk/src/kevm_pyk/kevm.py#LL32C12-L32C24), we have it inheriting from `KRun` and `KProve`, which implements `pretty_print`. We can override `pretty_print` in KEVM to first walk the AST and do:
  - Print large numbers as hex strings.
  - Print bytes as hex strings.
  - Omit uninteresting components of the configuration.
- Then have `pretty_print` just call `super().pretty_print(...)`

anvacaru:
Additional ideas to improve the output:
   - replace the calldata selector bytes with the function signature fetched from ABI?
   - assign aliases to each node id to improve readability.

palinatolmach:
@anvacaru do you think any of these items are actionable?

anvacaru:
@palinatolmach Some of them are already implemented, like `assign aliases to each node id`. 

I had an open PR for changing the bytecode and large numbers to hex(linked above). I closed it because it became stale, but I can reopen it and get it merged.

ehildenb:
I haven't seen other requests for this yet, so closing for now.

palinatolmach:
Reopening in view of https://github.com/runtimeverification/kontrol/issues/473.

PR #1824, opened 2023-05-16
Pretty print KTokens in KCFGs

anvacaru:
part of: #1792 

Overloads the `pretty_print` method in `KEVM` to allow formating of `Int`s and `Bytes` as hexadecimal strings.


kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -360,3 +362,58 @@ def accounts(accts: list[KInner]) -> KInner:
             else:
                 wrapped_accounts.append(acct)
         return build_assoc(KApply('.AccountCellMap'), KLabel('_AccountCellMap_'), wrapped_accounts)
+
+    def pretty_print(self, kast: KAst, *, unalias: bool = True) -> str:
+        if unalias and isinstance(kast, KInner):
+            kast = undo_aliases(self.definition, kast)
+        if isinstance(kast, KInner):
+            kast = KEVM._int_token_to_hex(kast)
```
tothtamas28, line 370: My guess is this only does the rewriting if `kast` itself is `KInner`, but it won't recur into the `KInner` descendants of a `kast` that is not `KInner`. Is this intentional?

Also, are you sure you want to override the default `pretty_print`? This might result in a string that cannot be parsed back. (E.g. `"0x..."` is not a valid `Bytes` token.)

kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -360,3 +362,58 @@ def accounts(accts: list[KInner]) -> KInner:
...
+        return super().pretty_print(kast, unalias=False)
+
+    @staticmethod
+    def _int_token_to_hex(kast: KInner) -> KInner:
```
tothtamas28, line 374: This function is non-trivial, consider adding tests.

kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -360,3 +362,58 @@ def accounts(accts: list[KInner]) -> KInner:
...
+                return KToken('0x' + pretty_bytes(_kast).hex(), BYTES)
+            return _kast
+
+        conjuncts = flatten_label('#And', kast)
+        term = None
+        for c in conjuncts:
+            if type(c) is KApply and c.is_cell:
+                if term:
+                    raise ValueError(f'Found two configurations in pattern:\n\n{term}\n\nand\n\n{c}')
+                term = c
```
tothtamas28, line 388: Is this for matching the config part of the term?

kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -360,3 +362,58 @@ def accounts(accts: list[KInner]) -> KInner:
...
+                    'NEWORIGIN_CELL',
+                    'EXPECTEDREASON_CELL',
+                    'EXPECTEDADDRESS_CELL',
+                    'EXPECTEDVALUE_CELL',
+                    'EXPECTEDDATA_CELL',
+                    'EXPECTEDEVENTADDRESS_CELL',
+                    'ACTIVEACCOUNTS_CELL',
+                ]:
+                    try:
+                        kast = set_cell(kast, cell, bottom_up(_helper, get_cell(kast, cell)))
```
tothtamas28, line 416: I think this can be simplified into a single top-down traversal using `map_inner` (that should improve performance too)

You need two helper functions:

* `match_cell_name`: checks if the term is a cell with the name in the list. If yes, recur with `rewrite_bytes_to_hex`, otherwise, recur with `match_cell_name`.
* `rewrite_bytes_to_hex`: this is the same as `_helper`, basically.

kevm-pyk/src/kevm_pyk/kevm.py
anvacaru: Yes, duplicated logic from https://github.com/runtimeverification/pyk/blob/master/src/pyk/kast/manip.py#L213.

kevm-pyk/src/kevm_pyk/kevm.py
tothtamas28: So my understanding is that the goal is to rewrite all `Int` and `Bytes` tokens to something, but only if it is a descendant of one of the predefined cells.

In this case I think it is cleaner and more efficient to explicitly do the recursion, like

```python3
CELLS_TO_REWRITE = { ... }

def _int_token_to_hex(kast: KInner) -> KInner:
    def match_cells(kast: KInner) -> KInner:
        if isinstance(kast, KApply) and kast.is_cell and kast.label.name in CELLS_TO_REWRITE:
            return kast.map_inner(rewrite_bytes)
        return kast.map_inner(match_cells)

    def rewrite_bytes(kast: KInner) -> KInner:
        if type(kast) is KToken and kast.sort == INT:
            return KToken(hex(int(kast.token)), INT)
        if type(kast) is KToken and kast.sort == BYTES:
            return KToken('0x' + pretty_bytes(kast).hex(), BYTES)
        return kast.map_inner(rewrite_bytes)  # same as _helper, except for this line

    return kast.map_inner(match_cells)
```

(I did not test this.)

kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -360,3 +362,58 @@ def accounts(accts: list[KInner]) -> KInner:
...
+        return super().pretty_print(kast, unalias=False)
+
+    @staticmethod
+    def _int_token_to_hex(kast: KInner) -> KInner:
+        def _helper(_kast: KInner) -> KInner:
+            if type(_kast) is KToken and _kast.sort == INT:
+                return KToken(hex(int(_kast.token)), INT)
+            if type(_kast) is KToken and _kast.sort == BYTES:
+                return KToken('0x' + pretty_bytes(_kast).hex(), BYTES)
+            return _kast
```
ehildenb, line 380: Maybe we only want to do this for numbers taht are bigger than 256 when it's an `INT`? If the user writes `3` in their code, and it ends up on the stack, then they want to see `3` directly, not the hex encoding of it.

kevm-pyk/src/kevm_pyk/kevm.py
anvacaru: Hmm, I thought `set_cell(kast, cell, bottom_up` would get to all descendants. Launching the debugger displays all the values as intended.

kevm-pyk/src/tests/unit/test_unit.py
```diff
@@ -230,3 +231,20 @@ def test_contract_creation() -> None:
     assert method.arg_types == ('tuple[]',)
     assert method.contract_name == 'TestContract'
     assert method.payable
+
+
+def test_int_to_hex() -> None:
```
tothtamas28, line 236: Please refactor this to a [parametric test](https://docs.pytest.org/en/stable/how-to/parametrize.html#pytest-mark-parametrize-parametrizing-test-functions).

Test data can be a list of pairs `(input, expected)` so that you can assert equality with the actual value.

tothtamas28: Also, please add tests for composite `KAst` objects.

kevm-pyk/src/kevm_pyk/kevm.py
tothtamas28: What was the resolution of this thread?

anvacaru:
I added a screenshot with the current changes. image

kevm-pyk/src/tests/unit/test_unit.py
anvacaru: Done!

kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -369,3 +371,58 @@ def accounts(accts: list[KInner]) -> KInner:
             else:
                 wrapped_accounts.append(acct)
         return build_assoc(KApply('.AccountCellMap'), KLabel('_AccountCellMap_'), wrapped_accounts)
+
+    def pretty_print(self, kast: KAst, *, unalias: bool = True, sort_collections: bool = False) -> str:
```
tothtamas28, line 375: What's still missing is testing that a `KOuter` is printed as expected.

This function returns a `str`, so a better way to organize it would be:

1. Transform `KAst -> KAst`. You can write robust tests for this.
2. Call `super().pretty_print` on the transformed `KAst`. This is tested in `pyk`, so if the `str` output changes (e.g. we do not print as many whitespace characters), tests are updated there.

Right now, these two steps are intertwined in a non-trivial recursion scheme.

If you do not address this in this PR (I do not expect you to), please open an issue to mark this request.

kevm-pyk/src/tests/unit/test_unit.py
```diff
@@ -230,3 +239,43 @@ def test_contract_creation() -> None:
...
+    ),
+]
+
+
+@pytest.mark.parametrize(
+    'test_id,input,result',
+    TEST_DATA,
+    ids=[test_id for test_id, *_ in TEST_DATA],
+)
+def test_int_to_hex(test_id: str, input: KApply, result: KApply) -> None:
```
tothtamas28, line 278: Nice! :sparkles:

anvacaru:
Closing for now

PR #1836, opened 2023-05-18
Makefile: Update KOMPILE_OPTS to always use at least -O1 optimization level

Scott-Guest:
As noted in runtimeverification/k/issues/3365, there is a fairly significant performance difference between `-O0` and `-O1` (20 minutes vs 3 minutes for the test in that issue).

This PR then ensures we always compile with at least `-O1`


ehildenb:
@Scott-Guest does this still allow us to build with debug symbols for running with gdb? If so, feel free to add automerge!

Scott-Guest:
@ehildenb Yup, building with debug symbols actually already implies -O1 in the LLVM backend, so this should be fine

PR #1837, opened 2023-05-18
Port `kevm prove` to Python

Scott-Guest:
Move the `kevm prove` out of bash and into Python. The only remaining logic is profiling, which, per discussion with @ehildenb, isn't worth porting.

A few minor things worth noting:
- `--concrete-rules-file` was dropped entirely. `--concrete-rules` isn't an argument that's actually accepted by either the pyk `kevm prove` or the `kprove` binary, it's a `kompile` argument.
- `--branching-allowed` is now passed to `kprove` directly instead of converting it to `--breadth` in the haskell backend - this forwarding logic already exists in `kprove` itself


kevm-pyk/src/kevm_pyk/cli.py
```diff
@@ -70,6 +70,52 @@ def kprove_args(self) -> ArgumentParser:
...
+            action='store_true',
+            help='Generate a haskell-backend bug report for the execution.',
+        )
+        args.add_argument(
+            '--debugger',
+            dest='debugger',
+            default=False,
+            action='store_true',
+            help='Launch proof in an interactive debugger.',
+        )
```
ehildenb, line 88: Did you havea  chance to check if `--debugger` is working as expected? Main concern is that launching the `kore-repl` from Python may behave weird in terms of input/output?

Scott-Guest: Good catch.

Actually, `--debugger` doesn't work at all currently because it can't be used with `--haskell-backend-command`, so I'll have to revert runtimeverification/pyk/pull/451. 

After fixing this issue though (runtimeverification/pyk/pull/455), `--debugger` seems to work fine and there's no problem with the I/O.

Scott-Guest:
Blocked on https://github.com/runtimeverification/pyk/pull/455

PR #1839, opened 2023-05-19
Reuse KoreServer between proof initialization and execution

### LINKED ISSUE #1830: Initializing prover starts up RPC server twice
With the booster backend, we're getting down to sub-minute runtimes for many proofs. This means that the extra 5-10s for starting up the RPC server twice is actually noticeable overhead.

We should find a way to only start teh RPC server once on initializing new proofs, and reuse it when actually executing the proof.

Here we start it for building the initial proof and simplifying it: https://github.com/runtimeverification/evm-semantics/blob/f5c206d1859efd33007708385d24734fb00f54a0/kevm-pyk/src/kevm_pyk/foundry.py#L773
And here we start it again for actually discharging the proof: https://github.com/runtimeverification/evm-semantics/blob/f5c206d1859efd33007708385d24734fb00f54a0/kevm-pyk/src/kevm_pyk/utils.py#L114

These are both in their own parallel blocks, which poses a challeng because the RPC server must be started _after_ stepping into parallel execution, so each thread is not trying to talk to the same RPC server.

We should probably instead:

- Remove `parallel_kcfg_explore` method, have it just handle a single proof (maybe just call it `kevm_apr_prove` instead).
- Start up the parallel block at the beginning, then immeditaately start up the RPC server.
- Pass the started RPC server into the `_method_to_apr_proof`, and into `kevm_apr_prove`.

nwatson22:
Closes #1830

- Moves KCFGExplore creation in `kevm foundry-prove` to new function, `_init_and_run_proof()` and executes this function in parallel, so the same RPC server can be used for both the simplification of the initial state and the execution of the proof
- Similarly for `kevm prove`, but uses a different `_init_and_run_proof()` function because the initialization logic is different between foundry and non-foundry proofs (one is generated from K claims and the other is generated from solidity functions).


kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -458,7 +458,7 @@ def foundry_prove(
                 )
         method.update_digest(foundry.out / 'digest')
 
-    def _init_apr_proof(_init_problem: tuple[str, str]) -> APRProof | APRBMCProof:
+    def _init_apr_proof(_init_problem: tuple[str, str], kcfg_explore: KCFGExplore) -> APRProof | APRBMCProof:
```
ehildenb, line 461: It looks like this method is only used in one place, inside `_init_and_run_proof`. It used to be its own method so initialization could happen in parallel, but your PR is even better at making that happen. So I'd just inline it's code into `_init_and_run_proof`.

tests/failing-symbolic.pyk
```diff
@@ -20,7 +20,6 @@ tests/specs/functional/lemmas-no-smt-spec.k
 tests/specs/functional/lemmas-spec.k
 tests/specs/functional/merkle-spec.k
 tests/specs/functional/storageRoot-spec.k
-tests/specs/mcd/functional-spec.k
```
ehildenb, line 23: Why this test in particular?

ehildenb (approved):

tests/failing-symbolic.pyk
nwatson22: I just happened to use it to test the non-foundry test parallelization and found that it was passing with the pyk prover.

kevm-pyk/src/kevm_pyk/foundry.py

Issue #1842, opened 2023-05-19
Problem with stBadOpcode/invalidAddr_d21g0v0_Berlin conformance test


anvacaru:
**Summary**: *`extcodecopy`usage when the account does not exist.*

retesteth state execution of the test: http://retesteth.ethdevops.io/web/index.php?gtests=stBadOpcode&singletest=invalidAddr&clients=t8ntool&snetwork=Berlin&vmtrace=vmtrace&verbosity=STATE&trd=21&trg=0&trv=0

After debugging the test in KEVM, I've discovered that KEVM fails this test because it consumes an extra `20000` gas from running `SSTORE 0 FE01FE01FE01FE01` instead of `SSTORE 0 0`.

The memory ops that take place before the first `SSTORE` are:
```
MSTORE 0x2000 0x2
MLOAD 0x2000
MSTORE 0x2040 0x2
MSTORE 0x2020 0xff00ff00ff00ff00
MSTORE 0x2060 0xff00ff00ff00ff
MLOAD 0x2000
EXTCODECOPY 0x2 0x2020 0x0 0x20
MLOAD 0x2040
EXTCODECOPY 0x2 0x2060 0x0 0x20
MSTORE 0x2080 0x0
MLOAD 0x2060
MLOAD 0x2020
```
And so, the memory should look like this:
```
[0x2000]0000000000000000000000000000000000000000000000000000000000000002
[0x2020]000000000000000000000000000000000000000000000000ff00ff00ff00ff00
[0x2040]0000000000000000000000000000000000000000000000000000000000000002
[0x2060]00000000000000000000000000000000000000000000000000ff00ff00ff00ff
[0x2080]0000000000000000000000000000000000000000000000000000000000000000
```

But in retesteth, the last two `MLOAD`s return `0`. 
<img width="654" alt="Screenshot 2023-05-19 at 14 52 25" src="https://github.com/runtimeverification/evm-semantics/assets/16517508/434f2504-7538-4e9f-ae1d-6cc608e04584">

The locations have also been used by `EXTCODECOPY`. I've looked at other implementations of the `EXTCODECOPY`. [An empty account is created when the address is not in the state,](https://github.com/ethereumjs/ethereumjs-monorepo/blob/e6a66e05ce0c71ed479a68b14fa5ae1bc68a44d4/packages/evm/src/opcodes/functions.ts#L496) which explains why the values are overwritten in the local memory. 

[...]

Issue #1671, opened 2023-03-28
Conformance test failing list 


anvacaru:
KEVM failing conformance tests
------------------------------
#### Current
- [ ] tests/ethereum-tests/BlockchainTests/GeneralStateTests/stBadOpcode/invalidAddr.json
- [ ] tests/ethereum-tests/BlockchainTests/GeneralStateTests/stRandom2/randomStatetest649.json

#### v11
- [ ] tests/ethereum-tests/BlockchainTests/GeneralStateTests/stCreate2/CREATE2_HighNonceDelegatecall.json

#### v11.1
- [ ] tests/ethereum-tests/BlockchainTests/GeneralStateTests/stCallCodes/touchAndGo.json
- [ ] tests/ethereum-tests/BlockchainTests/GeneralStateTests/stPreCompiledContracts/delegatecall09Undefined.json

#### v11.2
These tests fail because there is an expected schedule string that does not match `Merge+3540+3670`.
EIP3540 and 3670 need to be investigated/implemented.

- [ ] tests/ethereum-tests/BlockchainTests/GeneralStateTests/stEIP3540/CREATE2_EOF1.json
- [ ] tests/ethereum-tests/BlockchainTests/GeneralStateTests/stEIP3540/CREATE2_EOF1Invalid.json
- [ ] tests/ethereum-tests/BlockchainTests/GeneralStateTests/stEIP3540/CREATE2_EOF1Invalid_FromEOF.json
- [ ] tests/ethereum-tests/BlockchainTests/GeneralStateTests/stEIP3540/CREATE2_EOF1_FromEOF.json
- [ ] tests/ethereum-tests/BlockchainTests/GeneralStateTests/stEIP3540/CREATE_EOF1.json
- [ ] tests/ethereum-tests/BlockchainTests/GeneralStateTests/stEIP3540/CREATE_EOF1_FromEOF.json
- [ ] tests/ethereum-tests/BlockchainTests/GeneralStateTests/stEIP3540/CREATE_EOF1Invalid.json
- [ ] tests/ethereum-tests/BlockchainTests/GeneralStateTests/stEIP3540/CreateTransactionEOF1.json
- [ ] tests/ethereum-tests/BlockchainTests/GeneralStateTests/stEIP3540/CREATE_EOF1Invalid_FromEOF.json
- [ ] tests/ethereum-tests/BlockchainTests/GeneralStateTests/stEIP3540/CreateTransactionInvalidEOF1.json
[...]

PR #1843, opened 2023-05-20
EXTCODECOPY when account is not existent

anvacaru:
Fixes:#1842 Currently, `EXTCODECOPY ACCT MEMSTART PGMSTART WIDTH `:
 - copies `#padToWidth(.Bytes, WIDTH)` if the `ACCT` has no code
 - If you copy past the end of an existing accounts code, you get zeros
 - acts as a no-op if the target address is not in the network state

This PR changes the third point's behavior to copy the zero value if the account is non-existing.

Part of #1671


tests/failing.llvm
```diff
@@ -1,3 +1,2 @@
-tests/ethereum-tests/BlockchainTests/GeneralStateTests/stBadOpcode/invalidAddr.json
 tests/ethereum-tests/BlockchainTests/GeneralStateTests/stRandom2/randomStatetest649.json
 tests/ethereum-tests/BlockchainTests/GeneralStateTests/stCreate2/CREATE2_HighNonceDelegatecall.json
```
ehildenb, line 2: Is the test that is removed here failing before this change and passing afterwards? And if you provide the extra rule, with a `#TODO` linking it to the issue in K, do all the tests pass?

ehildenb (approved):

PR #1854, opened 2023-05-24
Trigger compilation when required files change

### LINKED ISSUE #1788: User still must pass `--regen` on changing the input K lemma files
As part of the  kompilation digest, we should be detecting if the user supplied `--requires` files are changed, and rekompile if so.

nwatson22:
Closes #1788.

~~This PR changes how the foundry digest is computed. Every `--require`ed file is hashed and then these hashes are concatenated onto the contract digests previously used to compute the digest, and the digest is the hash of that string. This way, whenever the required lemmas files are changed recompilation is triggered automatically.~~

~~I tested to make sure that there is now recompilation when lemma files are changed, and that running the command to compile repeatedly with no changes results in recompilation being skipped subsequent times.~~

This PR adds a separate digest file associated with kompilation, so if either of the generated files, `foundry.md` or `contracts.md` is changed, or any of the `--require`ed files are changed, kompilation will be triggered. We call regeneration the step where the K modules are generated from the compiled solidity JSON files  and kompilation is the kompilation of those K files along with any user-specified helper files.

I tested to make sure that there is now rekompilation when lemma files are changed, that running `kevm foundry-kompile` repeatedly with no changes results in rekompilation and regeneration being skipped subsequent times, and that running `kevm foundry-kompile` after changes to the foundry contracts results in both regeneration and rekompilation being triggered.


kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -48,6 +48,7 @@ class Foundry:
     _root: Path
     _toml: dict[str, Any]
     _bug_report: BugReport | None
+    _requires: Iterable[str] | None
```
ehildenb, line 51: Why do we have this and the property `requires`? Can we just have the field named `requires` then directly?

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -56,11 +57,17 @@ def __init__(
         self,
         foundry_root: Path,
         bug_report: BugReport | None = None,
+        requires: Iterable[str] | None = None,
```
ehildenb, line 60:
```suggestion
        bug_report: BugReport | None = None,
        requires: Iterable[str] = (),
```

Does it simplify things if we just have the default be empty? Then later, you don't need a branch on whether it's `None` or not (in `Foundry.digest`).

nwatson22: I intended `[]` to mean there are no `requires`, but `None` to mean there may be `requires` but we don't know about them (since the constructor for `Foundry` is sometimes invoked in a context when the dependencies aren't listed/relevant as they are already compiled into the definition). Is there some other pattern I could use to guard against this? Maybe it would be better to not have `requires` be a member of `Foundry` and instead just a parameter for `digest()`? Or rework it so that it's always tracking the required files.

kevm-pyk/src/kevm_pyk/foundry.py
nwatson22: There is just a field now.

kevm-pyk/src/kevm_pyk/foundry.py
nwatson22: The kompilation and foundry digests are now tracked separately so there is no need for us to store the `requires` files in the Foundry object.

nwatson22:
> Maybe go for something simpler:
> 
>     * If the `Foundry.digest` changes (because the solidity code changes), then regenerate the K files. This will retrigger kompilation because the requires digest will change.
> 
>     * `foundry_kompile` hardcodes the logic of checking the K file digests directly. That digest should take into account: (i) the required files `contracts.k` and `foundry.k`. Ideally, it should also include all of KEVM's includes (eg. `evm.md`, `driver.md`, etc..., and `domains.md`, for example). If that digest changes, then we rekompile.

I have now:
- `kompilation_digest()` - computes the digest including the `contracts.k`, `foundry.k`, and any `--requires` files
- `kompilation_up_to_date()` - checks `kompiled/digest` against `kompilation_digest()`
- `update_kompilation_digest()` - writes `kompilation_digest()` to `kompiled/digest`

ehildenb (approved):

ehildenb:
@nwatson22 can you update the main comment body to reflect the new approach?

Also, I assume you tested this locally to make sure the appropriate rekompilations are (or are not) triggered?

PR #1856, opened 2023-05-25
make storageLayout optional in solc-to-k

anvacaru:


kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -251,7 +251,7 @@ def digest(self) -> str:
 
     @cached_property
     def storage_digest(self) -> str:
-        return hash_str(f'{self.name} - {json.dumps(self.contract_json["storageLayout"], sort_keys=True)}')
+        return hash_str(f'{self.name} - {json.dumps(self.contract_json.get("storageLayout", []), sort_keys=True)}')
```
tothtamas28, line 254:
```suggestion
        storage_layout = self.contract_json.get("storageLayout") or {}
        return hash_str(f'{self.name} - {json.dumps(storage_layout, sort_keys=True)}')
```

This is to ensure that

```json
{"storageLayout": null}
```

```json
{"storageLayout":{}}
```

```json
{}
```

all get the same hash.

tothtamas28 (approved):

PR #1860, opened 2023-05-26
Use EqualityProof for functional claims

### LINKED ISSUE #1858: Use `EqualityProof` for functional proofs
- Detect functional-style proofs during proof loading with kevm
- Execute them as `EqualityProof`s
- Use this system for existing `runLemma -> doneLemma` proofs so `runLemma -> doneLemma` syntax is no longer required

nwatson22:
Per #1858, this PR changes proof loading and execution so that if a claim is functional, i.e. there is a function symbol at the top:
- It is loaded as an `EqualityProof`
- An appropriate prover, `EqualityProver`, is generated for the proof instead of the normal `APRBMCProver or APRProver`.
- The proof is advanced/executed

It also converts some of the existing claims to functional claims.


ehildenb:
Let's:

- Update the main comment body with usmmary of the changes.
- Give the ones that could not be converted names.
- make an issue listing the names of the ones that cannot be converted.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -296,36 +308,42 @@ def _init_and_run_proof(claim: KClaim) -> bool:
...
             trace_rewrites=trace_rewrites,
         ) as kcfg_explore:
-            proof_problem: APRProof
-
-            if save_directory is not None and not reinit and APRProof.proof_exists(claim.label, save_directory):
-                proof_problem = APRProof.read_proof(claim.label, save_directory)
-
+            proof_problem: Proof
+            if is_functional(claim):
+                proof_problem = EqualityProof.from_claim(claim, kevm.definition)
```
ehildenb, line 313: We should still be checking if the proof state is already saved on disk here, rather than reloading it every time.

Basically teh same logic as below:

```suggestion
            proof_problem: Proof
            if is_functional(claim):
                if save_directory is not None and not reinit and EqualityProof.proof_exists(claim.label, save_directory):
                    proof_problem = EqualityProof.read_proof(claim.label, save_directory)
                else:
                    proof_problem = EqualityProof.from_claim(claim, kevm.definition)
```

kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -111,33 +114,50 @@ def kevm_apr_prove(
...
     if type(_apr_proof) is APRBMCProof:
         assert same_loop, f'BMC proof requires same_loop heuristic, but {same_loop} was supplied'
         prover = APRBMCProver(
             _apr_proof, is_terminal=is_terminal, extract_branches=extract_branches, same_loop=same_loop
         )
-    else:
+    elif type(_apr_proof) is APRProof:
         prover = APRProver(_apr_proof, is_terminal=is_terminal, extract_branches=extract_branches)
+    elif type(_apr_proof) is EqualityProof:
+        prover = EqualityProver(_apr_proof)
```
ehildenb, line 126: We should probably have an `else:` case:

```suggestion
    elif type(_apr_proof) is EqualityProof:
        prover = EqualityProver(_apr_proof)
    else:
        raise ValueError(f'Do not know how to build prover for proof: {_apr_proo}')
```

And I guess we should rename variable `_apr_proof` to just `proof` or `_proof`.

tests/specs/bihu/functional-spec.k
```diff
@@ -3,20 +3,20 @@ requires "verification.k"
...
-    rule <k> runLemma(S) => doneLemma(S) ... </k>
-
-    syntax Step ::= Bytes | Int | Bool
- // ----------------------------------
+//     syntax KItem ::= runLemma ( Step ) | doneLemma ( Step )
+//  // -------------------------------------------------------
+//     rule <k> runLemma(S) => doneLemma(S) ... </k>
+// 
+//     syntax Step ::= Bytes | Int | Bool
+//  // ----------------------------------
```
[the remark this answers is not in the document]
nwatson22: Deleted.

kevm-pyk/src/kevm_pyk/utils.py
nwatson22: Changed this, and changed `kevm_apr_prove` to just `kevm_prove`.

kevm-pyk/src/kevm_pyk/__main__.py
nwatson22: Since we are creating the proof with `EqualityProof.from_claim`, which always initializes `proof_dir` to `None`, they are never getting saved in the first place. I made a PR here:
https://github.com/runtimeverification/pyk/pull/475
which allows us to specify a save directory.

nwatson22:
This is now failing because the production for `#if_#then_#else_#fi` is parametric, but this is not supported by pyk as far as I can tell, meaning the sort for this production is determined to be `Sort`. This means for a rule like:
```
    claim [gas.arith.0]: (#if B #then (VGas -Int Csstore(ISTANBUL, V1, V2, V3)) +Int -435 #else (VGas -Int Csstore(ISTANBUL, V1, V2, V3)) +Int -4342 #fi)
           => ((VGas -Int Csstore(ISTANBUL, V1, V2, V3)) +Int #if B #then -435 #else -4342 #fi)
```
The sort will be reported as none since the LHS sort is determined to be `Sort` and the RHS to be `Int`.
When we attempt to `EqualityProof.from_claim()` to convert this claim to an EqualityProof, it calls `sort_strict` on the body of the claim, which then crashes because the sort is `None`.

nwatson22:
Blocked on https://github.com/runtimeverification/k/issues/3436

kevm-pyk/src/kevm_pyk/__main__.py
nwatson22: This is fixed now and `EqualityProof`s are being written to files if requested.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -284,6 +286,16 @@ def exec_prove(
...
 
+    def is_functional(claim: KClaim) -> bool:
+        _LOGGER.warning(f'claim:{claim}')
+        return not (
+            type(claim.body) is KApply
+            and claim.body.label.name == '<generatedTop>'
+            or type(claim.body) is KRewrite
+            and type(claim.body.lhs) is KApply
+            and claim.body.lhs.label.name == '<generatedTop'
+        )
```
ehildenb, line 297:
```suggestion
        claim_lhs = claim.body
        if type(claim_lhs) is KRewrite:
            claim_lhs = claim_lhs.lhs
        return not (type(claim_lhs) is KApply and claim_lhs.label.name == '<generatedTop>')
```

nwatson22: Done.

ehildenb (approved):

kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -64,9 +67,10 @@ def get_apr_proof_for_spec(  # noqa: N802
     return apr_proof
 
 
-def kevm_apr_prove(
+def kevm_prove(
     kprove: KProve,
-    proof: APRProof | APRBMCProof,
+    cfgid: str,
+    proof: Proof,
```
ehildenb, line 73: Do we actually need to pass in `cfgid` again as well? Can we just use `proof.id` below wherever `cfgid` is used?

nwatson22: Removed it.

tests/foundry/golden/AssertTest.testFail_assert_true.expected
```diff
@@ -882,7 +882,13 @@ module SUMMARY-ASSERTTEST.TESTFAIL-ASSERT-TRUE:197B53D8A0C19A3B62F2C833DF85695B4
 endmodule
 1 Failure nodes. (0 pending and 1 failing)
 
+<<<<<<< HEAD
+1 Failure nodes. (0 frontier and 1 stuck)
+
+Stuck nodes:
+=======
 Failing nodes:
+>>>>>>> origin/master
```
ehildenb, line 892: What's going on here?

nwatson22: I guess these files are no longer used, so running the update didn't overwrite the failed automerge version. I removed all these expected files in favor of the ones at the location used by the pytest tests.

PR #1865, opened 2023-05-26
Refactor blockchain-k-plugin build

goodlyrottenapple:
As we are starting to test the booster with more and more kevm proofs, we are hitting new corner cases with the LLVM dynamic library, which the booster uses for concrete execution. Specifically, we have come across the following error:

```
/tmp/hs-booster-proxy-bfeaddc9bbbd2ef1/.log: withFile: user error (dlopen: out/kompiled-llvm/interpreter.so: undefined symbol: _ZTVN8CryptoPP6SHA512E)
```

It turns out the linker is being too aggressive when pruning seemingly unused symbols from the dynamic library which results in the above error. The fix is to compile the offending library statically, producing a`.a` file and then instead of passing e.g. `-lcryptopp -L .../libcryptopp/lib`, we instead provide the static `.../libcryptopp/lib/libcryptopp.a` via `-ccopts` in `kevm-pyk`.

Since we now rely on static libraries for `libcryptopp`, `libff` and `libsecp256k1`, we build them from source, insterad of relying on the nix/system provided shared libraries.


flake.nix
```diff
@@ -2,14 +2,16 @@
...
+      "github:NixOS/nixpkgs/b01f185e4866de7c5b5a82f833ca9ea3c3f72fc4";
     k-framework.url = "github:runtimeverification/k/v5.6.110";
     k-framework.inputs.nixpkgs.follows = "nixpkgs";
     #nixpkgs.follows = "k-framework/nixpkgs";
     flake-utils.follows = "k-framework/flake-utils";
     rv-utils.url = "github:runtimeverification/rv-nix-tools";
     poetry2nix.follows = "pyk/poetry2nix";
-    blockchain-k-plugin.url = "github:runtimeverification/blockchain-k-plugin/23fb5e8fdb4a87451fd40a0e49118eaab481931b";
+    blockchain-k-plugin.url =
+      "github:runtimeverification/blockchain-k-plugin/23fb5e8fdb4a87451fd40a0e49118eaab481931b";
```
ehildenb, line 14: This must stay formatted how it was before, or `.github/workflows/update-version.yml` must be updated to recognize the new format here: https://github.com/runtimeverification/evm-semantics/blob/cb789ee9d856216e9b4dbc3fe5bbe3e775ce61fd/.github/workflows/update-version.yml#L69

kevm-pyk/src/kevm_pyk/kompile.py
```diff
@@ -121,9 +121,11 @@ def kevm_kompile(
     llvm_kompile_type: LLVMKompileType | None = None,
     enable_llvm_debug: bool = False,
     debug: bool = False,
+    md_selector: str | None = None,
 ) -> Path:
     backend = target.backend
-    md_selector = 'k & ! standalone' if target == KompileTarget.NODE else 'k & ! node'
+    if not md_selector:
+        md_selector = 'k & ! standalone' if target == KompileTarget.NODE else 'k & ! node'
```
ehildenb, line 128: Hmmmm, are we sure we need to thread through this option to the user? I think the target actually fully determines what the `md_selector` needs to be, so we can just eliminate the option to `kevm_kompile`

kevm-pyk/src/kevm_pyk/kompile.py
```diff
@@ -173,8 +175,11 @@ def kevm_kompile(
         raise
 
 
-def _lib_ccopts(kernel: Kernel) -> list[str]:
-    ccopts = ['-g', '-std=c++14', '-lff', '-lcryptopp', '-lsecp256k1', '-lssl', '-lcrypto']
+def _lib_ccopts(kernel: Kernel, llvm_kompile_type: LLVMKompileType | None = None) -> list[str]:
+    ccopts = ['-g', '-std=c++14']
+
+    if llvm_kompile_type != LLVMKompileType.C: ccopts += ['-lff', '-lcryptopp', '-lsecp256k1', '-lssl', '-lcrypto']
+
```
ehildenb, line 182:
```suggestion
    if llvm_kompile_type != LLVMKompileType.C:
        ccopts += ['-lff', '-lcryptopp', '-lsecp256k1', '-lssl', '-lcrypto']
    
```

flake.nix
goodlyrottenapple: oh my bad, i ran nixfmt formatter out of habit

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -375,12 +360,12 @@ def update_kompilation_digest() -> None:
...
-        _kompile(foundry_definition_dir, KompileBackend.HASKELL, md_selector=md_selector)
+        _kompile(foundry_definition_dir, KompileTarget.HASKELL, md_selector=md_selector)
         if llvm_library:
             _LOGGER.info(f'Kompiling definition to LLVM dy.lib: {foundry_main_file}')
             _kompile(
                 foundry_llvm_dir,
-                KompileBackend.LLVM,
+                KompileTarget.LLVM,
                 llvm_kompile_type=LLVMKompileType.C,
                 md_selector=('k & ! symbolic' if md_selector is None else f'{md_selector} & ! symbolic'),
```
goodlyrottenapple, line 370: @ehildenb we pass custom md_selector here

Makefile
```diff
@@ -166,7 +124,15 @@ $(plugin_k_include)/%: $(PLUGIN_SUBMODULE)/plugin/%
...
 	install $< $@
 
-plugin-deps: $(plugin_includes) $(plugin_c_includes)
+
+$(KEVM_LIB)/%: $(PLUGIN_SUBMODULE)/build/%
+	@mkdir -p $(dir $@)
+	cp -r $< $@
+
+$(PLUGIN_SUBMODULE)/build/lib%:
+	cd $(PLUGIN_SUBMODULE) && make $*
```
jberthold, line 133: This [causes a `No rule to make target 'LIBRARYNAME'`](https://github.com/runtimeverification/evm-semantics/actions/runs/5136360689/jobs/9243059585?pr=1865#step:5:22) because the pattern cuts of the `lib` prefix.
Easiest fix is probably to put the `lib` prefix back, since we want to only consider the targets files that start with `lib` here.
```suggestion
$(PLUGIN_SUBMODULE)/build/lib%:
	cd $(PLUGIN_SUBMODULE) && make lib$*
```

jberthold (commented):
I like that we move the logic to build `libff` and `libcrypto` out of the `evm-semantics/Makefile`. We are still compiling the `cpp` files of the plugin here, one could go further and have a `make` target in `blockchain-k-plugin` that produces a static library (and links to the dependencies) but I _guess_ this variant should work.

.github/actions/with-docker/action.yml
```diff
@@ -15,6 +15,7 @@ runs:
       set -euxo pipefail
 
       Z3_VERSION=4.12.1
+      LLVM_VERSION=12
```
ehildenb, line 18: I think we should be able to use at least LLVM14 right? Maybe even LLVM15?

.github/workflows/Dockerfile
```diff
@@ -15,13 +15,16 @@ COPY --from=FOUNDRY /usr/local/bin/cast /usr/local/bin/cast
...
     && apt-get install --yes software-properties-common \
     && add-apt-repository ppa:ethereum/ethereum
 
 RUN    apt-get update         \
     && apt-get upgrade --yes  \
     && apt-get install --yes  \
+            clang-${LLVM_VERSION} \
             cmake             \
             curl              \
             debhelper         \
```
ehildenb, line 30: All of the `llvm-*` should also get `${LLVM_VERSION}` number here. Below, it's `llvm-14-dev` and `llvm-14-tools`. So we should default `LLVM_VERSION=14` in the `with-docker` action, and make sure to use that same version for all LLVM related tools.

.github/workflows/test-pr.yml
```diff
@@ -115,6 +115,8 @@ jobs:
           tag: kevm-ci-concrete-${{ github.sha }}
       - name: 'Build kevm-pyk'
         run: docker exec -u user kevm-ci-concrete-${GITHUB_SHA} /bin/bash -c 'make poetry'
+      - name: 'Build blockchain-k-plugin-deps'
+        run: docker exec -u user kevm-ci-concrete-${GITHUB_SHA} /bin/bash -c 'make CXX=clang++-12 plugin-deps'
```
ehildenb, line 119: Let's use v14.

.github/workflows/Dockerfile
ehildenb: Also let's try to keep the original formatting as much as possible, everything lined up nicely. Makes the files much more "scannable".

cmake/node/CMakeLists.txt
```diff
@@ -72,7 +74,8 @@ target_include_directories(kevm-vm
 	PUBLIC $ENV{NODE_DIR}/vm/kevm
 	PUBLIC ${CMAKE_SOURCE_DIR}/..
 	PUBLIC $ENV{KEVM_LIB_ABS}/libff/include
-	PUBLIC $ENV{KEVM_LIB_ABS}/cryptopp/include
+	PUBLIC $ENV{KEVM_LIB_ABS}/libcryptopp/include
+  PUBLIC $ENV{KEVM_LIB_ABS}/libsecp256k1/include
```
ehildenb, line 78: What's up with the formatting here?

kevm-pyk/src/kevm_pyk/foundry.py
ehildenb: I think we should instead modify the `class KompileTarget` to have a new property (in file `kompile.py`):

```
@property
def md_selector(self) -> str:
        match self:
            case self.LLVM:
                return 'k & ! node & ! symbolic'
            case self.NODE:
                return 'k & ! symbolic & ! standalone'
            case self.HASKELL:
                return 'k & ! node & ! concrete'
            case self.HASKELL_STANDALONE:
                return 'k & ! node & ! concrete'
            case self.FOUNDRY:
                return 'k & ! node & ! concrete'
            case _:
                raise AssertionError()
```

Then have function `kevm_kompile` (also defined in `kompile.py`) just use this method. That way we don't have to pass in custom `md_selector`, since it's fully a configuration value dependent on the compilation target, not something the user should really be messing with directly for the moment.

cmake/node/CMakeLists.txt
goodlyrottenapple: tab vs space... that looked aligned in my editor

kevm-pyk/src/kevm_pyk/foundry.py
goodlyrottenapple: so just to check, we want to completely remove the md_selector as a flag that the user can specify as a command line arg and instead just pass the right flags based on the compile target?

flake.nix
```diff
@@ -82,6 +79,15 @@
...
             };
+            #  src = prev.lib.cleanSource
+            #     (prev.nix-gitignore.gitignoreSourcePure [
+            #       ./.gitignore
+            #       ".github/"
+            #       "result*"
+            #       "*.nix"
+            #       "deps/"
+            #       "kevm-pyk/"
+            #     ] ./.);
```
Baltoli, line 90: Why is this commented out?

kevm-pyk/src/kevm_pyk/foundry.py
Baltoli: That seems sensible to me; arbitrary changes to the `md_selector` that aren't sanctioned by the design of the semantics are likely to just break things, I think.

kevm-pyk/src/kevm_pyk/kompile.py
```diff
@@ -173,11 +175,19 @@ def kevm_kompile(
         raise
 
 
-def _lib_ccopts(kernel: Kernel) -> list[str]:
-    ccopts = ['-g', '-std=c++14', '-lff', '-lcryptopp', '-lsecp256k1', '-lssl', '-lcrypto']
+def _lib_ccopts(kernel: Kernel, llvm_kompile_type: LLVMKompileType | None = None) -> list[str]:
```
Baltoli, line 178: The argument `llvm_kompile_type` is unused here - is that intentional?

Baltoli (approved):
A couple of small comments but this makes sense to me following our discussions re. the design of the static build system for the blockchain plugin.

flake.nix
goodlyrottenapple: ah my bad... i was trying something out and made a copy of src which i dint end up using...

kevm-pyk/src/kevm_pyk/kompile.py
goodlyrottenapple: ah sorry, it was used briefly but then changed the code and this was left in. thanks for the spot.

PR #1869, opened 2023-05-30
Override storageLayout extra output while building

iFrostizz:
It seems like foundry (or ethers-rs) is not defaulting anymore to including the storageLayout key in the artifacts or the build.
This has the effect to cause a crash when solc_to_k builds an identifier of the contract, by hasing the name and a json structure of the storage layout from the artifacts.

```sh
ifrostizz@kevm-dev-personal-01:~/tmp/my_kevm_project$ poetry run -C ~/rv/evm-semantics/kevm-pyk kevm foundry-kompile

WARNING 2023-05-30 07:41:00,700 __main__ - Ignoring command-line option: -O0
WARNING 2023-05-30 07:41:00,700 __main__ - Ignoring command-line option: -O1
WARNING 2023-05-30 07:41:00,700 __main__ - Ignoring command-line option: -O2
WARNING 2023-05-30 07:41:00,700 __main__ - Ignoring command-line option: -O3
Traceback (most recent call last):
  File "/usr/lib/python3.10/runpy.py", line 196, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/lib/python3.10/runpy.py", line 86, in _run_code
    exec(code, run_globals)
  File "/home/ifrostizz/rv/evm-semantics/kevm-pyk/src/kevm_pyk/__main__.py", line 905, in <modu
le>
    main()
  File "/home/ifrostizz/rv/evm-semantics/kevm-pyk/src/kevm_pyk/__main__.py", line 72, in main
    execute(**vars(args))
  File "/home/ifrostizz/rv/evm-semantics/kevm-pyk/src/kevm_pyk/__main__.py", line 178, in exec_
foundry_kompile
    foundry_kompile(
  File "/home/ifrostizz/rv/evm-semantics/kevm-pyk/src/kevm_pyk/foundry.py", line 287, in foundr
y_kompile
    _imports: dict[str, list[str]] = {contract.name: [] for contract in foundry.contracts.value
s()}
  File "/usr/lib/python3.10/functools.py", line 981, in __get__
    val = self.func(instance)
  File "/home/ifrostizz/rv/evm-semantics/kevm-pyk/src/kevm_pyk/foundry.py", line 102, in contra
cts
    _contracts[contract_name] = Contract(contract_name, contract_json, foundry=True)
  File "/home/ifrostizz/rv/evm-semantics/kevm-pyk/src/kevm_pyk/solc_to_k.py", line 231, in __in
it__
    msig, mid, method, method_ast, self.name, self.digest, self.storage_digest, self.sort_metho
d
  File "/usr/lib/python3.10/functools.py", line 981, in __get__
    val = self.func(instance)
  File "/home/ifrostizz/rv/evm-semantics/kevm-pyk/src/kevm_pyk/solc_to_k.py", line 254, in stor
age_digest
    return hash_str(f'{self.name} - {json.dumps(self.contract_json["storageLayout"], sort_keys=
True)}')
KeyError: 'storageLayout'
```

The solution is to explicitely choose the storageLayout variable when building the project with forge.


ehildenb (approved):

ehildenb:
If we specify `--extra-output ...` manually like this in the call to `forge ...`, does that override the `foundry.toml` that the user supplies? For example: https://github.com/runtimeverification/evm-semantics/blob/a68c97a2280b7ac9f1fc81f142464d0a23793b8d/tests/foundry/foundry.toml#LL4C32-L4C32.

If you check that `forge build ...` is adding those extra arguments instead of overriding them, and indeed foundry is adding to those arguments, then you can add `automerge` here.

iFrostizz:
@ehildenb , the arguments are indeed overridden by the command.
For instance, two `--extra-output` arguments that are apparently not mandatory for the Foundry internal to work and thus are optional are `storageLAyout` and `evm.gasEstimates`.

```toml
#foundry.toml
[profile.default]
src = "src"
out = "out"
libs = ["lib"]
extra_output = ['storageLayout', 'abi', 'evm.methodIdentifiers', 'evm.deployedBytecode.object', 'evm.gasEstimates']
```

Running `forge b`, the `storageLayout` and `evm.gasEstimates` are present in the build artifacts.
Running `forge b --extra-output storageLayout` will suppress the `evm.gasEstimates` in the new artifacts.

iFrostizz:
As mentioned by @anvacaru this was fixed by making it optional, but then contract whose name is the same would produce the same digest. We may want to use something like the absolute path, as two contract of the same name cannot be declared in the same file.

ehildenb:
Let's ask the Foundry maintainers about this in their Telegram (or maybe @gakonst can chime in directly here). The problem is that we need to sometimes pass arguments both via `foundry.toml` and `forge build ...`, but it seems that if we pass `--extra-output` via `forge build`, it doesn't _append_ to the values in `foundry.toml`, it _overwrites_ them. Is it possible to get appending behavior?

gakonst:
cc @evalir who was looking at this (and @mattsse for viz)

mattsse:
> If we specify --extra-output ... manually like this in the call to forge ..., does that override the foundry.toml that the user supplies

this is the expected behaviour imo

@evalir This could be solved with the `figment::Provider` impl

but I'm actually not sure if this would be the right fix because if we merge them then you can no longer override them via cli

 > It seems like foundry (or ethers-rs) is not defaulting anymore to including the storageLayout key in the artifacts or the build.

I'm not aware of any change regarding this

iFrostizz:
> I'm not aware of any change regarding this

Msg from @Evalir on tg:
image
For following the convo:
https://t.me/foundry_rs/29110

But this may have always been default, and not introduced at all

If that is a no-fix for Foundry config, and only affecting the `--extra-output`, then we could purposely parse the arguments from the `foundry.toml` and append them ?

Evalir:
Ah right, that comment is on the issue—it's not a change, and I suspect this behavior has always been the default as we haven't introduced any changes regarding solc output.

On the config side, this might be a no-fix indeed :( I think what you propose could be your way to go!

Issue #1870, opened 2023-05-30
Allow the user to specify signature to disambiguate


iFrostizz:
Currently, when for instance running `foundry-prove`, we have to specify one or more tests using the `--test` flag.
The issue is that when test methods are overloaded, we have no idea which one is ran.
We could for instance output an error message asking the user to disambiguate the test method.

Let's take as an example this test contract:

```solidity
contract MyTest is Test {
    function testSet() public {
        //
    }
    
    function testSet(uint256 x) public {
        //
    }
}
```

Running `kevm foundry-prove --test MyTest.testSet` would work, and we don't really control what function is proven. Furthermore, if one function is proven, there is no way other than proving the full contract to prove the overloaded function.

One suggestion would be to check for the matching function signatures.

If there is more than one, then ask the user to disambiguate the function to test by adding more informations.
e.g

`MyTest.testSet` => ambiguous
`MyTest.testSet(` => ambiguous
`MyTest.testSet()` => not ambiguous
`MyTest.testSet(u` => not ambiguous

It would not require the user to type the full function signature, especially useful for verbose ones.

iFrostizz:
This is the part where the [`--mt` or `--match-test` flag is parsed](https://github.com/foundry-rs/foundry/blob/master/cli/src/cmd/forge/test/filter.rs#L46-L68)
They are using [some regex](https://github.com/foundry-rs/foundry/blob/master/cli/src/cmd/forge/test/filter.rs#L16)
And for each [test signature](https://github.com/foundry-rs/foundry/blob/master/forge/src/runner.rs#L250), it checks if this one is a match with the regex. [It first checks if it's a match with the `--mt` and remove those which matches with `--no-match-test`](https://github.com/foundry-rs/foundry/blob/master/cli/src/cmd/forge/test/filter.rs#L103-L113)

PR #1873, opened 2023-05-31
Allow to specify test by signature

iFrostizz:
Closes: https://github.com/runtimeverification/evm-semantics/issues/1870

The issue with specifying tests by method name is that in the case of overloaded functions, they can be ambiguous (e.g having a function whose signature is `func()` and another whose signature is `func(uint8)`).
This PR should not break the past behaviour of entering a method name for the test, ~~but if it matches more than one test, then we raise an exception asking the user to be more precise until there is only 1 matching test.~~
Edit: we now use foundry-like `--match-test` flag with regex. If `>` 1 test matches, we run them all at once.

This is probably breaking a lot of other utils relying on method by name (not only foundry-prove)

Checklist of foundry-* commands that are not broken

- [x] kompile
- [x] prove
- [x] show
- [x] to-dot (getting another error related to  an internal)
- [x] list
- [x] remove-node
- [x] simplify-node
- [x] step-node
- [x] section-edge


ehildenb:
Back out the K submodule change please, and merge to latest master.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -684,8 +744,9 @@ def foundry_list(foundry_root: Path) -> list[str]:
 def foundry_remove_node(foundry_root: Path, test: str, node: NodeIdLike) -> None:
     foundry = Foundry(foundry_root)
     apr_proofs_dir = foundry.out / 'apr_proofs'
-    contract_name, test_name = test.split('.')
-    proof_digest = foundry.proof_digest(contract_name, test_name)
+    contract_name = test.split('.')[0]
+    test_sig = foundry.matching_sig(test).split('.')[1]
+    proof_digest = foundry.proof_digest(contract_name, test_sig)
```
anvacaru, line 749: This logic seems to be duplicated a few times. Can it be extracted in a new function?

iFrostizz: Thanks!

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -684,8 +746,8 @@ def foundry_list(foundry_root: Path) -> list[str]:
 def foundry_remove_node(foundry_root: Path, test: str, node: NodeIdLike) -> None:
     foundry = Foundry(foundry_root)
     apr_proofs_dir = foundry.out / 'apr_proofs'
-    contract_name, test_name = test.split('.')
-    proof_digest = foundry.proof_digest(contract_name, test_name)
+    (contract_name, test_sig) = foundry.unique_sig(test)
+    proof_digest = foundry.proof_digest(contract_name, test_sig)
```
anvacaru, line 750: This logic seems duplicated across some `foundry_*` commands. Can it be extracted?

anvacaru:
Looks good to me, but I request another review on this one.

kevm-pyk/src/kevm_pyk/foundry.py
iFrostizz: Thanks! Done

anvacaru (approved):

Issue #1871, opened 2023-05-30
Unsuported evm base sort type: int128


anvacaru:
`@lowprivuser` presented this issue on our discord server. When calling `foundry-kompile` on a contract that's using `int128`, `solc_to_k` will raise a `ValueError: Unsupported evm base sort type: int128`.

I replicated the issue with this minimal example 
```solidity

// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.13;

contract Counter {
   int128 public number;

   function setNumber(int128 newNumber) public {
       number = newNumber;
   }
}
```

fasapa:
@anvacaru (u)int24 too. Got the same error: `ValueError: Unsupported evm base sort type: int24`

Issue #1771, opened 2023-04-24
Functions with array arguments give "Unsupported ABI type for method` error


lucasmt:
The function [`_range_predicate`](https://github.com/runtimeverification/evm-semantics/blob/3f37fc99a9cf2d7adf41fa255a8d4a66674780cf/kevm-pyk/src/kevm_pyk/solc_to_k.py#L418) in `solc_to_k.py` does not support array types. As a result, if we try to `foundry-kompile` a function such as
```
function testArray(address[] memory array) public {
	assert(true);
}
```
that has an array as argument, we get the message `Unsupported ABI type for method TestContract.method_TestContract_testArray_address[], will not generate calldata sugar: address[]`, and no rule of the form
```
rule  ( TestContract . testArray ( ... ) => #abiCallData ( ... ) )
```
is added to `foundry.k`. This makes it so that the function cannot be evaluated during symbolic execution. If we try to run `kevm foundry-prove` on this function, the execution just starts branching on conditions like
```
lengthBytes ( TestContract . testArray ( VV0_array_114b9705:K : address[] ) ) <Int 4
```
without ever stepping into the function.

ehildenb:
From slack discussion with @iFrostizz :


Currently, when we have a function signature with types, we generate a rule that simplifies that function signature to its abi encoding, while at teh same time using an ensures clause to make sure the symbolic variables present are in the correct range: https://github.com/runtimeverification/evm-semantics/blob/a68c97a2280b7ac9f1fc81f142464d0a23793b8d/tests/foundry/contracts.k.check.expected#L80

This is an example generated rule for the KEVM foundry test-suite at tests/foundry
Those predicates in the ensures clause are generated by the `_range_predicate` thing you have linked above.
Currently, the logic there is "if it's a type we don't recognize directly, then return `None`, and don't generate the corresponnding rule with syntactic sugar".
Which means we can't write property tests about that function.

Here: https://github.com/runtimeverification/evm-semantics/blob/a68c97a2280b7ac9f1fc81f142464d0a23793b8d/kevm-pyk/src/kevm_pyk/solc_to_k.py#L15

We call `_range_predicate` to get the predicate used to constrain the symbolic value in the type.
And if it gives back `None`, then we just give up, and don't generate the overall rule.

I think there are two things we can do here:
[...]

PR #1883, opened 2023-06-02
Add support for ABI types

iFrostizz:
Advances towards a fix for: https://github.com/runtimeverification/evm-semantics/issues/1771

ref: https://solidity-fr.readthedocs.io/fr/latest/abi-spec.html#types

- Refactor
- `bytes<M>` support
- dynamic / fixed length arrays support
- fixed point (this one is parsed but not supported, Solidity itself has a very limited support for it at the moment)


iFrostizz:
I will aim this PR to at least support all static types and write an Issue explaining the issue I encountered for dynamic types .

kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -543,38 +543,78 @@ def _evm_base_sort_int(type_label: str) -> bool:
...
+
+def _range_predicate_bytes(term: KInner, type_label: str) -> tuple[bool, KInner | None]:
+    if type_label.startswith('bytes') and not type_label.endswith(']'):
+        str_width = type_label[5:]
+        if str_width != '':
+            width = int(str_width)
+            if not (0 < width and width <= 32):
+                raise ValueError(f'Unsupported range predicate bytes<M> type: {type_label}')
+            return (True, KEVM.range_bytes(intToken(width), term))
+    return (False, None)
```
iFrostizz, line 605: Any `bytes<M>` (`bytes1`..` bytes32`) now produces a `#rangeBytes(_, _)`.
This looks correct semantically, but can I have a confirmation please ?

kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -543,38 +543,78 @@ def _evm_base_sort_int(type_label: str) -> bool:
...
+
+
+# TODO write semantics that supports static arays
+def _range_predicate_static_array(term: KInner, type_label: str) -> tuple[bool, KInner | None]:
+    if type_label.endswith(']') and not type_label.endswith('[]'):
+        parts = type_label.rsplit('[', 1)
+        if len(parts) > 1:
+            number = int(parts[1].rstrip(']'))
+            return (True, KEVM.range_uint(((number + 1) * 256), intToken(32)))
+    return (False, None)
```
iFrostizz, line 615: Should we remove this or is it worth keeping it for later work on a better support of dynamic types, two-dimensional arrays, and structs ?

kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -543,38 +543,78 @@ def _evm_base_sort_int(type_label: str) -> bool:
...
 
     _LOGGER.info(f'Unknown range predicate for type: {type_label}')
     return None
 
 
 def _range_predicate_uint(term: KInner, type_label: str) -> tuple[bool, KInner | None]:
     if type_label.startswith('uint') and not type_label.endswith(']'):
-        width = int(type_label[4:])
+        if type_label == 'uint':
+            width = 256
```
anvacaru, line 574: So are you automatically set a witdh of `256` for all `uints`? What happens to `uint8` for example?

kevm-pyk/src/kevm_pyk/solc_to_k.py
```diff
@@ -543,38 +543,78 @@ def _evm_base_sort_int(type_label: str) -> bool:
...
+    else:
+        return (False, None)
+
+
+def _range_predicate_bytes(term: KInner, type_label: str) -> tuple[bool, KInner | None]:
+    if type_label.startswith('bytes') and not type_label.endswith(']'):
+        str_width = type_label[5:]
+        if str_width != '':
+            width = int(str_width)
+            if not (0 < width and width <= 32):
```
anvacaru, line 602:
```suggestion
            if not (0 < width and width <= 32):
            if not (0 < width <= 32):
```

kevm-pyk/src/kevm_pyk/solc_to_k.py
anvacaru: Looks good so far. Do you have any tests for these changes?

kevm-pyk/src/kevm_pyk/solc_to_k.py
anvacaru: If this code is not executed anywhere and tested, then I don't see a reason to keep it. I suggest adding it to an issue to support dynamic types as a starting point instead.

kevm-pyk/src/kevm_pyk/solc_to_k.py
iFrostizz: Not really, it should only set the width of `256` if the type label is strictly `uint`, because it's a shorthand of `uint256`. If it's `uint256` or `uint8` it should fall in the `else` clause where the `width` is dynamically set.

kevm-pyk/src/kevm_pyk/solc_to_k.py
iFrostizz: Looks very good, thanks !

kevm-pyk/src/kevm_pyk/solc_to_k.py
iFrostizz: I will add some, thanks for the suggestion!

kevm-pyk/src/kevm_pyk/solc_to_k.py
iFrostizz: Will do. Thanks !

iFrostizz:
Note for a future fix for including arrays and / or structs, this is how we could write a function to support static arrays

```python
def _range_predicate_bytes(term: KInner, type_label: str) -> tuple[bool, KInner | None]:
    if type_label.startswith('bytes') and not type_label.endswith(']'):
        str_width = type_label[5:]
        if str_width != '':
            width = int(str_width)
            if not (0 < width and width <= 32):
                raise ValueError(f'Unsupported range predicate bytes<M> type: {type_label}')
            return (True, KEVM.range_bytes(intToken(width), term))
    return (False, None)
```

iFrostizz:
Also closes https://github.com/runtimeverification/evm-semantics/issues/1871 from https://github.com/runtimeverification/evm-semantics/pull/1883/commits/1e9b175bd9ab991d5c24f6c422f8018cbc7087c3 !

palinatolmach:
In view of this PR, shall we also close the one introducing a similar `int` types fix: https://github.com/runtimeverification/evm-semantics/pull/1921?

iFrostizz:
@palinatolmach we could include @nwatson22 changes in this one. It makes me notice that we need to add the `rangeSInt` other than `256` and `128`

anvacaru (approved):

iFrostizz:
Status of this PR:
We first planned to support arrays but for some issues to create arbitrary size arrays because existentials are not permitted in decorators in K. I would need help to support those.

```k
   rule  ( CounterTest . testArray ( V0_numbers : uint256[] ) => #abiCallData ( "testArray", #array ( V0_numbers , ?SIZE , ?ARRAY ) , .TypedArgs ) )
      ensures #rangeUInt ( 128 , ?SIZE )
 ```

So what was done instead:
- Support all static `bytes1..32` variants
- Support all static signed ints `int8..256` variants

We still need to support:
- Dynamic arrays
- Fixed length arrays
- Two dimensional arrays (including static arrays or dynamic for one dimension)
- Structs

PR #1894, opened 2023-06-06
Port prover tests to Python

tothtamas28:
Switches the enumeration and execution of the prover tests on CI to happen in pytest instead of via Makefile logic.


ehildenb:
This looks really nice overall. I would like to make sure of a few things:

- Check that when running locally the same number of tests are run as with `make test-prove` on `master`.
- Check that when running on CI, the same number of tests are run as on `master`.
- Check that if you break a test _not_ listed in `tests/failing-symbolic.pyk`, the test harness registers the breakage correctly (and fails). You can break a test by just modifying one of the output cells that it's actually asserting something about.
- Check that if you break a test that _is_ listed in `tests/failing-symbolic.pyk` (so it uses the legacy prover), then the test harness registers the breakage correctly (and fails).

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -197,7 +197,7 @@ def exec_prove_legacy(
     bug_report: bool = False,
     save_directory: Path | None = None,
     spec_module: str | None = None,
-    claim_labels: Iterable[str] = (),
+    claim_labels: Iterable[str] | None = None,
```
ehildenb, line 200: Why this change?

bin/kevm
```diff
@@ -129,20 +129,20 @@ run_prove() {
...
+        if ${profile_haskell}; then
+            eventlog_name="${run_file}.eventlog"
+            prove_args+=("--haskell-backend-arg=\"+RTS -l -ol${eventlog_name} -RTS\"")
+            cur_dir=$(pwd)
+            run_script="cd "${cur_dir}" && run_kevm_pyk "${prove_args[@]}" "$@""
+                timeout -s INT "${profile_timeout}" bash -c "${run_script}" || true
+                execute kore-prof "${eventlog_name}" ${kore_prof_arg} > "${eventlog_name}.json"
+        else
+            run_kevm_pyk "${prove_args[@]}" "$@"
+        fi
```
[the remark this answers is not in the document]
tothtamas28: Indentation is mixed for this function on `master`: https://github.com/runtimeverification/evm-semantics/blob/bc5943c7550787d1c01aa0b92f66c95d5f1baefe/bin/kevm#L126

I noticed and fixed it.

kevm-pyk/src/kevm_pyk/__main__.py
tothtamas28: `claim_labels` is passed directly to `KProve.get_claims`, where `()` is interpreted as "filter out all claims". So if you call without explicitly specifying all labels, something will be left out.

When calling through the CLI this bug does not occur, as the `Namespace` contains `claim_labels=None`, which in `get_claims` gets interpreted as "filter out no claims". This is obviously violating the original type hint, but the part of code where this value comes from is practically not type checked.

tothtamas28:
> Check that when running locally the same number of tests are run as with `make test-prove` on `master`.

Done. I actually did a bit more: I listed `kprove` invocations with `make --dry-run` and test cases with `pytest --collect-only` (after adjusting for skipped tests), and paired them up. Also confirmed that legacy / pyk tests are matched with the right command line call (i.e. the `--pyk` flag). ([test-cases.txt](https://github.com/runtimeverification/evm-semantics/files/11687699/test-cases.txt))

> Check that when running on CI, the same number of tests are run as on master.

Done.

tothtamas28:
> Check that if you break a test not listed in `tests/failing-symbolic.pyk`, the test harness registers the breakage correctly (and fails).

Done.

> Check that if you break a test that is listed in `tests/failing-symbolic.pyk` (so it uses the legacy prover), then the test harness registers the breakage correctly (and fails).

Done.

ehildenb (approved):

kevm-pyk/src/tests/integration/test_prove.py
```diff
@@ -0,0 +1,168 @@
...
+    module_name = KOMPILE_MODULE.get(spec_id, 'VERIFICATION')
+
+    kevm_kompile(
+        target=KompileTarget.HASKELL,
+        main_file=spec_file,
+        output_dir=definition_dir,
+        main_module=module_name,
+        syntax_module=module_name,
+    )
+
```
ehildenb, line 114: Are we rebuilding the kompiled output every test? That seems needlessly expensive. For example, the MCD proof suite probably has 50+ specs all running against the same definition.

ehildenb: The makefile called `make build-prove-haskell` ahead of time, to build all the definitions once, then invoked the prover on those pre-built definitions. Can we do something similar with a fixture, where we build the definition once for each proof suite, and then execute the proofs?

tothtamas28: Good point.

I created a fixture similar to `pyk`'s `kompile` fixture that memoizes kompiled definitions for these tests. The cache is not shared between runners (I need to do some research to see whether it's possible at all), but it still saves a lot of work.

PR #1895, opened 2023-06-08
don't interpret placeholder as bytecode

iFrostizz:
@lucasmt had some issues during an engagement when using `foundry-show` to analyze a proof that was passing way too fast. The function that calculates the bytecode source maps was making the command crash because it was trying to convert underscore `__` which are the starting characters of library placeholders added by solc for linking purposes. 
This is because it was testing using a deployed library which had at least one visible function.
This PR fixes this issue by checking if the next byte is the start of these characters, if so then jump over it. Note: [the placeholder is 34 characters long](https://docs.soliditylang.org/en/v0.8.19/using-the-compiler.html#library-linking)


iFrostizz:
Note:
We will probably raise an error to the user before that it gets to that point and linking the library manually long term if requested so this PR won't be relevant anymore.

Issue #1826, opened 2023-05-17
Shanghai schedule update


anvacaru:
Schedule spec: https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/shanghai.md

PR #1907, opened 2023-06-15
Shanghai Schedule Update

anvacaru:
Fixes: #1826

Changes included in the Network Upgrade.

- [x] [EIP-3651: Warm COINBASE](https://eips.ethereum.org/EIPS/eip-3651)
- [x] [EIP-3855: PUSH0 instruction](https://eips.ethereum.org/EIPS/eip-3855)
- [x] [EIP-3860: Limit and meter initcode](https://eips.ethereum.org/EIPS/eip-3860)
- [x] [EIP-4895: Beacon chain push withdrawals as operations](https://eips.ethereum.org/EIPS/eip-4895)


anvacaru:
Conformance passing, need to update expected files and have build passing.

tothtamas28 (commented):
Python changes look good.

iFrostizz (approved):

include/kframework/state-utils.md
```diff
@@ -183,6 +186,30 @@ The `"rlp"` key loads the block information.
...
+         <extraData>         _ => #parseByteStackRaw(HX)          </extraData>
+         <mixHash>           _ => #asWord(#parseByteStackRaw(HM)) </mixHash>
+         <blockNonce>        _ => #asWord(#parseByteStackRaw(HN)) </blockNonce>
+         <baseFee>           _ => #asWord(#parseByteStackRaw(HF)) </baseFee>
+         <withdrawalsRoot>   _ => #asWord(#parseByteStackRaw(WR)) </withdrawalsRoot>
+         <ommerBlockHeaders> _ => BU                              </ommerBlockHeaders>
+
     rule <k> load "rlp" : [ [ HP , HO , HC , HR , HT , HE , HB , HD , HI , HL , HG , HS , HX , HM , HN , .JSONs ] , BT , BU , .JSONs ]
           => load "transaction" : BT
          ...
```
ehildenb, line 215: Is the only difference here between the new rule and the old one the fact that now we also have `HF`, `WR`, and `BW`?

If so, perhaps the new rule could deal with just those new parameters, and then reduce to the old rule? It would be less code duplication I guess. Not sure though.

include/kframework/state-utils.md
```diff
@@ -226,6 +253,9 @@ The `"rlp"` key loads the block information.
          <baseFee>           _ => #asWord(#parseByteStackRaw(HF)) </baseFee>
          <ommerBlockHeaders> _ => BU                              </ommerBlockHeaders>
 
+    rule <k> load "genesisRLP": [ [ HP, HO, HC, HR, HT, HE:String, HB, HD, HI, HL, HG, HS, HX, HM, HN, HF, WR, .JSONs ], _, _, _, .JSONs ] => .K ... </k>
+         <blockhashes> .List => ListItem(#blockHeaderHash(HP, HO, HC, HR, HT, HE, HB, HD, HI, HL, HG, HS, HX, HM, HN, HF, WR)) ListItem(#asWord(#parseByteStackRaw(HP))) ... </blockhashes>
+
```
ehildenb, line 258: Similarly, could we reduce to the old rule here and have less code duplication? Maybe it makes less sense here since there is less in common anyway.

.github/workflows/test-pr.yml
```diff
@@ -99,7 +99,7 @@ jobs:
     name: 'Build and Test KEVM concrete execution'
     needs: kevm-pyk-code-quality-checks
     runs-on: [self-hosted, linux, normal]
-    timeout-minutes: 45
+    timeout-minutes: 60
     steps:
       - name: 'Check out code'
         uses: actions/checkout@v3
```
ehildenb, line 105: Are these changes in timeout still needed? Seems like they're leftover from before?

F-WRunTime: Removed these changes.

include/kframework/state-utils.md
anvacaru: That's a great idea! I removed the code duplication.

Issue #1905, opened 2023-06-14
Cheatcodes for creating fresh symbolic values


lucasmt:
It would be good to have cheatcodes that can generate a symbolic value of a given type, for example:
- `freshBytes32()`
- `freshUInt256()`
- `freshBool()`
- `freshAddress()`

In the current engagement we are simulating these by using the `symbolicStorage` cheatcode:
```
contract UInt256Wrap {
    uint256 public value;
}

contract KevmUtil is KEVMCheats {
    function freshUInt256() internal returns (uint256) {
        UInt256Wrap x = new UInt256Wrap();
        kevm.symbolicStorage(address(x));
        return x.value();
    }
}
```
But this creates an overhead since it requires deploying a contract every time.

ehildenb:
How about `freshUInt(uint8)`, so that we can generate fresh uints in any range? Can also do `freshSInt(uint8)` too.

PR #1908, opened 2023-06-15
Implement freshWord cheatcode

iFrostizz:
Closes: https://github.com/runtimeverification/evm-semantics/issues/1905

Adds a new `freshWord` kevm cheatcode for returning a symbolic 32 bytes word in a foundry test.
~~Support for other types is then done by bit masking at the evm level.~~
We will support narrower types by writing the expected bytes width as the argument.
Signed int support is done by explicitly declaring a new cheatcode.


include/kframework/foundry.md
```diff
@@ -552,6 +552,23 @@ This rule then takes the address using `#asWord(#range(ARGS, 0, 32))` and makes
...
+`foundry.call.freshWord` will match when the `freshWord` cheat code function is called.
+This rule returns a symbolic 32 bytes word.
+
+```k
+    rule [foundry.call.freshWord]:
+         <k> #call_foundry SELECTOR _ => . ... </k>
+         <output> _ => #bufStrict(32, WORD) </output>
+      requires SELECTOR ==Int selector ( "freshWord()" )
+       andBool 0 <=Int WORD
+```
```
ehildenb, line 570:
```suggestion
```k
    rule [foundry.call.freshWord]:
         <k> #call_foundry SELECTOR _ => . ... </k>
         <output> _ => #bufStrict(32, WORD) </output>
      requires SELECTOR ==Int selector ( "freshWord()" )
       andBool 0 <=Int WORD
     ensures #rangeUInt(256, WORD)
```
```

**Comment by iFrostizz** [comment]
Added a new `freshBool` cheatcode because it had less overhead, but maybe that we would only create a `kevm.freshBool()` that returns `true` if the output is `> 0 ` and `false` otherwise ?

File: include/kframework/foundry.md [hunk 2]
```diff
@@ -1418,6 +1471,9 @@ If the production is matched when no prank is active, it will be ignored.
     rule ( selector ( "expectEmit(bool,bool,bool,bool,address)" )  => 2176505587 )
     rule ( selector ( "sign(uint256,bytes32)" )                    => 3812747940 )
     rule ( selector ( "symbolicStorage(address)" )                 => 769677742  )
+    rule ( selector ( "freshSInt(uint8)" )                         => 625253732  )
```
lucasmt, line 1474:
```suggestion
    rule ( selector ( "freshSInt(uint8)" )                         => 978336391  )
```
Probably a copy-paste mistake, `625253732` is the selector for `freshUInt(uint8)` below

include/kframework/foundry.md
```diff
@@ -552,6 +552,59 @@ This rule then takes the address using `#asWord(#range(ARGS, 0, 32))` and makes
...
+function freshSInt(uint8) external returns (uint256);
+```
+
+`foundry.call.freshSInt` will match when the `freshSInt` cheat code function is called.
+This rule returns a symbolic integer of up to the bit width that was sent as an argument.
+
+```k
+    rule [foundry.call.freshSInt]:
+         <k> #call_foundry SELECTOR ARGS => . ... </k>
+         <output> _ => #bufStrict(32, ?WORD) </output>
```
lucasmt, line 567: This doesn't work because `#bufStrict(SIZE, DATA)` requires `0 <=Int DATA`. I don't know if it makes sense to have a signed version of this cheatcode, instead it might be better to just implement `freshSInt...()` like
```
function freshSInt256() external returns (int256) {
    return int256(kevm.freshUInt256());
}
```

iFrostizz: Right, nice catch
I don't know if there is less overhead with a native cheatcode ? It seems like there isn't any, we might defer to @ehildenb

ehildenb: Native cheatcodes will usually have lower overhead, because they don't require executing any code aside from calling into the KEVMCheats contract.

Either approach seems fine.

kevm-pyk/poetry.lock
```diff
@@ -1,9 +1,10 @@
-# This file is automatically @generated by Poetry 1.5.1 and should not be changed by hand.
+# This file is automatically @generated by Poetry 1.4.2 and should not be changed by hand.
```
anvacaru, line 1: Is there a reason to downgrade the Poetry version and modify this file?

iFrostizz: This was not intentional, will back out

iFrostizz: thank you !

include/kframework/foundry.md
```diff
@@ -194,7 +194,7 @@ The configuration of the Foundry Cheat Codes is defined as follwing:
     - `<storageSlotSet>` - stores the storage whitelist containing pairs of addresses and storage indexes.
 
 ```k
-module FOUNDRY-CHEAT-CODES
+module FOUNDRY-CHEAT-CODES [symbolic]
```
anvacaru, line 197: Declaring the `FOUNDRY-CHEAT-CODES` module `symbolic` will block the booster compilation. Even if the `symbolic` tag is removed, the compilation is still blocked by existential vars.

tests/foundry/test/FreshInt.t.sol
```diff
@@ -0,0 +1,32 @@
...
+        assertLe(fresh_uint256, 1);
+    }
+
+    function test_int128() public {
+        kevm.infiniteGas();
+        uint128 num = uint128(kevm.freshUInt(16));
+        int128 val = int128(num);
+        assertGe(val, min);
+        assertLe(val, max);
+    }
```
ehildenb, line 31: Can we add a failing test, which asserts something like `assertLe(val, max + 1)`, and then actually check the output of it, by adding it to the `tests/foundry/checkoutput` list?

include/kframework/foundry.md
```diff
@@ -552,6 +552,41 @@ This rule then takes the address using `#asWord(#range(ARGS, 0, 32))` and makes
...
+
+`foundry.call.freshBool` will match when the `freshBool` cheat code function is called.
+This rule returns a symbolic boolean value being either 0 (false) or 1 (true).
+
+```{.k .symbolic}
+    rule [foundry.call.freshBool]:
+         <k> #call_foundry SELECTOR _ => . ... </k>
+         <output> _ => #bufStrict(32, ?WORD) </output>
+      requires SELECTOR ==Int selector ( "freshBool()" )
+        ensures #rangeBool(?WORD)
```
ehildenb, line 587:
```suggestion
    rule [foundry.call.freshBool]:
         <k> #call_foundry SELECTOR _ => . ... </k>
         <output> _ => #bufStrict(32, ?WORD) </output>
      requires SELECTOR ==Int selector ( "freshBool()" )
       ensures #rangeBool(?WORD)
```

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -24,6 +24,16 @@ module BYTES-SIMPLIFICATION [symbolic]
         orBool lengthBytes(B) ==Int lengthBytes(D)
       [simplification]
 
+    rule [signextend-word-less-pow128]:
+      ( ( (signextend ( 15 , WORD:Int )) s<Word (115792089237316195423570985008687907853099843482180094807725896704197245534208) ) ==Int 0 ) => true
+      requires WORD <Int pow128
+      [simplification, symbolic(WORD)]
```
ehildenb, line 30: What number is this? We should give it a name:

image

It's 32 `F` followed by one `8` followed by 31 `0` in hex.

anvacaru: @iFrostizz discovered that it is `maxUInt256 - maxUInt128 / 2`.

iFrostizz: This is also `not` `type(int128).max`
image
So it's `type(int128).min`
image

iFrostizz: In the signed form, it's `typeint(128).min`
image

include/kframework/lemmas/bytes-simplification.k
```diff
@@ -24,6 +24,14 @@ module BYTES-SIMPLIFICATION [symbolic]
         orBool lengthBytes(B) ==Int lengthBytes(D)
       [simplification]
 
+    rule [signextend-min]:
+      (signextend( _ , W:Int) s<Word minSInt128Word ) => 0
+      [simplification, symbolic(W)]
+
+    rule [signextend-max]:
+      (maxSInt128 s<Word signextend( _ , W:Int) ) => 0
+      [simplification, symbolic(W)]
```
anvacaru, line 33:
```suggestion
    rule [signextend-min]: (signextend( _ , W:Int) s<Word minSInt128Word ) => 0 [simplification, symbolic(W)]
    rule [signextend-max]: (maxSInt128 s<Word signextend( _ , W:Int) ) => 0 [simplification, symbolic(W)]
```

anvacaru (approved):

PR #1918, opened 2023-06-21
HOTFIX fix issues found while trying to update HS backend regression tests

jberthold:
* **Restore the ability to run single prover tests** This was broken by https://github.com/runtimeverification/evm-semantics/pull/1894 and is necessary for [Haskell backend test update automation](https://github.com/runtimeverification/haskell-backend/issues/3577) and a custom CI workflow using target `test-prove-smoke`.
* **Fix bug report name** `rstrip` was used but is doing more than `removesuffix`


tothtamas28 (approved):

ehildenb:
Hmmmm, it seems that we should just make the testing harness make it easier to run individual tests? We can already do that, right? I usually run (for example, to run proof `erc20/hkg/transferFrom-success-2-spec.k`):

```
export KEVM_LIB=$(pwd)/.build/usr/lib/kevm
make -C kevm-pyk test-integration TEST_ARGS='-k erc20/hkg/transferFrom-success-2-spec'
```

Is this sufficient?

Also, the `proj` attribute should already be taken care of on master branch I think.

jberthold:
> Hmmmm, it seems that we should just make the testing harness make it easier to run individual tests? We can already do that, right? I usually run (for example, to run proof `erc20/hkg/transferFrom-success-2-spec.k`):
> ```
> export KEVM_LIB=$(pwd)/.build/usr/lib/kevm
> make -C kevm-pyk test-integration TEST_ARGS='-k erc20/hkg/transferFrom-success-2-spec'
> ```
> Is this sufficient?

Yes, that works, but we need to retrieve the bug-report (`kevm-bug-something.tgz`) - it will probably land in a different place now. I could try to use an environment variable (instead of letting `kevm-pyk` come up with a name) to pass the option.

The other Q is for how long this is going to still work... it depends on running proofs with `kprove` (and that is what the extracted regression tests are checking). The restored `Makefile` variables are an easy way to get the functionality back so I went that way first. Can try to use the `test-integration` way instead.
 
> Also, the `proj` attribute should already be taken care of on master branch I think.

Yes, that change was there because the `k` version we run with is different (newer) in Haskell-backend. The `proj` will be gone when evm-semantics is updated to newer `k`.

jberthold:
So, there are two problems when I try to use `make -C kevm-pyk test-integration ...` in my script (which is [btw here](https://github.com/runtimeverification/haskell-backend/pull/3610/files))

1. When I use custom `KORE_EXEC_OPTS` to pass the `--bug-report` flag, the variable gets overwritten. I fixed that [in `pyk` here](https://github.com/runtimeverification/pyk/compare/allow-external-KORE_EXEC_OPTS) (maybe we want to merge that anyway?)
2. Some of the tests that we have chosen as regression tests are not run using the legacy `kprove` any more, they use the `pyk_prove` flavour instead. The regression test is for `kprove` so I would like to call `test_legacy_prove` on those tests in some way if possible. (which gets us back to the question how long this will continue to work...)

@tothtamas28 @ehildenb is there a way to call that legacy test method, or should I keep the `Makefile`  way to run them alive instead?

tothtamas28:
> is there a way to call that legacy test method, or should I keep the `Makefile` way to run them alive instead?

Would it be an option to make the backend regression test generation scripts standalone, i.e. copy the minimal set of artifacts necessary to run `kprove` to the backend repo? Test `test_legacy_prove` basically just kompiles a definition and calls `kprove`, so it's easy to reproduce.  This way, we avoid making the `evm-semantics` test suite effectively part of the public API.

jberthold:
State is kept as branch `hs-backend/test-extraction-branch`, to use in future (manual) test extraction runs - closing this

PR #1923, opened 2023-06-23
Fix automatic gas abstraction re-applying

nwatson22:
This is a fix to the automatic infinite gas abstraction (`--auto-abstract-gas`). It was causing infinite loops where the abstracted node would again be abstracted to a new node, which would then be abstracted, and so on.
Thanks to @lucasmt for finding this.


ehildenb:
@nwatson22 can we add a test of this? Something where we run a test with this option and check the output?

lucasmt:
Ok, so does the gas abstraction require the same amount of communication back-and-forth with the backend? Or does it require more? If it requires more, is it possible that it might counteract the performance gains of not carrying a large expression in the `gas` cell?

nwatson22:
> @nwatson22 can we add a test of this? Something where we run a test with this option and check the output?

I added a test in `test_foundry_prove.py` that uses the option on `GasTest.testInfiniteGas`.

kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -378,10 +378,14 @@ def _replace(term: KInner) -> KInner:
...
                 if type(gas_term) is KApply and gas_term.label.name == 'infGas':
+                    if type(gas_term.args[0]) is KVariable:
+                        return term
                     result = KApply('<gas>', KApply('infGas', abstract_term_safely(term)))
                     return result
                 return term
             elif type(term) is KApply and term.label.name == '<refund>':
+                if type(term.args[0]) is KVariable:
+                    return term
                 return KApply('<refund>', KVariable('ABSTRACTED_REFUND', KSort('Int')))
```
ehildenb, line 389:
```suggestion
                    if type(gas_term.args[0]) is KVariable:
                        return term
                    result = KApply('<gas>', KApply('infGas', abstract_term_safely(term, base_name='VGAS', sort=INT)))
                    return result
                return term
            elif type(term) is KApply and term.label.name == '<refund>':
                if type(term.args[0]) is KVariable:
                    return term
                return KApply('<refund>', abstract_term_safely(term, base_name='VREFUND', sort=INT))
```

kevm-pyk/src/tests/integration/test_foundry_prove.py
```diff
@@ -177,6 +177,27 @@ def test_foundry_bmc(test_id: str, foundry_root: Path, use_booster: bool) -> Non
...
+    show_res = foundry_show(
+        foundry_root,
+        test='GasTest.testInfiniteGas',
+        to_module=True,
+        sort_collections=True,
+        omit_unstable_output=True,
+        pending=True,
+        failing=True,
+        failure_info=True,
+    )
```
ehildenb, line 196:
```suggestion
    show_res = foundry_show(
        foundry_root,
        test='GasTest.testInfiniteGas',
        to_module=True,
        minimize=False,
        sort_collections=True,
        omit_unstable_output=True,
        pending=True,
        failing=True,
        failure_info=True,
    )
```

kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -378,11 +378,17 @@ def _replace(term: KInner) -> KInner:
             if type(term) is KApply and term.label.name == '<gas>':
                 gas_term = term.args[0]
                 if type(gas_term) is KApply and gas_term.label.name == 'infGas':
-                    result = KApply('<gas>', KApply('infGas', abstract_term_safely(term)))
+                    if type(gas_term.args[0]) is KVariable:
+                        return term
+                    result = KApply(
+                        '<gas>', KApply('infGas', abstract_term_safely(term, base_name='VGAS', sort=KSort('Int')))
+                    )
                     return result
```
ehildenb, line 386:
```suggestion
                    return KApply(
                        '<gas>', KApply('infGas', abstract_term_safely(term, base_name='VGAS', sort=KSort('Int')))
                    )
```

ehildenb: Also, you can use `INT` instead of `KSort('Int')` if you do `from pyk.prelude.kint import INT`

ehildenb:
I guess the expected outpput needs to be updated.

PR #1925, opened 2023-06-26
Capture logging for `test_pyk_prove`

tothtamas28:
Captures log messages and writes them to a file `log.txt` in the `pytest` folder.


kevm-pyk/src/tests/integration/test_prove.py
```diff
@@ -210,25 +210,34 @@ def test_pyk_prove(
...
     tmp_path: Path,
+    caplog: LogCaptureFixture,
 ) -> None:
+    caplog.set_level(logging.INFO)
+
     if spec_file in SKIPPED_PYK_TESTS:
         pytest.skip()
 
     # Given
+    log_file = tmp_path / 'log.txt'
```
anvacaru, line 221: Can you define the log file's name somewhere instead of hardcoding it?

anvacaru (approved):
Looks good to me 👍. 
Left a single comment.

kevm-pyk/src/tests/integration/test_prove.py
tothtamas28: It's not hard to implement that. For now, I think hardcoding the name is good enough as it is unique for the test run.

PR #1926, opened 2023-06-26
Utilize dependency claims when discharging KClaim style proofs

h0nzZik:
The function `exec_prove` now builds a dependency graph among all claims of the specified module and attempts to prove the claims in the order from leaves to the main claim. In particular:

- `utils.py` gets a function `claims_dependency_graph` which computes a dependency graph of the claims (and uses qualified names to do so).
- Method `exec_prove` uses `claims_dependency_graph` to compute this dependency graph, and discharges subproof obligations in dependency order using the worker pool.
- Two tests are modified to be enabled on the RPC provers, one which has a dependency in the same module sa the current one and one which has dependencies in a different module (to test dependency name resolution).


test-frob.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+
+KEVM_PREFIX=/nix/store/z30l75z7l3m2nynf52n13cwqz2586rb1-kevm-8ed6754edb8e7e8cadbcaa1fa08f255241dd34b6/
+echo 'Kompiling'
+kevm-pyk kompile --read-only-kompiled-directory --target haskell --definition "$KEVM_PREFIX/lib/kevm/haskell" --syntax-module VERIFICATION-SYNTAX --main-module VERIFICATION ./tests/specs/mcd/vat-frob-diff-zero-dart-pass-rough-reuse-spec.k -o frob-out.d
+
+
+echo 'Proving'
+kevm-pyk prove --verbose --workers 4 --smt-timeout 1000 --smt-retry-limit 5 --max-depth 100 --definition frob-out.d/ -I include/kframework/ -I "$KEVM_PREFIX/lib/kevm/include/kframework/" ./tests/specs/mcd/vat-frob-diff-zero-dart-pass-rough-reuse-spec.k --save-directory ./frob-save.d
```
ehildenb, line 9: This should be removed.

ehildenb:
Going through the CI output here, I only see that the `vat-frob...reuse` spec is skipped, but not that this one is run: 
image


Perhpas it needs to be added to an appropriate list somewhere?

See: https://github.com/runtimeverification/evm-semantics/actions/runs/5795199176/job/15724171072?pr=1926

h0nzZik:
@ehildenb  Going through the outputs in the link you shared, I found the slip test there:
```
[gw6] [ 70%] PASSED src/tests/integration/test_prove.py::test_pyk_prove[mcd/vat-slip-pass-rough-spec.k] 
```
However, I had to use the 'Search logs' Github feature, because I could not find it using the browser's (Firefox) search functionality. Only after finding it with Github's 'Search logs', the browser's search could find it. It is weird; maybe some javascript imaging is happening.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -365,26 +362,50 @@ def _init_and_run_proof(claim: KClaim) -> tuple[bool, list[str] | None]:
...
+            ready = topological_sorter.get_ready()
+            _LOGGER.info(f'ready: {ready}')
+            curr_claim_list = [all_claims_by_label[label] for label in ready]
+            results = process_pool.map(_init_and_run_proof, curr_claim_list)
+            _LOGGER.info(f'done: {ready}')
+            for label in ready:
+                topological_sorter.done(label)
+            selected_results.extend(results)
+            selected_claims.extend(curr_claim_list)
+    _LOGGER.info('finished.')
```
ehildenb, line 397: It looks like we lost some of the new logic from before which checked if there are multiple proofs to run or not. If not, then we don't start a process pool (overhead).

ehildenb: It also looks like there  are some added logging messages for debugging that can be removed.

tests/specs/mcd/vat-slip-pass-rough-spec.k
```diff
@@ -109,9 +109,13 @@ module VAT-SLIP-PASS-ROUGH-SPEC
        andBool #lookup(ACCT_ID_ORIG_STORAGE, #Vat.gem[ABI_ilk][ABI_usr]) ==Int Junk_1
        andBool #Vat.wards[CALLER_ID] =/=Int #Vat.gem[ABI_ilk][ABI_usr]
 
+      [depends(Vat.addui.pass)]
+
+    // Vat_addui
+
```
ehildenb, line 115:
```suggestion
```

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -365,26 +362,50 @@ def _init_and_run_proof(claim: KClaim) -> tuple[bool, list[str] | None]:
...
+    graph = {c.label: [f'{spec_module_name}.{d}' for d in c.dependencies] for c in all_claims}
+    _LOGGER.info(f'graph: {graph}')
+    topological_sorter = graphlib.TopologicalSorter(graph)
+    topological_sorter.prepare()
+    with ProcessPool(ncpus=workers) as process_pool:
+        selected_results = []
+        selected_claims = []
+        while topological_sorter.is_active():
+            ready = topological_sorter.get_ready()
+            _LOGGER.info(f'ready: {ready}')
```
ehildenb, line 389:
```suggestion
            _LOGGER.info(f'Discharging proof obligations: {ready}')
```

h0nzZik:
After the last merge from `master`, I am getting various branchings that IMHO should be discharged by the backend (possibly using the smt solver,) but are not. Both locally and [in the CI](https://github.com/runtimeverification/evm-semantics/actions/runs/5937681998/job/16104643518?pr=1926); in `test_pyk_prove[mcd/vat-slip-pass-rough-spec.k]`, in the top-level proof obligation.

For example:
```
Found 2 branches for node VAT-SLIP-PASS-ROUGH-SPEC.Vat.slip.pass.rough: 1: ['( #Ceil ( #bufStrict ( 32 , ABI_wad:Int ) )\n#And #Ceil ( #encodeArgsAux ( .TypedArgs , 96 , #buf ( 32 , ABI_ilk:Int ) +Bytes #buf ( 32 , ABI_usr:Int ) +Bytes #bufStrict ( 32 , ABI_wad:Int ) , b"" ) ) )', '#Not ( ( #Ceil ( #bufStrict ( 32 , ABI_wad:Int ) )\n#And #Ceil ( #encodeArgsAux ( .TypedArgs , 96 , #buf ( 32 , ABI_ilk:Int ) +Bytes #buf ( 32 , ABI_usr:Int ) +Bytes #bufStrict ( 32 , ABI_wad:Int ) , b"" ) ) ) )']
``` 
(`#bufStrict` on a non-negative integer is always defined)

```
pyk.kcfg.explore:explore.py:379 Found 2 branches using heuristic for node 5: 5: ['{ true #Equals ( lengthBytes ( #bufStrict ( 32 , ABI_wad:Int ) ) +Int lengthBytes ( CD:Bytes ) ) <Int -64 }', '{ true #Equals ( notBool ( lengthBytes ( #bufStrict ( 32 , ABI_wad:Int ) ) +Int lengthBytes ( CD:Bytes ) ) <Int -64 ) }']
```
(`lengthBytes` is always non-negative, therefore greater than -64)

and similar. I have a suspicion that something is not being imported when it should be.

h0nzZik:
So far I have tried:
1. commenting out `include_dirs += [INCLUDE_DIR]` - this would help if that directory shadowed something important
2. increasing smt-timeout to 1000

with no difference. But it is weird to me that the dependency (`Vat.addui.pass`) is proved without such spurious branching - it branches on `['{ true #Equals 0 <=Int ABI_y:Int }', '{ true #Equals ( notBool 0 <=Int ABI_y:Int ) }']`, twice, but that seems reasonable to me. Also, all other test-cases pass, too. Maybe there is some mutable state somewhere, so that when proving something in a second run, things are different? But `kore-rpc` is started new. for every claim. I want to try:
3. "smoke test" - prove `addui` using a copy of `addui` as a dependency. That should work.
4. "admitted/truested lemma" test - what if `addui` is admitted instead of proved, and the `_init_and_run_proof` function for `addui` is never called - only for the top-level theorem? If it goes through, then it indicates the "mutable state" problem.
5. what if the top-level theorem is proved directly, without a specified dependency?

h0nzZik:
When trying to use the current `master` (v1.0.286) to prove its version of  `VAT-SLIP-PASS-ROUGH-SPEC.Vat.slip.pass.rough`, I am getting these debugging messages:

```
kore-rpc: [660158396] Warning (WarnSymbolSMTRepresentation):
    Cannot translate symbol despite being given an SMT-LIB expression
        Lbl'UndsColnUndsUnds'EVM-TYPES'Unds'WordStack'Unds'Int'Unds'WordStack{}
Context:
    (DebugEvaluateCondition) while evaluating predicate
        (DebugAttemptEquation) while applying equation at /nix/store/bxxrklkmfh5vvk9wcwc2ycray94zdyly-kevm-eb5f2a46e8d7d10f3276c2505029b500e0a6d548/lib/kevm/include/kframework/lemmas/lemmas.k:38:10-38:65
```
which might be the reason why the branches are not pruned. For the record, I am using the following testing script (`test-slip.sh`):
```
#!/usr/bin/env bash

KEVM_PREFIX=$(dirname $(which kevm))/..
SPEC="./tests/specs/mcd/vat-slip-pass-rough-spec.k"
echo 'Kompiling'
kevm-pyk kompile --read-only-kompiled-directory --target haskell --definition "$KEVM_PREFIX/lib/kevm/haskell" --syntax-module VERIFICATION-SYNTAX --main-module VERIFICATION "$SPEC" -o slip-out.d


echo 'Proving'
kevm-pyk prove --verbose --workers 4 --smt-timeout 1000 --smt-retry-limit 5 --max-depth 100 --definition slip-out.d/ -I include/kframework/ -I "$KEVM_PREFIX/lib/kevm/include/kframework/" "$SPEC" --save-directory ./slip-save.d
```

ehildenb:
Let's remove the `vat-frob...reuse-spec.k` proof that is added here, because it's not quite working yet. Unelss with newest changes it does work now?

tests/specs/mcd/vat-frob-diff-zero-dart-pass-rough-reuse-spec.k
```diff
@@ -0,0 +1,697 @@
...
+          <output> .Bytes </output>
+          <statusCode> _ => EVMC_SUCCESS </statusCode>
+          <callStack> _VCallStack </callStack>
+          <interimStates> _ </interimStates>
+          <touchedAccounts> _ => ?_ </touchedAccounts>
+          <callState>
+            <program> Vat_bin_runtime </program>
+            <jumpDests> #computeValidJumpDests(Vat_bin_runtime) </jumpDests>
+            <id> ACCT_ID </id>
+            <caller> CALLER_ID </caller>
```
ehildenb, line 23: Let's delete this file

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -316,17 +329,39 @@ def _init_and_run_proof(claim: KClaim) -> tuple[bool, list[str] | None]:
...
+        claim_labels=None,
+        exclude_claim_labels=exclude_claim_labels,
+    )
+
+    if all_claims is None:
+        raise ValueError(f'No claims found in file: {spec_file}')
+
+    all_claims_by_label: dict[str, KClaim] = {c.label: c for c in all_claims}
+    graph = {c.label: [f'{spec_module_name}.{d}' for d in c.dependencies] for c in all_claims}
+    topological_sorter = graphlib.TopologicalSorter(graph)
```
ehildenb, line 348: This seems to me that it won't work for spces that are split into multiple modules? Because it assumes that the `depends` clause name not qualified with the module name?

ehildenb (approved):

PR #1929, opened 2023-06-27
Adjustments to concrete attributes, Makefile updates

ehildenb:
This PR contains the changes that are blocking merging the latest K: https://github.com/runtimeverification/evm-semantics/pull/1919

Basically:

- There were rules with inconsistent application of the `concrete` attribute which need to be updated. Now, no concrete rule set is stored as configuration, instead the attribute is directly on the rules in the semantics.
- The Makefile did not allow controlling parallelism or other arguments from the CLI when running pytest tests, which now it does.
- Support for `--pending` and `--failing` arguments are added to `kevm kcfg-show ...` instead of just `kevm foundry-show-kcfg ...`.
- A lemma is added for `#memoryUsageUpdate` which is needed because of the changes to which rules are `concrete`.


anvacaru (approved):

nwatson22 (approved):

include/kframework/lemmas/lemmas.k
```diff
@@ -91,12 +91,14 @@ module LEMMAS [symbolic]
       [simplification]
 
  // Hardcoded #addrFromPrivateKey simplifications, see: https://github.com/runtimeverification/haskell-backend/issues/3573
-    rule #addrFromPrivateKey("0x0000000000000000000000000000000000000000000000000000000000000001") => 721457446580647751014191829380889690493307935711 [priority(40)]
+    rule #addrFromPrivateKey("0x0000000000000000000000000000000000000000000000000000000000000001") => 721457446580647751014191829380889690493307935711 [priority(40), concrete]
```
Scott-Guest, line 94: The LEMMAS module is symbolic, so if this rule is marked concrete, it'll be ignored by both backends and is effectively dead code. Or am I misunderstanding something?

Scott-Guest (dismissed):

include/kframework/lemmas/lemmas.k
ehildenb: No, `concrete` doesn't mean that as an attribute on rules. It instead means "only apply this rule if all the arguments are `concrete`". Doesn't affect whichbackends see the rule.

Scott-Guest: Ah okay - we should update the K User Manual to better document this. The current documentation only mentions that behavior specifically for simplification rules: https://kframework.org/docs/user_manual/#concrete-and-symbolic-attributes-(haskell-backend)

PR #1934, opened 2023-06-30
Add foundry-merge-nodes option

nwatson22:
Adds an option `foundry-merge-nodes` used with:
```
kevm foundry-merge-nodes [Contract.test_name] --node [node id] --node [node id]
 ```
which creates a new node which is the `anti_unify_with_constraints` of all the listed nodes and makes this node a cover of all the listed nodes.
Optionally, include the `--include-disjunct` flag to keep the abstracted nodes as an `orBool` disjunction, e.g.
```
<k> 1 </k>
```
merged with
```
<k> 2 </k>
```
would yield:
```
<k> N:Int </k>
#And (N ==K 1 orBool N ==K 2)
```
Additionally, the PR changes `foundry-remove-node` so that it no longer prunes the target node if it is a child of the removed node, and fixes `foundry-step-node`, `foundry-section-edge`, `foundry-remove-node`, `foundry-list` and `foundry-to-dot`, which were assuming that the proof being operated on was an `APRProof` and not an `APRBMCProof`, so they converted the proof back to an `APRProof`.

### CODE REVIEW

File: kevm-pyk/src/kevm_pyk/foundry.py [hunk 1]
```diff
@@ -727,6 +727,40 @@ def foundry_simplify_node(
...
+
+    _nodes = [apr_proof.kcfg.node(int(node_id)) for node_id in nodes]
+
+    anti_unification: KInner
+    for i, node in enumerate(_nodes):
+        cterm = node.cterm
+        if i < 1:
+            anti_unification = cterm.kast
+            continue
+        anti_unification = anti_unify_with_constraints(anti_unification, cterm.kast, disjunct=False)
```
ehildenb, line 751:
```suggestion
    if len(nodes) < 2:
        raise ValueError(f'Must supply at least 2 nodes to merge, got: {nodes}')

    _nodes = [apr_proof.kcfg.node(int(node_id)) for node_id in nodes]

    anti_unification = _nodes[0]
    for node in _nodes[1:]:
        anti_unification = anti_unify_with_constraints(anti_unification, node.cterm.kast, disjunct=False)
```

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -727,6 +727,40 @@ def foundry_simplify_node(
...
+
+    new_cterm = CTerm.from_kast(anti_unification)
+    new_node = apr_proof.kcfg.create_node(new_cterm)
+
+    for node in _nodes:
+        apr_proof.kcfg.create_cover(node.id, new_node.id)
```
ehildenb, line 757:
```suggestion
    new_node = apr_proof.kcfg.create_node(CTerm.from_kast(anti_unification))
    for node in _nodes:
        apr_proof.kcfg.create_cover(node.id, new_node.id)
```

ehildenb:
Let's reduce whitespace a bit, no need to define some of the intermediates (eg. intermeditae for `new_cterm`, can just directly create it an insert it as a node). Also, instead of naming the variables `apr_proof`, I've been trying to switch to just the more compact `proof` name everywhere.

ehildenb:
We should also be putting a sanity check on the nodes the user has asked to merge. Specifically, they all must:

- Have the exact same content in the following cells: `<k>`, `<program>`, `<pc>`, `<callDepth>`.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -727,6 +730,50 @@ def foundry_simplify_node(
...
+                return False
+        return True
+
+    node_ids = [int(node) for node in node_ids]
+    br = BugReport(Path(f'{test}.bug_report')) if bug_report else None
+    foundry = Foundry(foundry_root, bug_report=br)
+    proofs_dir = foundry.out / 'apr_proofs'
+    contract_name, test_name = test.split('.')
+    proof_digest = foundry.proof_digest(contract_name, test_name)
+    proof = APRProof.read_proof(proof_digest, proofs_dir)
```
ehildenb, line 755: Should we be using `Proof.read_proof` instead, so that it correctly determines if it's an `APRProof` or an `APRBMCProof`?

I think for the engagement, we're using `APRBMCProof` instead of `APRProof`, so we should probably make sure it works with taht.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -763,6 +810,21 @@ def foundry_step_node(
...
+    bug_report: bool = False,
+) -> APRProof:
+    br = BugReport(Path(f'{test}.bug_report')) if bug_report else None
+    foundry = Foundry(foundry_root, bug_report=br)
+
+    apr_proofs_dir = foundry.out / 'apr_proofs'
+    contract_name, test_name = test.split('.')
+    proof_digest = foundry.proof_digest(contract_name, test_name)
+    apr_proof = APRProof.read_proof(proof_digest, apr_proofs_dir)
+    return apr_proof
```
ehildenb, line 825: Let's:

- Use `proof` variable name instead of `apr_proof`.
- Make sure it works with both `APRProof` and `APRBMCProof`, by using `Proof.read_proof` instead. We can make an assertion at the end of this function that `isinstance(proof, APRProof)` to make sure it's one of those subclasses if you want/need.

kevm-pyk/src/kevm_pyk/foundry.py

kevm-pyk/src/kevm_pyk/foundry.py
nwatson22: I renamed it here, but also for most of the lesser-used commands like `foundry-step-node`, `foundry-remove-node`, etc., those commands were converting `APRBMCProof`s back into `APRProof`s, and I fixed these so I could test the merging feature with BMC proofs.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -713,6 +725,7 @@ def exec_foundry_section_edge(
     trace_rewrites: bool = False,
     **kwargs: Any,
 ) -> None:
+    print(edge)
```
nwatson22, line 728:
```suggestion
```

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -470,6 +470,9 @@ def foundry_prove(
         for method in contract.methods
         if f'{contract.name}.{method.name}' not in all_tests
     ]
+
+    print([method.name for method in foundry.contracts['AssertTest'].methods])
+
```
nwatson22, line 475:
```suggestion
```

nwatson22:
Pulled out `foundry_get_apr_proof` and `foundry_get_proof` into #2002.

nwatson22:
I think this is now failing because the `#lookup` is not working because the merge is deleting the sort information on the two values it merges. It should be able to infer the weakest sort that covers both. I can't explain why [5f99e71](https://github.com/runtimeverification/evm-semantics/pull/1934/commits/5f99e71695d7c9408aaafafdfd65b5ac456ec2eb) passed CI.

kevm-pyk/src/kontrol/foundry.py
```diff
@@ -813,6 +813,53 @@ def foundry_simplify_node(
...
+    foundry = Foundry(foundry_root, bug_report=br)
+    proof = foundry.get_apr_proof(test)
+
+    if not isinstance(proof, APRProof):
+        raise ValueError('Specified proof is not an APRProof.')
+
+    if len(list(node_ids)) < 2:
+        raise ValueError(f'Must supply at least 2 nodes to merge, got: {node_ids}')
+
+    nodes = [proof.kcfg.node(int(node_id)) for node_id in node_ids]
```
ehildenb, line 844:
```suggestion
    br = BugReport(Path(f'{test}.bug_report')) if bug_report else None
    foundry = Foundry(foundry_root, bug_report=br)
    proof = foundry.get_apr_proof(test)

    if not isinstance(proof, APRProof):
        raise ValueError('Specified proof is not an APRProof.')

    if len(list(node_ids)) < 2:
        raise ValueError(f'Must supply at least 2 nodes to merge, got: {node_ids}')

    nodes = [proof.kcfg.node(nid) for nid in node_ids]
```

Why do we need these type-casts?

ehildenb (approved):

kevm-pyk/src/kontrol/foundry.py
nwatson22: It probably had to do with how I was reading them in from the command line. It looks like it doesn't matter anymore so I removed it.

PR #1935, opened 2023-07-03
Update build instructions for Mac Silicon

palinatolmach:
Updates instructions for building KEVM on Apple Silicon. Tested on a Mac with M1 running Ventura 13.4.1.

The changes are:
- Removes `make libsecp256k1` from the building instructions
- Removes `APPLE_SILICON=true` from the `make k-deps`command
- Adds `APPLE_SILICON=true` to the `make plugin-deps` command
- Removes extra `K_MVN_ARGS` when `make k-deps` is run with `APPLE_SILICON=true` (it seems to be only used for building Haskell backend, and I could only build it without `APPLE_SILICON=true`).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,11 +90,6 @@ endif
 ifneq ($(SKIP_HASKELL),)
     K_MVN_ARGS += -Dhaskell.backend.skip
 endif
-
-ifneq ($(APPLE_SILICON),)
-    K_MVN_ARGS += -Dstack.extra-opts='--compiler ghc-8.10.7 --system-ghc'
-endif
-
 ifneq ($(RELEASE),)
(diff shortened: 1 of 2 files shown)


palinatolmach:
@Baltoli I have updated some of the instructions for building `evm-semantics` on an M1 Mac according to what worked for me last week. If you still have access to an M1/M2 system, it would be great if you help me verify that the updated instructions work, when you have a moment. Or I would also appreciate an advice on who can I approach with a similar request :). Thank you!

Baltoli:
I'm a little suspicious of a couple of these changes; let me try them on a clean M1 machine and get back to you. Thanks @palinatolmach!

PR #1946, opened 2023-07-07
Output counterexamples for failed Foundry proofs

### LINKED ISSUE #1927: Output concrete counter-example for failed proofs
Related: https://github.com/runtimeverification/evm-semantics/issues/1700 and https://github.com/runtimeverification/evm-semantics/pull/1716

~Blocked on: https://github.com/runtimeverification/haskell-backend/issues/3616~

Following https://github.com/runtimeverification/evm-semantics/pull/1716, we now have the path-feasibility constraints being output on failing proofs. We want to improve on this by providing the user with concrete assignment to variables which demonstrates the failed test. This is now possible, because we have implemented the `get-model` endpoint in the Haskell backend: (https://github.com/runtimeverification/haskell-backend/commit/06c9515d802093b7e3267e2d3e7460f4a2cf8ede and https://github.com/runtimeverification/pyk/pull/509).

We should expose this functionality in the KEVM foundry integration in these ways:

- First, add `KCFGExplore.get_model(cterm: CTerm) -> Subst | None`, which takes a constrained term and returns either a `Subst` (model found), or `None`, no model found. This should happen in `KCFGExplore` class, and can have psuedocode like:

```
diff --git a/src/pyk/kcfg/explore.py b/src/pyk/kcfg/explore.py
index 4ad3e09a..e99dc4d7 100644
--- a/src/pyk/kcfg/explore.py
+++ b/src/pyk/kcfg/explore.py
@@ -168,6 +168,13 @@ class KCFGExplore(ContextManager['KCFGExplore']):
         kast_simplified = self.kprint.kore_to_kast(kore_simplified)
         return kast_simplified, logs
 
+    def cterm_get_model(self, cterm: CTerm) -> Subst | None:
+        # convert the cterm to Kore
+        # call get_model in the RPC server
+        # convert the Kore result back to Kast
+        # convert the kast to a substitution
+        # return the substitution
+
     def kast_simplify(self, kast: KInner) -> tuple[KInner, tuple[LogEntry, ...]]:
         _LOGGER.debug(f'Simplifying: {kast}')
         kore = self.kprint.kast_to_kore(kast, GENERATED_TOP_CELL)
```

- Then, make a testing harness for this `KCFGExplore.get_model`, which tests it with IMP. That can look something like:

```
diff --git a/src/tests/integration/kcfg/test_imp.py b/src/tests/integration/kcfg/test_imp.py
index 925b1ddf..cfe39e46 100644
--- a/src/tests/integration/kcfg/test_imp.py
+++ b/src/tests/integration/kcfg/test_imp.py
@@ -518,6 +518,22 @@ APRBMC_PROVE_TEST_DATA: Iterable[
     ),
 )
 
+GET_MODEL_TEST_DATA = (
+    (
+        'model-works',
+        'int $n;',
+        '$n |-> N',
+        ['N >=Int 3'],
+        Subst({'N', token(4)}), # True
+    ),
+    (
+        'model-doesnt-work',
+        'int $n;',
+        '$n |-> N',
+        ['N >=Int 3', 'N <Int 2'],
+        None, # False
+    ),
+)
 
 def leaf_number(proof: APRProof) -> int:
     non_target_leaves = [nd for nd in proof.kcfg.leaves if not proof.is_target(nd.id)]
@@ -847,3 +863,24 @@ class TestImpProof(KCFGExploreTest):
 
         assert proof.status == proof_status
         assert leaf_number(proof) == expected_leaf_number
+
+    @pytest.mark.parametrize(
+        'test_id,k_cell,state_cell,model_status',
+        GET_MODEL_TEST_DATA,
+        ids=[test_id for test_id, *_ in GET_MODEL_TEST_DATA],
+    )
+    def test_get_model(
+        self,
+        kprove: KProve,
+        kcfg_explore: KCFGExplore,
+        test_id: str,
+        k_cell: str,
+        state_cell: str,
+        constraints: Iterable[str],
+        expected_model: Subst | None,
+    ) -> None:
+        cterm = config(k_cell, state_cell, *constraints)
+
+        model = kcfg_explore.cterm_get_model(cterm)
+
+        assert model == expected_model
```

- Then, test this with KEVM. We should modify `foundry_prove` (here: https://github.com/runtimeverification/evm-semantics/blob/93e5e06800d88663bc206fcdafb15c7df917357b/kevm-pyk/src/kevm_pyk/foundry.py#L537), to check if there are any failing nodes, and if so, call `get_model` on them, and output that information to the failure log. That can look something like this:

```
diff --git a/kevm-pyk/src/kevm_pyk/foundry.py b/kevm-pyk/src/kevm_pyk/foundry.py
index 6a6173a1a..5c30f9c32 100644
--- a/kevm-pyk/src/kevm_pyk/foundry.py
+++ b/kevm-pyk/src/kevm_pyk/foundry.py
@@ -535,6 +535,9 @@ def foundry_prove(
             failure_log = None
             if not passed:
                 failure_log = print_failure_info(proof, kcfg_explore)
+                for node in proof.failing:
+                    model = kcfg_explore.cterm_get_model(node.cterm)
+                    failure_log.append(model)
 
             return passed, failure_log
```

- Finally, we can expose a direct command `kevm foundry-get-model TEST_NAME NODE_ID`, which gets a model for a specific node, and can take options `--failing` and `--pending`, to pre-populate the list of nodes to include in the get-model call (similar to how `exec_foundry_show` works). We can also modify `exec_foundry_show` to include get-model information directly perhaps.

palinatolmach:
Closes #1927.

1. Using `get-model` request, adds concrete counterexamples (i.e., models) of failing and pending nodes to `print_failure_info(proof, kcfg_explore)`. 
UPDATE: after a discussion with @nwatson22, we decided that we should probably only show counterexamples for failing (not pending) nodes by default.
This information is shown in the output of `foundry-prove` if any of the proofs has failed, or `foundry-show --failure-information`. Also, prints the invitation to the Discord channel once in `print_failure_info(print_failure_info(proof, kcfg_explore)` (instead of printing it with every failing node info), and fixes the link to the channel.
UPDATE: the counterexample is generated if `foundry-prove` and `foundry-show` are called with the `--counterexample-information` option.
The output looks like this: 
```
> `kevm foundry-prove --test ExampleTest.testBackdoor`

... 

PROOF FAILED: ExampleTest.testBackdoor
1 Failure nodes. (0 pending and 1 failing)

Failing nodes:

  Node id: 20
  Failure reason:
    Implication check failed, the following is the remaining implication:
    ( ( { true #Equals 0 <=Int CALLER_ID:Int }
    ...
  Path condition:
    { true #Equals chop ... ==Int 0 }
  Model:
    CALLER_ID = 0
    NUMBER_CELL = 0
    ORIGIN_ID = maxUInt160
    VV0_x_114b9705 = 6912213124124532

Join the Runtime Verification Discord server for support: https://discord.com/invite/CurfmXNtbN

Access documentation for KEVM foundry integration at https://docs.runtimeverification.com/kevm-integration-for-foundry/
```

4. Adds `foundry-get-model` command that shows models of nodes from a specific test. Can be called with `--node NODE_ID`, `--failing`, and `--pending` options. If none of these arguments are provided, the models of failing and pending nodes will be shown (i.e., missing `--node`, `--failing` and/or `--pending` is equivalent to `--failing --pending`), and a corresponding warning is shown to the user. The output looks like this:
```
> kevm foundry-get-model ExampleTest.testBackdoor --node 20

Node id: 20
  Model:
    CALLER_ID = 0
    NUMBER_CELL = 0
    ORIGIN_ID = maxUInt160
    VV0_x_114b9705 = 6912213124124532
```


kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -226,6 +230,20 @@ def print_failure_info(proof: Proof, kcfg_explore: KCFGExplore) -> list[str]:
...
+    result_subst = kcfg_explore.cterm_get_model(node.cterm)
+    if type(result_subst) is Subst:
+        res_lines.append('  Model:')
+        for var, term in result_subst.to_dict().items():
+            term_kast = KInner.from_dict(term)
+            res_lines.append(f'    {var} = {kcfg_explore.kprint.pretty_print(term_kast)}')
+    else:
+        res_lines.append('  Failed to generate a model.')
+
+    return res_lines
```
nwatson22, line 244: We rewrote the `print_failure_info` function in pyk to separate the printing from the RPC calls here: https://github.com/runtimeverification/pyk/commit/6c28b6613ea058a46fd4f838982a88bc2e957863
Possibly we would want to do something similar here, since it makes testing cleaner. I realized we haven't updated kevm to use that function yet and we still just use the kevm version, so actually we can probably just refactor this when we make those changes later.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -795,6 +795,45 @@ def foundry_section_edge(
...
+
+    res_lines = []
+
+    with KCFGExplore(foundry.kevm, id=proof.id) as kcfg_explore:
+        for node_id in nodes:
+            res_lines.append('')
+            res_lines.append(f'Node id: {node_id}')
+            node = proof.kcfg.node(node_id)
+            model_info = print_model(node, kcfg_explore)
+            res_lines.extend(model_info)
```
nwatson22, line 832:
```suggestion
            res_lines.extend(print_model(node, kcfg_explore))
```
I'm not sure which is better (same for other places this occurs).

nwatson22 (approved):

tothtamas28 (approved):

kevm-pyk/src/kevm_pyk/foundry.py
palinatolmach: Thanks! I agree, done in https://github.com/runtimeverification/evm-semantics/commit/13c0f97d313913202f9a79bdd796b15915959bab.

kevm-pyk/src/kevm_pyk/utils.py
palinatolmach: Thanks Noah! I can probably try integrating the new `failure_info` function into kevm to include it in this PR, if this works--will update you on that tomorrow.

palinatolmach:
Updates: (1) added the temporary `--counterexample-information` option to `foundry-prove` and `foundry-show`, which generates the counterexample for each failing node and adds it to the output. Since it's currently part of the `print_failure_info()` implementation, in `foundry-show`, it should be called with the `--failure-information` option (it's reflected in its description), i.e., `foundry-show --failure-information --counterexample-information`. The purpose of adding this option is to make sure this change (and the SMT solver requests it includes) doesn't cause major performance issues, while we test it a bit more.

(2) when `foundry-prove --counterexample-information` is called, the counterexamples are only generated for failing (and not pending) nodes.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -795,6 +797,44 @@ def foundry_section_edge(
     apr_proof.write_proof()
 
 
+def foundry_get_model(
```
tothtamas28, line 800: This function should produce a data structure (maybe `FoundryModel | None`). Function `print_model` can be a method on this class (without the `"Failed to ..."` branch). Function `print_failure_info` then can be redefined as

```python3
def print_failure_info(kprint: KPrint, proof: Proof, *, model: FoundryModel | None = None) -> list[str]:
    ...
```

palinatolmach: Thanks, I agree! Will add the `FoundryModel` class and modify the printing functions accordingly (there's some more refactoring to be done, as suggested by @nwatson22), but I need a bit more time to work on it. Would you recommend to merge the current PR and refactor it later, or should I include this change in this PR? Thank you!

tothtamas28: It's fine either way. If you decide to merge now, please open an issue to document the requested changes.

palinatolmach: Issue created: https://github.com/runtimeverification/evm-semantics/issues/1955

Issue #1916, opened 2023-06-19
Add unit test for `foundry-list`


tothtamas28:
To replace `test-foundry-list`.

https://github.com/runtimeverification/evm-semantics/blob/5734d0f5512794b6c07239e2f886bca4b2317b77/Makefile#L465

PR #1952, opened 2023-07-12
foundry-list unit test

iFrostizz:
Add a unit test for testing the output of `foundry-list`.
This is at the moment mocking `foundry-list` by calling `APRProof.summary`, but will be rendered more robust by refactoring the output of `APRProof.summary` to return a proper data structure.


iFrostizz:
Ref: https://github.com/runtimeverification/evm-semantics/issues/1916

kevm-pyk/src/tests/unit/test_foundry_list.py
```diff
@@ -0,0 +1,50 @@
...
+import pytest
+from pyk.proof import APRProof
+
+from tests.utils import REPO_ROOT
+
+if TYPE_CHECKING:
+    from pathlib import Path
+    from typing import Final
+
+TEST_DATA = REPO_ROOT / 'kevm-pyk/src/tests/unit/test-data'
```
tothtamas28, line 15: Create a file `utils.py` in `src/test/unit`. There, define `TEST_DATA_DIR` relative to that file. (See `integration/utils.py` for an example.)

kevm-pyk/src/tests/unit/test_foundry_list.py
```diff
@@ -0,0 +1,50 @@
...
+lists = foundry_list_expected.split('\n\n')
+assert len(lists) == len(APR_PROOFS)
+
+list_outputs = {list_output.split('\n')[0].split(': ')[1]: list_output.rstrip('\n') for list_output in lists}
+
+
+@pytest.mark.parametrize(
+    'apr_proof', APR_PROOFS, ids=[str(apr_proof.relative_to(TEST_DATA)) for apr_proof in APR_PROOFS]
+)
+def test_foundry_list(apr_proof: Path) -> None:
```
tothtamas28, line 40: This test performs logic. What it should do instead is just call `foundry_list` on a folder that contains a minimal KEVM-Foundry setup (prepared somewhere under `test-data`), and compare to the expected output.

tothtamas28 (changes requested):

kevm-pyk/src/tests/unit/test_foundry_list.py
iFrostizz: Much better, thank you !

kevm-pyk/src/tests/unit/test_foundry_list.py
iFrostizz: Thanks, there is now a setup foundry repo in `unit/test-data/foundry-list`

iFrostizz:
@tothtamas28  As it's a unit test, the foundry repo is there and ready for use. But as you can see it adds a lot of files.
Should we let foundry at least handle the `forge build` to install forge-std and produce the artifacts to air up a bit on space or leave as-is ?

tothtamas28:
> As it's a unit test, the foundry repo is there and ready for use. But as you can see it adds a lot of files. 

Right, I agree, please try to minimize this. A lot of files are necessary for constructing the proofs, but for `foundry_prove` only a small portion of these files will ever be touched.

Once the test setup is minimized, we can still mock some dependencies to reduce it further.

iFrostizz:
I don't think that I can reduce this much more, the next step would be to foundry-kompile but it may not be a unit test anymore

kevm-pyk/.gitignore
```diff
@@ -1,3 +1,11 @@
...
 __pycache__/
 .coverage
+src/tests/unit/test-data/foundry-list/cache/
+src/tests/unit/test-data/foundry-list/lib/
+src/tests/unit/test-data/foundry-list/out/
+!src/tests/unit/test-data/foundry-list/out/apr_proofs/
+!src/tests/unit/test-data/foundry-list/out/kompiled/
+!src/tests/unit/test-data/foundry-list/out/kompiled-llvm/
+!src/tests/unit/test-data/foundry-list/out/tmp/
+!src/tests/unit/test-data/foundry-list/out/digest
```
tothtamas28, line 11:
```suggestion
```

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -664,6 +664,7 @@ def foundry_to_dot(foundry_root: Path, test: str) -> None:
 def foundry_list(foundry_root: Path) -> list[str]:
     foundry = Foundry(foundry_root)
     apr_proofs_dir = foundry.out / 'apr_proofs'
+    print(apr_proofs_dir)
```
tothtamas28, line 667: Probably leftover from debugging.

kevm-pyk/src/tests/unit/test-data/foundry-list/out/kompiled/allRules.txt
tothtamas28, line 1: I'm pretty sure we don't need a kompiled definition to test this functionality. If necessary we can create a simple mock instead.

kevm-pyk/src/tests/unit/test_foundry_list.py
```diff
@@ -0,0 +1,36 @@
...
+        foundry_list_expected = f.read().rstrip()
+    assert foundry_list_expected
+
+    try:
+        run_process(['forge', 'install', '--no-git', f'foundry-rs/forge-std@{FORGE_STD_REF}'], cwd=LIST_DATA_DIR)
+    except CalledProcessError:
+        pass
+
+    run_process(['forge', 'build'], cwd=LIST_DATA_DIR)
+    list_out = '\n'.join(foundry_list(LIST_DATA_DIR))
```
tothtamas28, line 31: Just add the compiled `json` files statically if they're needed.

kevm-pyk/src/tests/unit/test-data/foundry-list/out/kompiled/allRules.txt
iFrostizz: Indeed looks like we don't need it

kevm-pyk/.gitignore
iFrostizz: Should we remove all those ignored folders ? We probably don't need to commit cache for instance, as well as {kompiled, kompiled-llvm, tmp} as they have shown to be unnecessary

tothtamas28: Right, but they shouldn't be there in the local tree either.

iFrostizz: Totally makes sense! Though, how do we prevent other people from committing them  once they launch the unit test ?

tothtamas28: My point is, unit tests should not generate these files. Whatever you need to run `foundry-list` has to be provided as a minimal set of files as test data.

iFrostizz: Of course, Thanks!

tothtamas28 (commented):
I think this can be minimized further. Let me do some experiments with the setup, then we can discuss.

kevm-pyk/src/tests/unit/test_foundry_list.py
```diff
@@ -0,0 +1,63 @@
...
+if TYPE_CHECKING:
+    from typing import Final
+
+LIST_DATA_DIR: Final = TEST_DATA_DIR / 'foundry-list'
+LIST_OUT: Final = LIST_DATA_DIR / 'out/'
+LIST_APR_PROOF: Final = LIST_OUT / 'apr_proofs/'
+LIST_EXPECTED: Final = LIST_DATA_DIR / 'foundry-list.expected'
+
+
+def contracts() -> dict[str, Contract]:
```
tothtamas28, line 21: This logic should not be replicated here. Instead, consider mocking `Foundry`.

iFrostizz: Much better, thanks !

kevm-pyk/src/tests/unit/test-data/foundry-list/foundry.toml
tothtamas28, line 1: Now that the mock is set up, is this file necessary?

kevm-pyk/src/tests/unit/test_foundry_list.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+LIST_DATA_DIR: Final = TEST_DATA_DIR / 'foundry-list'
+LIST_OUT: Final = LIST_DATA_DIR / 'out/'
+LIST_APR_PROOF: Final = LIST_OUT / 'apr_proofs/'
+LIST_EXPECTED: Final = LIST_DATA_DIR / 'foundry-list.expected'
+
+
+class FoundryMock:
+    out = LIST_OUT
```
tothtamas28, line 32: Consider

```suggestion
class FoundryMock:
   out: Path

    def __init__(self, out: Path)
        sef.out = out
```

kevm-pyk/src/tests/unit/test_foundry_list.py
```diff
@@ -0,0 +1,74 @@
...
+            proof_dict = json.loads(Path(proof_path).read_text())
+            pid = proof_dict['id']
+            if pid.startswith(f'{contract}.{test}'):
+                return pid
+        return ''
+
+
+def test_foundry_list(mocker: MockerFixture, update_expected_output: bool) -> None:
+    foundry_mock = mocker.patch('kevm_pyk.foundry.Foundry')
+    foundry_mock.return_value = FoundryMock()
```
tothtamas28, line 63: I think this should be

```suggestion
    foundry_mock.return_value = FoundryMock
```

kevm-pyk/src/tests/unit/test_foundry_list.py
```diff
@@ -0,0 +1,74 @@
...
+
+    with LIST_EXPECTED.open() as f:
+        foundry_list_expected = f.read().rstrip()
+    assert foundry_list_expected
+
+    list_out = '\n'.join(foundry_list(LIST_DATA_DIR))
```
tothtamas28, line 69: If you add `__init__` to the mock, you can add

```suggestion
    list_out = '\n'.join(foundry_list(LIST_DATA_DIR))
    
    assert mock.called_once_with(LIST_DATA_DIR)
```

This verifies the test setup to some extent.

kevm-pyk/src/tests/unit/test_foundry_list.py
```diff
@@ -0,0 +1,74 @@
...
+LIST_OUT: Final = LIST_DATA_DIR / 'out/'
+LIST_APR_PROOF: Final = LIST_OUT / 'apr_proofs/'
+LIST_EXPECTED: Final = LIST_DATA_DIR / 'foundry-list.expected'
+
+
+class FoundryMock:
+    out = LIST_OUT
+
+    @cached_property
+    def contracts(self) -> dict[str, Contract]:
```
tothtamas28, line 35: It would be good to keep mock implementation to the minimum, i.e. have a very simple implementations for `contracts` and `proof_digest`. In order to achieve that try to use simple static values. You can also simplify test data (basically rename files or their content) if necessary.

For example we don't really care if `digest` is realistic, we just want to match the value in the expected output.

kevm-pyk/src/tests/unit/test-data/foundry-list/foundry.toml
iFrostizz: It is not ! Thanks

kevm-pyk/src/tests/unit/test_foundry_list.py
iFrostizz: Thank you!

kevm-pyk/src/tests/unit/test_foundry_list.py
iFrostizz: Well then considering the other comment it should now be `FoundryMock(LIST_OUT)`, right ?

kevm-pyk/src/tests/unit/test_foundry_list.py
iFrostizz: Thanks!

kevm-pyk/src/tests/unit/test_foundry_list.py
tothtamas28: Well, we're mocking a class with another, so I'd expect it to work with just `FoundryMock`. Can you check to make sure?

iFrostizz: Oh I get it, I think that in the above review you meant

```python
class FoundryMock:
   out: Path

    def __init__(self)
        sef.out = LIST_OUT
```

iFrostizz: Calling `FoundryMock` does not seem to invoke the `__init__`

tothtamas28: Oh yeah, that makes sense. Maybe we should use `monkeypatch` then, that's what we're using `mocker` for anyways:

```python3
from pytest import MonkeyPatch

import foundry

def test(monkeypatch: MonkeyPatch):
    monkeypatch.setattr(foundry, 'Foundry', FoundryMock)
```

But the current approach is fine as well.

kevm-pyk/src/tests/unit/test_foundry_list.py
iFrostizz: Thank you so much !!

kevm-pyk/src/tests/unit/test_foundry_list.py
iFrostizz: Interesting, please tell me if the current implementation fits

iFrostizz:
The kevm library is still lazily requested from the import of the `foundry_list` and so the unit test fails on CI

deps/k
tothtamas28, line 1: Please restore these changes.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -121,6 +120,8 @@ def digest(self) -> str:
 
     @cached_property
     def llvm_dylib(self) -> Path | None:
+        from .config import Kernel
```
tothtamas28, line 123: I think you only need these local imports to make the unit test work, the code movements is unnecessary.

```suggestion
        from .kompile import Kernel
```

tothtamas28 (approved):

iFrostizz:
@tothtamas28 I'd require your input on simplifying those digest.
With the new proof system, the proofs are stored as `Contract.Method:digest`.
In that test, the digest was replaced by some static data according to your review so that it would just return `Contract.Method`.
Does that make the test too brittle, or is it okay to not test the digest ?

tothtamas28:
> With the new proof system, the proofs are stored as `Contract.Method:digest`.
In that test, the digest was replaced by some static data according to your review so that it would just return `Contract.Method`.
Does that make the test too brittle, or is it okay to not test the digest ?

It's fine both ways. If computing the digest does not introduce dependencies that are problematic, let's compute it. Otherwise, let's just stick with the mock.

tothtamas28 (approved):

PR #1957, opened 2023-07-14
Pass llvm_definition_dir to KCFGExplore in foundry-prove

goodlyrottenapple:
Requires https://github.com/runtimeverification/pyk/pull/514/ to be merged into pyk

This change makes sure we include the llvm definition files in the bug_report.tar when using the booster via `--use_booster`


tothtamas28 (commented):
Looks good. Let's wait for the `pyk` update to test it on CI.

tothtamas28 (approved):

Issue #1958, opened 2023-07-18
Make performance improving options default


palinatolmach:
Related: https://github.com/runtimeverification/evm-semantics/pull/1520, https://github.com/runtimeverification/evm-semantics/pull/1963

We've observed that significant performance improvements (e.g., 2m50s instead of 4m30s) for Foundry proofs can be obtained by supplying the following set of flags/optimizations:

- [ ] `--max-depth` equal to 25000
- [ ] `--no-break-on-calls`
- [ ] ~`--use-booster`~  (making booster the default is postponed till its behavior is stabilized)
- [ ] `infiniteGas` cheatcode

Consider using these options by default.

palinatolmach:
Closed in favor of two separate issues, https://github.com/runtimeverification/kontrol/issues/39 and https://github.com/runtimeverification/kontrol/issues/38.

PR #1963, opened 2023-07-24
Make performance improving options default

palinatolmach:
This PR fixes https://github.com/runtimeverification/evm-semantics/issues/1958, which is aimed to improve the default observable performance of KEVM by making the following set of flags/optimizations the default:

- Setting default `--max-depth` to `25000` (instead of `1000`)
- Using ` --no-break-on-calls` by default
- Using `--use-booster` by default
- Using infinite gas by default; implemented based on https://github.com/runtimeverification/evm-semantics/pull/1524. With regards to the conversation in https://github.com/runtimeverification/evm-semantics/pull/1520, this PR also adds a cheatcode `setGas(uint256)` that sets gas to the provided value; the PR also adds a test for `setGas`.
- ~Using `--no-implication-every-block` by default~ 
(requires further investigation, as using it by default breaks `79182a0149fceb6cd86e2ca420e4afb4032fdb144618f53b819e3edbfa47a1bb` proof in `nix build --extra-experimental-features 'nix-command flakes' --print-build-logs .#kevm-test`)

In this configuration, the execution of `kevm foundry-prove` on the [MiniVat test](https://github.com/baolean/symexec-bench/blob/c79cbecb6fc9564332eb8693d1ff378425d9427d/SymTest/test/MiniVat.t.sol#L15) takes ~170s instead of ~250s as before.


kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -1021,7 +1021,7 @@ def _init_cterm(
         'MEMORYUSED_CELL': intToken(0),
         'WORDSTACK_CELL': KApply('.WordStack_EVM-TYPES_WordStack'),
         'PC_CELL': intToken(0),
-        'GAS_CELL': intToken(9223372036854775807),
+        'GAS_CELL': KEVM.inf_gas(KVariable('VGAS')),
```
anvacaru, line 1024: With `inf_gas` by default, we could remove the `kevm.infiniteGas()` calls from the test suite.

include/kframework/foundry.md
```diff
@@ -884,13 +884,29 @@ function stopPrank() external;
...
+
+### `infiniteGas` - Sets the remaining gas to an infinite value.
 
 ```
 function infiniteGas() external;
 ```
 
-Set the remaining gas to an infinite value.
-This is useful for running tests without them running out of gas.
+`infiniteGas` is useful for running tests without them running out of gas. It is applied by default.
```
anvacaru, line 909: Nice changes! One tidbit, we're trying to maintain a format of having one sentence per line in the markdown files.

kevm-pyk/src/kevm_pyk/foundry.py
palinatolmach: `kevm.infiniteGas()` is removed from tests in https://github.com/runtimeverification/evm-semantics/commit/b4335416a272d175c8a811e79ff0fc7dc430338e.

include/kframework/foundry.md
palinatolmach: Thank you! Right, sorry, I've split these two sentences into separate lines in https://github.com/runtimeverification/evm-semantics/commit/b4335416a272d175c8a811e79ff0fc7dc430338e.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -421,7 +421,8 @@ def foundry_prove(
     break_on_jumpi: bool = False,
     break_on_calls: bool = True,
     implication_every_block: bool = True,
-    bmc_depth: int | None = None,
+    bmc_depth: int = 3,
+    unbounded_bmc_depth: bool = False,
```
tothtamas28, line 425: I find this a bit less intuitive than the original interface with `int | None`. (`unbounded_bmc_depth` is a self-contradiction.)

tothtamas28: > Would it be better if this other flag (e.g., `unbounded_loops`) just set the value of `bmc_depth` to `None`?

Yes, I think so. Luckily, in this particular case the, the CLI can be modified without changing the internal implementation.

palinatolmach: TY! Done in https://github.com/runtimeverification/evm-semantics/pull/1963/commits/38239deaec847e0ee10bdcce31d04297c649ae4c.
Now `unbounded_loops` is referenced once in `exec_foundry_prove`, where it's used to set `bmc_depth` to `None`—hope this looks better.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -421,7 +421,7 @@ def foundry_prove(
     break_on_jumpi: bool = False,
     break_on_calls: bool = True,
     implication_every_block: bool = True,
-    bmc_depth: int | None = None,
+    bmc_depth: int | None = 3,
```
tothtamas28, line 424: I still wouldn't change this (and the one in `_method_to_apr_proof`) to avoid breaking callers. If you do, make sure to adjust all calls to the function (in particular in `test_foundry_prove.py`).

palinatolmach: Thanks for the explanation! I've changed the default `bmc_depth` value to `None` in all functions (https://github.com/runtimeverification/evm-semantics/pull/1963/commits/69082952fda2f053a132dbe9a654d75e5d46e832). You're right, having `3` as a default  value for `--bmc-depth` in CLI is sufficient.

ehildenb (approved):

ehildenb:
I still prefer unbounded loops by default, following the principle taht @d-xo mentioned of "be as sound as possible, and let users make it simpler if wanted".

But not having that, I would also suggest having a loud warning on passed proofs if any nodes are passed because they are bounded, something like "Proof passed, but NNN nodes were loop-depth bounded at MMM iterations, re-run with `--unbounded-loops` for full verification."

.github/workflows/master-push.yml
palinatolmach, line 1: Since `kevm foundry-prove` now uses the booster backend by default, I modified the `Build Ubuntu Jammy in DockerHub` job in CI to run `kompile` with `--with-llvm-library` in `master-push.yml` and `test-pr.yml`.

palinatolmach, line 1: Since `kevm foundry-prove` now uses the booster backend by default, I modified the `Build Ubuntu Jammy in DockerHub` job in CI to run `kompile with --with-llvm-library` in `master-push.yml` and `test-pr.yml`.

palinatolmach:
@anvacaru @tothtamas28 @ehildenb Can I ask for your opinion on making `--with-llvm-backend` default in `foundry-kompile` in https://github.com/runtimeverification/evm-semantics/pull/1963/commits/db6f2d5865abc557f19c6359e13abaa892a416ed? I realized that with the booster enabled by default, we should also generate the llvm library (otherwise everyone would have to debug the `no interpreter.dylib` error), but I didn't want to enable it for `kompile` in general, beyond the Foundry integration — at least, until we migrate all proofs (including normal Haskell, not just Foundry ones) to use booster, as @ehildenb suggested.

Besides, now that `kevm foundry-prove` uses the booster backend by default, I also modified the `Build Ubuntu Jammy` in DockerHub job in CI to run `kompile with --with-llvm-library` in `master-push.yml` and `test-pr.yml`: https://github.com/runtimeverification/evm-semantics/pull/1963/commits/ffce5a8aa920cff0f13f0d85a3b75fe302cfe9ee.

palinatolmach:
Update on using `--with-llvm-library` in `kevm foundry-kompile` by default (and not having it on by default for `kevm kompile`):
- The option `--with-llvm-library` is inherited as part of [`kompile_args`](https://github.com/runtimeverification/pyk/blob/733391586dc70b19411a009cf6a8d257ed3fe956/src/pyk/cli/args.py#L58) in `pyk`, which we don't want to modify;
- Introducing `foundry_kompile.set_defaults(llvm_library=True) ` to re-set the default value `foundry_kompile` options overrides the default value for all subparsers (i.e., for `kevm kompile`, which is then also ran with `--with-llvm-library` by default)
- As a way around this, I added separate `--with-llvm-library` and `--no-llvm-library` options to both `kevm_kompile_args` and `foundry_kompile_args`; the first overrides the `--with-llvm-library` option inherited from the parent `kompile_args`: https://github.com/runtimeverification/evm-semantics/pull/1963/commits/a8b012609a1a3ed2b7b8685049630163477931b0. For that, I added `conflict_handler='resolve'` argument to the initialization of both subparsers. 
(Strangely enough, adding an overriding `--with-llvm-library` option  and `conflict_handler='resolve'` to just one subparser, `foundry_kompile_args`, messes up this option in `kevm_kompile_args`).

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -800,6 +801,19 @@ def parse(s: str) -> list[T]:
...
+        dest='llvm_library',
+        default=False,
+        action='store_true',
+        help='Do not generate a dynamic llvm library during kompile.',
+    )
+    kevm_kompile_args.add_argument(
+        '--no-llvm-library',
+        dest='llvm_library',
+        action='store_false',
+        help='Make kompile generate a dynamic llvm library.',
```
anvacaru, line 815: Hmm, why do we need both arguments?

palinatolmach: Thanks @anvacaru, I agree, we don't need `--no-llvm-library` in `kevm kompile`. I removed it in https://github.com/runtimeverification/evm-semantics/pull/1963/commits/8e5e7f1d3abbde036752057de58e29789db541ed to follow the convention of "positive and negative flags if the default is true, just a positive flag if the default is false”, as mentioned in https://github.com/runtimeverification/evm-semantics/pull/1716#discussion_r1221932255.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -805,6 +806,13 @@ def parse(s: str) -> list[T]:
     kevm_kompile_args.add_argument(
         '-o', '--output-definition', type=Path, dest='output_dir', help='Path to write kompiled definition to.'
     )
+    kevm_kompile_args.add_argument(
+        '--with-llvm-library',
+        dest='llvm_library',
+        default=False,
+        action='store_true',
+        help='Do not generate a dynamic llvm library during kompile.',
```
RaoulSchaffranek, line 814: Copy&paste mistake?

```suggestion
        help='Generate a dynamic llvm library during compile.',
```

palinatolmach: It is, thanks @RaoulSchaffranek!

palinatolmach:
Closing this PR since it introduces performance issues and doesn't pass CI. I'll split the changes suggested here into several smaller PRs.

PR #1968, opened 2023-07-26
kompile both backends in parallel

anvacaru:
runs kompile on both the Legacy backend and the Booster backend in parallel using `multiprocessing`.


anvacaru:
needs work.

PR #1975, opened 2023-07-27
correctly recursively invoke make

dwightguth:
When invoking `make plugin-deps`, we were incorrectly recursively invoking make when building the dependencies of the blockchain plugin. This meant that wee were always building at -j1 even if we pass a higher degree of parallelism to Make. Fixing this significantly speeds up parallel builds of `make plugin-deps`.


ehildenb (approved):

dwightguth:
@ehildenb what is going on with the CI here?

ehildenb:
General flakiness when they're at high volume.

PR #1979, opened 2023-07-27
Add `concrete` attribute to lemma `range-buf-zero`

lucasmt:
When trying to run a proof I got an SMT error from the backend saying that the solver wasn't able to decide a predicate when trying to apply lemma `range-buf-zero` from `bytes-simplification.k`:
```
    rule [range-buf-zero]:
      #range (#buf(W1:Int, X:Int), S2:Int, W2:Int) => #buf(W2, 0)
      requires 0 <=Int X andBool X <Int 2 ^Int (8 *Int (W1 -Int (S2 +Int W2)))
       andBool 0 <=Int S2 andBool 0 <=Int W2 andBool S2 +Int W2 <=Int W1
      [simplification]
```
Based on the output, the query included an expression of the form `2 ^Int X` with a symbolic subexpression in the exponent. I assume we don't want to send such an expression to the solver, so I added the attribute `concrete(W1, S2, W2)` to the lemma above, and that seems to fix the issue. I assume @PetarMax can confirm if this makes sense.


PetarMax:
This would make sense, yes. What was in the exponent, do you remember, @lucasmt?

Separately, should the backend be failing like that if a simplification can't be applied due to an SMT error due to query complexity? It could just not apply it.

PetarMax (approved):

lucasmt:
@PetarMax After simplification, the expression was
```
2 ^Int ( 8 *Int ( 32 -Int ( 32 *Int #lookup ( STORAGE0 , 4 ) ) ) )
```
I agree that the backend shouldn't be failing like that. Once I have some time I'll try to make a minimal reproducible example and submit an issue.

PR #1985, opened 2023-07-31
Factor out `kontrol` module from kevm-pyk codebase

ehildenb:
Fixes: https://github.com/runtimeverification/kontrol/issues/4

The idea here is to factor out the `kontrol` codebase (now responsible for `kevm foundry-*` commands and also managing `solc` compiled artifacts) into its own codebase.

In particular:

- Files `solc_to_k.py` and `foundry.py` are factored out into `kontrol.*` module.
- File `kontrol.__main__.py` is added with all the `solc` related functionality and `foundry` related functionality.
- `pyproject.toml` is modified to know about the `kontrol` binary.
- Script `bin/kevm` is modified to call into `kontrol` binary on calls to `kevm solc*` or `kevm foundry-*`.
- Runs the booster tests on a `normal` runner.


package/docker/Dockerfile
```diff
@@ -46,3 +46,5 @@ RUN    sudo apt-get update        \
 
 ADD --chown=user:user ./kevm-pyk /home/user/kevm-pyk
 RUN cd /home/user/kevm-pyk && pip3 install .
+
+ENV PATH=${PATH}:/home/user/.local/bin
```
tothtamas28, line 50: Is this something that we should have done before, or there's a new file that is put there?

tothtamas28 (approved):

package/docker/Dockerfile
ehildenb: Instead of doing `python3 -m kevm-pyk ...`, we are just doing directly `kevm-pyk ...`. So the direct call to the binary means that we need to have it on PATH.

bin/kevm
```diff
@@ -193,6 +208,7 @@ fi
 case "$run_command-$backend" in
     run-@(llvm|haskell|haskell-standalone) ) run_krun "$@" ;;
     kast-@(llvm|haskell)                   ) run_kast "$@" ;;
-    @(kompile|prune-proof|view-kcfg|show-kcfg|solc-to-k|prove|prove-legacy|foundry)-* ) run_kevm_pyk ${run_command} "$@" ;;
+    @(solc|compile|foundry)* )               run_kontrol ${run_command} "$@" ;;
```
anvacaru, line 211: Here should be `kompile` instead of `compile`, I'll open a new pr to fix this typo.

Issue #1944, opened 2023-07-06
KEVM gets stuck on `#assume(false)`


lucasmt:
KEVM seems to get stuck when it hits an `#assume(false)`. For example, the following minimal example fails because it branches inside the `atLeastOneIsValidAddress` function, with two of the branches evaluating to `#assume(true)` and the last one evaluating to `#assume(false)`. The `#assume(false)` branch then gets stuck.
```
    function atLeastOneIsValidAddress(address alice, address bob)
        internal
        returns (bool)
    {
        if (alice != address(0)) {
            return true;
        } else if (bob != address(0)) {
            return true;
        } else {
            return false;
        }
    }

    function test_assume_return(address alice, address bob) public {
        vm.assume(atLeastOneIsValidAddress(alice, bob));

        assert(true);
    }
```
In a situation like this, we would want `#assume(false)` to add `#Bottom` to the constraints, which would cause the branch where it happens to be vacuously satisfied. This would be consistent with the behavior in Foundry, which ignores inputs for which the `assume` evaluates to `false`.

`#assume` is implemented as `rule <k> #assume(B) => . ... </k> ensures B` in `foundry.md`, but KEVM gets stuck on `#assume(false)`, so is this a restriction on the `ensures` clause? Is `ensures false` not allowed? If this is the case, is there a way to implement an alternative `rule` for `#assume(false)` that will have the behavior that we need (discarding the vacuous branch)?

On the other hand, as @ehildenb pointed out, this can cause issues where a proof might vacuously pass because the user accidentally added contradictory assumptions:

> But you could also have something like `vm.assume(pred1); vm.assume(pred2);` (separated by a large chunk of code, even) and as a user you don't realize that `pred1 /\ pred2 == false`, so it vacuously discards that branch and leads to the proof passing, even though it didn't actually prove anything.

To prevent this from happening accidentally, we might want to do one of the following:

- Create a separate cheatcode, like `vm.assume_allow_vacuous`, that allows this behavior, so that the user needs to explicitly say when they want to accept vacuous branches.
- Another option is to mark branches that passed because they are vacuous and report that to the user. This way the user will know if a branch became vacuous accidentally.

anvacaru:
I created a minimal definition outside of evm-semantics and looks like `kprove` and `kore-exec` can apply the `foundry-assume(false)` rule and simplify to `#Bottom`, unlike the example above where the `kore-rpc` will remain stuck.

[...]

PR #1987, opened 2023-08-01
Refactor #assume to handle vacuous cases

anvacaru:
Fixes: #1944 

 - Added a new status code `FOUNDRY_VACUOUSSUCCESS`
 - Refactored the `#assume(B)` rules to `#halt` if the term `B` is `false` and set the `FOUNDRY_VACUOUSSUCCESS` status code.
 - Updated the `foundry_success` predicate to pass with both `EVMC_SUCCESS` and `FOUNDRY_VACUOUSSUCCESS` status codes.
 - Added logic in `kevm_prove` Python function to identify nodes that have the new status code and emit a warning with the vacuous constraints.

Example output:
image


kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -559,9 +565,17 @@ def _init_and_run_proof(_init_problem: tuple[str, str]) -> tuple[bool, list[str]
                 trace_rewrites=trace_rewrites,
                 abstract_node=(KEVM.abstract_gas_cell if auto_abstract_gas else None),
             )
+
```
anvacaru, line 568:
```suggestion
```

include/kframework/foundry.md
```diff
@@ -246,7 +266,11 @@ First we have some helpers in K which can:
 ```k
     syntax KItem ::= #assume ( Bool ) [klabel(foundry_assume), symbol]
  // ------------------------------------------------------------------
-    rule <k> #assume(B) => . ... </k> ensures B
+    rule <k> #assume(B) ~> _ => #halt </k>
+         <statusCode> _ => FOUNDRY_VACUOUSSUCCESS </statusCode>
+     requires notBool B
+
```
anvacaru, line 272:
```suggestion
    rule <k> #assume(B) ~> _ => #halt </k> <statusCode> _ => FOUNDRY_VACUOUSSUCCESS </statusCode> requires notBool B
```

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -513,6 +513,12 @@ def _init_and_run_proof(_init_problem: tuple[str, str]) -> tuple[bool, list[str]
         proof_id = f'{_init_problem[0]}.{_init_problem[1]}'
         llvm_definition_dir = foundry.out / 'kompiled-llvm' if use_booster else None
 
+        def _check_vacuous(cterm: CTerm) -> bool:
```
tothtamas28, line 516: This can be a `staticmethod` of the class.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -562,6 +568,13 @@ def _init_and_run_proof(_init_problem: tuple[str, str]) -> tuple[bool, list[str]
             failure_log = None
             if not passed:
                 failure_log = print_failure_info(proof, kcfg_explore, counterexample_info)
+
+            vacuous_nodes = {node.id for node in proof.kcfg.nodes if _check_vacuous(node.cterm)}
```
tothtamas28, line 572: Constructing a `list` is cheaper, and `proof.kcfg.nodes` are already unique:

```suggestion
            vacuous_nodes = [node.id for node in proof.kcfg.nodes if _check_vacuous(node.cterm)]
```

tothtamas28 (commented):
Python part looks good, I let @ehildenb review the K part.

kevm-pyk/src/kevm_pyk/foundry.py
tothtamas28: Right, this is not in class scope. Then just consider moving it to the top level (there's no reason for it to be nested two levels.).

kevm-pyk/src/kevm_pyk/foundry.py
anvacaru: Done.

iFrostizz (dismissed):

include/kframework/foundry.md
```diff
@@ -103,6 +103,24 @@ module FOUNDRY
...
+ - `FOUNDRY_WHITELISTSTORAGE`, which signals that a storage index of an address outside the whitelist has been changed during the execution.
+
+```k
+module FOUNDRY-STATUS-CODES
+    syntax ExceptionalStatusCode ::= "FOUNDRY_UNIMPLEMENTED"
+                                   | "FOUNDRY_WHITELISTCALL"
+                                   | "FOUNDRY_VACUOUSSUCCESS"
+                                   | "FOUNDRY_WHITELISTSTORAGE"
+ // -----------------------------------------------------------
+ endmodule
```
ehildenb, line 121:
```suggestion
endmodule
```

include/kframework/foundry.md
```diff
@@ -246,7 +266,8 @@ First we have some helpers in K which can:
 ```k
     syntax KItem ::= #assume ( Bool ) [klabel(foundry_assume), symbol]
  // ------------------------------------------------------------------
-    rule <k> #assume(B) => . ... </k> ensures B
+    rule <k> #assume(B) ~> _ => #halt </k> <statusCode> _ => FOUNDRY_VACUOUSSUCCESS </statusCode> requires notBool B
+    rule <k> #assume(B) => . ... </k> requires B
```
ehildenb, line 270: This will trigger a vacuous node being created on _every_ precondition that is inserted with `vm.assume`, won't it? So the user will be flooded with warnings about vacuous nodes, and we'll get extra branches in the KCFG.

It would be better if we detected the vacuous nodes directly using the backend, https://github.com/runtimeverification/haskell-backend/issues/3630

Or, in `pyk`, we can check every node we get back from the backend to see if it simplifies to `#Bottom`, and thus is vacuous.

lucasmt, line 270: This is going to produce a branch every time there is an `assume`, right? I think this has some undesirable consequences:

1. It will impact performance somewhat. It's probably not a huge impact because one branch will be vacuous and will halt, but it means the overhead of branching and saving to the kcfg for every `assume` in the code.
2. It will make kcfgs harder to read, because instead of a linear execution we will have multiple branching points, and each branching point indents the kcfg further to the right.
3. It will probably be confusing to users, because there will be branching that they won't expect. Also, this means that every code with `assume` will have vacuous branches, which might make it harder for the user to notice when actual vacuous branches show up.

What would happen if we instead had 
```
rule <k> #assume(false) ~> _ => #halt </k> <statusCode> _ => FOUNDRY_VACUOUSSUCCESS </statusCode>
rule <k> #assume(B) => . ... </k> ensures B
```
Would this still cause branching, or does the fact that `ensures false` is not allowed prevent it? Are there other downsides?

If we end up keeping the current version, I think it would be best to put it behind a command-line option like `--allow-vacuous-assume`, so users don't have to deal with it unless they know they need it.

lucasmt: I didn't see @ehildenb's comment before writing this, but yeah, I think we agree on this.

anvacaru: Yes, having two rules like the ones you suggested would still cause branching. So I either keep the semantic changes and add a flag to enable/disable this behaviour or I take Everett's suggestion and do the check on the pyk side.

anvacaru:
Closing this.

PR #2002, opened 2023-08-08
Add proof loading methods

nwatson22:
Adds `foundry_get_apr_proof` and `foundry_get_proof`, which load the proof object from its name, and which will be useful for integration testing since currently we don't have a good way to check proof status, nodes, etc. in e.g., `test_foundry_prove.py`, except for the pass/fail result from `foundry_prove`. It also lets us factor out proof loading logic from the other `foundry.py` functions.


kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -768,6 +751,32 @@ def foundry_step_node(
...
+    bug_report: bool = False,
+) -> Proof:
+    br = BugReport(Path(f'{test}.bug_report')) if bug_report else None
+    foundry = Foundry(foundry_root, bug_report=br)
+
+    proofs_dir = foundry.out / 'apr_proofs'
+    contract_name, test_name = test.split('.')
+    proof_digest = foundry.proof_digest(contract_name, test_name)
+    proof = Proof.read_proof_data(proofs_dir, proof_digest)
+    return proof
```
ehildenb, line 777: All the call sites of this have a `Foundry` in hand, so maybe instead of passing in `foundry_root`, you just pass the `Foundry` object in directly?

Construction of a `Foundry` object is expensive, requires reading in all teh Solidity sources and processing them.

ehildenb:
Let's double-check that all the places where `proof_digest` is used _could_ be using this unctionality.

kevm-pyk/src/kevm_pyk/foundry.py
nwatson22: Fixed.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -768,6 +752,29 @@ def foundry_step_node(
...
+    test: str,
+    bug_report: bool = False,
+) -> Proof:
+    proofs_dir = foundry.out / 'apr_proofs'
+    contract_name, test_name = test.split('.')
+    proof_digest = foundry.proof_digest(contract_name, test_name)
+    proof = Proof.read_proof_data(proofs_dir, proof_digest)
+    return proof
+
+
```
ehildenb, line 777: Because these are both methods that take `Foundry` as the first argument, they might as well become methods in the class.

```
class Foundry:

    def get_apr_proof(test: str) -> APRProof:
        ...

    def get_proof(test: str) -> Proof
        ...
```

Also, as far as I can tell, neither needs tho `BugReport` option anymore.

ehildenb (approved):

PR #2005, opened 2023-08-09
Drop `pathos`

tothtamas28:
Replace `ProcesPool` by `ThreadPoolExecutor` in provers.


ehildenb:
Hmmmm, why are we doing this actually? My understanding is that starting up threadpool will be less work overall. But also, it seems that onyl one thread will be allowed to execute at a time because of teh Python interpreter GIL:

https://superfastpython.com/threadpoolexecutor-vs-processpoolexecutor/

image

So if my undertanding is correct, then only one Python thread will be allowed to execute at a time. This would be fine if the amount of time spent in executing python code was neglible compared ot backend time, but it's not the case.

tothtamas28:
> it seems that onyl one thread will be allowed to execute at a time because of teh Python interpreter GIL

You're right, I wasn't aware of this detail.

> why are we doing this actually

I wanted to avoid pickling between processes because in some cases it causes problems, e.g. exceptions not propagating to the main thread, tests not being able to capture logs, etc.

> This would be fine if the amount of time spent in executing python code was neglible compared ot backend time, but it's not the case.

I'll replace executors to a `ProcessPoolExecutor`. If it works as expected, then we can still drop the third party dependency.

tothtamas28:
> I'll replace executors to a `ProcessPoolExecutor`. If it works as expected, then we can still drop the third party dependency.

Nope.

```
concurrent.futures.process._RemoteTraceback:
"""
Traceback (most recent call last):
  File "/home/ttoth/.pyenv/versions/3.10.12/lib/python3.10/multiprocessing/queues.py", line 244, in _feed
    obj = _ForkingPickler.dumps(obj)
  File "/home/ttoth/.pyenv/versions/3.10.12/lib/python3.10/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
AttributeError: Can't pickle local object 'foundry_prove.<locals>.init_and_run_proof'
"""
```

PR #2008, opened 2023-08-10
Foundry/forge+solc nix fixes

### LINKED ISSUE #2006: nix/kup version of kevm foundry-* commands broken due to foundry/solc version issues
Some issues observed when trying to run foundry proofs using nix/kup version of kevm:

* Failed to reproduce this CI run locally: https://github.com/runtimeverification/evm-semantics/blob/62996ec292db48ea137802f559d813e979101cd3/.github/workflows/test-pr.yml#L185-L192
  I get failures related to forge (being called directly in kevm now), not finding solc. This is surprising because we bundle solc 0.8.19 with the nix version of kevm but @ehildenb points out we have a tight bound of 0.8.13
* Same failure occurs if i try to run kup version of kevm and manually `kevm foundry-kompile/foundry-prove`. THe kup installed version should come with the right versions of forge+solc.
* We have two nix derivations of solc at `nix/solc/default.nix` and `package/nix/solc.nix`. The first one is the incorrect version,t he second does not work

goodlyrottenapple:
This fixes #2006 by:
* removing stale solc derivation in `package/nix/solc.nix`, removing incorrect verion of solc in `nix/solc/default.nix` and adding a flake dependency on solc which allows us to specify the exact version, namely `0.8.13`
* bundling both `solc` and `forge` executable paths with the `kevm` bin when built with nix. This ensures a consistent version of both and means the user doesnt have to install them separately (make it much easier to run the foundry tests with e.g. a custom branch of the booster backend on a clean vm)


flake.nix
```diff
@@ -148,6 +152,12 @@
         kevm-pyk = prev.poetry2nix.mkPoetryApplication {
           python = prev.python310;
           projectDir = ./kevm-pyk;
+
+          postPatch = ''
+              substituteInPlace ./src/kevm_pyk/foundry.py \
+                --replace "'forge', 'build'," "'forge', 'build', '--no-auto-detect',"
+            '';
```
goodlyrottenapple, line 159: `forge` does not play nice with nix by default and throws an error for me, when trying to detect `solc`. Since we supply a pinned version via nix, we add the `--no-auto-detect` flag to make forge just use that version of `solc` (see https://github.com/foundry-rs/foundry/issues/545)

flake.nix
```diff
@@ -21,10 +21,14 @@
     pyk.inputs.flake-utils.follows = "k-framework/flake-utils";
     pyk.inputs.nixpkgs.follows = "k-framework/nixpkgs";
     foundry.url = "github:shazow/foundry.nix/monthly"; # Use monthly branch for permanent releases
+    solc = {
+      url = "github:goodlyrottenapple/solc.nix";
```
goodlyrottenapple, line 25: I've made a clone of this repo until this/something similar is merged: https://github.com/hellwolf/solc.nix/pull/1

h0nzZik (approved):
Looks good to me.

ehildenb (approved):

Issue #1939, opened 2023-07-04
Parallelize `kompile` runs when `--with-llvm-library` is used


palinatolmach:
Currently, the execution of `kevm foundry-kompile --with-llvm-library ...` performs two sequential invocations of the `kompile` command: one writing the output to `out/kompiled`, and the other—to `out/kompiled-llvm`.

The performance of `foundry-kompile` can potentially be improved by parallelizing these `kompile` executions.

PR #2010, opened 2023-08-11
Parallelization of kompile processes in foundry_kompile

anvacaru:
closes: https://github.com/runtimeverification/evm-semantics/issues/1939

 - Use `concurrent.futures` to run the kompilation processes for the haskell backend and the llvm shared library in parallel.


kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -410,20 +411,29 @@ def kompilation_up_to_date() -> bool:
...
-                foundry_llvm_dir,
-                KompileTarget.LLVM,
-                llvm_kompile_type=LLVMKompileType.C,
-            )
+        with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
+            executor.submit(kompile_haskell)
+            if llvm_library:
+                future2 = executor.submit(kompile_llvm)
+                if future2.exception():
+                    _LOGGER.error(f'Error in kompile_llvm: {future2.exception()}')
```
tothtamas28, line 436:
```suggestion
            futures: list[Future] = []
            futures.append(executor.submit(kompile_haskell))
            if llvm_library:
                futures.append(executor.submit(kompile_llvm))
            [future.result() for future in futures]
```

tothtamas28: Here, `future.result()` takes care of raising the exception on the thread.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -410,20 +411,29 @@ def kompilation_up_to_date() -> bool:
 
         return old_digest == kompilation_digest()
 
+    def kompile_haskell() -> None:
+        _LOGGER.info(f'Kompiling definition: {foundry_main_file}')
+        _kompile(foundry_definition_dir, KompileTarget.HASKELL)
+
+    def kompile_llvm() -> None:
+        _LOGGER.info(f'Kompiling definition to LLVM dy.lib: {foundry_main_file}')
```
tothtamas28, line 419: Is `dy.lib` the file name on all supported platforms?

tothtamas28 (commented):
How much faster is it this way? I'd expect the resulting executing time to be approximately the same as the slower of the two executions.

kevm-pyk/src/kevm_pyk/foundry.py
anvacaru: Done!

kevm-pyk/src/kevm_pyk/foundry.py
anvacaru: @tothtamas28  According to the haskell backend team, it is `interpreter.so` on linux and `interpreter.dylib`.

tothtamas28: I see, consider this then:

```
_LOGGER.info(f'Kompiling definition to LLVM dynamic library: {foundry_main_file}')
```

anvacaru: Updated the logging message.

kevm-pyk/src/kevm_pyk/foundry.py
```diff
@@ -410,20 +412,29 @@ def kompilation_up_to_date() -> bool:
...
-        _LOGGER.info(f'Kompiling definition: {foundry_main_file}')
-        _kompile(foundry_definition_dir, KompileTarget.HASKELL)
-        if llvm_library:
-            _LOGGER.info(f'Kompiling definition to LLVM dy.lib: {foundry_main_file}')
-            _kompile(
-                foundry_llvm_dir,
-                KompileTarget.LLVM,
-                llvm_kompile_type=LLVMKompileType.C,
-            )
+        with concurrent.futures.ThreadPoolExecutor(max_workers=2) as executor:
```
tothtamas28, line 432: Posting it here as well so that it's easier to unfold the conversation:

> How much faster is it this way? I'd expect the resulting execution time to be approximately the same as the slower of the two executions.

anvacaru: @tothtamas28 
 - with `--with-llvm-library` flag in parallel:  6:21.98
 - with `--with-llvm-library` flag on master: 8:40.81
 - without `--with-llvm-library` flag on master: 6:16.03

tothtamas28: Looks good, thanks!

tothtamas28 (approved):

PR #2015, opened 2023-08-11
make foundry_kompile be called only once for any given pytest invocation

dwightguth:
calling kompile 8 times in parallel is very expensive. As far as I can tell, all of these kompile invocations are identical, so we should call kompile once and then just copy the results into the temp directory of each test worker.


kevm-pyk/src/tests/conftest.py
```diff
@@ -28,6 +28,6 @@ def update_expected_output(request: FixtureRequest) -> bool:
     return request.config.getoption('--update-expected-output')
 
 
-@pytest.fixture(scope='module')
+@pytest.fixture(scope='session')
```
anvacaru, line 31: I see another fixture with the `module` scope [here](https://github.com/runtimeverification/evm-semantics/blob/master/kevm-pyk/src/tests/integration/test_prove.py#L165). Is this one ommited on purpose?

dwightguth: It might be able to be changed, but it's outside the scope of these changes. The reason this one was changed was because it was a dependency of the other one.

Issue #1703, opened 2023-04-06
`foundry-kompile` will throw an error for functions named with K reserved names


anvacaru:
Functions named after K reserved names, such as `claim`, will cause `foundry-kompile` to throw an error.

Contract example 
```solidity
contract TestNumber is DSTest{
    uint256 public testNumber ;

    constructor(uint256 initial){
        testNumber = initial;
    }

    function t(uint256 a) public returns (uint256) {
        uint256 b = 0;
        testNumber = a;
        emit log_string("here");
        return b;
    }

    function claim(uint256 a) public {}

}
```

error 
```bash
[Error] Inner Parser: Parse error: unexpected end of file following token '.'.
        Source(/home/anvacaru/rv/evm-semantics/tests/foundry/out/kompiled/foundry.k)
        Location(5104,25,5104,25)
        5104 |      rule  ( TestNumber . claim ( V0_a ) => #abiCallData ( "claim" ,
#uint256 ( V0_a ) , .TypedArgs ) )
             .                          ^
[Error] Inner Parser: Parse error: unexpected token ')' following token ')'.
        Source(/home/anvacaru/rv/evm-semantics/tests/foundry/out/kompiled/foundry.k)
        Location(5104,102,5104,103)
        5104 |      rule  ( TestNumber . claim ( V0_a ) => #abiCallData ( "claim" ,
#uint256 ( V0_a ) , .TypedArgs ) )
```

radumereuta:
Palina ran into a similar issue with a function called `Test_Foo`.
The K foundry translator needs to escape function names to avoid clashes with K keywords and make sure that any translation to sort names doesn't contain forbidden characters.

palinatolmach:
I keep running into `kompile` issues caused by the underscore symbol in the contract name when trying to run KEVM on existing Foundry test suites (e.g., [Sablier's](https://github.com/sablier-labs/v2-core/tree/main/test)), so bumping the priority to `high`.

iFrostizz:
We could bulk with: https://github.com/runtimeverification/evm-semantics/issues/1368

radumereuta:
I sent this message on slack, posting it here again so we don't lose it.

There is no way to add these characters to the sort names. The K outer parser doesn't allow them.
What I was suggesting is to have a function that encodes special characters into normal alphanums. Example
`My_contract` -> `KEVMMyz5Fcontract`
[...]

Issue #1368, opened 2022-09-09
Single-letter contract names fail frequently


david-k:
Assume we have a Foundry test in file `test/MyTest.t.sol`:
```solidity
// SPDX-License-Identifier: UNLICENSED
pragma solidity ^0.8.13;

import "forge-std/Test.sol";

contract F { }

contract MyTest is Test { }
```

If we run
```sh
forge build
kevm foundry-kompile out/ \
    --require lemmas/int-simplification.k --module-import INT-SIMPLIFICATION \
    --require lemmas/lemmas.k --module-import LEMMAS

```
the following error is generated:
```
[Error] Inner Parser: Parse error: unexpected token 'F' following token '('.
        Source(/home/david/RV/evm-semantics/.build/usr/lib/kevm/include/kframework/lemmas/int-simplification.k)
        Location(115,48,115,49)
        115 |       rule (E *Int A) +Int B +Int C +Int D +Int (F *Int A) => ((E +Int F)
*Int A) +Int B +Int C +Int D [simplification]
            .                                                  ^
[Error] Compiler: Had 1 parsing errors.

...
```
However, if we move the definition of contract `F` into its own file under `src/` then the error disappears.

ehildenb:
This is actually because of the contract named `F`, not because of he missing `Test` inheritance. `foundry-kompile` generates a K definition, with `syntax Contract ::= "F"`, and then when parsing the listed rule, the K parser is confused because it is not sure if `F` is a variable or the declared constant.

I think in this case, K's parser should be able to figure this one out. https://github.com/runtimeverification/k/issues/2867

PR #2016, opened 2023-08-15
Issues with `foundry-kompile` colliding with K internals

iFrostizz:
Closes https://github.com/runtimeverification/evm-semantics/issues/1368, closes https://github.com/runtimeverification/evm-semantics/issues/1703

Now supports:

- [ ] Single name contracts by escaping by adding the ~`KEVM`~ `S2K` prefix
- [ ] Contracts name with an underscore by escaping (adds the prefix + replace the underscore by its ASCII representation, can also unescape, see tests)
- [ ] Method names such as `claim()` won't clash anymore with a K claim, we ~add `method-` in front of it~ **escape everything**

For visualization, consider a project structure such as:

```sh
.
├── cache
│   └── solidity-files-cache.json
├── foundry.toml
├── out
│   ├── Counter.sol
│   │   ├── Counter.json
│   │   └── COuntER.json
│   ├── Counter.t.sol
│   │   ├── Counter_Test.json
│   │   ├── CounterTest.json
│   │   └── Counterz_Test.json
│   ├── F.sol
│   │   └── F.json
├── README.md
├── src
│   └── Counter.sol
└── test
    ├── Counter.t.sol
    └── F.sol
```

Before this fix, the `contracts.k` would look like:

[contracts.k.txt](https://github.com/runtimeverification/evm-semantics/files/12342668/contracts.k.txt)

Note that kompilation was failing because of the two `Counter` and `COuntER` with different casing, and now fails due to the `_` in 
```k
Location(40,32,40,33)
40 |        syntax Contract ::= Counter_TestContract
   .                                   ^
```
   
Now it's:

[contracts2.k.txt](https://github.com/runtimeverification/evm-semantics/files/12342679/contracts.k.txt)

 and does not fail.


radumereuta:
`$` -> `ZDlr`
`_` -> `ZUnd`
`Z` -> `ZZ`
Use capital Z because it's less likely to be found in a contract.
`contract_zero` -> `KEVMcontractZUndzero`
`Contract_Zero` -> `KEVMContractZUndZZero`
`KEVM` at the end would make it more readable but runs into issues with lowercase contract names.
We can do the same for module names.

kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -347,3 +347,53 @@ def legacy_explore(
                 id=id,
                 trace_rewrites=trace_rewrites,
             )
+
+
+def escaped_chars() -> list[str]:
```
radumereuta, line 352: `__` make the function private.

radumereuta:
I was thinking instead of the `KEVM` prefix, maybe `S2K` would be more appropriate because the translation is done in the `solidity-to-k` tool.

iFrostizz:
Yes, thank you @radumereuta this makes more sense.

kevm-pyk/src/kevm_pyk/utils.py
iFrostizz: I've done it a `@classmethod`, is it good enough ?

iFrostizz:
Waiting for https://github.com/runtimeverification/evm-semantics/pull/1883

kevm-pyk/src/kevm_pyk/utils.py
radumereuta: I'm not that familiar with Python. This is just a suggestion and is not important. Although documentation would be really nice.

iFrostizz: Will do, thanks!

tests/foundry/lemmas.k
```diff
@@ -60,8 +60,8 @@ module SUM-TO-N-INVARIANT
    andBool #rangeUInt(256, S)
    andBool GAS_AMT >=Int N *Int 178
    andBool CONDITION ==K bool2Word ( N:Int ==Int 0 )
-   andBool PROGRAM ==K #binRuntime(LoopsTest)
-   andBool JUMPDESTS ==K #computeValidJumpDests(#binRuntime(LoopsTest))
+   andBool PROGRAM ==K #binRuntime(KEVMLoopsTest)
+   andBool JUMPDESTS ==K #computeValidJumpDests(#binRuntime(KEVMLoopsTest))
```
radumereuta, line 64: KEVM here?

kevm-pyk/src/tests/unit/test_solc_to_k.py
```diff
@@ -0,0 +1,28 @@
...
+
+from kontrol.solc_to_k import Contract
+
+TEST_DATA: list[tuple[str, str, str, str]] = [
+    ('has_underscore', 'KEVM', 'My_contract', 'KEVMMyz5fcontract'),
+    ('no_change', '', 'mycontract', 'mycontract'),
+    ('starts_underscore', 'KEVM', '_method', 'KEVMz5fmethod'),
+    ('with_escape', '', 'z_', 'z7az5f'),
+    ('with_escape', 'KEVM', 'zKEVM_', 'KEVMz7aKEVMz5f'),
+]
```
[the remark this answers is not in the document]
iFrostizz: Not necessarily, the prefix is defined as the second argument. It's just testing for some edge cases, sometimes without a prefix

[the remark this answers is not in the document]
iFrostizz: But for clarity I'll update to `S2K` , thank you!

tests/foundry/lemmas.k
iFrostizz: Good one, thanks!

kevm-pyk/src/kontrol/solc_to_k.py
```diff
@@ -342,17 +352,69 @@ def test_to_claim_name(t: str) -> str:
     def name_upper(self) -> str:
         return self.name[0:1].upper() + self.name[1:]
 
+    @staticmethod
+    def escaped_chars() -> list[str]:
+        return [Contract.prefix_code, '_']
+
+    @staticmethod
+    def escape_char(char: str) -> str:
+        return f'z{hex(ord(char)).removeprefix("0x")}'
```
radumereuta, line 361: We also agreed to make the encoding more readable by using this encoding:
```
$ -> ZDlr
_ -> ZUnd
Z -> ZZ
```

radumereuta (commented):
With those minor comments, this looks fine.
But it should be reviewed by someone more familiar with the evm-semantics as well.

kevm-pyk/src/kontrol/solc_to_k.py
```diff
@@ -342,17 +352,69 @@ def test_to_claim_name(t: str) -> str:
     def name_upper(self) -> str:
         return self.name[0:1].upper() + self.name[1:]
 
+    @staticmethod
+    def escaped_chars() -> list[str]:
+        return [Contract.prefix_code, '_']
```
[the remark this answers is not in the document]
iFrostizz: Thank you very much! I'll update and write tests

iFrostizz:
Currently updating the expected output.

kevm-pyk/src/tests/integration/test-data/show/AssumeTest.test_assume_false(uint256,uint256).expected
```diff
@@ -2263,12 +2263,7 @@ Failing nodes:
     #And { true #Equals ( notBool VV0_a_114b9705:Int ==Int VV1_b_114b9705:Int ) ==Bool true } ) ) ) ) ) ) ) ) ) ) ) #Implies { true #Equals foundry_success ( ... statusCode: EVMC_SUCCESS , failed: #lookup ( ( 46308022326495007027972728677917914892729792999299745830475596687180801507328 |-> 1 ) , 46308022326495007027972728677917914892729792999299745830475596687180801507328 ) , revertExpected: false , opcodeExpected: false , recordEventExpected: false , eventExpected: false ) } )
   Path condition:
     #Top
-  Model:
-    CALLER_ID = 0
-    NUMBER_CELL = 0
-    ORIGIN_ID = 0
-    VV0_a_114b9705 = 1
-    VV1_b_114b9705 = 0
+  Failed to generate a model.
```
anvacaru, line 2266: Is this related to these changes?

kevm-pyk/src/kontrol/solc_to_k.py
```diff
@@ -228,6 +237,7 @@ def application(self) -> KInner:
     raw_sourcemap: str | None
     methods: tuple[Method, ...]
     fields: FrozenDict
+    prefix_code: str = 'Z'
```
anvacaru, line 240: If this is supposed to be a constant, I guess it should be marked accordingly. `PREFIX_CODE: Final = 'Z'`.

kevm-pyk/src/kontrol/solc_to_k.py
```diff
@@ -342,17 +352,86 @@ def test_to_claim_name(t: str) -> str:
...
+        """
+        escape all the chars that would cause issues once kompiling and add a prefix to mark it as "escaped"
+        """
+        escaped = prefix
+        for char in iter(name):
+            if char in Contract.escaped_chars():
+                escaped += Contract.escape_char(char)
+            else:
+                escaped += char
+        return escaped
```
anvacaru, line 394: Maybe something like this to have less string concatenations?
```suggestion
        escaped = [Contract.escape_char(char) if char in Contract.escaped_chars() else char for char in name]
        return prefix + ''.join(escaped)
```

kevm-pyk/src/kontrol/solc_to_k.py
```diff
@@ -342,17 +352,86 @@ def test_to_claim_name(t: str) -> str:
...
+
+    @staticmethod
+    def unescaped(name: str, prefix: str = '') -> str:
+        if not name.startswith(prefix):
+            raise RuntimeError(f'name {name} should start with {prefix}')
+        unescaped = name.removeprefix(prefix)
+        res = ''
+        unes_iter = iter(unescaped[:-1])
+        skipped = 0
+        for i, char in enumerate(unes_iter):
```
anvacaru, line 404: Rather than using an iterator with explicit counting, I would iterate the string indices.

kevm-pyk/src/kontrol/solc_to_k.py
```diff
@@ -342,17 +352,86 @@ def test_to_claim_name(t: str) -> str:
...
+            else:
+                escaped += char
+        return escaped
+
+    @staticmethod
+    def unescaped(name: str, prefix: str = '') -> str:
+        if not name.startswith(prefix):
+            raise RuntimeError(f'name {name} should start with {prefix}')
+        unescaped = name.removeprefix(prefix)
+        res = ''
```
anvacaru, line 401: Instead of a resulting string, perhaps you should use a list to build the result and then `join` it at the end.

kevm-pyk/src/tests/integration/test-data/show/AssumeTest.test_assume_false(uint256,uint256).expected
iFrostizz: Most likely the local machine flakyness, thanks I missed that.

kevm-pyk/src/kontrol/solc_to_k.py
iFrostizz: Right, thanks!

kevm-pyk/src/kontrol/solc_to_k.py
iFrostizz: Looks good! Thanks!

kevm-pyk/src/kontrol/solc_to_k.py
iFrostizz: I'm not sure to understand what you mean by that @anvacaru

kevm-pyk/src/kontrol/solc_to_k.py
iFrostizz: It looks more expensive, because instead of concatenating the string on the fly, we would have to push to the array for every iteration, and then loop over every element to build this final concatenated string.

anvacaru: Well, strings in Python are immutable. This means that every time you concatenate two strings with `+`, a new string object is created, and the content of the two strings is copied over to the new string. This can be inefficient in a loop as you end up creating multiple intermediate objects. 

But I guess it depends on the size of the `unescaped` string. The difference may not be noticeable.

kevm-pyk/src/kontrol/solc_to_k.py
anvacaru: Using something like  ` while i < len(unescaped):`

iFrostizz: Right, looks much better, thank you!

kevm-pyk/src/kontrol/solc_to_k.py
iFrostizz: True, let's apply your suggestion. Thank you

anvacaru (approved):

PR #2017, opened 2023-08-15
use one server per thread in test_foundry_prove.py

dwightguth:
This PR modifies the test_foundry_prove test suite in order to ensure that each individual test is not starting and stopping multiple KoreServer objects. Instead, a single server is started per thread when the python module for the tests is initialized, and then that server is used for all requests in that thread across all the proofs. This ought to significantly reduce the time we are spending starting and then stopping servers, since we were previously starting and then stopping the server between 1-2 times for each proof.


nwatson22 (commented):
Possibly the `port` option could be added to all the other command functions that call `legacy_explore` as well, and used in the tests where they appear, particularly `foundry_step_node`, since this is also used in one of the `test_foundry_prove` tests.

nwatson22 (approved):

dwightguth:
The only test where it's not being used is the merge_nodes test because I wasn't entirely clear on which function calls I needed to pass it to. If you want to help me figure that out I can definitely do it though.

nwatson22 (approved):

PR #2023, opened 2023-08-17
Remove k submodule, rely on release tags for updates

ehildenb:
This PR removes the K submodule dependency from the chain, instead relying directly on teh release tags for K. In particular:

- The submodule at `deps/k` is removed.
- The `Makefile` targets for building that submodule are removed.
- The `flake.nix` handling of that submodule is removed.
- The `update-deps.yml` workflow is updated to not change/check taht submodule.
- The `README` is updated with new instructions for installing K or starting K shell with the correct version of K based on the release tag.


F-WRunTime (approved):

F-WRunTime:
Would be good to make a note to devs once this is gone to remove the deps/k folder as it gets cut off and untracked.. To avoid accidentally trying to check in files to that folder with a 'git add --all'

PR #2024, opened 2023-08-17
Fixing incorrect passing proof reporting

PetarMax:
A proof that had no failing nodes but was not passing (e.g., it had pending nodes) was incorrectly reported as passing.

This PR corrects this by understanding this using `proof.status`, not `proof.failing`.


kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -125,8 +125,7 @@ def kevm_prove(
                 cut_point_rules=cut_point_rules,
             )
             assert isinstance(proof, APRProof)
-            failure_nodes = proof.failing
-            if len(failure_nodes) == 0:
+            if proof.status == ProofStatus.PASSED:
```
ehildenb, line 128:
```suggestion
            if proof.passed:
```

We can use this helper, defined here: https://github.com/runtimeverification/pyk/blob/949163ea8cc7f168490c84a7daa5df5481d1af9c/src/pyk/proof/proof.py#L182

ehildenb: It looks like there are other places in this file where `ProofStatus.PASSED` or `ProofStatus.FAILED` is compared to directly instead of using `Proof.passed` and `Proof.failed`, can you adjust those as well?

PetarMax: @ehildenb Done, please check if you can - I've also removed some dead code.

ehildenb:
I believe we can get rid of the final mention of `ProofStatus.PENDING` in `utils.py` by using an `if ... elif ... else` instead of an `if ... if ... if ...`, which means we can remove an import from the top! (faster loading times)

PetarMax:
There seems to be a problem with the [AssumeTest file](https://github.com/runtimeverification/evm-semantics/blob/reporting-correction/tests/foundry/test/AssumeTest.t.sol).

In the CI, the test `test_assume_false` is now failing, I believe correctly, because it has one pending node and the proof is, therefore, not complete.

Separately, to me it appears that the bodies of `test_assume_false` and `testFail_assume_false` are syntactically the same, as well as those of `test_assume_false` and `testFail_assume_true`. As far as I understand how this testing works, if the test name starts writes `test_`, the test is supposed to pass, and if it starts with `testFail_`, the test is supposed to fail. I am not sure how the same code can pass and fail at the same time, what am I missing?

@anvacaru @ehildenb

anvacaru:
> There seems to be a problem with the [AssumeTest file](https://github.com/runtimeverification/evm-semantics/blob/reporting-correction/tests/foundry/test/AssumeTest.t.sol).
> 
> In the CI, the test `test_assume_false` is now failing, I believe correctly, because it has one pending node and the proof is, therefore, not complete.
> 
> Separately, to me it appears that the bodies of `test_assume_false` and `testFail_assume_false` are syntactically the same, as well as those of `test_assume_false` and `testFail_assume_true`. As far as I understand how this testing works, if the test name starts writes `test_`, the test is supposed to pass, and if it starts with `testFail_`, the test is supposed to fail. I am not sure how the same code can pass and fail at the same time, what am I missing?
> 
> @anvacaru @ehildenb

@PetarMax Yes, the naming is not ideal. The prefix `test_` or `testFail_` indicates the expected outcome of the test (is the test expected to pass or fail?), and the suffixes `_true` and `_false` are indicators of whether the execution of the test succeeds or fails, and not related to the conditions being checked.

Breaking down the naming of the tests:
1. `test_assume_true`: This expects the execution to pass, and the test does pass (execution is successful) because it assumes `a == b` and then asserts `a == b`.

2. `test_assume_false`: This expects the execution to pass, but it fails because it assumes `a != b` and then asserts `a == b`.

3. `testFail_assume_true`: This expects the execution to fail, and the execution is unsuccessful ( in the end the test passes, hence the `_true`) because it assumes `a != b` and then asserts `a == b`.

4. `testFail_assume_false`: This expects the execution to fail, but instead it passes because it assumes `a == b` and then asserts `a == b`, and so the test fails, hence the `_false`.

PetarMax:
Yes, and what do you think is the way forward to make the tests pass in the
CI? Apologies if they already are, I'm on my phone.

ehildenb:
@PetarMax this is pretty consistently failing the foundry tests, are we sure it actually works? Did you run the tests locally and check that they pass with these changes on yoru machine?

PetarMax:
> @PetarMax this is pretty consistently failing the foundry tests, are we sure it actually works? Did you run the tests locally and check that they pass with these changes on yoru machine?

No, the PR has revealed problems with some of the tests. I've flagged this to @anvacaru, but I'm not sure what the status is at the moment.  The relevant comment is [here](https://github.com/runtimeverification/evm-semantics/pull/2024#issuecomment-1685237342).

PR #2026, opened 2023-08-21
Foundry digest rehaul

iFrostizz:
Closes https://github.com/runtimeverification/kontrol/issues/16

Adding a human-readable incrementing number for the proof ID and moving the proof digest in a `proof_digest` field in the `proof.json`.
Changes the way kontrol deals with digests that was causing annoyances if the current digest went out of date, for instance before we couldn't `foundry-show` a proof whose digest changed.
When we `foundry-prove`, Either prove the proof ID has been provided and the proof is up to date, then we can just resume the proof, or multiple up-to-date proofs are here and in this case we ask the user to provide an ID.
It should be retro-compatible with being able to use the "old" IDs.

~TODO: change branch of `pyk` to `master` once https://github.com/runtimeverification/pyk/pull/613 is merged.~

edit: we don't make any change on `pyk`, see https://github.com/runtimeverification/pyk/pull/613 for context.


kevm-pyk/src/kontrol/__main__.py
```diff
@@ -516,6 +542,7 @@ def parse(s: str) -> list[T]:
         ],
     )
     foundry_show_args.add_argument('test', type=str, help='Display the CFG for this test.')
+    foundry_show_args.add_argument('--id', type=str, default=None, required=False, help='ID of the test')
     foundry_show_args.add_argument(
         '--omit-unstable-output',
         dest='omit_unstable_output',
```
ehildenb, line 548: Let's factor out a `foundry_test_args` in `cli.py`, and add the arguments `'test'` and `--id` to it, then use that everywhere here that both arguments are used.

kevm-pyk/src/kontrol/foundry.py
```diff
@@ -84,6 +85,14 @@ def profile(self) -> dict[str, Any]:
     def out(self) -> Path:
         return self._root / self.profile.get('out', '')
 
+    @property
+    def proofs_dir(self) -> Path:
+        return self.out / 'apr_proofs'
```
ehildenb, line 90:
```suggestion
    @property
    def proofs_dir(self) -> Path:
        return self.out / 'proofs'
```

Let's just use the word `proofs`, because we'll be eventually storing other types of proofs too.

kevm-pyk/src/kontrol/__main__.py
```diff
@@ -468,15 +494,15 @@ def parse(s: str) -> list[T]:
     )
     foundry_prove_args.add_argument(
         '--test',
-        type=str,
+        type=_test_type,
         dest='tests',
```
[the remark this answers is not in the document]
iFrostizz: in the context of `foundry-prove`, we need to pass the tests with multiple `--test` flags, this will be in the form of `--test test,id` and can also be passed as `--test test` for having `id` to `None`. We need it for `foundry-prove` because `id` has to be grouped with `test`

kevm-pyk/src/kontrol/__main__.py
iFrostizz: Looks much better, thanks!

kevm-pyk/src/kontrol/foundry.py
iFrostizz: Understood, thank you!

iFrostizz:
Surprisingly, the `test_foundry_resume_proof` finds that two proofs are matching with the `AssumeTest.test_assume_false(uint256,uint256)` test.
The ID for now is hardcoded and should be fine if test names are not colliding between themselves but please see the note if this has to ever be an issue in the future for better scalability, to return the `test_id` (containing the id) from `foundry_prove`: https://github.com/runtimeverification/evm-semantics/pull/2026/commits/a52327fcdeecd715b2ef3584b0d8cb87cb6b09a5

iFrostizz:
> Surprisingly, the `test_foundry_resume_proof` finds that two proofs are matching with the `AssumeTest.test_assume_false(uint256,uint256)` test on CI (cannot reproduce locally). Maybe some proof leftovers it might be the `foundry_root` fixture copying old proof but it seems to start again from scratch ? The ID for now is hardcoded and should be fine if test names are not colliding between themselves but please see the note if this has to ever be an issue in the future for better scalability, to return the `test_id` (containing the id) from `foundry_prove`: [a52327f](https://github.com/runtimeverification/evm-semantics/commit/a52327fcdeecd715b2ef3584b0d8cb87cb6b09a5)

This was implemented instead

kevm-pyk/src/kontrol/__main__.py
```diff
@@ -426,6 +439,19 @@ def parse(s: str) -> list[T]:
     solc_to_k_args.add_argument('contract_file', type=file_path, help='Path to contract file.')
     solc_to_k_args.add_argument('contract_name', type=str, help='Name of contract to generate K helpers for.')
 
+    def _test_type(value: str) -> tuple[str, str | None]:
+        # Make sure that it matches the regex if the input is incomplete
+        if value.find(',') == -1:
```
anvacaru, line 444: Could you insert this find in the regex directly?

kevm-pyk/src/kontrol/__main__.py
```diff
@@ -426,6 +439,19 @@ def parse(s: str) -> list[T]:
     solc_to_k_args.add_argument('contract_file', type=file_path, help='Path to contract file.')
     solc_to_k_args.add_argument('contract_name', type=str, help='Name of contract to generate K helpers for.')
 
+    def _test_type(value: str) -> tuple[str, str | None]:
```
anvacaru, line 442: Perhaps a more suggestive name? Like `_parse_test_id_tuple` or `_parse_test_type`.

anvacaru:
Can you double check that you updated every function that uses `test: str` (`foundry-show`, `foundry-view-kcfg`, `foundry-prove`, `prune-proof`, `remove-node` and so on)?
Does this change also affect the tests not related to foundry (`prove`,`view-kcfg`, ...)?

kevm-pyk/src/kontrol/__main__.py
iFrostizz: Makes sense! Thanks

kevm-pyk/src/kontrol/__main__.py
iFrostizz: Right, that's a good idea, thank you!

iFrostizz:
Thanks for your review @anvacaru , I have tested all foundry-related functions, and they kinda all comply to the same interface for making sure that the test won't match another one for `foundry-*` commands that require a specific test to be ran but haven't checked with legacy commands since they don't make the use of this automated proof system.
Though, I wanted to attract your attention on one point which is that the `--test` flag semantics changed a bit. It used to mimic Foundry's `--match-test` which matches any regex submitted on the flag. In this case, and to avoid footguns, the `--test` matches a single test. So I would propose to drop the `--exlude-test` flag.

anvacaru (approved):

PR #2031, opened 2023-08-22
Setting fetch-depth

F-WRunTime:
I think this may resolve the issues we've been seeing @dwightguth with corrupted submodule checkouts. 

With a related issue noted here: This could be a fix for it. 
https://stackoverflow.com/questions/66460369/git-submodule-update-failure


ehildenb:
Do we need to make the same change to `master-push.yml` and `foundy-update-output.yml`?

ehildenb (approved):

F-WRunTime:
> Do we need to make the same change to `master-push.yml` and `foundy-update-output.yml`?

Lets do it to avoid this issue if it were to affect those locations as well. Will commit in a moment

ehildenb (approved):

PR #2032, opened 2023-08-23
Default `max_depth` = 25000, no `break-on-calls`

palinatolmach:
Partially addresses https://github.com/runtimeverification/evm-semantics/issues/1958.

This PR is aimed to improve the default observable performance of KEVM by making `no-break-on-calls` default and increasing default `--max-depth` to `25000`.


PetarMax:
There seems to be a proof failing because an output obtained from the is different - I am confused why this could be showing up with this change - could it be that the output depends on `break-on-calls`? @ehildenb @palinatolmach

PetarMax:
It could be flakiness, I just saw it fail twice in the same way, so I flagged.

PetarMax:
It just failed in the same way right now...

PetarMax:
I can confirm that there is a problem with the output, here is a snippet:

OBTAINED:
```
(2819 steps)\n├─ 5\n│   k: CALL 9223372036854765663 645326474426547203313410069153905908525362434349 0 388  ...\n│   pc: 2389\n│   callDepth: 0\n│   statusCode: STATUSCODE:StatusCode\n│\n│  (1 step)\n├─ 6\n│   k: #checkCall 728815563385977040452943777879061427756277306518 0 ~> #call 728815563 ...\n│   pc: 2389\n│   callDepth: 0\n│   statusCode: STATUSCODE:StatusCode\n│\n│
```
EXPECTED:
```
(1000 steps)\n├─ 5\n│   k: #gas [ MSTORE , MSTORE 192 46308022326495007027972728677917914892729792999299745 ...\n│   pc: 2271\n│   callDepth: 0\n│   statusCode: STATUSCODE:StatusCode\n│\n│  (1000 steps)\n├─ 6\n│   k: #gasExec ( LONDON , JUMP 2621 ) ~> #deductGas ~> #access [ JUMP , JUMP 2621 ] ~> ...\n│   pc: 2691\n│   callDepth: 0\n│   statusCode: STATUSCODE:StatusCode\n│\n│  (819 steps)\n├─ 7\n│   k: CALL 9223372036854765663 645326474426547203313410069153905908525362434349 0 388  ...\n│   pc: 2389\n│   callDepth: 0\n│   statusCode: STATUSCODE:StatusCode\n│\n│  (1 step)\n├─ 8\n│   k: #checkCall 728815563385977040452943777879061427756277306518 0 ~> #call 728815563 ...\n│   pc: 2389\n│   callDepth: 0\n│   statusCode: STATUSCODE:StatusCode\n│\n│
```

from which you can see two things:
1. OBTAINED does 2819 steps immediately, EXPECTED does 1000+1000+819; but also, more worryingly
2. they both do 1 step afterwards, which means we *still break on calls* and the PR needs further work, I think.

@palinatolmach @ehildenb

palinatolmach:
@PetarMax thanks for the investigation, I'll look into it now! I wanted to check if the test is failing locally too, but had problems running tests on my machine. It's strange that it doesn't fail when the booster is on though.

ehildenb:
Out of date.

PR #2035, opened 2023-08-24
Hotfix/gh pages

F-WRunTime:
Think I figured out what's wrong with the gh-pages build. This should fix it. I have a temp repository built on my local box I can try to push the gh-pages with if we'd like to try it first before releasing this into CI.


.github/workflows/master-push.yml
```diff
@@ -148,12 +148,13 @@ jobs:
         run: |
           git checkout -B gh-pages
           cd web
+          git submodule init k-web-theme && git submodule update --recursive
```
ehildenb, line 151: Hmmmm, why would this be needed? That submodule already exists in the repository. Shouldn't it be already included in the recursive submodule checkout above?

Also, I'm not even sure we're still using this submodule. @shd101wyy do we still need this submodule?

F-WRunTime: I wass seeing it as part of the package.json definition for 'build install run'. 

The submodule you're right should be there. However when going through these steps. it's not being checked out when calling 'git submodule update --recursive' ... this could just be a local issue I was having each time I started with a fresh clone. I'll pull it for now and lets push this through with the gitmodules getting added back.

ehildenb (approved):

F-WRunTime:
To answer the k-web-theme submodule was still being used as part of the npm build definition in web/package.json

PR #1866, opened 2023-05-27
Upstream lemmas from Wormhole engagement

qian-hu:
Part of: https://github.com/runtimeverification/evm-semantics/issues/1588

This PR includes lemmas from previous Wormhole engagement.

PR #2037, opened 2023-08-24
Lemmas on Boolean reasoning, set reasoning, map lookup

PetarMax:
This PR upstreams further lemmas from Lukso/BlockSwap concerning
- Boolean reasoning;
- set reasoning; and
- map lookup.


include/kframework/lemmas/lemmas.k
```diff
@@ -76,16 +96,47 @@ module LEMMAS [symbolic]
     rule 0 <=Int keccak( _ )             => true [simplification]
     rule         keccak( _ ) <Int pow256 => true [simplification]
 
+    // Further keccak assumptions: these assumptions are not sound in principle, but are
+    // required for verification - they should be collected at the end of execution
+
+    // keccak does not equal a concrete value
+    rule [keccak-eq-conc-false]: keccak(_A)  ==Int _B => false [symbolic(_A), concrete(_B), simplification]
+    rule [keccak-neq-conc-true]: keccak(_A) =/=Int _B => true  [symbolic(_A), concrete(_B), simplification]
```
ehildenb, line 104: Hmmmmm, these could be problematic. Say you have `keccak(X) ==Int SOME_HASH` in your constraints, but then somewhere else you have `X ==Int 3`, and then we have that `keccak(3) ==Int SOME_HASH` is true. This could happen, for example, if we looked up a symbolic storage location `X`, which later resolved to `3`, and `SOME_HASH` was the actual slot for storage location `3`.

But if these rules fire before the substitution `X ==Int 3` is applied and then `keccak` is simplified, we'll get the incorrect result. So it seems we're relying on the order that the backend tries things here.

ehildenb: If they're useful for the engagement, then we should keep them there, but I don't think we should upstream them.

include/kframework/lemmas/lemmas.k
```diff
@@ -76,16 +96,47 @@ module LEMMAS [symbolic]
...
+    rule [keccak-inj]: keccak(A) ==Int keccak(B) => A ==K B [simplification]
+
+    // keccak has no "fixpoint"
+    rule [keccak-no-fix-eq-false]: #buf(32, keccak(X))  ==K X => false [simplification]
+    rule [keccak-no-fix-neq-true]: #buf(32, keccak(X)) =/=K X => true  [simplification]
+
+    // disequality of keccak under shifting
+    rule ( ( keccak ( _X ) +Int A ) modInt pow256 ) ==Int keccak ( _Y ) => false
+      requires 0 <Int A andBool A <Int pow256
+      [simplification, symbolic(_X, _Y), concrete(A)]
```
ehildenb, line 121: Again, relying on the order things happen in here I think.

PetarMax:
All of the `keccak` lemmas are indeed problematic (unsound, really), but needed for the engagements. I will remove them from this PR.

include/kframework/lemmas/lemmas.k
```diff
@@ -14,6 +14,23 @@ module LEMMAS [symbolic]
...
+
+    rule B ==Bool true  => B         [simplification, comm]
+    rule B ==Bool false => notBool B [simplification, comm]
+
+    rule notBool notBool X => X [simplification]
+
+    rule { true  #Equals X     ==K   Y     } => { X #Equals Y } [simplification]
+    rule { true  #Equals X:Int ==Int Y:Int } => { X #Equals Y } [simplification]
+    rule { false #Equals X     ==K   Y     } => #Not ( { X #Equals Y } ) [simplification]
+    rule { false #Equals X:Int ==Int Y:Int } => #Not ( { X #Equals Y } ) [simplification]
```
ehildenb, line 29:
```suggestion
    rule { true  #Equals X ==K   Y } => { X #Equals Y } [simplification]
    rule { true  #Equals X ==Int Y } => { X #Equals Y } [simplification]
    rule { false #Equals @B } => #Not { true #Equals @B } [simplification]
```

Sorts are inferred automatically for `X` and `Y` because they are under the `_==Int_` symbol. Also, would adding one lemma to handle `{ false #Equals B }` help here?

Another thing to look at: https://kframework.org/k-distribution/include/kframework/builtin/domains/#implementation-of-booleans

Notice that we're using set variables there (prefixed with `@` instead).

Also notice, module `INT-KORE` seems to have many of these lemmas already? https://kframework.org/k-distribution/include/kframework/builtin/domains/#implementation-of-integers. Are we including them here instead of rewriting them?

PetarMax: Aha, ok, excellent, I didn't notice them in `INT-SYMBOLIC`. I think we should probably have a list of them somewhere so that all of the relevant libraries are included. I'll go through and cross-check.

PetarMax: @ehildenb I've looked at all of the `[...]-SYMBOLIC` modules, as well as the `lemmas.k` file, so now there should be no duplicates.

include/kframework/lemmas/lemmas.k
```diff
@@ -28,14 +28,18 @@ module LEMMAS [symbolic]
     rule         chop(_V) <Int pow256 => true        [simplification]
 
     rule X *Int Y <Int pow256 => true requires Y <=Int maxUInt256 /Int X [simplification]
-    rule #if B #then C +Int C1 #else C +Int C2 #fi => C +Int #if B #then C1 #else C2 #fi [simplification] 
+    rule #if B #then C +Int C1 #else C +Int C2 #fi => C +Int #if B #then C1 #else C2 #fi [simplification]jj
```
ehildenb, line 31:
```suggestion
    rule #if B #then C +Int C1 #else C +Int C2 #fi => C +Int #if B #then C1 #else C2 #fi [simplification]
```

include/kframework/serialization.md
```diff
@@ -21,7 +21,7 @@ Address/Hash Helpers
 -   `keccak` serves as a wrapper around the `Keccak256` in `KRYPTO`.
 
 ```k
-    syntax Int ::= keccak ( Bytes ) [function, total, smtlib(smt_keccak)]
+    syntax Int ::= keccak ( Bytes ) [function, total, injective, smtlib(smt_keccak)]
  // -------------------------------------------------------------------------
```
ehildenb, line 25:
```suggestion
    syntax Int ::= keccak ( Bytes ) [function, total, injective, smtlib(smt_keccak)]
 // ------------------------------------------------------------------------------------
```

PetarMax:
I'm having some formatting issues here as well with the expected outputs. For example, for the I have additional spacing, different order of constraints (ACTUAL top, EXPECTED bottom):
```
statusCode: STATUSCODE_FINAL\n\n\n\n\n    \n    \n    rule [BASIC-BLOCK-1-TO-3]:
statusCode: STATUSCODE_FINAL\n\n\n\n\n\n\n    rule [BASIC-BLOCK-1-TO-3]:

[label(BASIC-BLOCK-1-TO-3)]\n    \n    rule [BASIC-BLOCK-3-TO-4]:
[label(BASIC-BLOCK-1-TO-3)]\n\n    rule [BASIC-BLOCK-3-TO-4]:

[label(BASIC-BLOCK-3-TO-4)]\n    \n    rule [BASIC-BLOCK-4-TO-5]:
[label(BASIC-BLOCK-3-TO-4)]\n\n    rule [BASIC-BLOCK-4-TO-5]:

[label(BASIC-BLOCK-4-TO-5)]\n    \n    rule [BASIC-BLOCK-6-TO-8]:
[label(BASIC-BLOCK-4-TO-5)]\n\n    rule [BASIC-BLOCK-6-TO-8]:

andBool ( VV0_n_114b9705:Int <=Int 51816696836262767\n       andBool ( CALLER_ID:Int <Int pow160\n       andBool ( ORIGIN_ID:Int <Int pow160\n       andBool ( NUMBER_CELL:Int <=Int maxSInt256\n       andBool ( VV0_n_114b9705:Int <Int pow256\n               ))))))))))\n
andBool ( CALLER_ID:Int <Int pow160\n       andBool ( ORIGIN_ID:Int <Int pow160\n       andBool ( NUMBER_CELL:Int <=Int maxSInt256\n       andBool ( VV0_n_114b9705:Int <Int pow256\n       andBool ( VV0_n_114b9705:Int <=Int 51816696836262767\n               ))))))))))\n

[label(BASIC-BLOCK-6-TO-8)]\n    \n    rule [BASIC-BLOCK-7-TO-9]:
[label(BASIC-BLOCK-6-TO-8)]\n\n    rule [BASIC-BLOCK-7-TO-9]:

[label(BASIC-BLOCK-7-TO-9)]\n    \n    rule [BASIC-BLOCK-8-TO-10]:
[label(BASIC-BLOCK-7-TO-9)]\n\n    rule [BASIC-BLOCK-8-TO-10]:

[label(BASIC-BLOCK-8-TO-10)]\n    \n    rule [BASIC-BLOCK-9-TO-11]:
[label(BASIC-BLOCK-8-TO-10)]\n\n    rule [BASIC-BLOCK-9-TO-11]: 

* another re-ordering of constraints, too long *

[label(BASIC-BLOCK-9-TO-11)]\n    \n    rule [BASIC-BLOCK-10-TO-12]:
[label(BASIC-BLOCK-9-TO-11)]\n\n    rule [BASIC-BLOCK-10-TO-12]:

[label(BASIC-BLOCK-10-TO-12)]\n    \n    rule [BASIC-BLOCK-11-TO-13]:
[label(BASIC-BLOCK-10-TO-12)]\n\n    rule [BASIC-BLOCK-11-TO-13]:

* another re-ordering of constraints, too long *
```
I'm confused, because I would think the way things are printed hasn't been affected by the PR...

include/kframework/lemmas/lemmas.k
```diff
@@ -28,14 +28,18 @@ module LEMMAS [symbolic]
...
+    rule #if B #then C +Int C1 #else C +Int C2 #fi => C +Int #if B #then C1 #else C2 #fi [simplification]
 
   // ########################
   // Set Reasoning
   // ########################
 
-    rule X in (SetItem(Y) _   ) => true      requires X  ==Int Y [simplification]
-    rule X in (SetItem(Y) REST) => X in REST requires X =/=Int Y [simplification]
+    rule X in SetItem ( Y ) S:Set => ( X ==K Y andBool notBool ( X in S ) ) orBool ( X =/=K Y andBool X in S ) [simplification]
+    rule X in S:Set SetItem ( Y ) => ( X ==K Y andBool notBool ( X in S ) ) orBool ( X =/=K Y andBool X in S ) [simplification]
```
lucasmt, line 38: Is the use of `==K` instead of `==Int` to make the lemma more general? Are those operators always identical when applied to Ints?

include/kframework/lemmas/lemmas.k
```diff
@@ -80,12 +82,19 @@ module LEMMAS [symbolic]
...
-    rule #lookup ( _M:Map [ K1 <- V1 ] , K2 ) => #lookup ( K1 |-> V1 , K1 )  requires K1 ==Int  K2 [simplification]
-    rule #lookup (  M:Map [ K1 <- _  ] , K2 ) => #lookup ( M         , K2 )  requires K1 =/=Int K2 [simplification]
+    rule keys(.Map)             => .Set                    [simplification]
+    rule keys(M [ K <- _V ])    => keys(M) |Set SetItem(K) [simplification]
+    rule keys(M [ K <- undef ]) => keys(M) -Set SetItem(K) [simplification]
 
     rule 0 <=Int #lookup( _M:Map , _ )             => true [simplification, smt-lemma]
     rule         #lookup( _M:Map , _ ) <Int pow256 => true [simplification, smt-lemma]
 
+    rule #lookup(M:Map, X:Int) => 0 requires notBool (X:Int in keys(M):Set) [simplification]
```
lucasmt, line 92: What's the difference between this lemma and [this rule](https://github.com/runtimeverification/evm-semantics/blob/ba06470f78f9b37ff9058b62a87c749a5c7aee13/include/kframework/evm-types.md?plain=1#L414) in the definition of `#lookup`?

include/kframework/lemmas/lemmas.k
PetarMax: Yes, it will work for arbitrary sets, including sets of integers, in which case the backend may automatically convert `==K` to `==Int`, I believe.

include/kframework/lemmas/lemmas.k
PetarMax: The difference is that `in_keys` is a `K` built-in. I wasn't able to use it well in Lukso for some reason and so I created `keys` instead.

include/kframework/lemmas/lemmas.k
lucasmt, line 38: The version of this rule that we have in `gateway-lemmas.k` is
```
rule X in SetItem ( Y ) S:Set => ( X ==Int Y ) orBool ( X in S )
  [simplification]
```
which seems better because we might have cases where `X ==K Y` but we don't know if `X in S` or not.

PetarMax: By construction of sets, I believe that we have to have the stronger one, otherwise the set is malformed. This is similar to maps, where `(K -> _) M` means that `M` does not have `K` as a key. Am I correct, @ehildenb?

PetarMax:
On hold for the moment, waiting for #1866.

PetarMax:
I am having severe difficulties passing CI because it seems that the set simplifications increase the running times over the limit. These simplifications, however, are extremely simple, and are now just with concrete values. @jberthold , how can I do a performance analysis with and without them? They just might get deleted for this PR. @ehildenb, @lucasmt, if you've got any further thoughts on this PR, please share, it would be good to have it merged.

PetarMax:
I am also seeing @dwightguth's PR failing on the booster tests with the same percentage compleeted (97%) as this one (in the last run, at least). Perhaps it is not the simplifications that are causing the CI failure. @ehildenb, should we increase the CI times so that we are not that stochastic in pass/fail and not that much on the limit? What is the actual performance of all of these tests and what is the standard deviation observed on CI?

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/lemmas.k
```diff
@@ -37,6 +37,9 @@ module LEMMAS [symbolic]
     rule X in (SetItem(Y) _   ) => true      requires X  ==Int Y [simplification]
     rule X in (SetItem(Y) REST) => X in REST requires X =/=Int Y [simplification]
 
+    rule   ( S:Set |Set SetItem ( X )                      ) |Set SetItem( X ) => S [simplification]
+    rule ( ( S:Set |Set SetItem ( X ) ) |Set SetItem ( Y ) ) |Set SetItem( X ) => S [simplification]
```
PetarMax, line 41: Are these too ad-hoc now? @ehildenb

lucasmt: I'm confused, I don't see why these would be valid. There's nothing here guaranteeing that `X` or `Y` are already in `S`

lucasmt: Was it supposed to be `S SetItem(X)` instead of `S |Set SetItem(X)`?

PetarMax: Yes, oh wow, so many changes, and I've lost track of what's happening, it's the LHS

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/serialization.md
```diff
@@ -21,7 +21,7 @@ Address/Hash Helpers
 -   `keccak` serves as a wrapper around the `Keccak256` in `KRYPTO`.
 
 ```k
-    syntax Int ::= keccak ( Bytes ) [function, total, smtlib(smt_keccak)]
+    syntax Int ::= keccak ( Bytes ) [function, total, injective, smtlib(smt_keccak)]
  // -------------------------------------------------------------------------
```
ehildenb, line 25:
```suggestion
    syntax Int ::= keccak ( Bytes ) [function, total, injective, smtlib(smt_keccak)]
 // ------------------------------------------------------------------------------------
```

ehildenb:
Please add:

- Tests of these new lemmas, added to relevant files in `tests/specs/functional/...`.
- Performance numbers on both KEVM and Kontrol, comparing against master KEVM and Kontrol (instructinos: https://github.com/runtimeverification/hs-backend-booster/issues/355). We need to make sure this doesn't cause a regression.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/serialization.md
PetarMax: How strange, I thought this was taken care of...

PetarMax:
I've added tests for the lemmas, one for `#lookup` and one for a combination of set reasoning and Boolean reasoning.

There should be no need to test `keccak`, since that would mean testing the `injective` attribute for functions.

The CI times have gone down substantially since the limites were increased; this PR should not introduce any slowdowns. I am not able to confirm this 100% on my machine since I cannot build the `plugin` dependency due to M1 issues.

palinatolmach:
@PetarMax now that the [blocking booster PR](https://github.com/runtimeverification/hs-backend-booster/pull/405) has been merged, shall we be able to merge it?

PetarMax:
Not yet, the PR is still failing in the booster and passing in the Haskell backend on some tests. I'm trying to figure out what's happening and liaising with @geo2a.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/serialization.md
```diff
@@ -21,8 +21,8 @@ Address/Hash Helpers
 -   `keccak` serves as a wrapper around the `Keccak256` in `KRYPTO`.
 
 ```k
-    syntax Int ::= keccak ( Bytes ) [function, total, smtlib(smt_keccak)]
- // -------------------------------------------------------------------------
+    syntax Int ::= keccak ( Bytes ) [function, total, injective, smtlib(smt_keccak)]
+ // --------------------------------------------------------------------------------
```
ehildenb, line 25: I guess we've largely decided that this is unsound and we shouldn't do it (and will take another approach): https://github.com/runtimeverification/kontrol/issues/165#issuecomment-1860803741.

So maybe we can just pull this out of this PR, so the other good lemmas here can be merged?

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/int-simplification.k
```diff
@@ -13,35 +13,77 @@ module INT-SIMPLIFICATION-HASKELL [symbolic, kore]
...
+
+    rule A   <Int B -Int C => A +Int C  <Int B [concrete(A, C), symbolic(B), simplification(45)]
+    rule A  <=Int B -Int C => A +Int C <=Int B [concrete(A, C), symbolic(B), simplification(45)]
+    rule A   >Int B -Int C => A +Int C  >Int B [concrete(A, C), symbolic(B), simplification(45)]
+    rule A  >=Int B -Int C => A +Int C >=Int B [concrete(A, C), symbolic(B), simplification(45)]
+
+    rule A +Int B  ==Int C => A ==Int C -Int B [concrete(B, C), symbolic(A), simplification(45), comm]
+    rule A +Int B  ==Int C => B ==Int C -Int A [concrete(A, C), symbolic(B), simplification(45), comm]
+    rule A -Int B  ==Int C => A ==Int C +Int B [concrete(B, C), symbolic(A), simplification(45), comm]
+    rule A -Int B  ==Int C => A -Int C ==Int B [concrete(A, C), symbolic(B), simplification(45), comm]
```
ehildenb, line 81: Should we add at least a handful of tests of these symbols? Just so we have a test-harness for them in place?

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/serialization.md
PetarMax: Yes

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/serialization.md
```diff
@@ -22,7 +22,7 @@ Address/Hash Helpers
 
 ```k
     syntax Int ::= keccak ( Bytes ) [function, total, smtlib(smt_keccak)]
- // -------------------------------------------------------------------------
+ // --------------------------------------------------------------------------------
```
ehildenb, line 25:
```suggestion
 // -------------------------------------------------------------------------
```

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/lemmas.k
```diff
@@ -72,20 +73,21 @@ module LEMMAS [symbolic]
...
 
-    rule #lookup ( _M:Map [ K1 <- V1 ] , K2 ) => #lookup ( K1 |-> V1 , K1 )  requires K1 ==Int  K2 [simplification]
-    rule #lookup (  M:Map [ K1 <- _  ] , K2 ) => #lookup ( M         , K2 )  requires K1 =/=Int K2 [simplification]
-
     rule 0 <=Int #lookup( _M:Map , _ )             => true [simplification, smt-lemma]
     rule         #lookup( _M:Map , _ ) <Int pow256 => true [simplification, smt-lemma]
 
+    rule #lookup ( _M:Map [ K1 <- V1 ] , K2 )    => #lookup ( K1 |-> V1 , K1 ) requires K1 ==Int  K2 [simplification]
+    rule #lookup (  M:Map [ K1 <- _  ] , K2 )    => #lookup ( M         , K2 ) requires K1 =/=Int K2 [simplification]
+    rule #lookup ( (K1:Int |-> _) M:Map, K2:Int) => #lookup ( M,          K2 ) requires K1 =/=Int K2 [simplification]
```
ehildenb, line 89:
```suggestion
    rule #lookup ( (K1:Int |-> _) M:Map, K2:Int) => #lookup ( M         , K2 ) requires K1 =/=Int K2 [simplification]
```

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/int-simplification.k
```diff
@@ -13,35 +13,77 @@ module INT-SIMPLIFICATION-HASKELL [symbolic, kore]
...
+    rule A   <Int B +Int C => A -Int B  <Int C [concrete(A, B), symbolic(C), simplification(45)]
+    rule A  <=Int B +Int C => A -Int B <=Int C [concrete(A, B), symbolic(C), simplification(45)]
+    rule A   >Int B +Int C => A -Int B  >Int C [concrete(A, B), symbolic(C), simplification(45)]
+    rule A  >=Int B +Int C => A -Int B >=Int C [concrete(A, B), symbolic(C), simplification(45)]
+
+    rule A   <Int B +Int C => A -Int C  <Int B [concrete(A, C), symbolic(B), simplification(45)]
+    rule A  <=Int B +Int C => A -Int C <=Int B [concrete(A, C), symbolic(B), simplification(45)]
+    rule A   >Int B +Int C => A -Int C  >Int B [concrete(A, C), symbolic(B), simplification(45)]
+    rule A  >=Int B +Int C => A -Int C >=Int B [concrete(A, C), symbolic(B), simplification(45)]
+
```
ehildenb, line 67: I notice that some of these move the concrete to the left and symbolic to the rgiht, and some the other direction. I wonder if there is benefit to always picking one side of the comparison for the concrete and teh other for the symbolic?

I think there are also some rules somewhere which normalize all `_>=_` and `_>_` to the other operators (`_<=_` and `_<_`), so I wonder if we actually don't need all the duplicate rules? Though I can't find those rules now, so maybe they would need to be added.

ehildenb (approved):

PR #2043, opened 2023-08-25
Throw an error in the case of contracts with the same name

iFrostizz:
Closes https://github.com/runtimeverification/kontrol/issues/18

### DIFF
--- a/kevm-pyk/src/kontrol/foundry.py
+++ b/kevm-pyk/src/kontrol/foundry.py
@@ -111,6 +111,9 @@ class Foundry:
             contract_name = json_path.split('/')[-1]
             contract_json = json.loads(Path(json_path).read_text())
             contract_name = contract_name[0:-5] if contract_name.endswith('.json') else contract_name
+            if _contracts.get(contract_name) is not None:
+                raise RuntimeError('Project has some duplicated contract names that may clash with K definitions.')
+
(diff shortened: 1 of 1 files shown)


kevm-pyk/src/kontrol/foundry.py
```diff
@@ -111,6 +111,9 @@ def contracts(self) -> dict[str, Contract]:
             contract_name = json_path.split('/')[-1]
             contract_json = json.loads(Path(json_path).read_text())
             contract_name = contract_name[0:-5] if contract_name.endswith('.json') else contract_name
+            if _contracts.get(contract_name) is not None:
+                raise RuntimeError('Project has some duplicated contract names that may clash with K definitions.')
```
palinatolmach, line 115:
```suggestion
                raise RuntimeError('Project contains duplicated contract names that may clash in K definitions.')
```
What do you think of the slightly paraphrased formulation? The original version works too, but I read it as "it might clash with K keywords", which you've already fixed. Feel free to ignore this suggestion :).

palinatolmach (approved):

iFrostizz:
let's wait for https://github.com/runtimeverification/evm-semantics/pull/2026 to get merged (some testing annoyances)

kevm-pyk/src/kontrol/foundry.py
iFrostizz: That works well thank you!

PR #2044, opened 2023-08-25
Make booster default in Foundry proofs

palinatolmach:
Partially addresses https://github.com/runtimeverification/evm-semantics/issues/1958.

This PR is aimed to improve the default observable performance of Kontrol by using the Booster backend in `foundry-prove` and generating an LLVM library during `foundry-kompile` by default.


kevm-pyk/src/kontrol/__main__.py
```diff
@@ -451,6 +452,19 @@ def parse(s: str) -> list[T]:
...
+        default=True,
+        action='store_true',
+        help='Make kompile generate a dynamic llvm library.',
+    )
+    foundry_kompile.add_argument(
+        '--no-llvm-library',
+        dest='llvm_library',
+        action='store_false',
+        help='Do not generate a dynamic llvm library during kompile.',
+    )
```
ehildenb, line 467: We're already actually building with the llvm library every time, so this isn't needed. Also, this is made more explicit here: https://github.com/runtimeverification/evm-semantics/pull/2034

So we can just remove this option altogether (and the `conflict_handler` option above).

kevm-pyk/src/kontrol/__main__.py
```diff
@@ -499,9 +513,15 @@ def parse(s: str) -> list[T]:
...
         action='store_true',
-        help="Use the booster RPC server instead of kore-rpc. Requires calling foundry-kompile with the '--with-llvm-library' flag",
+        help='Use the booster RPC server instead of kore-rpc.',
+    )
+    foundry_prove_args.add_argument(
+        '--no-booster',
+        dest='use_booster',
+        action='store_false',
+        help='Do not use the booster RPC server instead of kore-rpc.',
     )
```
ehildenb, line 525: We should name this option `--no-use-booster`, not `--no-booster`. The convention is to prefix the exact same argument name with `no-`.

ehildenb:
We should also adjust here: https://github.com/runtimeverification/evm-semantics/blob/ba06470f78f9b37ff9058b62a87c749a5c7aee13/.github/workflows/test-pr.yml#L172, to add the option `--no-use-booster`, so that we're still testing non-booster on CI.

ehildenb:
This PR shuold probably wait for https://github.com/runtimeverification/evm-semantics/pull/2034 to merge, which simplifies the booster building process a bit and build sit by default every time we build Foundry proofs anyway.

kevm-pyk/src/kontrol/__main__.py
palinatolmach: Thanks! Removed in https://github.com/runtimeverification/evm-semantics/pull/2044/commits/18f4e7722736dfa9f112aceb21b2d26c6e2adb19.

kevm-pyk/src/kontrol/__main__.py
palinatolmach: Renamed to `--no-use-booster` in https://github.com/runtimeverification/evm-semantics/pull/2044/commits/18f4e7722736dfa9f112aceb21b2d26c6e2adb19.

.github/workflows/test-pr.yml
```diff
@@ -198,7 +198,7 @@ jobs:
       - name: 'Build Foundry'
         run: docker exec -u github-user kevm-ci-foundry-${GITHUB_SHA} /bin/bash -c 'make build-foundry -j2'
       - name: 'Foundry Prove'
-        run: docker exec -u github-user kevm-ci-foundry-${GITHUB_SHA} /bin/bash -c 'make test-foundry-prove PYTEST_ARGS=-vv'
+        run: docker exec -u github-user kevm-ci-foundry-${GITHUB_SHA} /bin/bash -c 'make test-foundry-prove PYTEST_ARGS=-vv --no-use-booster'
```
iFrostizz, line 201: This `--no-use-booster` flag logic should be added somewhere, you could either:
- rename all uses of `use_booster` in pytest tests
- handle the logic in the `Makefile` by writing `--use-booster` if the flag `NO_USE_BOOSTER` is false
As you want to make the booster the default, 1. might be a better choice and a less confusing one

iFrostizz (changes requested):

.github/workflows/test-pr.yml
iFrostizz: Actually, what about writing 
```diff
diff --git a/kevm-pyk/src/tests/conftest.py b/kevm-pyk/src/tests/conftest.py
index 7a2b381d3..58cd4dcd6 100644
--- a/kevm-pyk/src/tests/conftest.py
+++ b/kevm-pyk/src/tests/conftest.py
@@ -30,4 +30,4 @@ def update_expected_output(request: FixtureRequest) -> bool:

 @pytest.fixture(scope='session')
 def use_booster(request: FixtureRequest) -> bool:
-    return request.config.getoption('--use-booster')
+    return not request.config.getoption('--no-use-booster')
```

palinatolmach: Thanks @iFrostizz for the suggestion, looks very good! Done in https://github.com/runtimeverification/evm-semantics/pull/2044/commits/393e42b820e6c1c6e668220889980383711e8a0b.

iFrostizz: I agree that it's okay to leave the CI commands as they are @palinatolmach , because they are not user facing, and because we *may* drop non booster CI jobs in the future. Also, the `--use-booster` option is more straightforward, having `--no-use-booster` features a double negation unless you edit further the tests, it looks like unnecessary effort!

palinatolmach:
Decided not to merge this PR and enable booster by default until its behavior is stabilized. Moving to Blocked.

palinatolmach:
> @palinatolmach What is this blocked by?

The instability of the booster backend — technically, we can merge it and there's no specific issue that should be fixed first, but it just doesn't make sense to do it now as the booster is quite problematic now.

palinatolmach:
Closing as we decided not to use booster by default until the booster stabilizes. Will open a similar PR against Kontrol repository when that happens.

PR #2049, opened 2023-08-26
Place skipped tests first

iFrostizz:


iFrostizz:
Marginal gains with the haskell backend, but more than 30% of time saved with the booster because it only runs the foundry tests, making an heavy use of multithreading.

PR #2060, opened 2023-09-06
Re-adding GitModule for ethereum-tests; Fix Submodule checkout in CI

F-WRunTime:
After many a commit / testing. Removing the module and re-adding it appears to be the solution to fixing the revision missing issue. 

Additional Testing still required. 

The original modules defined the branch to watch as develop. This is redundant since it is the default branch and possibly the root cause of the issue... Unclear why specifically, git?


F-WRunTime:
This runner had previously failed consistently to checkout submodule 
image

F-WRunTime:
Don't think the failing test are related to this change, but last thing to investigate.

ehildenb:
The failures in the `concrete` tests are probably due to this change, because the submodule which contains the tests (`tests/ethereum`) changed revisions. So likely we need to update the file `tests/failing.llvm` to include any new tests that have been added which are now failing.

PR #2065, opened 2023-09-07
Automatically add subproofs to parallelized execution once they become available

nwatson22:


kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -82,6 +82,7 @@ def kevm_prove(
     break_on_calls: bool = True,
     extract_branches: Callable[[CTerm], Iterable[KInner]] | None = None,
     abstract_node: Callable[[CTerm], CTerm] | None = None,
+    max_branches: int | None = 1,
```
tothtamas28, line 85: To preserve backward compatibility for now:

```suggestion
    max_branches: int | None = None,
```

kevm-pyk/src/kontrol/foundry.py
```diff
@@ -771,18 +775,71 @@ def _split_test(test: tuple[str, str | None]) -> tuple[str, str, str | None]:
...
+                port=port,
+                command=kore_rpc_command,
+                bug_report=br,
+                smt_timeout=smt_timeout,
+                smt_retry_limit=smt_retry_limit,
+            )
+
+        results = {}
+        futures: dict[str, Future] = {}
+        with KoreServerPool(create_server=create_server, max_workers=workers) as pool:
```
tothtamas28, line 795: `KoreServerPool` starts a server per worker, Unfortunately, servers are not stateless due to calls to `add-module`.

There are two solutions:

1. Do not use `KoreServerPool`, but have a separate server for each test. As the server can handle concurrent requests, having more than one active workers for a single test shouldn't cause an issue.
2. Add an endpoint `reset-module` and call it every time a worker picks up a new task. Then add back the modules necessary for the given test.

nwatson22:
Waiting on https://github.com/runtimeverification/pyk/pull/654

kevm-pyk/src/kevm_pyk/utils.py
nwatson22: It's now an option available for the `foundry-prove` command, and is None by default.

kevm-pyk/src/kontrol/foundry.py
nwatson22: There is now a system where it starts one server for each original/base test, and passes in the port for the appropriate server when starting subproofs. However, I am now realizing that we still want the servers to be started inside the worker threads, since startup is non-trivial.

kevm-pyk/pyproject.toml
```diff
@@ -17,7 +17,8 @@ packages = [
 [tool.poetry.dependencies]
 python = "^3.10"
 pathos = "*"
-pyk = { git = "https://github.com/runtimeverification/pyk.git", tag="v0.1.434" }
+pyk = { path = "/home/noah/dev/pyk", develop=true }
```
nwatson22, line 20:
```suggestion
```

kevm-pyk/pyproject.toml
```diff
@@ -17,7 +17,7 @@ packages = [
 [tool.poetry.dependencies]
 python = "^3.10"
 pathos = "*"
-pyk = { git = "https://github.com/runtimeverification/pyk.git", tag="v0.1.434" }
+pyk = { git = "https://github.com/runtimeverification/pyk.git", branch="noah/subproof-split" }
```
tothtamas28, line 20: Make sure not to accidentally merge this.

kevm-pyk/src/kontrol/foundry.py
```diff
@@ -413,10 +416,13 @@ def help_info() -> list[str]:
         return res_lines
 
     def proofs_with_test(self, test: str) -> list[Proof]:
+        #          print([pid.split(':')[0] for pid in listdir(self.proofs_dir)])
+        #          print(single(self._escape_brackets([test])))
```
tothtamas28, line 420: Make sure to remove these debug comments for the final version.

kevm-pyk/src/kontrol/foundry.py
```diff
@@ -758,37 +800,87 @@ def _init_and_run_proof(_init_problem: tuple[str, str, str | None]) -> tuple[boo
...
+                definition_dir=foundry.kevm.definition_dir,
+                llvm_definition_dir=llvm_definition_dir,
+                module_name=foundry.kevm.main_module,
+                command=kore_rpc_command,
+                bug_report=bug_report,
+                smt_timeout=smt_timeout,
+                smt_retry_limit=smt_retry_limit,
+            )
+
+        class Task:
```
tothtamas28, line 821:
```suggestion
        class Task(ABC):
```

kevm-pyk/src/kontrol/foundry.py
```diff
@@ -758,37 +800,87 @@ def _init_and_run_proof(_init_problem: tuple[str, str, str | None]) -> tuple[boo
...
+                self._test = test
+                self._id = id
+
+        class RunTask(Task):
+            _proof: Proof
+            _port: int
+
+            def __init__(self, proof: Proof, port: int):
+                self._proof = proof
+                self._port = port
```
tothtamas28, line 838:
```suggestion
        @dataclass(frozen=True)
        class CreateAndRunTask(Task):
            test: str
            id: str | None

        @dataclass(frozen=True)
        class RunTask(Task):
            proof: Proof
            port: int
```

kevm-pyk/src/kontrol/foundry.py
```diff
@@ -758,37 +800,87 @@ def _init_and_run_proof(_init_problem: tuple[str, str, str | None]) -> tuple[boo
...
 
-        apr_proofs = dict(zip(tests, _apr_proofs, strict=True))
-        return apr_proofs
+                if type(task) is CreateAndRunTask:
+                    contract, method = task._test.split('.')
+
+                    futures[task._test] = executor.submit(_init_and_run_proof, (contract, method, task._id))
+
+                elif type(task) is RunTask:
+                    futures[task._proof.id] = executor.submit(_run_proof, (task._port, task._proof))
```
tothtamas28, line 861: Consider something like

```suggestion
                futures[task.name] = task.submit(executor)
```

where `name` and `submit` are abstract members of `Task`.

kevm-pyk/src/kontrol/foundry.py
```diff
@@ -758,37 +800,87 @@ def _init_and_run_proof(_init_problem: tuple[str, str, str | None]) -> tuple[boo
...
+                    proof = future.result()
+                    results.append(proof)
+
+                    subproofs = proof.subproofs
+                    for subproof in subproofs:
+                        base_proofs[subproof.id] = base_proofs[_proof]
+                        port = servers[base_proofs[subproof.id]].port
+                        tasks.append(RunTask(subproof, port))
+
+            wait(futures.values())
```
tothtamas28, line 876: If I understand correctly, at this point all futures are waited before pending tasks are submitted. If so, this is not good for utilization.

tothtamas28: Btw, are two lists (`futures`, `tasks`) necessary? Given a single list of `pedning` futures, in each iteration you can try something like
1. If `pending` is empty then `break`.
2. Otherwise, if there are no `done` futures in `pending` then `continue`.
3. Otherwise, get a `done` future from `pending` and check it's results.
4. If based on the results there are new tasks to perform, submit them, and add all new futures to `pending`.
5. Remove the future from `pending`.

kevm-pyk/src/kontrol/foundry.py
nwatson22: Done.

kevm-pyk/src/kontrol/foundry.py
nwatson22: Done.

kevm-pyk/src/kontrol/foundry.py
nwatson22: Done.

kevm-pyk/src/kontrol/foundry.py
```diff
@@ -758,37 +800,87 @@ def _init_and_run_proof(_init_problem: tuple[str, str, str | None]) -> tuple[boo
...
+
+        tasks: list[Task] = [CreateAndRunTask(test, id) for test, id in tests]
+        base_proofs = {test: test for test, _ in tests}
+        servers = {}
+
+        for test, _ in tests:
+            servers[test] = create_server()
+
+        while len(tasks) > 0 or len(futures) > 0:
+            if len(tasks) > 0:
```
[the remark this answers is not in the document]
nwatson22: Done.

kevm-pyk/src/kontrol/foundry.py
nwatson22: I didn't realize this was what `wait` was doing. I've added the parameter `return_when='FIRST_COMPLETED'`, so it should wait for any, not all now.
I've also applied your suggestion and now there is only one list, and we just immediately add tasks to it rather than needing to wait for the next iteration.

kevm-pyk/src/kontrol/foundry.py
nwatson22: Done.

ehildenb:
Closing due to age.

PR #2068, opened 2023-09-13
Replace `Makefile` based build by `kevm-dist` command

tothtamas28:
Enables `kevm-pyk` to become a dependecy for downstream projects (e.g. `kontrol`).

`evm-semantics` and `plugin` sources become part of the Python distribution:
* Moves `include/kframework` into `kevm-pyk/src/kevm-pyk/kproj/evm-semantics`.
* Moves `deps/plugin` into `kevm-pyk/src/kevm-pyk/kproj/plugin`.

Targets can be built using a module `kevm_pyk.dist` and a corresponding command `kevm-dist`:
* Exposes functions to build, clean and look up targets `llvm`, `haskell`, `haskell-standalone`, `foundry` and `plugin`.
* Installs targets into a location within `$XDG_CACH_HOME` unique to the module installation.
* Dependent modules can look up targets from that location.

For local development,

```
make plugin-deps
make build-llvm
make build-haskell
make build-haskell-standalone
make build-foundry
```

become

```
poetry -C kevm-pyk run kevm-dist build plugin
poetry -C kevm-pyk run kevm-dist build llvm
poetry -C kevm-pyk run kevm-dist build haskell
poetry -C kevm-pyk run kevm-dist build haskell-standalone
poetry -C kevm-pyk run kevm-dist build foundry
```

---

TODOs before merge:

- [x] Fix Test jobs
- [x] Fix Nix jobs
- [x] Clean up `Makefile`
- [x] Add `RELEASE=true` mode for `kevm-dist build` for `test-concrete-execution`
- [x] Fix version update (set `VERSION` in `kevm_pyk/__init__,py`)
- [x] Update docs
- [ ] Collect feedback from auditors

TODOs after merge:

- [ ] Adjust `deps/plugin` submodule update ([here](https://github.com/runtimeverification/devops/blob/ce75d17e2de4bc9632d5de8f35694e75fbe542ca/deps/runtimeverification/blockchain-k-plugin.json#L6))


.github/workflows/test-pr.yml
```diff
@@ -112,11 +107,9 @@ jobs:
...
       - name: 'Build kevm-pyk'
         run: docker exec -u github-user kevm-ci-concrete-${GITHUB_SHA} /bin/bash -c 'make poetry'
-      - name: 'Build blockchain-k-plugin-deps'
-        run: docker exec -u github-user kevm-ci-concrete-${GITHUB_SHA} /bin/bash -c 'make CXX=clang++-14 plugin-deps'
-      - name: 'Build kevm'
-        run: docker exec -u github-user kevm-ci-concrete-${GITHUB_SHA} /bin/bash -c 'make build-kevm build-haskell build-llvm -j`nproc` RELEASE=true'
-      - name: 'Test kevm-pyk'
+      - name: 'Build targets'
+        run: docker exec -u github-user kevm-ci-concrete-${GITHUB_SHA} /bin/bash -c 'CXX=clang++-14 poetry -C kevm-pyk run kevm-dist --verbose build -j`nproc` plugin llvm haskell'
+      - name: 'Test integration'
```
ehildenb, line 112: We have kinda inconsistent usage of `-j...` across the different tests. Can we just use ``-j`nproc``` everywhere?

ehildenb:
It seems we probably need to update the README with updated instructions? Also, should we remove the `Makefile` recipes altogether, to make sure that users are going through the new flow.

.gitmodules
```diff
@@ -4,7 +4,7 @@
 	branch = develop
 	ignore = untracked
 [submodule "deps/plugin"]
-	path = deps/plugin
+	path = kevm-pyk/src/kevm_pyk/kproj/plugin
 	url = https://github.com/runtimeverification/blockchain-k-plugin
 	ignore = untracked
```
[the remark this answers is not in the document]
ehildenb: Ok, let's make sure to follow up on the devops repo after this is merged then. Specifically here: https://github.com/runtimeverification/devops/blob/ce75d17e2de4bc9632d5de8f35694e75fbe542ca/deps/runtimeverification/blockchain-k-plugin.json#L6

tothtamas28:
> It seems we probably need to update the README with updated instructions? Also, should we remove the `Makefile` recipes altogether, to make sure that users are going through the new flow.

Yes, I agree. There's a list under the PR description that collects all the tasks I could think of that need to be done before merging.

.github/workflows/test-pr.yml
tothtamas28: Done.

tothtamas28:
The PR does not pass the Docker packaging job, but the Docker image (as well as the Debian package, probably) is already broken:

```
$ docker run --rm runtimeverificationinc/kevm:ubuntu-jammy-1.0.291 kevm kompile
/usr/bin/kevm: line 11: kevm-pyk: command not found
$ docker run --rm runtimeverificationinc/kevm:ubuntu-jammy-1.0.291 kevm foundry-kompile
/usr/bin/kevm: line 10: kontrol: command not found
```

@ehildenb, is it an option to bypass this step and fix it once project structure settles a bit? (I.e. `kontrol` is extracted, the Python package is moved to the top level, etc.)

tothtamas28:
@lucasmt, @palinatolmach, @PetarMax, please check if these changes are compatible with the way you install, update and interact with the tool (especially during engagements). Let me know if you run into any difficulties. Thank you!

tothtamas28:
Preview of how the `kontrol` project can be organized after this change; https://github.com/runtimeverification/kontrol/pull/26

kevm-pyk/src/tests/integration/test_prove.py
```diff
@@ -236,7 +236,8 @@ def test_pyk_prove(
         exec_prove(
             spec_file=spec_file,
             definition_dir=target.definition_dir,
-            includes=[str(config.INCLUDE_DIR)] + target.includes,  # TODO are target.includes required?
+            includes=[str(include_dir) for include_dir in config.INCLUDE_DIRS]
+            + target.includes,  # TODO are target.includes required?
```
ehildenb, line 240: We can probably remove this `TODO`.

kevm-pyk/src/kevm_pyk/dist.py
```diff
@@ -0,0 +1,340 @@
...
+def _dist_dir() -> Path:
+    dist_dir_env = os.getenv('KEVM_DIST_DIR')
+    if dist_dir_env:
+        return Path(dist_dir_env).resolve()
+
+    digest = hash_str({'module-dir': config.MODULE_DIR})[:7]
+    return xdg_cache_home() / f'evm-semantics-{digest}'
+
+
+DIST_DIR: Final = _dist_dir()
```
ehildenb, line 46: Is this just a shim for Nix install, so it can set `KEVM_DIST_DIR`? Maybe it would be good to have a comment here about it. At first I thought we were requiring to set `KEVM_DIST_DIR` every time we called this script, but it looks like it's just for Nix.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -511,15 +512,15 @@ def exec_kast(
...
     save_directory: Path | None = None,
     **kwargs: Any,
 ) -> None:
     if target is None:
-        target = KompileTarget.LLVM
+        target = DistTarget.LLVM
 
     _ignore_arg(kwargs, 'definition_dir', f'--definition: {kwargs["definition_dir"]}')
-    kevm = KEVM(target.definition_dir, use_directory=save_directory)
+    kevm = KEVM(target.check(), use_directory=save_directory)
```
ehildenb, line 523: I am not sure how I feel about `_.check()` as the name, because it really just checks if the path exists. Is the intention to eventually check the integrity/up-to-date of the build as well, and roll that into this `_.check()` call?

ehildenb (approved):

ehildenb:
Maybe we should make an issue to factor out the `kevm-dist` functionality into a module in `pyk` instead. It may make sense as an extension to `kbuild`, so that many of the `kevm-dist` commands can just become `kbuild ...` commands with arguments? Either way, a lot of the functionality there seems generic.

kevm-pyk/src/kevm_pyk/__main__.py
tothtamas28: Good point, renamed to `get`.

> Is the intention to eventually check the integrity/up-to-date of the build as well

Yes, this should be done (ideally by the generic `kbuild` feature).

kevm-pyk/src/kevm_pyk/dist.py
tothtamas28: Done.

kevm-pyk/src/tests/integration/test_prove.py
tothtamas28: Done.

tothtamas28:
> Maybe we should make an issue to factor out the `kevm-dist` functionality into a module in `pyk` instead. It may make sense as an extension to `kbuild`, so that many of the `kevm-dist` commands can just become `kbuild ...` commands with arguments?

Definitely, this functionality should be upstreamed to `kbuild` (it is actually very similar to the original `kbuild` repository feature that just got dropped): https://github.com/runtimeverification/pyk/issues/655.

We just need to figure out a way to represent these targets as `kbuild` targets. I think it would make sense to generalize / extend `kbuild.toml` to `kbuild.py` (similarly to how `setuptools` works with `pyproject.toml` and `setup.py`): https://github.com/runtimeverification/pyk/issues/585.

goodlyrottenapple:
> We just need to figure out a way to represent these targets as `kbuild` targets. I think it would make sense to generalize / extend `kbuild.toml` to `kbuild.py` (similarly to how `setuptools` works with `pyproject.toml` and `setup.py`): [runtimeverification/pyk#585](https://github.com/runtimeverification/pyk/issues/585).

I think we need to consider how kbuild is meant to interact with kup/nix. If the goal is to only write a `kbuild.toml` file which nix can parse and build a kup package with little to no custom nix code, then changing to `kbuild.py` might make this harder. However, I am still unclear as to how these two tools should interact and what the UX priorities for the new kbuild based setup should be.

tothtamas28:
> I think we need to consider how kbuild is meant to interact with kup/nix

I agree. How I see it, Nix is a way to package and distribute K projects (i.e. K source files, kompiled K definitions and the Python scripts written to interact with them), and `kbuild` can be used to produce the definitions during the Nix build phase. `kbuild` should accomodate the Nix way of distribution, but it should also enable distribution without Nix.

> If the goal is to only write a `kbuild.toml` file which nix can parse and build a kup package with little to no custom nix code, then changing to `kbuild.py` might make this harder.

In my opinion, if we go with a single `kbuild.toml` configuration file, in order to handle semantics like `evm-semantics` it will eventually become too complex. I think using Python hooks is a better way to make the build process more flexible.

But whatever the input for `kbuild` is, we can add a command to interface it for Nix, i.e. to process all the inputs and produce a JSON descriptor with the data the Nix pipeline requires.

> However, I am still unclear as to how these two tools should interact and what the UX priorities for the new kbuild based setup should be.

I'd like to keep `kbuild` simple. In particular, use it only to describe when and how `kompile` should be called, and use Nix and Python for proper packaging and distribution.

PR #2069, opened 2023-09-14
Use an llvm backend library in `exec_prove` with `--use-booster`

jberthold:
Set `llvm_definition_dir` in `exec_prove` when `use_booster == True`.
Otherwise we might run with booster but will fall back to `kore-rpc` on every concrete term that needs hooks.

With this change, test `address00-spec.k` runs in 76 sec with `--use-booster`, while it takes 110 sec without (This includes the compilation time for the verification module, the actual speedup for running the proof is higher).


jberthold:
The activated LLVM backend causes issues in proof `mcd/dai-symbol-pass-spec.k`, the function `#asByteStackInWidthaux` appears to be missing. I will investigate.

ehildenb:
Wow, good find!

jberthold:
I tried to remove the `[symbolic]` attributes from the verification modules in `tests/specs/erc20` and `tests/specs/mcd` but that causes compilation issues because `include/kframework/lemmas/lemmas.k` is `[symbolic]`, too, and its definitions are missing when compiling the now-not-symbolic `mcd/verification.k`.
Not sure about the best way forward here, probably we need to refactor more module headers, or craft a dedicated syntax module for llvm-targeted compilation.

ehildenb:
There are  some more changes on [this PR](https://github.com/runtimeverification/evm-semantics/pull/2067), which may help this one to pass. In particular:

- https://github.com/runtimeverification/evm-semantics/pull/2067/commits/ec44f50c7f04f686d5a1ed3d1044582ba5c074bc
- https://github.com/runtimeverification/evm-semantics/pull/2067/commits/b7466e5b6b41d70934f676c8849b0a79c652c8d6
- https://github.com/runtimeverification/evm-semantics/pull/2067/commits/928afb5878c4270293c86ff1d6a0e673ba78b6d0
- https://github.com/runtimeverification/evm-semantics/pull/2067/commits/83ecad6db386a543699b0f88d8bf320002f35383
- https://github.com/runtimeverification/evm-semantics/pull/2067/commits/ee6dad0cb531f0abed6010a5c09ba926a1922035

ehildenb:
This looks ready to merge to me, any other blockers? The last require check is removed here: https://github.com/runtimeverification/evm-semantics/pull/2076

ehildenb (approved):

PR #2071, opened 2023-09-15
Fixes to ABI type handling on signed integers

h0nzZik:
- `#range` is now alias instead of a macro
-  `#rangeSInt`: use <= for upper bound
-  `#getValue(#int*(X))` chops X (fixes a bug introduced in https://github.com/runtimeverification/evm-semantics/pull/1883


tests/specs/functional/lemmas-spec.k
```diff
@@ -288,6 +288,9 @@ module LEMMAS-SPEC
 
     claim <k> runLemma ( #bufStrict( #ceil32(LEN) -Int LEN, 0 ) ) => doneLemma( #buf( #ceil32(LEN) -Int LEN, 0 ) ) ... </k> requires 0 <=Int LEN
 
+    claim [bufstrict-simplify]:
+          <k> runLemma ( #bufStrict(32, #getValue(#int256(X))) ) => doneLemma ( #buf(32, X) ) ... </k> requires #rangeSInt(256, X)
```
ehildenb, line 292:
```suggestion
          <k> runLemma ( #bufStrict(32, #getValue(#int256(X))) ) => doneLemma ( #buf(32, chop(X)) ) ... </k> requires #rangeSInt(256, X)
```

I think this should be the correct simplification now, right?

ehildenb:
@anvacaru this bug probably should have been caught in code-review, FYI!

ehildenb (approved):

PR #2079, opened 2023-09-19
Use safe map lookups

RaoulSchaffranek:
When using the bracket notation to look up a value in a dictionary, Python raises an exception if the key is not defined. This PR uses the safer `dct.get(key, default)` function to look up values.


ehildenb:
These keys should always be present in the namespace passed to these functions. Did you run into an error where they weren't, or was this suggested by a codestyle checker?

Also may just consider removing these calls to `_ignore_args` altogether instead.

RaoulSchaffranek:
@ehildenb I am now calling this function directly from the Symbolic Debugger, and I'm still figuring out how to call it correctly. Chances are I am doing something wrong. But if these keyword args MUST be present, why have this `_ignore_args` call at the beginning?

ehildenb:
They must be present if you use the argeparse mechanism (via kwargs), but if you're calling thefunction directly, then they won't be present.

I would recommend just removing these `_ignore_args` lines then.

kevm-pyk/src/kontrolx/__main__.py
```diff
@@ -48,14 +48,6 @@
...
 
-
-def _ignore_arg(args: dict[str, Any], arg: str, cli_option: str) -> None:
-    if arg in args:
-        if args[arg] is not None:
-            _LOGGER.warning(f'Ignoring command-line option: {cli_option}')
-        args.pop(arg)
-
-
 def main() -> None:
```
palinatolmach, line 51:
```suggestion

def main() -> None:
```

palinatolmach (approved):

ehildenb:
Calling `make format` in the `kevm-pyk` subdirectory will cause it to automatically reformat all the python code according to our style guide, FYI. Calling `make check` will run our styleguide and report errors, as well as type-checking errors, and run the unit tests we have (very fast, <10s).

anvacaru:
Closing this since the `kontrol` logic has been factored out in its own repo [here](https://github.com/runtimeverification/kontrol). @RaoulSchaffranek are these changes still required? If so, I can open a new PR in the kontrol repo.

PR #2084, opened 2023-09-22
DRAFT: Optimize symbolic execution using depth-based call graphs

anvacaru:
- [x] Extend `Contract.Method` class to contain the list of the external functions that the current method calls.


ehildenb:
There are several Solidity call-graph generators that exist already, maybe we could use one of those instead?

Example:

- https://github.com/crytic/slither (written in Python, so can be included as a library)
- https://github.com/Consensys/surya (written in nodejs, so can be called as an executable)

Have we tried those ones?

Also, since this feature will take a while to imlpement, let's implement it against the new Kontrol repo here, to avoid having more features we have to migrate over: https://github.com/runtimeverification/kontrol

anvacaru:
> Have we tried those ones?

@ehildenb  No, I haven't tried those yet. We already have classes for `Contract` and `Method` in `solc_to_k.py` and we use them in `foundry-prove`. The `Method` class holds the method AST, so I just extended the class with a list of external calls identified in the said AST. So instead of having a single call-graph, we would parse the external calls of the target functions.

> let's implement it against the new Kontrol repo here

Sounds good. I will migrate the changes there.

anvacaru:
Moving this to https://github.com/runtimeverification/kontrol/pull/62.

PR #2086, opened 2023-09-25
Add '--smt-tactic <s-expression>' command-line option

geo2a:
~Blocked on: https://github.com/runtimeverification/pyk/pull/660~
~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/2087~

Setting `--smt-tactic (check-sat-using smt)` is not possible as part of `kore-rpc-command`, as the command is split by the whitespace character and it results in treating "(check-sat-using" and "smt)" as separate arguments, which causes an invalid call to `kore-rpc-booster`/`kore-rpc` down the line.

Separating the tactic into a separate argument solves this problem.


ehildenb:
The relevant portion of thi schange should also be applide to the runtimeverification/kontrol repo: https://github.com/runtimeverification/kontrol

PR #2096, opened 2023-09-27
Support for init bytecode

nwatson22:
#2020 is mostly moved to https://github.com/runtimeverification/kontrol/pull/54, but there are a few changes still required for kevm for this change to kontrol. Adds `#initBytecode` syntax and method to generate kast for it.


palinatolmach (approved):

kevm-pyk/src/kevm_pyk/kproj/plugin
anvacaru, line 1: Is there a reason for downgrading the plugin submodule?

Issue #2088, opened 2023-09-25
Re-executing a kclaim style proof does not correctly detect out of date proof


ehildenb:
Currently, if I run a KEVM proof with `--save-directory`, and it passes, then I modify the proof in a way that causes it to be failing, rerunning it does not report failure. Instead, it loads the proof from disk, and sees that it's passing.

yale-vinson:
@nwatson22 what is the relative priority of this? I am assuming that [#2009](https://github.com/runtimeverification/evm-semantics/issues/2009) had priority, but I want to make sure things are aligned across Kontrol.

yale-vinson:
Closing [evm-semantics PR #2099 ](https://github.com/runtimeverification/evm-semantics/pull/2099)closes the first bullet point of this one. The others will be addressed once this PR closes.

palinatolmach:
@anvacaru @PetarMax do you think these changes might help order proofs for compositional symbolic execution? Or do you have another solution in mind?

yale-vinson:
Issue blocked by Booster #383 [(https://github.com/runtimeverification/hs-backend-booster/issues/383)]

palinatolmach:
Moved to backlog with Medium High priority, since we haven't got to work on it since November, and it doesn't seem to be a priority now. Please bring it back if I'm wrong!

PR #2099, opened 2023-09-30
Rerun claims when the claim body or those of dependent claims changes

nwatson22:
Part of #2088

- Adds `KClaimJob` class, which wraps `KClaim` and computes a digest associated with each claim (to change whenever the claim changes on disk). Digest also depends on the digest of all dependent claims.
- Adds `init_claim_jobs` to set up the `KClaimJobs` so they can easily access the digests of their dependencies to compute their own digest.


kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -201,6 +202,69 @@ def wrap_process_pool(workers: int) -> Iterator[ZeroProcessPool | ProcessPool]:
...
+
+def init_claim_jobs(spec_module_name: str, claims: list[KClaim]) -> list[KClaimJob]:
+    labels_to_claims = {claim.label: claim for claim in claims}
+    labels_to_claim_jobs: dict[str, KClaimJob] = {}
+
+    def get_or_load_claim_job(claim_label: str) -> KClaimJob:
+        if claim_label not in labels_to_claim_jobs:
+            if claim_label in labels_to_claims:
+                claim = labels_to_claims[claim_label]
+            if f'{spec_module_name}.{claim_label}' in labels_to_claims:
```
anvacaru, line 258: Is there any chance for `labels_to_claims` to not contain either of `claim_label` and `f'{spec_module_name}.{claim_label}'`?

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -201,6 +202,69 @@ def wrap_process_pool(workers: int) -> Iterator[ZeroProcessPool | ProcessPool]:
...
+        return json.JSONEncoder.default(self, obj)
+
+
+@dataclass
+class KClaimJob:
+    claim: KClaim
+    dependencies: list[KClaimJob]
+
+    @property
+    def digest(self) -> str:
```
anvacaru, line 218: The `digest` property computes the hash of dependencies every time it's accessed. If the dependencies don't change frequently, maybe you should cache them.

nwatson22:
I've realized there is an awkwardness when using this functionality. When a claim doesn't have an explicit label, we use the `UNIQUE_ID` attribute, which I'm assuming is some sort of hash of the claim body, and I don't have a better way to track claim identity. So if the claim is not labeled and changes, it will simply be treated as a new claim, and the system I implemented here will not kick in.

kevm-pyk/src/kevm_pyk/__main__.py
nwatson22: It is now a cached property, and to make sure it doesn't change, I've made the class a frozen dataclass.

kevm-pyk/src/kevm_pyk/__main__.py
nwatson22: Currently it could be anything that someone tries to add as a `[depends(...)]` attribute. The current setup assumes this is either the claim label itself or the claim label with the module name at the beginning followed by a dot. I tried to copy this behavior from [how a similar thing with dependencies is handled in pyk](https://github.com/runtimeverification/pyk/blob/d263b25ee713b51eb1189723581d800784c1b801/src/pyk/proof/reachability.py#L250). I've now added an exception in the case where it doesn't match one of these.

nwatson22:
Change this to not use digests, but just always use the unique id and store under directory that has the same name as the user-provided label

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -200,6 +202,71 @@ def wrap_process_pool(workers: int) -> Iterator[ZeroProcessPool | ProcessPool]:
...
+        deps_digest = ''.join([dep.digest for dep in self.dependencies])
+        claim_hash = hash_str(json.dumps(self.claim.to_dict(), sort_keys=True, cls=JSONEncoder))
+        return hash_str(f'{claim_hash}{deps_digest}')
+
+    def up_to_date(self, digest_file: Path | None) -> bool:
+        if digest_file is None:
+            return False
+        if not digest_file.exists():
+            return False
```
anvacaru, line 227: Could you merge these two together into something like this? (not sure if `flake8` accepts this).

```suggestion
        if not isinstance(digest_file, Path) or not digest_file.exists(): 
            return False
```

anvacaru (approved):
Looks good to me! 👍  Left a single comment, but feel free to ignore it.

PR #2106, opened 2023-10-05
Add a new entry point for the debugger

RaoulSchaffranek:
Previously, the Solidity Debugger relied on syscalls to call Kontrol. In our continuous effort to optimize, we're transitioning to embed Kontrol directly as a Python library. To facilitate this transition, we've introduced a new entry point for the debugger to access the Prover instances. This PR introduces the `kevm_debug` entry point, which mirrors `kevm_prove` in functionality but returns the prover instance directly rather than the final result.


kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -168,6 +168,61 @@ def kevm_prove(
...
+                'EVM.staticcall',
+                'EVM.create',
+                'EVM.create2',
+                'FOUNDRY.foundry.call',
+                'EVM.end',
+                'EVM.return.exception',
+                'EVM.return.revert',
+                'EVM.return.success',
+            ]
+        )
```
anvacaru, line 204: I am not entirely sure that duplicating this code is a good idea. Perhaps you could extract this into its own function and use it both here and in `kevm_prove`. 

Something along the lines of :
```python
def generate_rules(break_every_step: bool, break_on_jumpi: bool, break_on_calls: bool):
    terminal_rules = ['EVM.halt']
    cut_point_rules = []
    if break_every_step:
        terminal_rules.append('EVM.step')
    if break_on_jumpi:
        cut_point_rules.extend(['EVM.jumpi.true', 'EVM.jumpi.false'])
    if break_on_calls:
        cut_point_rules.extend(
            [
                'EVM.call',
                'EVM.callcode',
                'EVM.delegatecall',
                'EVM.staticcall',
                'EVM.create',
                'EVM.create2',
                'FOUNDRY.foundry.call',
                'EVM.end',
                'EVM.return.exception',
                'EVM.return.revert',
                'EVM.return.success',
            ]
        )
    return terminal_rules, cut_point_rules
```

anvacaru (approved):

kevm-pyk/src/kevm_pyk/utils.py
ehildenb: Is this code being duplicated? It looks like it's being moved to here. Maybe this comment is resolved?

ehildenb (approved):

kevm-pyk/src/kevm_pyk/utils.py
RaoulSchaffranek: Yes, I followed Andrei's recommendation and refactored this part.

RaoulSchaffranek:
I'm closing this because the logic is better implemented in the debugger.

PR #2112, opened 2023-10-12
WIP: Check valid jump dests on demand

RaoulSchaffranek:
Don't merge. I'm just profiling.


palinatolmach:
I've brought the PR up to date since constructors with symbolic parameters are needed for the engagement: https://github.com/runtimeverification/kontrol/issues/451.

Some considerations regarding JUMPDEST calculation from a discussion with @ehildenb in October:
- precomputing valid jump dests ahead of time was actually done to speed up the verification — the corresponding claims are fully concrete and should be sent to the concrete LLVM backend which is supposed to be fast; so two things to check here are 1) if we're using the booster, and 2) if this ahead-of-time validity checking is indeed using LLVM backend and, if not, why
- not pre-computing JUMPDESTs might slow the execution down if there’s a lot of loop unrolling, so it’d be good to include some loop-containing code in profiling
- conformance tests can also help measure the potential speedup and determine if we’re indeed executing the relevant parts concretely
- one reason why the execution might take a long time is that a set of valid JUMPDESTs is a large tree (and the bigger the bytecode, the larger the tree gets, probably); so, in the Haskell Backend, every simplification includes walking this large tree, and one of the potential optimizations is perhaps saving some relevant information for the backend instead of having it walk through the tree every time

ehildenb:
Subsumed by: https://github.com/runtimeverification/evm-semantics/pull/2471

Re-open if still relevant.

PR #2118, opened 2023-10-18
Add infrastructure to KEVM for Maude Backend

dwightguth:
This PR modifies the kevm_pyk package so it is capable of building and using the maude backend when sending RPC requests. This code is still unused; the kontrol codebase needs a separate PR in order to start the maude backend and make use of it when invoking kontrol.


kevm-pyk/src/kevm_pyk/kompile.py
```diff
@@ -106,6 +107,28 @@ def kevm_kompile(
                 )
                 return kompile(output_dir=output_dir, debug=debug, verbose=verbose)
 
+            case KompileTarget.MAUDE:
+                kompile_maude = MaudeKompile(
+                    base_args=base_args,
+                )
+                kompile_haskell = HaskellKompile(base_args=base_args)
```
anvacaru, line 114: I'm unfamiliar with how the Maude backend works, but from this, I understand that it depends on the Haskell backend similar to what the booster backend does.

dwightguth: For the time being, the Maude backend does not implement `get-model` or `check-implication`, so it still requires the Haskell backend to implement those endpoints. This will eventually change, but that's not needed for a minimum viable product, so it has not been the focus.

ehildenb (approved):

dwightguth:
Okay, this is a problem. The test suite is so flaky that I've been spending days trying to get this merged. Days.

ehildenb:
I agree. Maybe the frontend could be made to not launch so many Z3s? That was an initial guess  at a solution. Other things are being implemented, like better retrying in haskell backend on Z3 failures. We've also been reconfiguring parallelism and runner memory allocation.

Investigation welcome!

dwightguth:
I mean, if we can't solve this immediately, what we really need to do is to implement a policy that only the oldest approved PR in the repo is allowed to be merged. Otherwise we end up with some PRs getting starved and never getting merged, ever.

PR #2126, opened 2023-10-20
Implement dependency handling for `kdist`

tothtamas28:
New feature:

* A target now can specify dependencies: `Target.deps()`
* Before building a target, `kdist` builds all its dependencies
* Dependency paths are passed to `Target.build()`

Refactorings:

* Eliminate dependency of `kompile` on `kdist` by requiring `plugin_dir` to be passed to `kevm_kompile`
* Extract module `kdist.api` for plugins to import in order to eliminate potential circular dependency on `_kdist`
* Move `build` logic from `__main__` to `_kdist`, reimplement
* Eliminate hard-coded dependency handling (use `graphlib` instead)


kevm-pyk/src/kevm_pyk/kdist/plugin.py
```diff
@@ -79,6 +85,7 @@ def _build_env(self) -> Iterator[Path]:
             'syntax_module': 'ETHEREUM-SIMULATION',
             'optimization': 2,
         },
+        deps=('plugin',),
```
ehildenb, line 88: I believe that haskell target may also require `plugin`? If the booster is enabled?

ehildenb (approved):

kevm-pyk/src/kevm_pyk/kdist/plugin.py
tothtamas28: This entry defines kompilation of `ETHEREUM-SIMULATION` via  `kevm_kompile(target=KompileTarget.HASKELL, ...)`, for which `plugin` is not needed. It is indeed necessary for `kevm_kompile(target=KompileTarget.HASKELL_BOOSTER, ...)` though.

PR #2128, opened 2023-10-21
Use `Bytes` to represent byte strings instead of `String`

Scott-Guest:
runtimeverification/blockchain-k-plugin#163 is going to be merged shortly, changing all of the `KRYPTO` hooks to use `Bytes` rather than `String`. 

To make that change a bit smoother, this PR temporarily adds wrapper functions around the existing `String`-based `KRYPTO` hooks to match the upcoming `Bytes`-based interface, then updates the semantics to use these wrapped versions everywhere. When runtimeverification/blockchain-k-plugin#163 is merged, we can then just textually find-and-replace every instance of the wrappers.

More pervasively, I also made changes to ensure that we use `Bytes` rather than `String` for things that are conceptually raw byte arrays (e.g. `#rlpEncode` should operate on and return `Bytes`).


kevm-pyk/src/kevm_pyk/kproj/evm-semantics/serialization.md
```diff
@@ -260,90 +259,91 @@ Encoding
...
 
     syntax JSON ::= Bytes
     rule #rlpEncode( [ J:JSONs ] ) => #rlpEncodeLength( #rlpEncode(J, .StringBuffer) , 192 )
 
-    rule #rlpEncode( .JSONs                   , BUF ) => StringBuffer2String(BUF)
-    rule #rlpEncode( (J:Int,       REST:JSONs), BUF ) => #rlpEncode(REST, BUF +String #rlpEncodeInt(J)   )
-    rule #rlpEncode( (J:String,    REST:JSONs), BUF ) => #rlpEncode(REST, BUF +String #rlpEncodeString(J))
-    rule #rlpEncode( (J:Bytes,     REST:JSONs), BUF ) => #rlpEncode(REST, BUF +String #rlpEncodeBytes(J) )
-    rule #rlpEncode( ([ J ],       REST:JSONs), BUF ) => #rlpEncode(REST, BUF +String #rlpEncode([ J ])  )
+    rule #rlpEncode( .JSONs                   , BUF ) => String2Bytes(StringBuffer2String(BUF))
```
Scott-Guest, line 292: The conversions here are necessary so that we can use StringBuffer, avoiding quadratic complexity when concatenating.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/serialization.md
```diff
@@ -260,90 +259,91 @@ Encoding
...
+    rule #rlpEncode( (J:Int,       REST:JSONs), BUF ) => #rlpEncode(REST, BUF +String Bytes2String(#rlpEncodeInt(J)   ))
+    rule #rlpEncode( (J:String,    REST:JSONs), BUF ) => #rlpEncode(REST, BUF +String Bytes2String(#rlpEncodeString(J)))
+    rule #rlpEncode( (J:Bytes,     REST:JSONs), BUF ) => #rlpEncode(REST, BUF +String Bytes2String(#rlpEncodeBytes(J) ))
+    rule #rlpEncode( ([ J ],       REST:JSONs), BUF ) => #rlpEncode(REST, BUF +String Bytes2String(#rlpEncode([ J ])  ))
```
Scott-Guest, line 296: There is no `BytesBuffer`, so we need to convert here to use `StringBuffer` to avoid quadratic complexity.

PR #2130, opened 2023-10-23
Refactor the way pranks are injected into calls

anvacaru:
Fixes https://github.com/runtimeverification/kontrol/issues/120.


hjorthjort (commented):
Won't the rules as they stand lead to an infinite loop?
If you do `CALL ...`, that becomes `#injectPrank ~> CALL ...`. And there is no check on whether the prank has already been injected once, so `#injectPrank` will just be injected again.

Otherwise I think the design makes sense.

anvacaru:
Closing this because `foundry.md` was moved to kontrol in the meanwhile.

PR #2103, opened 2023-10-04
Remove kontrolx and associated tests/files

ehildenb:
~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/2135~
~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/2152~

Fixes #2113

This removes the Kontrolx modules completely.

- Removes the code in `kevm-pyk/src/kontrolx`
- Adjusts the tests in `kevm-pyk/src/tests/` to remove all references to `kontrolx`.
- Updates the Nix flake and lock file and Docker image to not include Foundry/solc anymore.
- Removes the kontrolx CLI arguments that existed in KEVMCLIArgs.
- Removes the reference to a `Foundry.` rule in `cut_point_rules`.

PR #2135, opened 2023-10-24
Remove testing infrastructure related to kontrolx

ehildenb:
This is factored out of https://github.com/runtimeverification/evm-semantics/pull/2103. `solc-to-k` is a Kontrol utility, and here we were testing it still, and the kontrolx proof harnesses were still present. In particular:

- The generated K files that were being produced by `solc_to_k` before are now committed directly.
- The various tests in `kevm-pyk/src/tests/integration/` related to kontrolx are deleted, and the associated test data too.
- The git submodule `forge-std` for the foundry tests is removed.
- The `test_prove` module is simplified now that no K code gen is required for those tests.


.gitmodules
```diff
@@ -14,6 +14,3 @@
 [submodule "web/k-web-theme"]
 	path = web/k-web-theme
 	url = https://github.com/runtimeverification/k-web-theme
-[submodule "tests/foundry/lib/forge-std"]
```
anvacaru, line 17: You might also need to remove the `tests/foundry/lib/forge-std` submodule directly from the repo.

nwatson22 (approved):

PR #2144, opened 2023-10-30
Simplify and compact kompilation for Haskell vs Booster

ehildenb:
We currently have two targets `HASKELL` and `HASKELL_BOOSTER` for kompilation, and really the differences between them are also that the main module changes. Instead here, we:

- Always kompile the LLVM backend definition in place when building haskell backend, so that at runtime the user can specificy whether they want the booster or not (by changing the RPC server).
- Explicitly list the `plugin` as a dependency for the two haskell builds, since they now build the LLVM backend.
- Remove the explicit list of `plugin` as a dependency in the CI job.
- Change the names of the targets to `HASKELL` and `HASKELL_STANDALONE` to reflect that it's more about which is the main module being used.
- Simplify the kompilation logic used in the tests based on this change.


kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -600,16 +599,13 @@ def exec_run(
     schedule: str,
     mode: str,
     chainid: int,
-    target: str | None = None,
+    target: KompileTarget,
```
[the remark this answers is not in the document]
ehildenb: `KompileTarget` is now just used for `kevm run` and `kevm kast`, because it allows specifying which KEVM precompiled definition to use for those commands. It's no longer used for kompiling specifications (`kevm kompile ...` which became `kevm kompile-spec ...`).

kevm-pyk/src/kevm_pyk/cli.py
```diff
@@ -37,7 +38,7 @@ class KEVMCLIArgs(KCLIArgs):
     @cached_property
     def target_args(self) -> ArgumentParser:
         args = ArgumentParser(add_help=False)
-        args.add_argument('--target', choices=['llvm', 'haskell', 'haskell-standalone', 'foundry'])
+        args.add_argument('--target', type=KompileTarget, default=KompileTarget.LLVM, choices=list(KompileTarget))
```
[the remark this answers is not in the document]
ehildenb: This is now used not for kompiling (`kevm kompile ...` which is now `kevm kompile-spec ...`), but just for `kevm kast ...` and `kevm run ...`, for selecting which prebuilt backend to use for executing those commands. In that case, it doesn't make sense to use `__TARGET__`, because doing `kevm kast --target plugin ...` wouldn't make sense.

kevm-pyk/src/kevm_pyk/kompile.py
```diff
@@ -64,7 +63,7 @@ def kevm_kompile(
     if llvm_library is None:
         llvm_library = output_dir / 'llvm-library'
 
-    if target in [KompileTarget.LLVM, KompileTarget.HASKELL_BOOSTER] and not plugin_dir:
+    if target == KompileTarget.LLVM and not plugin_dir:
```
tothtamas28, line 66: `plugin_dir` should now become a mandatory parameter, and checks and assertions that it is not `None` removed.

kevm-pyk/src/kevm_pyk/cli.py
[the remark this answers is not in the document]
tothtamas28: `KompileTarget` values correspond to templates for kompilation (with values `haskell` and `llvm`). Here, we want to refer to one of the pre-built definitions (with values `haskell`, `haskell-standalone` and `llvm`.

kevm-pyk/src/kevm_pyk/kdist/plugin.py
```diff
@@ -32,7 +32,6 @@ def build(self, output_dir: Path, deps: dict[str, Path], args: dict[str, Any]) -
             output_dir=output_dir,
             enable_llvm_debug=enable_llvm_debug,
             verbose=verbose,
-            plugin_dir=deps.get('plugin'),
```
tothtamas28, line 35: Letting the framework inject the path for `plugin` has a few advantages:
* it allows for flexibility in target location, e.g. in case we want to add a local build feature
* without it, there's a conceptual circular dependency: `kompile -> kdist -> plugin -> kompile`

kevm-pyk/src/kevm_pyk/cli.py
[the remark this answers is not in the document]
ehildenb: Done.

kevm-pyk/src/kevm_pyk/__main__.py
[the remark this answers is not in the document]
ehildenb: Back to `str` which is among hardcoded values.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -637,15 +633,12 @@ def exec_kast(
...
```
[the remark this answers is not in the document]
ehildenb: Back to `str` among hardcoded values.

kevm-pyk/src/kevm_pyk/kompile.py
ehildenb: Done.

kevm-pyk/src/kevm_pyk/kdist/plugin.py
ehildenb: The disadvantage is that `run_kompile` then becomes KEVM specific, instead of semantics generic. I could move `run_kompile` into a different module if it would break the dependency?

ehildenb: Though I guess the calculation of `ccopts` in `run_kompile` is semantics specific...

ehildenb: I've made it so that `ccopts` is passed into all calls to `run_kompile` instead, so that `run_kompile` is semantics-generic, and `plugin_dir` is calculated outside that function.

tothtamas28: `run_kompile` (originally `kevm_kompile`) was meant to be specific to `evm-semantics`. It's a thin wrapper around  the semantics-generic `kompile` that consolidates kompilation logic by computing some arguments (like `md_selector` and `ccopts`) and passing them along.

I agree that semantic-specific logic should be kept to the minimum, but we have to calculate these arguments somewhere. I prefer a single location, which this function was intended to be.

ehildenb: Yeah, but there is no need now, the amount of semantics specific stuff is so small (just the location of plugin and thus ccopts), that this `run_kompile` should be upstreamed into pyk even.

I can open a PR doing that, and then refactor this to depend on that one.

tothtamas28: What's the logic that you'd like to upstream? Kompilation for the Booster? I don't have anything against that, but I'd still like to have a function that does what this function is aimed to do so that we don't have to pass the same `md_selector`, `ccopts`, etc. in:
* `kevm-pyk` / `kevm_pyk.__main__`
* `kevm-pyk` / `kevm_pyk.kdist.plugin`
* `kevm-pyk` / `tests.integration.test_prove.py`
* `kontrol` / `kontrol.kompile`
* `kontrol` / `tests.integration.test_solc_to_k`

tothtamas28 (approved):

PR #2094, opened 2023-09-27
Proof queue

h0nzZik:
Subsumed by https://github.com/runtimeverification/evm-semantics/pull/2147/

In this PR we want to add a queue containing proof tasks; a worker, whenever is ready, will read a proof task from the queue and calls the prover on it.

PR #2147, opened 2023-10-31
Proof queue

h0nzZik:
We add a queue containing proof tasks; a worker process, whenever is ready, will read a proof task from the queue and calls the prover on it. This mechanism is disabled when there is only one proof task to perform.

Subsumes https://github.com/runtimeverification/evm-semantics/pull/2094/


Makefile
```diff
@@ -5,7 +5,7 @@ all: poetry
 # --------
 
 PYTHON_BIN   := python3.10
-KEVM_PYK_DIR := ./kevm-pyk
+KEVM_PYK_DIR := $(abspath ./kevm-pyk)
```
ehildenb, line 8: Why is this change needed? Also, why is `poetry.lock` updated? These changes should probably be backed out, should only be updated by the auto-updater.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -432,23 +458,95 @@ def _init_and_run_proof(claim_job: KClaimJob) -> tuple[bool, list[str] | None]:
...
+                        )
+
+    def coordinator(env: MyEnvironment, jobs_to_do: list[KClaimJob]) -> list[tuple[bool, list[str] | None]]:
+        id_to_job: dict[str, KClaimJob] = {job.claim.label: job for job in jobs_to_do}
+        remaining_job_labels: set[str] = {job.claim.label for job in jobs_to_do}
+        finished_job_labels: dict[str, tuple[bool, list[str] | None]] = {}
+
+        def get_a_ready_id() -> str | None:
+            for job_label in remaining_job_labels:
+                if set({d.claim.label for d in id_to_job[job_label].dependencies}).issubset(finished_job_labels.keys()):
```
ehildenb, line 485: What if the dependency was finished in a prior job, and we just need to read from disk that it's the case? Will this catch that? It seems here that it will only work if it was proved _in this session_, not if it was proved in a prior session. Is that the case?

Makefile
h0nzZik: My version of Poetry (I think it is 1.6.1) complained about a relative path and this fixed it. I will try reverting `poetry.lock` and see whether it still works.

kevm-pyk/src/kevm_pyk/__main__.py
h0nzZik: I didn't think about that; you are probably right.

ehildenb:
CLosing due to age.

PR #2164, opened 2023-11-08
Inject `ccopts` directly into `kevm_kompile` and fix some tests

ehildenb:
This PR is pulled out of: https://github.com/runtimeverification/evm-semantics/pull/2144

In particular:

- A few proof tests that do not kompile properly with LLVM backend library because of shadowed module imports (due to `symbolic` attribute) are fixed.
- The command `kevm kompile ...` is renamed to `kevm kompile-spec ...`, which is more representative of what it does. It defaults to `--target haskell-booster`, but only accepts `haskell,haskell-booster,maude`.
- The `kevm_kompile` command no longer takes `plugin_dir` as an argument, but instead relies on user to pass in `ccopts` directly, which is all `plugin_dir` was used to calculate. This calculation is done at the call-site now instead, which makes `kevm_kompile` more generic (removes semantics specific component), and barely changes the call-sites.
- The `Makefile` system is adjusted to make sure that `evm-optimizations-spec.md` is correctly updated on changes.
- Adds a test to `package/test-package.sh` that was previously failing of using the booster on a proof. This was failing because we weren't correctly setting the `plugin_dir` (and thus `ccopts`) when using the booster at spec compile time. This was missed because the pytest testing harness did inject this option.


kevm-pyk/src/kevm_pyk/kdist/plugin.py
```diff
@@ -27,12 +27,15 @@ def __init__(self, kompile_args: Mapping[str, Any], *, deps: Iterable[str] | Non
     def build(self, output_dir: Path, deps: dict[str, Path], args: dict[str, Any]) -> None:
         verbose = args.get('verbose', False)
         enable_llvm_debug = args.get('enable_llvm_debug', False)
+        debug_build = args.get('debug_build', False)
+        ccopts = args.get('ccopts', [])
 
+        ccopts = ccopts + lib_ccopts(kdist.get('evm-semantics.plugin'), debug_build=debug_build)
```
tothtamas28, line 33:
```suggestion
        ccopts = ccopts + lib_ccopts(deps.get('evm-semantics.plugin'), debug_build=debug_build)
```

(Of course the `None` case should be handled.)

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -123,6 +128,8 @@ def exec_kompile(
     if debug_build:
         optimization = 0
 
+    ccopts = list(ccopts) + lib_ccopts(kdist.get('evm-semantics.plugin'), debug_build=debug_build)
```
tothtamas28, line 131: There are thee occurrences of the following pattern in this repo (and there will be two in `kontrol` if this is merged):

```python3
ccopts = ccopts + lib_ccopts(plugin_dir)
kevm_kompile(
    ...,
    ccopts=ccopts,
    ...,
)
```

Furhtermore, this is the only way `kevm_kompile` is called.

What's the advantage of pulling out this detail?

ehildenb: Because it brings semantics-specific things (like the fact that there is a plugin, and the location of it) _out_ of `kevm_kompile`, so that the same method can be upstreamed and using by, for example, the MIR semantics.

kevm-pyk/src/kevm_pyk/kdist/plugin.py
ehildenb: Done.

kevm-pyk/src/kevm_pyk/kdist/plugin.py
```diff
@@ -27,12 +27,18 @@ def __init__(self, kompile_args: Mapping[str, Any], *, deps: Iterable[str] | Non
     def build(self, output_dir: Path, deps: dict[str, Path], args: dict[str, Any]) -> None:
         verbose = args.get('verbose', False)
         enable_llvm_debug = args.get('enable_llvm_debug', False)
+        debug_build = args.get('debug_build', False)
+        ccopts = args.get('ccopts', [])
 
+        plugin_dir = deps.get('evm-semantics.plugin')
+        if plugin_dir is None:
+            raise ValueError(f'Cannot build target without dependency evm-semantics.plugin already built: {output_dir}')
```
tothtamas28, line 35: This can be ensured by making `evm-semantics.plugin` a dependency of all `KEVMTarget` instances:

```python3
def deps(self) -> tuple[str]:
    return ('evm-semantics.plugin',)
```

and then in `build` simply:

```python3
plugin_dir = deps['evm-semantics.plugin']
```

kevm-pyk/src/kevm_pyk/__main__.py
tothtamas28: In that case, consider bringing out the semantic independent details instead by extracting a function:

```python3
def kevm_kompile(..., plugin_dir, ccopts, ...):
    ccopts = list(ccopts) + _lib_ccopts(plugin_dir)
    return generic_kompile(..., ccopts, ...)

def generic_kompile(..., ccopts, ...):
    ...
```

kevm-pyk/src/kevm_pyk/kdist/plugin.py
ehildenb: Done.

kevm-pyk/src/kevm_pyk/kdist/plugin.py
```diff
@@ -27,17 +27,20 @@ def __init__(self, kompile_args: Mapping[str, Any], *, deps: Iterable[str] | Non
...
             verbose=verbose,
-            plugin_dir=deps.get('evm-semantics.plugin'),
+            ccopts=ccopts,
+            debug_build=debug_build,
             **self._kompile_args,
         )
 
     def deps(self) -> tuple[str, ...]:
-        return self._deps
+        return tuple(unique(['evm-semantics.plugin'] + list(self._deps)))
```
tothtamas28, line 43: Consider

```suggestion
        return ('evm-semantics.plugin',)
```

and dropping `self._deps` altogether.

tothtamas28 (approved):

kevm-pyk/src/kevm_pyk/kdist/plugin.py
ehildenb: Done.

PR #2167, opened 2023-11-09
Correction: semantics of #setLocalMem

PetarMax:
The `#setLocalMem` auxiliary function was implemented under the assumption that the return value is never longer than `RETWIDTH` bytes. This assumption is not correct. Specifically:

1. the return valus always corresponds to the entire contents of the `<output>` cell;
2. if a call completes successfully, the length of the return value is exactly `RETWIDTH`, and that is reflected in the `requires` clause;
3. Otherwise, the length of the return value is arbitrary.

I have observed all of these behaviours in my recent work on compositional symbolic execution, where I was analysing a call with `RETWIDTH ==Int 32`, which had reverting branches with return value length 0 (`REVERT 0 0`) and 36 (`REVERT 0 36`).

N.B. 1: As this PR corrects the semantics and the changes are minimal, it should not affect the performance in any way, and even if it does, these changes are required.
N.B. 2: Similar changes will have to be made in Kontrol when this PR is upstreamed.


PetarMax:
On closer inspection, this does not appear to be the desired behaviour.

PR #2169, opened 2023-11-10
Fix build instructions in `README.md`

tothtamas28:


anvacaru (approved):
Should we modify the instructions in kontrol as well?

tothtamas28:
> Should we modify the instructions in kontrol as well?

Kontrol docs is up to date with the changes previously missed here.

PR #2182, opened 2023-11-17
Add change detection to `kdist`

tothtamas28:
* Adds methods `context` and `source` to `Target`. From these a build manifest is computed that is used to check for changes between builds.
* Target `evm-semantics.plugin` is rebuilt on changes to the `plugin` submodule. K targets are rebuilt on changes to the package source (including K files), as well as K version change.
* Adds class `TargetCache` that encapsulates loading and accessing targets defined in plugins. (Refactoring)


kevm-pyk/src/kevm_pyk/kdist/_kdist.py
```diff
@@ -50,118 +52,158 @@ def _dist_dir() -> Path:
...
+        if not _valid_id(target):
+            raise ValueError(f'Invalid target identifier: {target!r}')
 
-        res[plugin_name] = _load_targets(module)
+        return TargetId(plugin, target)
 
-    return res
+    @property
+    def fqn(self) -> str:
+        return f'{self.plugin}.{self.target}'
```
ehildenb, line 77: I assume this is an abbreviation for `fully_qualified_name`? Is it typical to use this abbreviation? If not, maybe we could just use the fully qualified name?

kevm-pyk/src/kevm_pyk/kdist/_kdist.py
```diff
@@ -50,118 +52,158 @@ def _dist_dir() -> Path:
...
-    except KeyError:
-        raise ValueError(f'Undefined plugin: {plugin}') from None
+        return res
 
-    if not target in targets:
-        raise ValueError(f'Plugin {plugin} does not define target: {target}')
 
-    return TargetId(plugin, target)
+_DIST_DIR: Final = _dist_dir()
+_CACHE: Final = TargetCache.load()
```
ehildenb, line 188: Does this mean we're loading the target cache every time this process starts up? Is that only for calls to `kevm-dist` or also for calls to `kevm ...`?

ehildenb (approved):

kevm-pyk/src/kevm_pyk/kdist/_kdist.py
tothtamas28: I think it's typical, but good point. I renamed it to `full_name`.

kevm-pyk/src/kevm_pyk/kdist/api.py
```diff
@@ -16,3 +18,24 @@ def build(self, output_dir: Path, deps: dict[str, Path], args: dict[str, Any]) -
 
     def deps(self) -> Iterable[str]:
         return ()
+
+    def source(self) -> Iterable[str | Path]:
+        return ()
```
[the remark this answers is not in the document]
tothtamas28: This method is called by the framework, where `str | Path` is converted to `Path`. The idea is to make implementing the interface as simple as possible.

kevm-pyk/src/kevm_pyk/kdist/_kdist.py
tothtamas28: The cache is populated on the first import to `kdist`. Although there's some extra processing involved, I don't think it's significantly more overhead then importing `plugin` directly.

If it turns out to be a bottleneck, we can separate "producers" like `kdist.build` from "consumers" like `kdist.get` so that code that's only interested in accessing definitions don't have to wait for the cache to be populated, though It's still useful for error handling.

PR #2194, opened 2023-11-28
Ensure #checkCall branch exploration

anvacaru:
@PetarMax reported that branchings on `#checkCall` do not expand further and get stuck with the `<k>` cell looking like :
```k
#end #if VALUE >Int BAL #then EVMC_BALANCE_UNDERFLOW #else #if CD >=Int 1024 #then EVMC_CALL_DEPTH_EXCEEDED #else EVMC_NONCE_EXCEEDED #fi #fi
```
 The `#if` statements cannot be evaluated further.

In this PR I refactor the `#checkCall` rules to rewrite into individual checks for balance underflow, exceeded call depth, and out-of-bounds nonce to ensure all branchings are explored.


PetarMax (changes requested):
Minor changes to the descriptions, otherwise appears ok.

PetarMax (approved):

PR #2195, opened 2023-11-28
Extract class `KDist`

tothtamas28:
* Extract class `KDist` with methods `which`, `clean`, `get`, and `build`
* Move target cache to new module `kdist._cache`
* Lazy load target cache on first access
* `KDist.get` no longer requires the target cache


kevm-pyk/src/kevm_pyk/config.py
```diff
@@ -20,3 +22,6 @@
 
 
 NIX_LIBS: Final = os.getenv('NIX_LIBS')
+
+KEVM_DIST_DIR: Final = os.getenv('KEVM_DIST_DIR')  # Used by Nix flake to set the output
+kdist: Final = KDist(KEVM_DIST_DIR)
```
ehildenb, line 27: And if this is not set in environment, what happens?

ehildenb: Looks like it's handled in the `KDist` constructor.

ehildenb (approved):

PR #2227, opened 2023-12-19
Upstreaming lemmas for symbolic calldata reasoning

palinatolmach:
This PR adds lemmas that are useful for reasoning about symbolic calldata.


ehildenb:
Closing due to age.

PR #2229, opened 2023-12-19
De-duplication of use-gas-related rules

### LINKED ISSUE #2228: De-duplicate rules introduced as part of the no-gas PR
Some of the rules introduced in the [PR that allows gas to not be calculated](https://github.com/runtimeverification/evm-semantics/pull/2207) are currently duplicated because the booster backend cannot handle `#if` constructs efficiently.

Once this has been made possible, they should be de-duplicated.

PetarMax:
This PR de-duplicates some of the rules related to handling the use of gas.

~Blocked on: [runtimeverification/hs-backend-booster/#437](https://github.com/runtimeverification/hs-backend-booster/issues/437)~

Resolves #2228.


kevm-pyk/src/kevm_pyk/kproj/evm-semantics/optimizations.md
```diff
@@ -81,41 +53,10 @@ module EVM-OPTIMIZATIONS [kore]
       </ethereum>
       ...
     </kevm>
-    requires ( Gbase < SCHED > <=Gas GAVAIL )
+    requires ( notBool USEGAS orBool Gbase < SCHED > <=Gas GAVAIL )
```
ehildenb, line 56: Any reason to use `notBool ... orBool ...` here instead of `#if ... #then ... #else ... #fi`? Given that we'll have direct implementation of `#ite` in the backend, that will likely perform better, as it will get directly reduced to one branch or the other and won't have to evaluate both sides of the `orBool`.

geo2a:
@PetarMax @anvacaru we have a [PR](https://github.com/runtimeverification/hs-backend-booster/pull/442) in Booster that implements `KEQUAL.ite`. could you try it out?

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/optimizations.md
PetarMax: Given `domains.md`, the RHS of `orBool` will be evaluated only when the LHS evaluates to false, so the only extra computation will be that of `notBool USEGAS`, for a concrete `USEGAS: true`. Happy to revert this to `#if`.

PetarMax:
@geo2a I can confirm that the proof that was previously hanging now passes with the `#ite` hook implemented. Specifically, I ran:
```
kup install kevm --version . --override k-framework/booster-backend 437-if-then-else-hook 

kevm kompile-spec tests/specs/examples/erc20-spec.md -o definition --main-module VERIFICATION
kevm prove tests/specs/examples/erc20-spec.md --definition definition --spec-module ERC20-SPEC  --verbose --use-booster --claim ERC20-SPEC.totalSupply.noGas --save-directory .temp
```

anvacaru (approved):

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/optimizations.md
ehildenb: I think we should use the `#ite` specifically because it's _hooked_ and we can guarantee the evaluation order of it with the backend (making sure to inspect the boolean condition before evaluating the arguments of each branch).

I'm not talking about the k-rules, I'm talking about how the backend simplifies terms. It does not pay attention to left/right, instead it decides bottom-up or top-down, and no guarantee on the order of arguments that it evaluates. Even with the k-rules which say things like `rule true orBool _ => true`, the backend can still evaluate the second argument before firing this rule.

Let's use `#ite`.

PetarMax: Yes, this is what's happening now.

geo2a:
The implementation of `KEQUAL.ite` https://github.com/runtimeverification/hs-backend-booster/pull/442 had been merged yesterday and is on its way to evm-semantics.

PR #2243, opened 2024-01-09
Enable options for controlling booster execution

ehildenb:
~Blocked on: https://github.com/runtimeverification/pyk/pull/804~
~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/2236~
~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/2269~

This PR takes advantage of a few new features introduced in pyk, and makes a few updates to CLI and testing options:

- Option `--fast-check-subsumption` allows a faster way to check that a node will definitely not be subsumed into the target.
- Option `--always-check-subsumption` allows checking even non-terminal node for subsumption.
- Options `--[no-]post-exec-simplify` allows telling the booster to not call the old backend to simplify states that it produces before handing them back to pyk (see https://github.com/runtimeverification/hs-backend-booster/pull/436).
- Option `--fallback-on` allows the user to specify the fallback reasons the booster will consider (see https://github.com/runtimeverification/hs-backend-booster/pull/436).
- Option `--interim-simplification` allows the user to specify how frequently to invoke the full simplifier on the execution state.
- By default, _all_ breaks are turned off for testing purposes (which mimics CLI behavior better). Before, we had `break_on_calls` set to `True` for testing/CI behavior, but `False` for CLI behavior, which made it hard to track down a bug.
- An option `--spec-name` is added to `make test-prove-pyk ...` pytest testsuite, which allows filtering for a specific specification to run in the testing harness environment (much faster to narrow down to specific test that way).


yale-vinson:
It looks like you and Andrei have had some back and forth on this one. Is there an ETA to merge this one?

kevm-pyk/src/kevm_pyk/cli.py
```diff
@@ -53,6 +54,32 @@ def kprove_args(self) -> ArgumentParser:
...
+            dest='fast_check_subsumption',
+            default=True,
+            action='store_true',
+            help='Use fast-check on k-cell to determine subsumption.',
+        )
+        args.add_argument(
+            '--no-fast-check-subsumption',
+            dest='fast_check_subsumption',
+            action='store_false',
+            help='Use fast-check on k-cell to determine subsumption.',
```
anvacaru, line 81:
```suggestion
            help='Do not use fast-check on k-cell to determine subsumption.',
```

anvacaru (approved):
Approved with a typo fix. :+1:

PR #2252, opened 2024-01-10
Don't use padRightBytes for MSTORE8 opcode

nwatson22:


kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm-types.md
```diff
@@ -323,7 +323,8 @@ A cons-list is used for the EVM wordstack.
     syntax Bytes ::= "#write" "(" Bytes "," Int "," Int ")" [function]
                    | Bytes "[" Int ":=" Bytes "]" [function, total, klabel(mapWriteRange)]
  // --------------------------------------------------------------------------------------
-    rule #write(WM, IDX, VAL) => padRightBytes(WM, IDX +Int 1, 0) [ IDX <- VAL ]
+    syntax Bytes ::= #buf ( Int , Int ) [function, total, smtlib(buf)]
+    rule #write(WM, IDX, VAL) => WM [ IDX := #buf(1, VAL) ]
```
ehildenb, line 327: I don't think that we should move the definition of `#buf` to here, but instead just include `BUF` module into `EVM-TYPES` module.

But also not sure that is really the best way. How about we just add `concrete` to the `rule #write(...) => padRightBytes(...)`, so it doesn't simplify out?

```
    rule #write(WM, IDX, VAL) => padRightBytes(WM, IDX +Int 1, 0) [ IDX <- VAL ] [concrete]
```

Then, if it's concrete it will simplify to a fully concrete byte array. But if it's symbolic, it will keep the `#write(...)` symbol.

ehildenb: We can add the rule:

```
rule #write(WM, IDX, VAL) => WM [ IDX := #buf(1, VAL) ] [symbolic(WM, IDX, VAL), simplification]
```

to the `BUF` module where `#buf` is defined.

nwatson22: Moving the definition here is not clean, yes, and I'd like to avoid if I can. We noticed that `#write` is only used for this one `MSTORE8` opcode and nowhere else so we were considering removing it and just inlining the logic here. My understanding is that the idea to use `#buf` here was based which lemmas we already have.

nwatson22: Maybe it would be cleaner, as you suggest, to leave `#write()` unevaluated here, but I'm just worried that would require adding a whole set of lemmas to work with `#write()`. I am trying it out.

nwatson22: Fixed.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/buf.md
```diff
@@ -41,6 +41,7 @@ module BUF
 
     rule 0    <Int #powByteLen(SIZE) => true requires 0 <=Int SIZE [simplification]
     rule SIZE <Int #powByteLen(SIZE) => true requires 0 <=Int SIZE [simplification]
+    rule #write(WM, IDX, VAL) => WM [ IDX := #buf(1, VAL) ] [symbolic(WM), simplification]
```
PetarMax, line 44: This could potentially not be needed, but should apply at all times really, independently of whether the memory/index/value are concrete or symbolic.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm-types.md
```diff
@@ -323,7 +323,7 @@ A cons-list is used for the EVM wordstack.
     syntax Bytes ::= "#write" "(" Bytes "," Int "," Int ")" [function]
                    | Bytes "[" Int ":=" Bytes "]" [function, total, klabel(mapWriteRange)]
  // --------------------------------------------------------------------------------------
-    rule #write(WM, IDX, VAL) => padRightBytes(WM, IDX +Int 1, 0) [ IDX <- VAL ]
+    rule #write(WM, IDX, VAL) => padRightBytes(WM, IDX +Int 1, 0) [ IDX <- VAL ] [concrete]
```
PetarMax, line 326: Why not just have `rule #write(WM, IDX, VAL) => WM [ IDX := #buf(1, VAL) ]` here and not use the one above?

PetarMax: Understood. Then could the simplification in `buf.md` be unconditional? Would that be enough to never see `padRightBytes` in the symbolic world?

ehildenb: Yes it could reasonably be unconditional.

ehildenb: Have to test with the use-case that inspired this, and see if it works.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm-types.md
```diff
@@ -323,8 +323,6 @@ A cons-list is used for the EVM wordstack.
     syntax Bytes ::= "#write" "(" Bytes "," Int "," Int ")" [function]
                    | Bytes "[" Int ":=" Bytes "]" [function, total, klabel(mapWriteRange)]
  // --------------------------------------------------------------------------------------
-    rule #write(WM, IDX, VAL) => padRightBytes(WM, IDX +Int 1, 0) [ IDX <- VAL ]
```
PetarMax, line 326: I think the idea was to keep this rule with `[concrete]` for the LLVM backend, and then use the simplification in `buf.md` when reasoning symbolically?

PetarMax:
@nwatson22 Are you still observing the issues you were observing initially?

PR #2255, opened 2024-01-12
Remove evm-int-simplifications.k module

### LINKED ISSUE #2250: Duplicate lemmas between two different files
_Originally posted by @ehildenb in https://github.com/runtimeverification/evm-semantics/pull/2247#discussion_r1447681130_

It appears we may have duplication of some lemmas in KEVM, which could be a source of slowdown.

jberthold:
All rules in `evm-int-simplifications.k` are duplicated in `int-simplifications.k` at the time of writing. Therefore the file can be deleted without losing any simplifications, all proofs should have the simplifications available as before.

Fixes #2250


PetarMax:
@ehildenb at some point wanted to separate EVM-specific simplifications from those that should be in K all the time. These rules are duplicated indeed and it is safe to remove the file entirely, but I'd like @ehildenb's opinion on the structure.

ehildenb:
The idea was that `int-simplifications.k` (and associated tests) should be upstreamed into the K prelude. So I think a better solution, perhaps, is to keep `evm-int-simplifications.k`, but remove the references to `up/Int` that exist in `int-simplifications.k` (since `up/Int` is defined here in KEVM).

anvacaru (approved):

PetarMax (approved):

PR #2201, opened 2023-12-01
[WIP] Add support for hardcoded symbolic calldata, dynamic array arguments in ABI calldata

palinatolmach:
WIP.

This PR adds `abi_type_dynamic_array` and `abi_type_bytes_dynamic_array`) to the list of `TypesArgs` as well as the associated rules needed to handle Solidity function arguments of the dynamic array type (`bytes`, `bytes[]`, `uint256[]`, etc.)

Experimental: this PR also adds hardcoded calldata for a particular test from the Optimism codebase.

PR #2265, opened 2024-01-22
[WIP] Calldata generation for `bytes`, `bytes[]` parameters

palinatolmach:
WIP. 
An alternative to the manual approach taken in https://github.com/runtimeverification/evm-semantics/pull/2201.
Depends on the https://github.com/runtimeverification/kontrol/pull/303 PR in Kontrol. 

In order to generate calldata for functions that contain `bytes`, `bytes[]` parameters, as well as structs with such elements, this PR adds the following to `abi.md`:
- `#bytesArray(ARRAY_LEN:Int , EACH_ELEMENT_LEN:Int , ELEMENTS:TypedArgs )` production, as well as `#typeName`, `#enc`, `#sizeOfDynamicType`, `#lenOfHead` rules for it
- `ensures lengthBytes(BS) <=Int 1073741824` clause to the rule `#enc(#bytes(BS))`, thereby limiting the possible length of a `bytes` variable by 1 Gb
- supplement`#encBytes ( Int , TypedArg )`, `#encodeArgsBytes( TypedArgs, Int )`, ` #encodeArgsAuxBytes`

**Remaining work** that has to be done wrt to this PR includes proper **struct** handling. Previously, our encoding of `#tuple` (i.e., `struct`) in `abi.md` was based on the assumption that `#tuple` is a static (not dynamic) elements, and, in this case, the following encoding is correct:
```
rule #encodeArgs(#tuple(ARGS, B)) => #encodeArgs(ARGS)
```
However, if a struct contains a dynamically-sized element (e.g., `bytes data`), it gets treated as a dynamic argument: its being stored at an offset, where the elements of a struct are encoded sequentially. For the encoding to work correctly, the encoding and supplementary rules (e.g., `#enc`, `#sizeOfDynamicType`, `#lenOfHeads`) for tuples should be implemented correctly. To make it easier to define different rules for static and dynamic tuples, this PR added another boolean argument to the `#tuple` production:
```
 #tuple   ( ARGS: TypedArgs , IS_DYNAMIC: Bool )
```
this value represents whether the struct contains dynamic elements (which is calculated in Kontrol), but there might be a better solution for it. 

Finally, this PR also adds `#dynArray` production with some dummy rules as a placeholder to avoid failing on (temporarily) unsupported array types, e.g., `tuple[]`.


anvacaru:
Migrated the `#tuple` implementation in #2273. The `#dynArray` and `#bytesArray` `TypedArg` productions are no longer needed since I managed to generate the calldata in terms of `#bytes` and `#array` in https://github.com/runtimeverification/kontrol/pull/321. If I missed anything, let's open new issues.

Issue #2143, opened 2023-10-27
Bound env variables (`NUMBER`, `TIMESTAMP`) by reasonable values


palinatolmach:
Currently, environmental variables such as `block.number` and `block.timestamp` are symbolic and are only bounded by the type range constraints. We can further restrict their values by adding some reasonable bounds, e.g., `#rangeUInt(32, NUMBER)` for `block.number`, which is ~3 orders of magnitude larger that current block number, as suggested by @ehildenb.

lucasmt:
I have PR #2277 open to add bounds to the `block.timestamp` variable, and `block.number` is already bounded [here](https://github.com/runtimeverification/evm-semantics/blob/6d6ec8f3f2edcc400deaf417ed17e589b0e06ad3/kevm-pyk/src/kevm_pyk/kevm.py#L285). But from what I understand this issue is proposing to have stricter bounds for them based on what values we can reasonably expect these variables to have. Is this still something we want?

anvacaru:
I'd say to merge it as it is. We can further constrain `block.number` if we see fit.

palinatolmach:
Moving it into backlog for it to be addressed if/when we run into this issue. As suggested by @lucasmt, an alternative is to use `vm.assume` to provide bounds on the value, e.g., `vm.assume(block.number < 65536);`.

PR #2277, opened 2024-01-30
Add range constraints for TIMESTAMP_CELL

lucasmt:
Closes runtimeverification/kontrol#315.

Adds range constraints for the `TIMESTAMP_CELL` symbolic variable (corresponding to `block.timestamp` in Solidity) to be in the range of a `uint256`.


anvacaru:
@lucasmt I found this issue([link here](https://github.com/runtimeverification/evm-semantics/issues/2143)) from October which is a duplicate of yours. Can you have a look and also bound the `blocknumber` range if you consider it should be bounded?

anvacaru (approved):

PR #2286, opened 2024-02-05
Additional arithmetic reasoning needed for symbolic bytes

PetarMax:
This PR brings in some additional arithmetic reasoning that is needed [for the symbolic bytes cheatcode of Kontrol](https://github.com/runtimeverification/kontrol/pull/254) to work properly. Specifically, the PR:
- strengthens the constraints around `&Int`, `|Int`, and `lengthBytes`;
- introduces several simplifications concerning idempotency of `chop`;
- introduces several simplifications that further normalise comparisons; and
- introduces a test case provided by @nwatson22 , which corresponds to compiler-generated constraints in the presence of byte arrays of symbolic length.


kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/int-simplification.k
```diff
@@ -83,6 +83,15 @@ module INT-SIMPLIFICATION-HASKELL [symbolic]
...
     rule { A -Int B #Equals C } => { A -Int C #Equals B } [concrete(A, C), symbolic(B), simplification(45), comm]
 
+    rule A +Int B  <Int C +Int D => A -Int C  <Int D -Int B [concrete(B, D), symbolic(A, C), simplification(45)]
+    rule A +Int B <=Int C +Int D => A -Int C <=Int D -Int B [concrete(B, D), symbolic(A, C), simplification(45)]
+    rule A +Int B  >Int C +Int D => A -Int C  >Int D -Int B [concrete(B, D), symbolic(A, C), simplification(45)]
+    rule A +Int B >=Int C +Int D => A -Int C >=Int D -Int B [concrete(B, D), symbolic(A, C), simplification(45)]
+
+    rule A +Int B ==Int C +Int D => A -Int C ==Int D -Int B [concrete(B, D), symbolic(A, C), simplification(45)]
+
+    rule { A +Int B #Equals C +Int D } => { A -Int C #Equals D -Int B } [concrete(B, D), symbolic(A, C), simplification(45), comm]
```
ehildenb, line 93: Dose this rule need the `comm` attribute? `comm` attribute causes the frontend to duplicate the rule with the arguments of the top-level symbol flipped. The rule that the frontend will create will be identical to this one anyway, so it just adds to th erules to try I think.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
```diff
@@ -52,4 +53,20 @@ module EVM-INT-SIMPLIFICATION-COMMON
       requires 0 <Int Y
       [simplification, comm, preserves-definedness]
 
+  // ###########################################################################
+  // chop
+  // ###########################################################################
+
+  rule [chop-idempotent]:
+    chop ( chop ( X:Int ) ) => chop ( X )
+    [simplification]
```
ehildenb, line 62: I'm surprised this rule is needed. I guess I would think that these three rules together would imply the same: https://github.com/runtimeverification/evm-semantics/blob/6f86c28def0ac17b0998352ec06875229d6bfee8/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/lemmas.k#L25.

Perhaps we need to mark the third rule there as `smt-lemma` then we won't need this one as well?

ehildenb: Maybe also those lemmas linked above could be moved over here.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bitwise-simplification.k
```diff
@@ -71,6 +74,10 @@ module BITWISE-SIMPLIFICATION [symbolic]
        andBool X +Int Z ==Int maxUInt256
       [simplification, concrete(X, Z), comm]
 
+    rule [notMaxUInt5-bit-and-nested]:
+      notMaxUInt5 &Int ( ( notMaxUInt5 &Int X ) +Int Y ) => chop ( ( notMaxUInt5 &Int X ) +Int ( notMaxUInt5 &Int Y ) )
+      [simplification, concrete(Y)]
+
```
ehildenb, line 80: This simplification seems pretty specific and the odd-one-out. Is it needed for the overall simplification (the test added here)? If not, maybe we can remove it, or add a test which specifically exercises it.

I also wonder if it can be made more general somehow....

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/int-simplification.k
PetarMax: Yes, this rule is fully commutative already, I will remove the attribute.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
PetarMax: I will move the lemmas and add `smt-lemma`, I think you are correct.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bitwise-simplification.k
PetarMax: It feels specific, yes, but it comes from the compiler. This pattern has appeared both in Optimism and as part of the tests for @nwatson22's [PR](https://github.com/runtimeverification/kontrol/pull/254). I can make it more general:
but my intuition is that this is sufficient. We do already have specific lemmas concerning bounds on `notMaxUInt5 &Int lengthBytes(X)`, which also come from the compiler. The more general lemma would be:
```
    rule [mask-bit-and-nested]:
      M &Int ( ( M &Int X ) +Int Y ) => chop ( ( M &Int X ) +Int ( M &Int Y ) )
      requires #rangeUInt(256, M)
       andBool 0 <=Int X andBool 0 <=Int Y
       andBool M +Int 1 ==Int 2 ^Int log2Int(M +Int 1)
      [simplification, concrete(M, Y)]
```
covering the cases when the mask is not just the first 26 bytes, but it's not clear there's a use case for it.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bitwise-simplification.k
```diff
@@ -71,6 +74,11 @@ module BITWISE-SIMPLIFICATION [symbolic]
        andBool X +Int Z ==Int maxUInt256
       [simplification, concrete(X, Z), comm]
 
+    rule [notMaxUInt5-bit-and-nested]:
+      notMaxUInt5 &Int ( ( notMaxUInt5 &Int X ) +Int Y ) => chop ( ( notMaxUInt5 &Int X ) +Int ( notMaxUInt5 &Int Y ) )
+      requires 0 <=Int X andBool 0 <=Int Y
+      [simplification, concrete(Y)]
+
```
ehildenb, line 81: This one still seems a bit off to me.

Is it true that:

```
rule notMaxUInt5 &Int ((notMaxUInt5 &Int X) +Int Y) => notMaxUInt5 &Int (X +Int Y) [simplification]
```

is true?

ehildenb: Hmmmm, it's probably not true. Counterexample is `X == 1` and `Y == 1` I think.

ehildenb: Ah wait, that's not a counterexample, since it's 0 in any case...

ehildenb (approved):

PR #2270, opened 2024-01-24
Adding further lemmas for bitwise reasoning

PetarMax:
The PR adds some further general lemmas about bitwise reasoning, coming from various engagements.

PR #2288, opened 2024-02-08
Additional lemmas for Kontrol

PetarMax:
The simplifications introduced by #2270 lead to a Kontrol CI test failing because we end up with an unreduced `#padRightToWidth`. This PR introduces the simplification needed for this reduction to happen, along with an appropriate test.

In addition, one of the conjuncts is removed from one of the #2270 simplifications, as it is not strictly necessary.


tests/specs/functional/lemmas-spec.k
```diff
@@ -602,6 +602,17 @@ module LEMMAS-SPEC
...
+ // #padRightToWidth
+ // ----------------
+
+    claim [pRTW-01]:
+      <k> runLemma (
+            notBool ( #asWord ( #padRightToWidth ( 32 , #buf ( 4 , VV0_x_114b9705:Int ) ) ) ==Int
+              #asWord ( #buf ( 4 , VV0_x_114b9705:Int ) +Bytes b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00" ) )
+          ) => doneLemma(
+            false
+          ) ... </k>
```
ehildenb, line 614: I guess there should be a side-condition on `VV0_x_...` from the original unsimplified term:

```suggestion
    claim [pRTW-01]:
      <k> runLemma (
            notBool ( #asWord ( #padRightToWidth ( 32 , #buf ( 4 , VV0_x_114b9705:Int ) ) ) ==Int
              #asWord ( #buf ( 4 , VV0_x_114b9705:Int ) +Bytes b"\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00" ) )
          ) => doneLemma(
            false
          ) ... </k>
        requires #rangeUInt(32, VV0_x_114b9705)
```

palinatolmach (approved):

PR #2299, opened 2024-02-13
Small cleanups/simplifications to run_prover

ehildenb:
The `run_prover` method has become simpler because we've unified the `APRProof` and `APRBMCProof` class hierarchies. This does some more minor cleanups:

- Factors out common logic between different proof types.
- Removes the only use of `kprove` from `run_prover`, which was for a use-case that was not exercised anyway (we aren't using `EqualityProof` anywhere anyway).
- Removes some now-defunct options.


ehildenb:
The changes needed for Kontrol to function with this update are here: https://github.com/runtimeverification/kontrol/pull/359

nwatson22 (approved):

PR #2306, opened 2024-02-20
fix bug in computation of miner balance after tx if miner == sender

dwightguth:
The implementation of the London hard fork introduced a bug that occurred during transaction finalization if the sender of a transaction and the miner are the same address. In this case, the balance in that account after receiving the refund to the sender and the transaction fee to the miner is incorrect.

We modify the rule to quite simply correct the relevant integer expression. The code added results from taking `balance + gavail*gprice + (glimit - gavail)*(gprice - basefee)` and simply redistributing it slightly.

Fixing this issue fixes 2 of the 3 failing tests on the llvm backend.


PetarMax:
Just for my understanding, what is the official source that tells us what the correct balance should be?

ehildenb:
Here is the London hardffork specification @PetarMax : https://github.com/ethereum/execution-specs/blob/master/network-upgrades/mainnet-upgrades/london.md

The official "source of truth" is largely the test-suite (which because we are able to enable 2 more tests without breaking other things, is a good indication of the correctness of this change!).

ehildenb (approved):

ehildenb:
Also here is the PR into KEVM that implemented this hardfork: https://github.com/runtimeverification/evm-semantics/pull/1187

gtrepta (approved):
Looks like the problem was that the base fee was being burned for unused gas.

LGTM.

anvacaru (approved):

PR #2307, opened 2024-02-21
Lemmas and tests needed for symbolic bytes manipulation

PetarMax:
This PR introduces a number of lemmas and tests needed for manipulating symbolic bytes in KEVM memory and storage.

These were crafted for the simple tests of [Kontrol PR 254](https://github.com/runtimeverification/kontrol/pull/254), but are likely to have a broader use.

The present simplifications are sufficient for all symbolic-bytes tests of [PR 254](https://github.com/runtimeverification/kontrol/pull/254) to pass except `test_symbolic_bytes_2`, which revealed substantial complexities when it comes to storing and manipulating `bytes` in contract storage.


kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bitwise-simplification.k
```diff
@@ -113,8 +113,8 @@ module BITWISE-SIMPLIFICATION [symbolic]
 
     rule (X <<Int Y) <Int pow256 => true
       requires 0 <=Int X andBool 0 <=Int Y andBool Y <Int 256 andBool X <Int 2 ^Int (256 -Int Y)
-      [simplification, preserves-definedness]
+      [simplification]
```
PetarMax, line 116: `preserves-definedness` is not required since the RHS does not contain partial functions.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bitwise-simplification.k
```diff
@@ -113,8 +113,8 @@ module BITWISE-SIMPLIFICATION [symbolic]
...
 
-    rule 0 <=Int (X <<Int Y) => true requires 0 <=Int X andBool 0 <=Int Y [simplification, preserves-definedness]
+    rule 0 <=Int (X <<Int Y) => true requires 0 <=Int X andBool 0 <=Int Y [simplification]
```
PetarMax, line 118: `preserves-definedness` is not required since the RHS does not contain partial functions.

ehildenb:
It looks good to me, would be good to have someone else also check over the lemmas and such!

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
```diff
@@ -88,12 +88,11 @@ module BYTES-SIMPLIFICATION [symbolic]
...
-      #buf(W:Int, X <<Int Y) => #buf(W -Int (Y /Int 8), X) +Bytes #buf(Y /Int 8, 0)
-      requires 0 <=Int W andBool W <=Int 32
-       andBool 0 <=Int X andBool X <Int 2 ^Int (256 -Int Y)
-       andBool 0 <=Int Y andBool ( Y <Int W *Int 8 ) andBool Y modInt 8 ==Int 0
-      [simplification, preserves-definedness]
+    rule [shift-as-asWord]:
+      X <<Int Y => #asWord ( #buf ( 32 -Int (Y /Int 8) , X ) +Bytes #buf ( Y /Int 8 , 0 ) )
+      requires 0 <=Int X andBool X <Int 2 ^Int (256 -Int Y)
+       andBool 0 <=Int Y andBool Y <=Int 256 andBool Y modInt 8 ==Int 0
+      [simplification, concrete(Y), preserves-definedness]
```
lucasmt, line 95: Is it possible that this might get in the way of the other `X <<Int Y` rules, like `(X <<Int Y) <Int pow256 => true` above? If so, should we lower this rule's priority?

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
```diff
@@ -88,12 +88,11 @@ module BYTES-SIMPLIFICATION [symbolic]
       requires 0 <=Int W1 andBool 0 <=Int W2
       [simplification]
 
-    rule [buf-shift]:
-      #buf(W:Int, X <<Int Y) => #buf(W -Int (Y /Int 8), X) +Bytes #buf(Y /Int 8, 0)
-      requires 0 <=Int W andBool W <=Int 32
-       andBool 0 <=Int X andBool X <Int 2 ^Int (256 -Int Y)
-       andBool 0 <=Int Y andBool ( Y <Int W *Int 8 ) andBool Y modInt 8 ==Int 0
-      [simplification, preserves-definedness]
```
lucasmt, line 96: This original rule covers the case where `W <Int 32`, as long as `X <<Int Y` fits in `W` bytes. I couldn't find a rule that could be combined with the new rule below to cover this case as well. For example, would `#buf(20, X <<Int 16)` get simplified, assuming `X <Int 2 ^Int 144`?

PetarMax: I have added more lemmas to take care of this case, together with the corresponding  claim.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
```diff
@@ -59,11 +59,17 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+    rule [buf-asWord-invert-lr-len-leq]:
       #buf (W:Int , #asWord(B:Bytes)) => #buf(W -Int lengthBytes(B), 0) +Bytes B
       requires lengthBytes(B) <=Int W andBool W <=Int 32
       [simplification]
 
+    rule [buf-asWord-invert-lr-len-gt]:
+      #buf (W:Int , #asWord(B:Bytes)) => #range(B, lengthBytes(B) -Int W, W)
+      requires W <Int lengthBytes(B) andBool lengthBytes(B) <=Int 32
+       andBool #asWord(B) <Int 2 ^Int (8 *Int W)
+      [simplification]
```
PetarMax, line 71: This simplification covers the case brought up by @lucasmt in [this comment](https://github.com/runtimeverification/evm-semantics/pull/2307/files#r1498050387).

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
```diff
@@ -65,6 +65,11 @@ module EVM-INT-SIMPLIFICATION-COMMON
     requires 2 ^Int ( lengthBytes(BA) *Int 8) <=Int X
     [concrete(X), simplification]
 
+  rule [asWord-lt-concat]:
+    #asWord ( BA1 +Bytes BA2 ) <Int X => #asWord ( BA1 ) <Int ( X -Int #asWord ( BA2 ) ) /Int ( 2 ^Int ( 8 *Int lengthBytes ( BA2 ) ) )
+    requires ( X -Int #asWord ( BA2 ) ) modInt ( 2 ^Int ( 8 *Int lengthBytes ( BA2 ) ) ) ==Int 0
+    [concrete(BA2, X), simplification]
```
PetarMax, line 71: This is an auxiliary simplification required for the interaction of `[shift-as-asWord]` and `[buf-asWord-invert-lr-len-gt]` so that the expressivity of `[buf-shift]` is preserved.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
```diff
@@ -59,11 +59,17 @@ module BYTES-SIMPLIFICATION [symbolic]
...
 
-    rule [buf-asWord-invert-lr]:
+    rule [buf-asWord-invert-lr-len-leq]:
       #buf (W:Int , #asWord(B:Bytes)) => #buf(W -Int lengthBytes(B), 0) +Bytes B
       requires lengthBytes(B) <=Int W andBool W <=Int 32
       [simplification]
 
+    rule [buf-asWord-invert-lr-len-gt]:
+      #buf (W:Int , #asWord(B:Bytes)) => #range(B, lengthBytes(B) -Int W, W)
+      requires W <Int lengthBytes(B) andBool lengthBytes(B) <=Int 32
```
lucasmt, line 69:
```suggestion
      requires 0 <=Int W andBool W <Int lengthBytes(B) andBool lengthBytes(B) <=Int 32
```
Not sure if strictly necessary.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
lucasmt, line 71:
```suggestion
      [simplification, concrete(W)]
```
Since `W` it's used in a non-linear constraint in the previous line. This maybe limits expressivity a bit, but I think most of the time our `#buf`s have concrete width anyway, so I wouldn't expect to make a big difference.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
lucasmt: Alright, if we haven't seen any issues and the expressions will still simplify anyway then it should be fine.

lucasmt (approved):
I've left some suggestions, but once those are addressed I think it can be merged.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
PetarMax: We always have constraints like this in `#buf` to ensure definedness of the LHS.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
PetarMax: I would like to keep `W` concrete for now and remove `concrete` only if we find a use case in the real-world code. There is a number of other simplifications that follow this pattern.

PR #2308, opened 2024-02-21
Add lemmas used in recent engagements

palinatolmach:
This PR adds a file with `ENGAGEMENT-LEMMAS`: a set of lemmas that were used in recent engagements.  

Some of these lemmas can probably be upstreamed to KEVM directly, while other engagement-specific ones can be kept in the [engagement-lemmas.k ](https://github.com/runtimeverification/evm-semantics/blob/f5b0821d8044083c05e13eb9323a90d125c9033b/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/engagement-lemmas.k) to be reused , when necessary.


palinatolmach:
@PetarMax may I ask for your help with deciding which of the lemmas should be upstreamed? Thank you!

PR #2310, opened 2024-02-21
only check nonce on CREATE and CREATE2

dwightguth:
There was a bug in the implementation of EIP 2681: we did not allow contracts whose nonce was 2^64-1 to make any calls at all, even ones which do not require incrementing the nonce. This relatively straightforward fix separates the checks we perform on CALL/CALLCODE/DELEGATECALL/STATICCALL from the checks we perform on CREATE/CREATE2 and disables the nonce check on the former. This appears to fix the failing test `tests/ethereum-tests/BlockchainTests/GeneralStateTests/stCreate2/CREATE2_HighNonceDelegatecall.json`


kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
```diff
@@ -1523,6 +1523,7 @@ For each `CALL*` operation, we make a corresponding call to `#call` and a state-
     syntax InternalOp ::= "#create"   Int Int Int Bytes
                         | "#mkCreate" Int Int Int Bytes
                         | "#incrementNonce" Int
+                        | "#checkCreate" Int Int
  // -------------------------------------------
```
anvacaru, line 1527: Minor formatting.
```suggestion
 // --------------------------------------------
```

anvacaru (approved):

PR #2315, opened 2024-02-28
Moving `#accessStorage` logic back to gas calculations

PetarMax:
I initially thought that maintaing `#accessStorage` logic would be beneficial for CSE, but it is creating substantial difficulties in that:
- accessed storage is calculated locally, per function, and the effects do not permeate across external calls
- the `#accessStorage` rules are such that if the `<accessedStorage>` cell is fully symbolic, the execution branches infinitely due to the sorting limitations of K.

This PR moves the `#accessStorage` logic back to gas calculations.


ehildenb:
What if we turned off the `#accessStorage ...` rules altogether with a different flag, similar to how we can turn off the gas calculations? My guess is the benefit here is because we are turning off gas calculations (by using the `<useGas>` cell), so by putting these here, we are able to get it to skip this reasoning too. So we could introduce a new cell `<useAccessStorage>`, and have that be able to be turned off by Kontrol as well, but still keep it separate from gas calculations.

PetarMax:
The only reason I kept the logic in non-gas mode was because I thought it would help us inform CSE about which locations in the storage are touched so that we could streamline the produced specifications (that is, limit them only to the part of the storage relevant for a particular function). However, this doesn't seem to be needed, plus there is the problem with sorting that causes infinite branching.

PetarMax:
Separately, there is the question whether or not `<accessedStorage>` should live as part of `<substate>`, since the [EIP that introduced it](https://eips.ethereum.org/EIPS/eip-2929) states that:
```
When executing a transaction, maintain a set accessed_addresses: 
Set[Address] and accessed_storage_keys: Set[Tuple[Address, Bytes32]] .

The sets are transaction-context-wide, implemented identically to other 
transaction-scoped constructs such as the self-destruct-list and global 
refund counter. In particular, if a scope reverts, the access lists should 
be in the state they were in before that scope was entered.
```
and `<substate>` is cleared on end of every external function call, not transaction.

PetarMax:
Just to clarify, `#accessStorage` logic is now reverted to how it was implemented originally.

ehildenb (approved):

PR #2319, opened 2024-02-29
Improve options passing and default options handling (do not merge)

nwatson22:
Building on top of https://github.com/runtimeverification/pyk/pull/916
This is what the changes to KEVM could look like with the design in pyk#916.

### DIFF
--- a/kevm-pyk/src/kevm_pyk/__main__.py
+++ b/kevm-pyk/src/kevm_pyk/__main__.py
@@ -15,14 +15,12 @@ from pathlib import Path
 from typing import TYPE_CHECKING
 
 from pathos.pools import ProcessPool  # type: ignore
-from pyk.cli.utils import file_path
 from pyk.cterm import CTerm
 from pyk.kast.outer import KApply, KRewrite, KSort, KToken
 from pyk.kcfg import KCFG
 from pyk.kdist import kdist
-from pyk.kore.tools import PrintOutput, kore_print
+from pyk.kore.tools import kore_print
 from pyk.ktool.kompile import LLVMKompileType
-from pyk.ktool.krun import KRunOutput
 from pyk.prelude.ml import is_top, mlOr
 from pyk.proof import APRProof
 from pyk.proof.implies import EqualityProof
@@ -31,7 +29,17 @@ from pyk.proof.tui import APRProofViewer
 from pyk.utils import FrozenDict, hash_str, single
 
 from . import VERSION, config
-from .cli import KEVMCLIArgs, node_id_like
+from .cli import (
+    KastOptions,
+    KompileSpecOptions,
+    ProveLegacyOptions,
+    ProveOptions,
+    PruneProofOptions,
+    RunOptions,
+    ShowKCFGOptions,
+    VersionOptions,
+    ViewKCFGOptions,
+)
 from .gst_to_kore import SORT_ETHEREUM_SIMULATION, gst_to_kore, kore_pgm_to_kore
 from .kevm import KEVM, KEVMSemantics, kevm_node_printer
 from .kompile import KompileTarget, kevm_kompile
@@ -53,6 +61,8 @@ if TYPE_CHECKING:
     from pyk.kcfg.kcfg import NodeIdLike
     from pyk.kcfg.tui import KCFGElem
     from pyk.kore.rpc import FallbackReason
+    from pyk.kore.tools import PrintOutput
+    from pyk.ktool.krun import KRunOutput
     from pyk.proof.proof import Proof
     from pyk.utils import BugReport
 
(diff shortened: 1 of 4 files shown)


ehildenb:
Wow, this looks really nice! @tothtamas28 if you have time, it would be good to get your opinion here too.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -699,147 +688,19 @@ def parse(s: str) -> list[T]:
...
-    )
+    command_parser = VersionOptions.parser(command_parser)
+    command_parser = KompileSpecOptions.parser(command_parser)
+    command_parser = ProveOptions.parser(command_parser)
+    command_parser = PruneProofOptions.parser(command_parser)
+    command_parser = ProveLegacyOptions.parser(command_parser)
+    command_parser = ViewKCFGOptions.parser(command_parser)
+    command_parser = ShowKCFGOptions.parser(command_parser)
+    command_parser = RunOptions.parser(command_parser)
+    command_parser = KastOptions.parser(command_parser)
```
ehildenb, line 703: I guess I'm not 100% thrilled about this design (pass the parser around and have it be modified). I would like a design where it was more like:

```
command_parser = CommandParser([VersionOptions, KompileSpecOptions, ProveOptions, ...])
```

for example. Or even have a super class `KEVMCLI`, which inherits alll the other options classes and puts them togethr in the way that's needed.

kevm-pyk/src/kevm_pyk/cli.py
```diff
@@ -32,382 +48,758 @@ def node_id_like(s: str) -> NodeIdLike:
...
+        case 'view-kcfg':
+            return ViewKCFGOptions(args)
+        case 'show-kcfg':
+            return ShowKCFGOptions(args)
+        case 'run':
+            return RunOptions(args)
+        case 'kast':
+            return KastOptions(args)
+        case _:
+            raise ValueError('Unrecognized command.')
```
ehildenb, line 73: This dispatch feels a bit weird to me as well. If we had one single `KEVMCLI` class, that inherited all the other classes and put together all the options, then we would know that we had handled enough of the options. As it stands, this must be kept in sync with the other place where all the options are put together.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -74,120 +85,98 @@ def main() -> None:
...
     logging.basicConfig(level=_loglevel(args), format=_LOG_FORMAT)
+    options = generate_command_options({key: val for (key, val) in vars(args).items() if val is not None})
 
     executor_name = 'exec_' + args.command.lower().replace('-', '_')
     if executor_name not in globals():
         raise AssertionError(f'Unimplemented command: {args.command}')
 
     execute = globals()[executor_name]
-    execute(**vars(args))
+    execute(options)
```
ehildenb, line 95: It would be cool if we instead had something like:

```
kevm_cli = KEVMCLI(args)
```

And it just kinda internalizes all the logic. Then all teh `exec_...` methods can be moved there, and they can be registered explicitely for each subcommand inside the class.

kevm-pyk/src/kevm_pyk/__main__.py
nwatson22: I've implemented a change to the `pyk` version here: https://github.com/runtimeverification/pyk/pull/916/commits/9f6574681d58d787c63ee484cbf02b26410096ad similar to what is suggested here which should handle the dispatch issue, and having to manually call `parser` for each subcommand, but it's still passing around the parser. I'll see if I can eliminate that as well.

kevm-pyk/src/kevm_pyk/cli.py
```diff
@@ -32,382 +49,783 @@ def node_id_like(s: str) -> NodeIdLike:
...
+        return {
+            'failure_info': False,
+            'to_module': False,
+            'pending': False,
+            'failing': False,
+            'counterexample_info': False,
+        }
+
+    @staticmethod
+    def args(parser: ArgumentParser) -> ArgumentParser:
```
tothtamas28, line 174:
```suggestion
    def args(parser: ArgumentParser) -> None:
```

(Unless there are cases it is expected to return a new instance.)

tothtamas28: Also, consider naming this `argument_parser`.

kevm-pyk/src/kevm_pyk/cli.py
```diff
@@ -32,382 +49,783 @@ def node_id_like(s: str) -> NodeIdLike:
...
+        parser.add_argument(
+            '--node',
+            type=node_id_like,
+            dest='nodes',
+            default=[],
```
tothtamas28, line 179: Should this be `None`, and the empty list added to `defaults`?

kevm-pyk/src/kevm_pyk/cli.py
```diff
@@ -32,382 +49,783 @@ def node_id_like(s: str) -> NodeIdLike:
...
+            type=arg_pair_of(node_id_like, node_id_like),
+            dest='node_deltas',
+            default=[],
+            action='append',
+            help='List of nodes to display delta for.',
+        )
+        parser.add_argument(
+            '--failure-information',
+            dest='failure_info',
+            default=None,
```
tothtamas28, line 194: All occurrances of `default=None` can be removed (default of `default` is `None`).

nwatson22: Fixed.

kevm-pyk/src/kevm_pyk/cli.py
nwatson22: Fixed.

kevm-pyk/src/kevm_pyk/cli.py
nwatson22: Removed return and renamed all occurrences to `update_args`.

kevm-pyk/src/kevm_pyk/cli.py
```diff
@@ -25,389 +24,419 @@ def parse(s: str) -> list[T]:
...
-    @cached_property
-    def kprove_args(self) -> ArgumentParser:
-        args = ArgumentParser(add_help=False)
-        args.add_argument(
+class KOptions(KDefinitionOptions):
+    depth: int | None
+
+    @staticmethod
+    def default() -> dict[str, Any]:
+        return {'depth': None}
```
ehildenb, line 49: This appears to just delete the defaults of the superclass `KDefinitionOptions`. Should we be doing something like:

```
def default() -> dict[str, Any]:
    return chain([super().default(), {'depth': None}])
```

ehildenb: Ah, I see that `Options` uses `type(...).mro()` to build up the overall defaults...

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -71,144 +88,214 @@ def _ignore_arg(args: dict[str, Any], arg: str, cli_option: str) -> None:
...
+            VersionCommand,
+            ViewKCFGCommand,
+        )
+    )
+    parser = kevm_cli.create_argument_parser()
     args = parser.parse_args()
+
     logging.basicConfig(level=_loglevel(args), format=_LOG_FORMAT)
+    command = kevm_cli.generate_command({key: val for (key, val) in vars(args).items() if val is not None})
+    command.exec()
```
ehildenb, line 110: I'm curious about the performance of this approach. Suppose we make it so that `KastCommand` is a no-op (take out the actual implementation in teh `exec()` method), how long does running `kevm-pyk kast ....` take? Basically, what is the overhead associated with this method of putting together the CLI?

nwatson22: As far as I can tell, any overhead is negligible. With or without a no-op `exec()`, and on master as well, it takes between 0.21 and 0.23 seconds, and the vast majority of that must be loading the code or something, because it fails to go down below that even when I exit right at the start of `main()`.

Is there a particular reason we'd expect a performance hit from code like this?

ehildenb: Just more complicated class structure that Python has to load, but I guess that's neglible compared to the overhead of just starting the python process up.

ehildenb: Thanks for checking!

PR #2321, opened 2024-02-29
Terminating execution on symbolic program

PetarMax:
This PR ensures that symbolic execution stops if the program to be executed is a symbolic variable. 

This happens, for example, when the backend infers the existence of an account in the `<accounts>` cell during CSE because we do not have the code of the contract being called into.


anvacaru (approved):

ehildenb:
Oh, that is annoying. We can turn on `always_check_subsumption` for the functional proofs?

ehildenb:
Or, we can mark `doneLemma(...)` as `terminal` as well, in this PR (by adjusting `is_terminal`).

PetarMax:
Unfortunately, it's not about the proof executing and not finishing because the program is symbolic, it's not being able to start it. Is there a way to set the program automatically to something concrete (`b""`, for example) for functional tests?

kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -55,6 +55,12 @@ def is_terminal(self, cterm: CTerm) -> bool:
             # <k> #halt ~> X:K </k>
             elif k_cell.arity == 2 and k_cell[0] == KEVM.halt() and type(k_cell[1]) is KVariable:
                 return True
+
+        program_cell = cterm.cell('PROGRAM_CELL')
+        # Fully symbolic program
+        if type(program_cell) is KVariable:
+            return True
```
ehildenb, line 62:
```suggestion
        program_cell = cterm.cell('PROGRAM_CELL')
        # Fully symbolic program
        if type(program_cell) is KVariable:
            if <k>-cell != <k> runLemma(...) </k> and <k>-cell != <k> runLemma(...) ~> _ </k>:
                return False
            return True
```

PetarMax: I have implemented this now and tested locally, it appears to be working.

PR #2328, opened 2024-03-04
Redesigining `#precompiledAccounts`

PetarMax:
Reasoning about `#precompiledAccounts` causes the Haskell backend to severely slow down if the account is symbolic. This is likely due to the complexity of set reasoning in the checked side-conditions.

This PR redesigns `#precompiledAccounts` to describe these accounts as a range, simplifying the reasoning and removing the slowdown. Range is favoured over any more complex data structure (such as a list) because of simplicity and solver support.


kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
```diff
@@ -1728,21 +1727,37 @@ Precompiled Contracts
     rule #precompiled(8) => ECPAIRING
     rule #precompiled(9) => BLAKE2F
 
-    syntax Set ::= #precompiledAccounts ( Schedule ) [klabel(#precompiledAccounts), function, total]
+    syntax Int ::= #precompiledAccounts ( Schedule ) [klabel(#precompiledAccounts), function, total]
```
PetarMax, line 1730: `#precompiledAccounts` now returns only the upper bound of the precompiled accounts, assuming they will remain contiguous, starting from 1.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
```diff
@@ -1333,12 +1333,11 @@ The various `CALL*` (and other inter-contract control flow) operations will be d
...
-    rule [precompile.false]: <k> #precompiled?(ACCTCODE, SCHED) => .K                                ... </k> requires notBool #isPrecompiledAccount(ACCTCODE, SCHED)
+    rule [precompile.false]: <k> #precompiled?(ACCTCODE, SCHED) => .K                               ... </k> requires notBool #isPrecompiledAccount(ACCTCODE, SCHED)
 
     syntax Bool ::= #isPrecompiledAccount ( Int , Schedule ) [klabel(#isPrecompiledAccount), function, total, smtlib(isPrecompiledAccount)]
  // ---------------------------------------------------------------------------------------------------------------------------------------
-    rule [isPrecompiledAccount.true]:  #isPrecompiledAccount(ACCTCODE, SCHED) => true  requires         ACCTCODE in #precompiledAccounts(SCHED)
-    rule [isPrecompiledAccount.false]: #isPrecompiledAccount(ACCTCODE, SCHED) => false requires notBool ACCTCODE in #precompiledAccounts(SCHED)
+    rule [isPrecompiledAccount]:  #isPrecompiledAccount(ACCTCODE, SCHED) => 0 <Int ACCTCODE andBool ACCTCODE <=Int #precompiledAccounts(SCHED)
```
PetarMax, line 1340: The rule for `#isPrecompiledAccount(ACCTCODE, SCHED)` now trivially rewrites to the account being in the appropriate range, leading to no side conditions being checked (which is what caused the original slowdown).

palinatolmach:
@PetarMax can you please add a test here in KEVM that demonstrates that there're no side conditions being checked/no slowdown now?

palinatolmach:
I can confirm that with this change the [Optimism-like test](https://github.com/runtimeverification/kontrol/blob/0c4ab95f45d7d290aa69c9c677a10b1fe4617385/src/tests/integration/test-data/foundry/test/OptimismPortalKontrol.t.sol) (to be upstreamed) takes 9 minutes instead of 12. It lgtm, but I'd wait for @geo2a and @ehildenb feedback for final approval.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
```diff
@@ -1728,21 +1727,37 @@ Precompiled Contracts
...
     rule #precompiledAccounts(CONSTANTINOPLE)    => #precompiledAccounts(BYZANTIUM)
     rule #precompiledAccounts(PETERSBURG)        => #precompiledAccounts(CONSTANTINOPLE)
-    rule #precompiledAccounts(ISTANBUL)          => #precompiledAccounts(PETERSBURG) SetItem(9)
+    rule #precompiledAccounts(ISTANBUL)          => 9
     rule #precompiledAccounts(BERLIN)            => #precompiledAccounts(ISTANBUL)
     rule #precompiledAccounts(LONDON)            => #precompiledAccounts(BERLIN)
     rule #precompiledAccounts(MERGE)             => #precompiledAccounts(LONDON)
     rule #precompiledAccounts(SHANGHAI)          => #precompiledAccounts(MERGE)
+
+    syntax Set ::= #precompiledAccountsAsSet ( Schedule ) [klabel(#precompiledAccountsAsSet), function, total]
```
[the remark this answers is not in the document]
ehildenb: Another implementation that wouldn't require any maintance:

```
rule #precompiledAccountsAsSet(SCHED) => #precompiledAccountsAsSetAux(#precompiledAccounts(SCHED))
rule #precompiledAccountsAsSetAux(N) => .Set requires N <Int 0
rule #precompiledAccountsAsSetAux(N) => SetItem(N) #precompiledAccountsAsSetAux(N -Int 1) [owise]
```

Since we'll always pass in a concrete `N`, should be fine.

ehildenb:
Good that we checked the Optimism proof, was there another proof with slowdown we should check?

palinatolmach:
There's also [`CSETest.test_identity`](https://github.com/runtimeverification/kontrol/blob/petar/cse-exploration/src/tests/integration/test-data/foundry/test/CSE.t.sol#L1C1-L34C2) that @PetarMax reported the slowdown on originally. I think Petar checked it too, but I also ran it on my machine now — the execution time for it drops back to 3 minutes (from almost 6).

PetarMax:
All of the proofs are back to normal, yes. With respect to a test for this specific PR, I should be able to write a functional test that would take substantially longer with the previous version of `#precompiledAccounts`, but I'm not sure how we will measure that or use it in CI effectively. I don't think it's possible to write a test that demonstrates there are no side conditions being checked, we don't have that level of reflection. In any case, this is guaranteed by the side conditions being removed from the `isPrecompiledAccount` rules.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
geo2a: > #precompiledAccounts now returns only the upper bound of the precompiled accounts, assuming they will remain contiguous, starting from address 1.
Perhaps it's worth adding this information as an inline comment into the definition?

geo2a (commented):
This is a great optimization! I think we need to make sure it is well-documented though. I think adding an inline comment explaining what the `#precompiledAccounts` function now computes would be enough.

PetarMax:
I've added a simple test that does nothing except having a bunch of `#isPrecompiledAccount` in the `requires` clause. It exhibits the following performance:
```
master: 
w/o precompiled in requires:  56.63user 1.53system 0:28.57elapsed 203%CPU (0avgtext+0avgdata 1217136maxresident)k
  w precompiled in requires: 204.17user 8.20system 3:00.51elapsed 117%CPU (0avgtext+0avgdata 1536000maxresident)k

petar/precompiled-acccounts: 
w/o precompiled in requires: 55.14user 1.49system 0:27.95elapsed 202%CPU (0avgtext+0avgdata 1071808maxresident)k
  w precompiled in requires: 61.99user 2.08system 0:33.45elapsed 191%CPU (0avgtext+0avgdata 1603792maxresident)k
```
which clearly demonstrates the slowdown. I am not sure how we would track this through CI, though.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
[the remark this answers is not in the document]
PetarMax: Yes, I was just thinking of doing something like that, that's the way to go.

palinatolmach:
> I am not sure how we would track this through CI, though.

Thank you @PetarMax! I believe we don't necessarily need to track the performance of these tests in KEVM CI, but they're still useful when profiling e.g. [backend changes](https://github.com/runtimeverification/hs-backend-booster/issues/355).

PetarMax:
I've now renamed the functions slightly to be aligned with their meaning, added some comments on their meaning, and added one more test for the correctness of `#precompiledAccountsSet`, which is now calculated as @ehildenb suggested.

ehildenb (approved):

ehildenb:
The test being present is enough in this case. The Haskell Backend team uses the KEVM test-suite directly when submitting PRs to their own repo in order to make sure tehy don't accidentally cause any regressions themselves. They don't do it on CI because CI machines are often having lots of contention, and so the timing numbers there cannot be trusted. Instead, they run it locally and post the numbers themselves to their PRs.

PR #2336, opened 2024-03-07
Add `libstdc++-12-dev` to dependencies in README

palinatolmach:
Closes https://github.com/runtimeverification/kontrol/issues/409.

This PR adds `libstdc++12-dev` to the list of system dependencies that need to be installed prior to KEVM and Kontrol installation on Ubuntu 22.04.


anvacaru (approved):
Approved! We should also add it in KEVM.

PR #2369, opened 2024-03-25
added toml processing capability

ovatman:
This PR adds toml config file input capability.

Following changes are provided:
- a parameter set is added to the command line arguments to input toml file name and profile
- `parse_toml_args` function reads all the options available in the toml file provided, taking into account the provided profile name
- `get_option_string_destination`function collects the mathing option string according to the provided key in toml file by utilizing the newly added `from_option_string` method in `*Options` classes.
- some methods and classes are moved to the `pyk.cli`package to provide better testability
- unit tests and test harness added for toml input sanity tests
- required modifications made on code regarding the new functionality and the ones that ar moved to another package.


kevm-pyk/src/kevm_pyk/cli.py
```diff
@@ -34,6 +54,269 @@ def node_id_like(s: str) -> NodeIdLike:
...
+        if opt_string[:3] == 'no_':
+            toml_adj_args[opt_string[3:]] = not v
+        elif k == 'optimization-level':
+            level = toml_args[k] if toml_args[k] >= 0 else 0
+            level = level if toml_args[k] <= 3 else 3
+            toml_adj_args['o' + str(level)] = True
+        else:
+            toml_adj_args[opt_string] = v
+
+    return toml_adj_args
```
nwatson22, line 317: This function is very similar to the one already in pyk. Is it possible to use that one instead and then to do any KEVM-specific adjustments to its output?

kevm-pyk/src/kevm_pyk/cli.py
```diff
@@ -155,6 +451,15 @@ def default() -> dict[str, Any]:
             'counterexample_info': False,
         }
 
+    @staticmethod
+    def from_option_string() -> dict[str, str]:
+        return {
+            'counterexample-information': 'counterexample_info',
+            'no-counterexample-information': 'no_counterexample_info',
+            'node': 'nodes',
```
nwatson22, line 459: How would it look to specify a list type option like `nodes` in a TOML file? I was just a bit confused since the toml option is called `node` implying a single node when I'm imagining it would look something like `node = [1, 2, 3]`. Could you add a test for this?

kevm-pyk/src/kevm_pyk/cli.py
```diff
@@ -155,6 +445,19 @@ def default() -> dict[str, Any]:
...
+            'node-delta': 'node_deltas',
+        }
+
+    @staticmethod
+    def get_argument_type() -> dict[str, Callable]:
+        return {'node': node_id_like, 'node-delta': arg_pair_of(node_id_like, node_id_like)}
```
nwatson22, line 459: I think this is still a type mismatch. `KCFGShowOptions.nodes` is expected to be a list of node ids, but the TOML reading system expects a single node ID, and will assign `KCFGShowOptions.nodes` to that `int`. With the `get_argument_type` system you added maybe you can just set the argument type to `list`. This should apply to all the options where the `add_argument` call for that option has `action='append'`. That setting makes it so that the typing function specified in `add_argument`'s `type` parameter no longer matches the type that we want to end up as a field in the options class.

nwatson22 (approved):

PR #2371, opened 2024-03-26
Add [owise] to last case of #isStaticType

jberthold:
(hopefully) Fixes https://github.com/runtimeverification/hs-backend-booster/issues/558 

The function definition was missing an `[owise]` on the last case (which  can apply to anything). 
Apparently, the LLVM backend was erroneously picking this case for a `#bytes` argument.


anvacaru (approved):

ehildenb:
@jberthold please open a follow-up PR with a test that demonstrates the simplification needed here. We should not be merging things without tests, when it's a regression especially!

jberthold:
> @jberthold please open a follow-up PR with a test that demonstrates the simplification needed here. We should not be merging things without tests, when it's a regression especially!

Turns out this is a bit tricky...  
I [wrote a `runLemma-doneLemma` test for this on branch `jb/test-for-isStaticType-owise`](https://github.com/runtimeverification/evm-semantics/compare/jb/test-for-isStaticType-owise) , checking the value of `#isStaticType(#bytes(_))` as well as whether the resulting `#encodeArgs` includes the length field in the beginning.

However, this test was supposed to _fail_ on the version before the `[owise]`  was added (the branch point for my branch) but it unexpectedly _passes_. 
This can be because of the The LLVM compilation making a luckier choice when the verification module for this test is compiled than in the compilation where we saw it failing (the original report in runtimeverification/hs-backend-booster#558)
I tried adding other imports to the `VERIFICATION` module to trigger the wrong behaviour but that was unsuccessful.

PR #2374, opened 2024-03-27
Add functional claim for `[range-memUpdate-before]` from `lemmas/evm-bytes-simplificaton.k`

geo2a:
The claim should by run with `kore-rpc-booster` as a regression test for https://github.com/runtimeverification/hs-backend-booster/issues/331


geo2a:
This has been stale in a while, closing.

PR #2376, opened 2024-03-28
Test lemmas for isStaticType and resulting encodeArgs values

jberthold:
In PR 2371, an `[owise]` attribute was added to `#isStaticType`.

The original reported issue was
[hs-backend-booster/issues/558](https://github.com/runtimeverification/hs-backend-booster/issues/558), where a wrong rule was applied because of the missing `[owise]` (the LLVM backend library evaluated the side condition to the wrong value).

This PR adds two lemmas to test the behaviour. On KEVM versions before PR 2371 the tests fail when using `--use-booster` but succeed without it.


tests/specs/functional/helpers-spec.k
```diff
@@ -0,0 +1,41 @@
+requires "abi.md"
+
+module VERIFICATION
+    imports EVM-ABI
+
+    syntax Step ::= Bool | Bytes
+ // ------------------------------
+
+    syntax KItem ::= runLemma ( Step ) [symbol(runLemma)]
```
ehildenb, line 9: We can probably call this file `abi-spec.k`.

tests/specs/functional/helpers-spec.k
```diff
@@ -0,0 +1,41 @@
...
+module HELPERS-SPEC
+    imports VERIFICATION
+
+
+ // EVM Type helper
+ // ---------------
+
+    // checking the function evaluation that went wrong in LLVM (issue booster-558)
+    claim [isStaticType-fix-2371]:
+//      <k> runLemma ( #isStaticType( #bytes(b"\229\230/\221") ) ) => doneLemma ( false ) ... </k>
```
ehildenb, line 25:
```suggestion
      <k> runLemma ( #isStaticType( #bytes(b"\229\230/\221") ) ) => doneLemma ( true ) ... </k>
```

Any harm in adding this test too?

jberthold: Well, the result should still be `false` since it is `#bytes` (no fixed length). The syntax `\229` is wrong, I would have to use `\xe6\xe5\xdd` but I can add that test. Or delete the comment.

jberthold: Added the test - pushing as soon as the `--use-booster` run of the integration tests is finished.

jberthold:
Failing for a (seemingly unrelated) SMT-error in `/test-arithmetictest-test_wmul_weakly_increasing_positive-uint256-uint256-0-spec.k`. This also fails on `master` for me locally. Going to investigate it on Tuesday.

anvacaru (approved):

PR #2378, opened 2024-03-28
Deprecate `symbol` and `klabel(_)`

Baltoli:
This PR makes a mechanical replacement of `klabel(_)` attributes with the equivalent `symbol(_)` attribute; this is to address a future deprecation.


tothtamas28 (approved):
There are a few symbols that the Python code refers to, e.g.: https://github.com/runtimeverification/evm-semantics/blob/a82be984e45ccd4cb0d49ba37c99bf63d0be52be/kevm-pyk/src/kevm_pyk/kevm.py#L402-L404

Before merging, please check if any of those have changed.

anvacaru (approved):

PR #2400, opened 2024-04-19
Use parallel prover branch

nwatson22:
Integrating changes from:
- https://github.com/runtimeverification/k/pull/4256.

Enables single-proof parallelism on KEVM. (Independent branches of APRProofs will be executed on separate threads)

Changes:
- To `run_prover`, a utility used here and by kontrol, adds a `max_frontier_parallel` parameter and a `create_kcfg_explore` parameter. Both of these parameters are optional to avoid breaking kontrol.
- `max_frontier_parallel` specifies the maximum number of worker threads to start up per proof.
- `create_kcfg_explore` is a `Callable` that should return fresh `KCFGExplore`s such that all can be used simultaneously on separate threads.
- If `max_frontier_parallel` is 1 and `kcfg_explore` is supplied, run with the `Prover.advance_proof` function as before.
- If it's greater than 1 and `create_kcfg_explore` is supplied, run with `parallel_advance_proof`.
- `exec_prove` starts the kore server with its `haskell_threads` parameter set to `max_frontier_parallel`.
- Adds `--max-frontier-parallel` option to `kevm-prove` with a default of 1.

To use on command line:
```
kevm prove [spec_file] --max-frontier-parallel [workers]
```

To use in python:
```py
def creatae_kcfg_explore() -> APRProver:
    # Return a fresh APRProver using its own KoreClient, initialized with your custom settings
run_prover(proof, create_kcfg_explore=create_kcfg_explore, max_frontier_parallel=[workers])
```


ehildenb:
Was this tested on an example and demonstrated to achieve a speedup?

nwatson22:
> Was this tested on an example and demonstrated to achieve a speedup?

@ehildenb Yes, I've tested it on a kontrol proof which I converted into a KEVM spec.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -397,7 +402,28 @@ def _init_and_run_proof(claim_job: KClaimJob) -> tuple[bool, list[str] | None]:
             fallback_on=options.fallback_on,
             interim_simplification=options.interim_simplification,
             no_post_exec_simplify=(not options.post_exec_simplify),
+            port=options.port,
+            haskell_threads=options.max_frontier_parallel,
         ) as kcfg_explore:
+
+            def create_kcfg_explore() -> KCFGExplore:
```
tothtamas28, line 409: When are these resources closed?

kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -102,22 +104,57 @@ def run_prover(
...
+                    execute_depth=max_depth,
+                    terminal_rules=terminal_rules,
+                    cut_point_rules=cut_point_rules,
+                    counterexample_info=counterexample_info,
+                    always_check_subsumption=always_check_subsumption,
+                    fast_check_subsumption=fast_check_subsumption,
+                )
+
+            if max_frontier_parallel > 1 and create_kcfg_explore is not None:
+                parallel_advance_proof(
```
tothtamas28, line 126: It'd be nice to abstract these implementation details using "strategy" classes with a common interface. Then here, you could pick a strategy based on input parameters, but otherwise handle the different implementations uniformly.

kevm-pyk/src/kevm_pyk/__main__.py
nwatson22: Would there be anything besides the `KoreClient` that needs closed? Currently that should be being done in `parallel_advance_proof`. When the `with _ProverPool as pool` is exited, the `_ProverPool` calls `close()` on its list of `Prover`s it has created, and the `APRProver` implementation of `close()` closes the `KoreClient` associated with its `KCFGExplore`. However, `parallel_advance_proof` creates a prover on the main thread as well, and I realized that is not getting closed, so I will fix that.

kevm-pyk/src/kevm_pyk/utils.py
nwatson22: Do you mean in pyk in the same module as `Proof`, or localized here to where we know we're working with `APRProof`s? Assuming the former, I'm not sure what the common interface would look like. I'm picturing something like:
```py
class ProvingStrategy(ABC):
    @abstractmethod
    def prove(self, proof: P, max_iterations: int | None = None, fail_fast: bool = False) -> None: ...


class MultithreadedProvingStrategy(ProvingStrategy):
    _create_prover: Callable[[], Prover]
    _max_workers = 1

    def __init__(self, create_prover: Callable[[], Prover], max_workers: int = 1) -> None:
        self._create_prover = create_prover
        self._max_workers = max_workers

    def prove(self, proof: P, max_iterations: int | None = None, fail_fast: bool = False) -> None:
        parallel_advance_proof(
            proof=proof,
            create_prover=self._create_prover,
            max_iterations=max_iterations,
            fail_fast=fail_fast,
            max_workers=self._max_workers,
        )


class SequentialProvingStrategy(ProvingStrategy):
    _prover: Prover

    def __init__(self, prover: Prover) -> None:
        self._prover = prover

    def prove(self, proof: P, max_iterations: int | None = None, fail_fast: bool = False) -> None:
        self._prover.advance_proof(fail_fast=fail_fast, max_iterations=max_iterations, proof=proof)
```
in pyk, and here we'd do something like
```py
#create_prover definition
if max_frontier_parallel > 1 and create_kcfg_explore is not None:
    strategy = MultithreadedProofStrategy(create_prover, max_workers)
elif kcfg_explore is not None:
    # init APRProver
    strategy = SequentialProofStrategy(prover)
...
strategy.prove(max_iterations, fail_fast)
``` 
But all this accomplishes is condensing `advance_proof` and `parallel_advance_proof` into a single call.
I'm also hesitant to add too much that relies on the current version of `parallel_advance_proof` because per discussions with @ehildenb I'm thinking we want to expand its proving loop to deal with a collection of proofs rather than a single proof (see https://github.com/runtimeverification/k/issues/4318)

tothtamas28: > Do you mean in `pyk` in the same module as `Proof`, or localized here to where we know we're working with `APRProof`s?

Just here. This is not a design suggestion, just a way to make the code more readable by making it explicit that the logic is about choosing between things that are interchangeable.

kevm-pyk/src/kevm_pyk/__main__.py
nwatson22: Opened a PR for this in pyk here: https://github.com/runtimeverification/k/pull/4323

kevm-pyk/src/kevm_pyk/utils.py
nwatson22: I've implemented something similar in this file. Can you see if you think this version is cleaner?

kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -102,22 +140,58 @@ def run_prover(
...
-                execute_depth=max_depth,
-                terminal_rules=terminal_rules,
-                cut_point_rules=cut_point_rules,
-                counterexample_info=counterexample_info,
-                always_check_subsumption=always_check_subsumption,
-                fast_check_subsumption=fast_check_subsumption,
-            )
-            prover.advance_proof(proof, max_iterations=max_iterations, fail_fast=fail_fast)
+
+            def create_prover() -> APRProver:
```
tothtamas28, line 149: As a next steps, you can move these initialization steps into the respective strategy classes. Then you can extract a function that picks a strategy for you:

```python3
if create_kcfg_explore is not None:
    return MultithreadedProvingStrategy(...) max_frontier_parallel > 1 else SequentialProvingStrategy(...)
elif kcfg_explore is not None:
    return SequentialProvingStrategy(...)
else:
   raise ValueError(...)
```

You can make `execute_depth`, etc. parameters of `__init__` on the abstract class. Cosider extracting a class to hold these parameters.

kevm-pyk/src/kevm_pyk/utils.py
tothtamas28: Looks pretty good!

kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -91,9 +95,43 @@ def get_apr_proof_for_spec(
     return apr_proof
 
 
+class APRProofRunStrategy(ABC):
+    @abstractmethod
+    def prove(self, proof: P, max_iterations: int | None = None, fail_fast: bool = False) -> None: ...
+
+
+class MultithreadedProvingStrategy(APRProofRunStrategy):
```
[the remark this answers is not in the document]
nwatson22: Fixed.

kevm-pyk/src/kevm_pyk/utils.py
nwatson22: I improved it by pushing the initialization steps into the strategies.

kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -102,22 +185,53 @@ def run_prover(
...
-                execute_depth=max_depth,
-                terminal_rules=terminal_rules,
-                cut_point_rules=cut_point_rules,
-                counterexample_info=counterexample_info,
-                always_check_subsumption=always_check_subsumption,
-                fast_check_subsumption=fast_check_subsumption,
-            )
-            prover.advance_proof(proof, max_iterations=max_iterations, fail_fast=fail_fast)
+
+            strategy: APRProofStrategy
```
tothtamas28, line 194: Consider extracting a a function for instantiating the strategy.

kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -91,9 +93,90 @@ def get_apr_proof_for_spec(
...
+    _terminal_rules: Iterable[str]
+    _cut_point_rules: Iterable[str]
+    _counterexample_info: bool
+    _always_check_subsumption: bool
+    _fast_check_subsumption: bool
+
+    @abstractmethod
+    def prove(self, proof: APRProof, max_iterations: int | None = None, fail_fast: bool = False) -> None: ...
+
+    def __init__(
```
tothtamas28, line 107: `__init__` should come first.

kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -91,9 +93,90 @@ def get_apr_proof_for_spec(
     return apr_proof
 
 
+class APRProofStrategy(ABC):
+    _execute_depth: int
+    _terminal_rules: Iterable[str]
+    _cut_point_rules: Iterable[str]
+    _counterexample_info: bool
+    _always_check_subsumption: bool
+    _fast_check_subsumption: bool
```
tothtamas28, line 102: This I assume is shared between both strategies, Consider extracting a class for passing them around in one object. That should get rid of `*args, **kwargs` as well.

tothtamas28 (approved):

kevm-pyk/src/kevm_pyk/utils.py
nwatson22: Added `select_apr_strategy(...) -> APRProofStrategy` function.

kevm-pyk/src/kevm_pyk/utils.py
nwatson22: Added `RunProverParams` dataclass for this.

PR #2417, opened 2024-05-08
Pre-kompile K definitions and specs for integration tests

geo2a:
This PR modifies the RPC prover test harness in `test_prove.py`, to allow in-advance kompilation of the K definitions. This greatly improves the reliability and precision of the Haskell Backend performance measurement in scripts when we run several proofs in parallel. By separating the kompilation and proving stages, we reduce the noise caused by the K frontend.

An overview of changes:
- add a new custom `--kompiled-targets-dir` command line option to `pytest`, allowing to set persistent cache directory for the kompiled definitions used by the testsl;
- add the `kompiled_target_cache` fixture to `test_prove.py` that will read the persitent cache from disk if `--kompiled-targets-dir` is provided;
- add the `test_kompile_targets` test to `test_prove.py` that forces the kompilation of all targets. If `--kompiled-targets-dir` is provided, the targets' kompiled directories will be stored there and could be re-used by the following pytest sessions that are given the same `--kompiled-targets-dir`. This test will be skipped if `--kompiled-targets-dir` is not given, making sure the CI configuration does not have to be changed to ignore it explicitly.

For example, after executing

```
pytest src/tests/integration/test_prove.py::test_kompile_targets --kompiled-targets-dir ./prekompiled
```

the `./prekompiled` will be populated with the following subdirectories:

```
benchmarks-verification
bihu-functional-spec
bihu-verification
erc20-verification
examples-erc20-spec
examples-solidity-code-spec
examples-storage-spec
functional-abi-spec
functional-evm-int-simplifications-spec
kontrol-verification
mcd-functional-spec
mcd-verification
```

The invocation
```
pytest src/tests/integration -k "test_prove_pyk" --kompiled-targets-dir ./prekompiled
```

will now be able to re-use the kompilation artefacts and would not have to make expensive and scheduling-disrupting calls to `kompile`.


nwatson22 (approved):

kevm-pyk/src/tests/integration/test_prove.py
```diff
@@ -118,31 +119,58 @@ class Target(NamedTuple):
...
+    cache_dir = kompiled_targets_dir / 'target'
+    cache: dict[str, Path] = {}
+    if cache_dir.exists():  # cache dir exists, populate cache
+        for file in cache_dir.iterdir():
+            if file.is_dir():
+                # the cache key is the name of the target, which is the filename by-construction.
+                cache[file.stem] = file / 'kompiled'
+    else:
+        cache_dir.mkdir(parents=True)
+    return cache_dir, cache
```
tothtamas28, line 155: Consider abstracting this in a class.

kevm-pyk/src/tests/integration/test_prove.py
```diff
@@ -118,31 +119,58 @@ class Target(NamedTuple):
...
+    cache_dir, cache = kompiled_target_cache
 
     def kompile(spec_file: Path) -> Path:
         target = _target_for_spec(spec_file)
 
-        if target not in cache:
-            output_dir = cache_dir / f'{target.main_file.stem}-{len(cache)}'
-            output_dir.mkdir()
-            cache[target] = target(output_dir)
+        if target.name not in cache:
```
tothtamas28, line 168: `Target` already uniquely identifies a kompilation target, it's probably not necessary to change the key to a `str` derived from it.

kevm-pyk/src/tests/integration/test_prove.py
```diff
@@ -118,31 +119,58 @@ class Target(NamedTuple):
...
+            output_dir = cache_dir / target.name
+            output_dir.mkdir(exist_ok=True)
```
tothtamas28, line 170: Is this to cover the case when the user points to an already existing cache to rekompile it?

kevm-pyk/src/tests/integration/test_prove.py
```diff
@@ -118,31 +119,58 @@ class Target(NamedTuple):
...
+    """
+    Populate the cache of kompiled definitions from an existing file system directory. If the cache is hot, the `kompiled_target_for` fixture will not containt a call to `kompile`, saving an expesive call to the K frontend.
+    """
+    cache_dir = kompiled_targets_dir / 'target'
+    cache: dict[str, Path] = {}
+    if cache_dir.exists():  # cache dir exists, populate cache
+        for file in cache_dir.iterdir():
+            if file.is_dir():
+                # the cache key is the name of the target, which is the filename by-construction.
+                cache[file.stem] = file / 'kompiled'
```
tothtamas28, line 152: Initializing a cache might be unnecessary, you can just do the directory lookup when you request the kompiled directory.

kevm-pyk/src/tests/integration/test_prove.py
geo2a: Right, `exist_ok=True` is indeed unnecessary as the directory should never exist if the target is not already cached.

kevm-pyk/src/tests/integration/test_prove.py
geo2a: I don't understand what you mean: could you elaborate?

 My goals are:
* keep using `cache` in so that the old behavior of the `kompiled_target_for` fixture is retained,
* make sure that `cache` is never actually updated when `--kompiled-targets-dir` is passed.

kevm-pyk/src/tests/integration/test_prove.py
geo2a: It is not possible to construct a `Target` when initializing the cache from a directory in `kompiled_target_cache`, as I have nothing to construct the `main_file` field from.

kevm-pyk/src/tests/integration/test_prove.py
tothtamas28: As I see it, tests basically require an interface that takes a `Target` and returns a `Path` to the kompiled target (in the actual tests there might be another indirection from spec files to targets, but I'll not consider this for simplicity).

So you just have to ensure that `Target -> Path` is reproducible between runs, then accommodating both use cases (prekompiled vs on-the-fly) is straightforward.

* If the `Path` does not exist, lock and kompile
* Otherwise, return the `Path`.

For prekompiled mode, you pass in a non-empty directory as base. For on-the-file mode, you pass in an empty temporary directory.

`cache`, etc. can be dropped, the actual cache is the file system.

geo2a: Yeah, that sound good, I understand what you mean. However, I'd rather get my incremental change merged quickly and now spend too much time refactoring the test suite. Would that be OK if I let you refactor it as you see fit?

kevm-pyk/src/tests/integration/test_prove.py
geo2a: Hmm, turns out it is actually easier if we keep `exist_ok=True`, and let the tests wait on the lock to populate/read it, rather then fail to create it if it exists.

kevm-pyk/src/tests/integration/test_prove.py
tothtamas28: I can live with that, but please open an issue for the refactoring and assign it to me.

geo2a: Thanks, will do!

PR #2421, opened 2024-05-09
Enable SimpleSub type inference mode for kompile and kprove

Scott-Guest:
Part of runtimeverification/k#4315

Sets `--type-inference-mode simplesub` for `kompile` and `kprove`.

CI passes with `--type-inference-mode checked`, which confirms that the SimpleSub and Z3 algorithms agree.


kevm-pyk/src/kevm_pyk/kompile.py
```diff
@@ -61,6 +62,7 @@ def kevm_kompile(
     debug_build: bool = False,
     debug: bool = False,
     verbose: bool = False,
+    type_inference_mode: str | TypeInferenceMode | None = TypeInferenceMode.SIMPLESUB,
```
tothtamas28, line 65: Consider

```suggestion
    type_inference_mode: str | TypeInferenceMode | None = None,
```

Then handle the default within the function.

tothtamas28 (approved):

PR #2430, opened 2024-05-16
Simplify kompilation for proof tests

### LINKED ISSUE #2419: Refactor kompilation cache in the intergration tests
As a follow-up to https://github.com/runtimeverification/evm-semantics/pull/2417, @tothtamas28 suggest we should do the following: 

> As I see it, tests basically require an interface that takes a Target and returns a Path to the kompiled target (in the actual tests there might be another indirection from spec files to targets, but I'll not consider this for simplicity).
>
> So you just have to ensure that Target -> Path is reproducible between runs, then accommodating both use cases (prekompiled vs on-the-fly) is straightforward.
>
>     If the Path does not exist, lock and kompile
>     Otherwise, return the Path.
>
> For prekompiled mode, you pass in a non-empty directory as base. For on-the-file mode, you pass in an empty temporary directory.
>
> cache, etc. can be dropped, the actual cache is the file system.

tothtamas28:
Fixes #2419

* Remove the `kompiled_target_cache` fixture
* Do not require directory passed as `--kompiled-targets-dir` to exist
* Simplify target directory layout
* Use a soft file lock
* Renamings


kevm-pyk/src/tests/conftest.py
```diff
@@ -31,8 +29,8 @@ def pytest_addoption(parser: Parser) -> None:
         help='Run only this specific specification (skip others)',
     )
     parser.addoption(
-        '--kompiled-targets-dir',
-        type=dir_path,
+        '--kompiled-target-dir',
```
geo2a, line 32: May I ask you to not change this name, as our performance script [uses](https://github.com/runtimeverification/haskell-backend/blob/6e5778899e91ff3c951221bdb620c69b5ec41aa3/scripts/performance-tests-kevm.sh#L105) the old one?

kevm-pyk/src/tests/integration/test_prove.py
```diff
@@ -189,22 +178,22 @@ def _target_for_spec(spec_file: Path) -> Target:
     ALL_TESTS,
     ids=[str(spec_file.relative_to(SPEC_DIR)) for spec_file in ALL_TESTS],
 )
-def test_kompile_targets(spec_file: Path, kompiled_target_for: Callable[[Path], Path], request: FixtureRequest) -> None:
+def test_kompile(
```
geo2a, line 181: Same request as above, would it be possible to not change this name too?

kevm-pyk/src/tests/conftest.py
tothtamas28: Done.

PR #2432, opened 2024-05-17
Updates to the conformance test suite

ehildenb:
~Blocked on: https://github.com/runtimeverification/k/pull/4370~
~Blocked on: https://github.com/runtimeverification/evm-semantics/pull/2439~

- Update all actions/checkout from v3 => v4 and Gr1N/setup-poetry from v8 => v9 to get rid of NodeJS deprecation warnings on actions runs.
- Updates Poetry to v1.8.3.
- Switch us off the `LegacyTests` directory in the Ethereum test-suite completely.
- Avoid cloning submodules recursively in almost all cases (saves 1-2min per job or so).
  - Interactive tests in unit/profiling/nix have the files they used copied directly into this repository instead.
  - Conformance test stage still does recursive submodule clone, because it needs the test-suite. In this case, it's still faster to let GH action runner manage the clone with caching.
  - Many stages only recursively clone the plugin submodule manually after doing a non-recursive code checkout.
- Break each conformance test into an individual run, instead of batch-running them. Logging messages for each as well.
- Does not run the VMTests twice (once in BChain tests and once on its own).
- Does not run each conformance test twice (once in normal tests, once in legacy).


ehildenb:
Before:

image

After:

image

Baltoli:
Blocked on: https://github.com/runtimeverification/k/pull/4370 reaching KEVM

anvacaru (approved):

flake.nix
```diff
@@ -15,20 +15,10 @@
...
     };
-    ethereum-tests = {
-      url = "github:ethereum/tests/6401889dec4eee58e808fd178fb2c7f628a3e039";
-      flake = false;
-    };
-    ethereum-legacytests = {
-      url =
-        "github:ethereum/legacytests/d7abc42a7b352a7b44b1f66b58aca54e4af6a9d7";
-      flake = false;
-    };
```
Baltoli, line 26: I presume the logic here is that we're never actually running these tests in a Nix build of KEVM, and so can drop them from being inputs?

Baltoli (approved):

flake.nix
ehildenb: Yep. All the data they needed (2 files) have been copied into teh repo.

PR #2437, opened 2024-05-21
Computing valid jump destinations for non-concrete programs

PetarMax:
This PR extends the computation of valid jump destinations for cases in which the program to be analysed is not fully concrete (e.g., it contains symbolic immutable variables). It is assumed that any symbolic content cannot in itself be/hold a valid jump destination.


PetarMax:
We have agreed on a different approach, closing PR.

PR #2440, opened 2024-05-21
Increase SMT timeout in integration tests

geo2a:
A [PR](https://github.com/runtimeverification/haskell-backend/pull/3870) to the Haskell backend will strengthen Booster's simplifications, potentially making the predicates we send to the SMT solver from Booster more complicated. Some of the recently [re-enabled tests ](https://github.com/runtimeverification/evm-semantics/pull/2431) fail with `kore-rpc-booster` due to Booster's SMT solver interface timing out. Until we have implemented a [retry](https://github.com/runtimeverification/haskell-backend/issues/3884) mechanism in Booster, I propose to increase the solver time-out.


geo2a:
This is not needed anymore, as we do now have a more robust SMT sovler interface in Booster.

PR #2441, opened 2024-05-22
Compute jumpdests using the KCFGSemantics.custom_step heuristic

anvacaru:
~-blocked on: https://github.com/runtimeverification/k/pull/4376~
~-blocked on: https://github.com/runtimeverification/evm-semantics/pull/2450~

Implementing a `custom_step` heuristics for KEVMSemantics to compute the jumpdests using kevm-pyk.
Adding the `#loadProgram` as a cut-rule.

PR #2460, opened 2024-06-03
Avoid PUSH data when computing jumpdests

anvacaru:
Fixing a mistake I made in my previous PR https://github.com/runtimeverification/evm-semantics/pull/2441. When computing the valid JUMPTESTS, I omitted the case in which the `0x5B` byte could have been part of a `PUSH(N)` opcode.

I've modified the `_compute_jumpdests` function to first check if the current byte is a `PUSH(N)` and skip `byte - PUSH1 +2` bytes.

i.e. if byte is `PUSH1`, it skips `0x60 - 0x60 + 2` bytes where one byte is the data to be pushed and one byte is the opcode itself.


kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -699,4 +699,16 @@ def _process_jumpdests(bytecode: bytes, offset: int) -> list[KToken]:
...
+    jumpdest = 0x5B
+    i = 0
+    jumpdests: list[KToken] = []
+    while i < len(bytecode):
+        if push1 <= bytecode[i] <= push32:
+            i += bytecode[i] - push1 + 2
+            continue
+        if bytecode[i] == jumpdest:
+            jumpdests.append(intToken(offset + i))
+        i += 1
```
ehildenb, line 713:
```suggestion
        if push1 <= bytecode[i] <= push32:
            i += bytecode[i] - push1 + 2
            continue
        else:
            if bytecode[i] == jumpdest:
                jumpdests.append(intToken(offset + i))
            i += 1
```

Perhaps like this instead? Whichever reads better to you.

anvacaru: Done.

PetarMax (approved):

PR #2469, opened 2024-06-06
First draft of the USL API

theo25:
This is a first draft for the low-level API that will be used to communicate with the DSL.

Note that many of the data types used in this draft are simply 64-bit integers. These are placeholders and I welcome your input as to what would be a recommended data type for each such placeholder.


ehildenb:
Are we sure this needs to be merged into KEVM? Would it be more appropriate as a repo on the pi2 side that includes KEVM?

ehildenb:
At the very least, if it's to be merged into KEVM, then it should come with a testing harness here as well, and some tests of basic functionality/sanity, because it will become the responsibility of our team to maintain this functionality over time (rather than pi2 team) when we need to make changes.

theo25:
@ehildenb I tend to agree that this should probably live somewhere in the pi2 side. I put it here for now because it was the faster way to have the kevm runtime available to link with.

And of course, I agree that we need significant test harness. This is just a draft so that we can come to an agreement as to what this API should look like and specifically what kind of information needs to be communicated between KEVM and the USL client for executing transactions.

By the way, I am happy to add you as reviewer and get your input if you have the time.

theo25:
I have added a similar PR to our Pi2 fork of the EVM semantics and I am closing this one. We will be working on this on the fork and eventually push here when things are finalized and well-tested.

PR #2471, opened 2024-06-07
Have <jumpDests> as Bytes instead of Set

anvacaru:
Having the valid jumpdests computed AOT and stored in a `Bytes` instead of `Set` increases the performance considerably.

In addition, the `--break-on-load-program` flag is added, to trigger the custom-step logic.


anvacaru:
Some conformance tests are currently failing with `stderr`
```
terminate called after throwing an instance of 'std::invalid_argument'
  what():  [hook_BYTES_get]: Buffer overflow on get: off=4294967295, len=77
```

from kevm-pyk:
```
make test-integration PYTEST_ARGS+="-k 'test_conformance.py and BlockchainTests/GeneralStateTests/stMemoryStressTest/JUMPI_Bounds.json'"
```

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
```diff
@@ -1039,14 +1039,14 @@ The `JUMP*` family of operations affect the current program counter.
...
+    rule <k> JUMP DEST => #endBasicBlock ... </k>
          <pc> _ => DEST </pc>
          <jumpDests> DESTS </jumpDests>
-      requires DEST in DESTS
+      requires DESTS[DEST] ==Int 1
 
     rule <k> JUMP DEST => #end EVMC_BAD_JUMP_DESTINATION ... </k>
          <jumpDests> DESTS </jumpDests>
-      requires notBool DEST in DESTS
+      requires DESTS[DEST] ==Int 0
```
PetarMax, line 1049: Same as above.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
```diff
@@ -1039,14 +1039,14 @@ The `JUMP*` family of operations affect the current program counter.
 
     syntax UnStackOp ::= "JUMP"
  // ---------------------------
-    rule <k> JUMP DEST => #endBasicBlock... </k>
+    rule <k> JUMP DEST => #endBasicBlock ... </k>
          <pc> _ => DEST </pc>
          <jumpDests> DESTS </jumpDests>
-      requires DEST in DESTS
+      requires DESTS[DEST] ==Int 1
```
PetarMax, line 1045: We need a check that `DEST` is within bounds of the buffer.

PetarMax, line 1045:
```suggestion
      requires DEST <Int lengthBytes(DESTS) andBool DESTS[DEST] ==Int 1
```
I think we're safe from this being less than zero, but if not, we can also add `0 <=Int DEST`.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
PetarMax, line 1049:
```suggestion
      requires DEST <Int lengthBytes(DESTS) andBool DESTS[DEST] ==Int 0
```

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
anvacaru: Right, great catch!

kevm-pyk/src/kevm_pyk/cli.py
```diff
@@ -1119,6 +1121,13 @@ def explore_args(self) -> ArgumentParser:
             action='store_true',
             help='Store a node for every EVM basic block (implies --break-on-calls).',
         )
+        args.add_argument(
+            '--break-on-load-program',
+            dest='break_on_load_program',
+            default=None,
+            action='store_true',
+            help='Store a node every time the jumpDests are computed, allowing the custom step to apply.',
```
PetarMax, line 1129: Ah, sorry, could we be a bit more informative here, what is the custom step now? Why would we want that, actually?

anvacaru: Done!

kevm-pyk/src/kevm_pyk/cli.py
```diff
@@ -1119,6 +1121,13 @@ def explore_args(self) -> ArgumentParser:
             action='store_true',
             help='Store a node for every EVM basic block (implies --break-on-calls).',
         )
+        args.add_argument(
+            '--symbolic-immutables',
+            dest='break_on_load_program',
+            default=None,
+            action='store_true',
+            help='Enable support for symbolic immutables in Solidity constructor code.',
```
palinatolmach, line 1129:
```suggestion
            help='Enable support for symbolic parameters in Solidity constructor code.',
```
I understand it's a nit and `--symbolic-immutables` looks good as the argument name, but technically that enables any symbolic arguments being passed to constructor, not necessarily for immutables. Please feel free to ignore!

kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -670,43 +674,43 @@ def compute_jumpdests(sections: list[KInner]) -> KInner:
...
             width_token = s.args[0]
             assert type(width_token) is KToken
-            offset += int(width_token.token)
+            mutable_jumpdests += bytes(int(width_token.token))
         elif type(s) is KToken and s.sort == BYTES:
             bytecode = pretty_bytes(s)
-            jumpdests.extend(_process_jumpdests(bytecode, offset))
-            offset += len(bytecode)
+            mutable_jumpdests += _process_jumpdests(bytecode)
+            len(bytecode)
```
palinatolmach, line 686: Is it a typo?

kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -670,43 +674,43 @@ def compute_jumpdests(sections: list[KInner]) -> KInner:
...
+def _process_jumpdests(bytecode: bytes) -> bytes:
     """Computes the location of JUMPDEST opcodes from a given bytecode while avoiding bytes from within the PUSH opcodes.
 
     :param bytecode: The bytecode of the contract as bytes.
     :type bytecode: bytes
     :param offset: The offset to add to each position index to align it with the broader code structure.
     :type offset: int
-    :return:  A list of intToken instances representing the positions of all found jump destinations in the bytecode adjusted by the offset.
-    :rtype: list[KToken]
+    :return: A bytes object with the same length as the input of zeros. Valid jumpdests are represented by the value 1.
```
palinatolmach, line 700: I'm not sure, what is "input of zeros" here? Isn't it same length than bytecode with the zero input, or something like that?

palinatolmach (approved):
Left a couple of comments, but looks good to me otherwise! If Kontrol tests work too and @PetarMax approves, I think it's good to merge.

kevm-pyk/src/kevm_pyk/kevm.py
palinatolmach: I mean line 686 and `len(bytecode)` specifically.

kevm-pyk/src/kevm_pyk/cli.py
anvacaru: Should I change to `--symbolic-constructor` instead?

palinatolmach: Yes, sounds great to me!

kevm-pyk/src/kevm_pyk/kevm.py
anvacaru: yup, good catch! `len(bytecdeo)` was used to increment the `offset`. I removed the `offset` but I guess I did not catch this.

kevm-pyk/src/kevm_pyk/kevm.py
anvacaru: Updated.

PR #2477, opened 2024-06-11
Pass remaining benchmarks, lemmas, and delete unused proofs

ehildenb:
Factored out of: https://github.com/runtimeverification/evm-semantics/pull/2472. Related: https://github.com/runtimeverification/k/issues/4359.

In particular:

- The last failing test in `benchmarks/ecrec` is passing with one additional lemma on booster and enabled.
- The last failing test in `benchmarks/functional-spec` is commented out and now that file is enabled on booster.
- Several tests that are not currently being run with any backend (and are not passing) are deleted:
  - A single spec in `tests/specs/kontrol`
  - 6 proofs in `tests/specs/mcd`
- The test `tests/specs/functional/lemmas-no-smt-spec.k` is enabled on booster.
- 6 specifications in `tests/specs/functional/lemmas-spec.k` that are failing on booster are commented, and that test-suite is now enabled on booster. That test-suite was (accidentally) not being run by any backend prior.

PR #2482, opened 2024-06-12
Improve performance of large lemmas proof suite

ehildenb:
https://github.com/runtimeverification/evm-semantics/pull/2477 introduced running `tests/specs/functional/lemmas-spec.k` on the RPC prover. But the RPC prover runs these sequentially, and has a lot of overhead for each individual claim, of which there are >250.

This enables running some proof-suites in parallel, in this case we run the `lemmas-spec.k` proof suite with 8 workers.

Related: https://github.com/runtimeverification/k/issues/4359


ehildenb:
This takes us from 90 minutes here: https://github.com/runtimeverification/evm-semantics/actions/runs/9468016968/job/26083573853

To 50 minutes on this PR.

nwatson22 (approved):

PR #2518, opened 2024-07-08
Add option to use sequential proof loop

nwatson22:
Adds `--force-sequential` option to `kevm prove` which will instantiate one prover and call `advance_proof` on it rather than the default behavior now which is to pass the prover creation function to `parallel_advance_proof`.


geo2a (approved):
Look good to me, thanks @nwatson22!

PR #2520, opened 2024-07-09
EIP-1153: Implement Transient Storage

anvacaru:
In this PR:

- implementing transient storage ( [eip-1153](https://eips.ethereum.org/EIPS/eip-1153#specification))
- adding a <transientStorage> cell that's similar to the persistent storage cell in the <accounts> cell.
- Implementing `TSTORE` and `TLOAD` opcodes.
TODO: Implement a way to clean transient storage after the transaction has finished executing or at the beginning of the transaction.
- All the tests except `eip1153_tstore/reentrant_selfdestructing_call` are passing. `eip1153_tstore/reentrant_selfdestructing_call` is blocked on [eip-6780](https://eips.ethereum.org/EIPS/eip-6780).

PR #2522, opened 2024-07-10
Add transient_storage parameter to KEVM.account_cell

nwatson22:
Follow-up to #2520 to allow integration of transient storage into kontrol


anvacaru (approved):
Nice!!

PR #2523, opened 2024-07-10
Fix indexed reference to nonce cell

nwatson22:
I found another issue in KEVM with the updated accounts cell configuration that I forgot to include with #2522. We were obtaining the `<nonce>` cell by index, and the index changed, making us confuse that cell with the `<transientStorage>` cell.


palinatolmach (approved):
thanks @nwatson22!

PR #2527, opened 2024-07-16
Add a new `chop` simplification lemma.

jberthold:
This lemma is required to make `vow-fess-fail-rough` pass once the backend uses the SMT solver to evaluate and simplify expressions (forthcoming change, https://github.com/runtimeverification/haskell-backend/pull/3904)

What happens with the SMT solver is that some calls to `chop` are simplified away, but others aren't. This leads to a spurious branch because a `JUMPI` condition cannot be evaluated to `false` any more. The SMT solver loses information when applying `chop-idempotent`.

For more context see [the slack discussion here](https://runtimeverification.slack.com/archives/CC360GUTG/p1721069197632439) and [here](https://runtimeverification.slack.com/archives/CC360GUTG/p1721062195080599?thread_ts=1721041777.289639&cid=CC360GUTG).


kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
```diff
@@ -94,4 +94,8 @@ module EVM-INT-SIMPLIFICATION-COMMON
     rule [chop-sum-left]:  chop ( chop ( X:Int ) +Int Y:Int ) => chop ( X +Int Y ) [simplification]
     rule [chop-sum-right]: chop ( X:Int +Int chop ( Y:Int ) ) => chop ( X +Int Y ) [simplification]
 
+    rule [chop-in-bounds]: X:Int <=Int chop ( X +Int Y:Int ) => X +Int Y <Int pow256
+                         requires #rangeUInt(256, X) andBool #rangeUInt(256, Y)
+                         [simplification]
```
PetarMax, line 99: Minor OCD-like suggestion.
```suggestion
    rule [chop-no-overflow-add]: 
      X:Int <=Int chop ( X +Int Y:Int ) => X +Int Y <Int pow256
      requires #rangeUInt(256, X) andBool #rangeUInt(256, Y)
      [simplification]
```

PetarMax:
Perhaps we could rename the misnamed `[chop-idempotent]` to `[chop-resolve]` as well as part of this PR?

PetarMax (approved):

PR #2541, opened 2024-07-29
Execute multiple blocks in conformance tests

anvacaru:
For context, there are 4 types of keys in the testing harness: `loadKeys`, `execKeys` ,`checkKeys` and `discardKeys` and there are rules that ensure the order of execution is `loadKeys ~> execKeys ~> checkKeys`.

`execKeys` start the execution of the evm with values loaded from `loadKeys`, while `checkKeys` validate the execution.

**Problem:**  Execution blocks are skipped in blockchain tests.

Previously, the concrete execution harness from `driver.md` would only execute the first block in a test.
The `"block"` key is defined as a `load` key, and once the first block is flattened, the remaining blocks are ignored. 

```k
    rule <k> run _TESTID : { "blocks" : [ { KEY : VAL , REST1 => REST1 }, _ ] , ( REST2 => KEY : VAL , REST2 ) } ... </k>
    rule <k> run  TESTID : { "blocks" : [ { .JSONs }, _ ] , REST } => run TESTID : { REST }                      ... </k>
```

**Changes:**
- "blocks" is now an `exec` key, to ensure that the `genesisRLP` has been decoded and the `genesisBlockHeader` has been validated first.
- there is a new `EthereumComand process JSON` that is used to load a block from the test (by decoding the `rlp` field), start the execution and validate all the block level checks: `blockHeader`, `transaction`, `blockNumber`.
- `clearTx` and `clearBlock` are used to reset the cells before a block is loaded.
- `lastblockhash` is discarded since it wasn't checked before.


anvacaru:
@ehildenb Yes, but it depends on how the change affects the test. Only the `blockheader` fields are checked after a block is executed. So if the change modifies the `gasUsed`, for example, it will be caught early. If the change is to an account's balance or storage value, and it doesn't affect the gas values, then it will be caught only at the end, when the `post` accounts checks take place. There are no intermediate account states to test.

tests/failing/ContractCreationSpam_d0g0v0.json.expected
```diff
@@ -1,6 +1,6 @@
...
...00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000" , "coinbase" : "0x2adc25665018aa1fe0e6bc666dac8fc2697ff9ba" , "difficulty" : "0x020000" , "extraData" : "0x42" , "gasLimit" : "0x174876e800" , "gasUsed" : "0x00" , "hash" : "0x24a30e4305ac41674b26493c800c05f507e98d3b8bceb0a314f9b9bc43622736" , "mixHash" : "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421" , "nonce" : "0x0102030405060708" , "number" : "0x00" , "parentHash" : "0x0000000000000000000000000000000000000000000000000000000000000000" , "receiptTrie" : "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421" , "stateRoot" : "0xd54ec35ca872c653659ad91eaa355665c286183b541a0519b3ab765a2d6733c3" , "timestamp" : "0x03b6" , "transactionsTrie" : "0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421" , "uncleHash" : "0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347" , .JSONs } , .JSONs } ~> clear ~> run { .JSONs } ~> success .EthereumSimulation ~> .K
```
gtrepta, line 3: Why do we check against the specific output for the failing tests? Wouldn't you normally just want to check that the K cell is `.K`?

gtrepta (approved):
LGTM. Just one question in the comments.

tests/failing/ContractCreationSpam_d0g0v0.json.expected
anvacaru: Yes, for conformance tests we check the `<k>` cell. However, these two expected files are part of the [test integration stage](https://github.com/runtimeverification/evm-semantics/blob/master/Makefile#L70-L71), that only runs two or three tests.

PR #2543, opened 2024-07-29
Add status bar support

nwatson22:
Adds parameters to `run_prover` which allow kontrol, or KEVM if we want later, to pass a `rich.progress.Progress` bar and task to send information to about a proof as it's running. Assumes that the status bar has a custom `summary` field which it passes the `one_line_summary` of the current proof.


palinatolmach (approved):
lgtm! If it works with https://github.com/runtimeverification/kontrol/pull/728, please feel free to merge, or can also wait for @anvacaru's review.

PR #2546, opened 2024-08-01
EIP-4788: Beacon Roots contract

anvacaru:
Specification: https://eips.ethereum.org/EIPS/eip-4788

- The beacon roots contract at `BEACON_ROOTS_ADDRESS = 0x000F3df6D732807Ef1319fB7B8bB8522d0Beac02` should be called when a new block is initialized in order to store the timestamp and beacon root hash.
- If no code exists at `BEACON_ROOTS_ADDRESS`, the call must fail silently.
- Instead of triggering a call to the `BEACON_ROOTS_ADDRESS`, I'm inserting the values directly in the storage, as mentioned in the spec: 
> Clients may decide to omit an explicit EVM call and directly set the storage values. Note: While this is a valid optimization for Ethereum mainnet, it could be problematic on non-mainnet situations in case a different contract is used.

Not all beacon_roots tests are passing yet because they depend on unimplemented eips.


palinatolmach (approved):
Looks great to me!

palinatolmach:
As a side thought, maybe we can enforce some relation between the schedule and timestamp in Kontrol? I.e., if `CANCUN` is used, then we add constraint on `block.timestamp` being `>=  1710338135` for consistency? But I'm not sure if we ever branch on `block.timestamp` value being too small.

anvacaru:
@palinatolmach Yes, that could work.We should also add the `BEACON_ROOTS_ADDRESS` account in the initial state, together with the storage values, as `#startBlock` is not triggered in Kontrol.

PR #2550, opened 2024-08-02
install-build-deps: Automate installation of required system packages

0cjs:
[Message edited to match commit #64e019b.]

install-build-deps: Automate installation of required system packages

These used to be listed in README.md; some notes on the various systems still remain there.

The framework for this script is taken from the same script in [k].

This has been tested on Ubuntu 22.04 and recent Arch. The MacOS code from the [k] version has been tested on MacOS, and this is so similar (including the use of `stack`) that there should be no issues with it. The Arch support is untested, but is so simple it's hard to see what could go wrong.

[k]: https://github.com/runtimeverification/k

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,102 +61,34 @@ This repository generates the build-products for each backend in `$XDG_CACHE_HOM
 
 ### System Dependencies
 
+Run `install-build-deps` to install the required OS-supplied dependencies.
 
-#### Ubuntu
+There are some additional notes for specific systems:
 
-On Ubuntu >= 22.04 (for example):
+Ubuntu:
+- The script works for ≥ 22.04.
+- On Ubuntu < 18.04, you'll need to skip `libsecp256k1-dev` and instead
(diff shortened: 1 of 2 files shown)


0cjs:
The Version Bump check seems to be failing due to configuration problems; it's complaining about a missing token. Perhaps a re-run would fix it?

The Proofs: Functional check is failing for reasons I don't understand, but I don't see how it can be related to this commit, which adds a new file that it can't possibly using, and removes/changes text from the README.

anvacaru:
@0cjs The `Proofs: Functional` check most likely fails because of the CI flakiness. It should pass after one or more retriggers. However, the `Version Bump` fails because the PR is external (cc: @F-WRunTime). I can open a proxy PR with your commit later today.

anvacaru:
Closing this in favor of https://github.com/runtimeverification/evm-semantics/pull/2555.

PR #2551, opened 2024-08-02
More opcode optimizations

ehildenb:
Inspired by the work at Pi2 in optimizing the semantics: https://github.com/runtimeverification/evm-semantics/compare/master...Pi-Squared-Inc:evm-semantics:master. In particular: https://github.com/runtimeverification/evm-semantics/commit/febfe1d8f4d1bbbd0e4d7da77f84a9224eaf0277.

One thing to note, in the past it was not advantageous to add more optimizations because of how the old haskell backend selected the next rule to apply. Each new optimization mean another rule that had to _fail_ to apply every execution step that the optimized rule wasn't applied. The new booster backend, with it's rule indexing, shouldn't consider the higher-priority rules that aren't possible due to the K cell not matching, and so I suspect that adding new optimization rules no longer is a slowdown for the booster at all. So we may consider adding even more optimizations.

This PR:

- Adds optimized variants of `ISZERO`, `JUMPDEST`, `JUMP`, and `JUMPI`.
- Partially evaluates the `#sizeWordStack` expressions that show up on these rules.


ehildenb:
It appears that these changes don't have an appreciable impact on performance of symbolic execution, and cause some bad looping behaviors in some cases. Going to try just the wordstack expression partial evaluation, otherwise close.

ehildenb:
Only the partial evaluation of stack operations also didn't have an appreciable effect, so closing this.

PR #2557, opened 2024-08-05
Enable Outside Collaborators to run test suite before PR's are reviewed / merged

F-WRunTime:
test-pr.yml: Drop use of secrets to enable outside collaborators to t…est their code in a PR


anvacaru (approved):

F-WRunTime:
To avoid breaking the release process. Leaving the version file as is on master for now. It will later only be revised on the release branch and on the master branch should only represent the desired major / minor versions. These are set in PRs merged to master. Where it will then overwrite the release branch with the new Major.Minor version and from there later merges to master will revise the patch version number on release on merges to master.

PR #2559, opened 2024-08-06
 EIP-4895: Beacon chain push withdrawals as operations

anvacaru:
blocked on https://github.com/runtimeverification/evm-semantics/pull/2696

Implementing withdrawal operations as specified in [EIP-4895](https://eips.ethereum.org/EIPS/eip-4895).

- adding a new `<withdrawals>` cell map subconfiguration to the `<network>` that is used to keep track of withdrawals.
Similar to how `<messages>` is implemented, `<withdrawals>` has two additional Lists:
    - `<withdrawalsPending>` to keep track of the withdrawals that need to be applied.
    - `<withdrawalsOrder>` to preserve the order of the withdrawals during the `check` rules after the execution has finished.
 - adding new `EthereumCommands` productions:
    - `load "withdraw"` that handles the rlp decoding of the withdrawals array. For each item, the following two productions will be applied: 
    - `mkWD` - that creates a new `<withdrawal>` cell map item.
    - `loadWithdraw` - that initialises the newly created `<withdrawal>` with values that are provided from the rlp payload.
    - `newWithdrawalID` - function that is used to generate a new index**.
- updating `.k` specs

** `newWithdrawalID` shouldn't be necesary, as according to the specs:
>Withdrawals provide key information from the consensus layer:
>     1. a monotonically increasing `index`, starting from 0, as a `uint64` value that increments by 1 per withdrawal to uniquely identify each withdrawal
>     4. a nonzero `amount` of ether given in Gwei (1e9 wei) as a `uint64` value.

So the provided withdrawal `index` should be enough to keep track of the withdrawals.
However, the conformance tests have multiple withdrawals with the `0` index and `0` amount ([example](https://github.com/ethereum/tests/blob/develop/BlockchainTests/GeneralStateTests/Pyspecs/shanghai/eip4895_withdrawals/zero_amount.json#L50-L60)).


kevm-pyk/src/kevm_pyk/kproj/evm-semantics/driver.md
```diff
@@ -382,14 +382,14 @@ Note that `TEST` is sorted here so that key `"network"` comes before key `"pre"`
  // ---------------------------------------
     rule <k> #halt ~> check J:JSON => check J ~> #halt ... </k>
 
-    rule <k> check DATA : { .JSONs } => .K ... </k> requires DATA =/=String "transactions"
-    rule <k> check DATA : [ .JSONs ] => .K ... </k> requires DATA =/=String "ommerHeaders"
+    rule <k> check DATA : { .JSONs } => .K ... </k> requires notBool DATA in (SetItem("transactions") SetItem("withdrawals"))
```
JuanCoRo, line 385: Given the size of the set, probably the impact is null, but wouldn't it be more efficient to evaluate a side condition like `DATA =/=String "transactions orBool DATA =/=String "withdrawals"` instead of a set-based one?

anvacaru: Actualy, I'm not sure how big is the overhead caused by using `Set` here, but given that `driver.md` is only used for the conformance test suite using the llvm backend, I'd say it wouldn't make a difference. I used it here to be consistent with already existing code ([sample](https://github.com/runtimeverification/evm-semantics/blob/master/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/driver.md?plain=1#L389)).

tests/failing.llvm
```diff
@@ -134,17 +131,9 @@ BlockchainTests/GeneralStateTests/Pyspecs/frontier/opcodes/double_kill.json,*
...
-BlockchainTests/GeneralStateTests/Pyspecs/shanghai/eip4895_withdrawals/self_destructing_account.json,*
-BlockchainTests/GeneralStateTests/Pyspecs/shanghai/eip4895_withdrawals/use_value_in_contract.json,*
-BlockchainTests/GeneralStateTests/Pyspecs/shanghai/eip4895_withdrawals/use_value_in_tx.json,*
-BlockchainTests/GeneralStateTests/Pyspecs/shanghai/eip4895_withdrawals/withdrawing_to_precompiles.json,*
-BlockchainTests/GeneralStateTests/Pyspecs/shanghai/eip4895_withdrawals/zero_amount.json,*
+BlockchainTests/GeneralStateTests/Pyspecs/shanghai/eip4895_withdrawals/withdrawing_to_precompiles.json,src/GeneralStateTestsFiller/Pyspecs/shanghai/eip4895_withdrawals/test_withdrawals.py::test_withdrawing_to_precompiles[fork_Cancun-precompile_0x000000000000000000000000000000000000000a-blockchain_test-amount_0]
+BlockchainTests/GeneralStateTests/Pyspecs/shanghai/eip4895_withdrawals/withdrawing_to_precompiles.json,src/GeneralStateTestsFiller/Pyspecs/shanghai/eip4895_withdrawals/test_withdrawals.py::test_withdrawing_to_precompiles[fork_Cancun-precompile_0x000000000000000000000000000000000000000a-blockchain_test-amount_1]
```
anvacaru, line 136: These two remain skipped as the `0x0a` precompiled contract has not yet been added.

dwightguth (approved):
I didn't review the whole PR in detail, but it largely looks fine and I did review the changes to block finalization in detail and they look correct.

You might consider using some kind of named constant to represent the conversion factor from GWEI to WEI though. It's not critical, but something to consider for readability.

PR #2571, opened 2024-08-09
Optimize how we run proofs with booster

ehildenb:
This turns on some flags by default that pass all the tests, and produce a nice speedup across the board for KEVM proofs (~ 15%). In particular:

- We turn on `--no-post-exec-simplify` for the `kore-rpc-booster`, which tells it to not fallback to Kore for simplification at the end of execute requests.
- We turn off fallback on `Branching`, only continuing to fallback on `Aborted` and `Stuck` reasons.


ehildenb:
Perhaps subsumed by: https://github.com/runtimeverification/haskell-backend/pull/4020

PR #2573, opened 2024-08-09
remove procps

dwightguth:
the blockchain-k-plugin is updated to no longer use this dependency, so it is no longer required to be passed anywhere.


dwightguth:
It's beyond obnoxious that I've run the integration tests on this PR four times and it's failed four times even though the failures have absolutely nothing to do with this pr. I don't have time to keep trying, so you'll have to fix this yourselves later.

PR #2575, opened 2024-08-12
Upstreaming of Engagement Lemmas and Test Suite Streamlining

PetarMax:
This PR:

1. Introduces a general mechanism for automatically handling storage slot updates, consisting of a number of simplifications on bitwise operators and byte arrays. This mechanism automates and obsoletes the axiomatic reasoning of `word-pack.k`.
2. Introduces several new support lemmas in the associated lemma-files, including the normalisation of multiplication (concrete-to-the-left, like we already do for addition and other commutatives), which required some adjustments across the development.
4. Removes unneeded lemmas and no-longer-relevant claims from the test suites (`benchmarks`, `examples`, `mcd`)
5. In the `mcd` test suite, rewrites the specifications to use structured instead of unstructured storage, like we do for CSE, allowing us to use efficient booster map matching. This also makes the specifications much more readable, removing all references of `#lookup`, the need for `Junk` variables, and the need for separation constraints, which are now inferred by well-formedness of maps.


tests/specs/mcd/vat-move-diff-rough-spec.k
```diff
@@ -76,7 +76,16 @@ module VAT-MOVE-DIFF-ROUGH-SPEC
...
+              <storage>
+                (#Vat.can[ABI_src][CALLER_ID] |-> May)
+                (#Vat.dai[ABI_src] |-> Dai_src)
+                (#Vat.dai[ABI_dst] |-> Dai_dst)
+                ACCT_ID_STORAGE =>
+                  (#Vat.can[ABI_src][CALLER_ID] |-> May)
+                  (#Vat.dai[ABI_src] |-> Dai_src -Int ABI_rad)
+                  (#Vat.dai[ABI_dst] |-> Dai_dst +Int ABI_rad)
+                  ACCT_ID_STORAGE
+              </storage>
```
ehildenb, line 88: This implicitely assumes that `#Vat.dai[ABI_src]` and `#Vat.dai[ABI_dst]` don't alias, which is true for this test, but not present here is teh `vat-move-same-rough-spec.k` test, which _does_ assume they alias. So just wondering if we need to have this change, because the previous way of expressing it didn't lead to having any questions about the aliasing.

PetarMax: Oh that's interesting, yes, they could be the same.

PR #2579, opened 2024-08-14
Streamlining smt-lemmas

PetarMax:
This PR moves and removes some SMT-lemmas present in the development, in an attempt to minimize the instantiation of uninterpreted functions that Z3 has to do when checking satisfiability.


PetarMax:
On second thought, there is no need for this right now.

PR #2587, opened 2024-08-20
General slot updates and test suite extension

PetarMax:
This PR brings in the general slot update mechanism used in engagements, and creates a structured version of the existing `mcd` test suite as a test suite for said mechanism.


kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/int-simplification.k
```diff
@@ -151,11 +155,15 @@ module INT-SIMPLIFICATION-COMMON
     rule A /Int 1 => A  [simplification, preserves-definedness]
 
     // safeMath mul check c / a == b where c == a * b
-    rule (A *Int B) /Int A => B requires A =/=Int 0 [simplification, preserves-definedness]
+    rule (A *Int B) /Int A => B requires notBool A ==Int 0 [simplification, preserves-definedness]
+    rule (B *Int A) /Int A => B requires notBool A ==Int 0 [simplification, preserves-definedness]
 
-    rule ((A *Int B) /Int C) /Int B => A /Int C requires B =/=Int 0 andBool C =/=Int 0 [simplification, preserves-definedness]
+    rule ((A *Int B) /Int C) /Int B => A /Int C requires notBool B ==Int 0 andBool notBool C ==Int 0 [simplification, preserves-definedness]
```
ehildenb, line 161: Hmmmm, is this one true in the `Int` domain? Could we have rounding errors in `(A *Int B) /Int C`, that result in a value that is no longer divisible by `B`?

say:

- `A == 15`
- `B == 4`
- `C == 7`
- `(A * B) / C == 60 / 7 == 8`
- `((A * B) / C) / B == 8 / 4 == 2`
- `A / C == 2`

Hmmm, I can't come up with an example where it's violated. Could we ask an SMT solver if it's true? For lemmas we weren't sure about, we used to provide inline solver transcripts: https://github.com/runtimeverification/evm-semantics/blob/b3198a80e462a13423cf64cf7d5a7f5e347415b4/tests/specs/mcd/verification.k#L69

ehildenb: Another caveat with division, should ew require the numbers to be non-negative? Is `/Int` rounding towards zero, or rounding towards nearest?

It looks like `/Int` and `%Int` use `t-division`, which rounds towards zero and `divInt` and `modInt` use Euclidian division, which rounds towards nearest. Does that affect these lemmas? https://github.com/runtimeverification/k/blob/master/k-distribution/include/kframework/builtin/domains.md#integer-arithmetic

PetarMax: `/Int`, as far as I know, is the mathematical floor function.

PetarMax: If it is not floor, then it's definitely safer to go non-negative.

PetarMax: Ok, so, Solidity rounds towards zero, presumably `/Int` does as well. What does Z3 integer division do?

I will go through lemmas with division later to enforce non-negativity where required.

ehildenb: Z3 seems to think taht we do need positivity constraints:

This query, which is setup as `assert REQ`, then `check-set NOT (LHS == RHS)`, returns `sat`, which we need it to return `unsat`. I asked for a model, and it gave back negative numbers:

```
(declare-const a Int)
(declare-const b Int)
(declare-const c Int)
(assert (not (= b 0)))
(assert (not (= c 0)))
(assert (not (= (div (div (* a b) c) b) (div a c))))
(check-sat)
(get-model) 
```

Output:

```
sat
(
  (define-fun c () Int
    (- 2))
  (define-fun b () Int
    (- 1))
  (define-fun a () Int
    (- 3))
  (define-fun div0 ((x!0 Int) (x!1 Int)) Int
    (ite (and (= x!0 (- 1)) (= x!1 (- 1))) 1
    (ite (and (= x!0 (- 3)) (= x!1 (- 2))) 2
      (- 1))))
  (define-fun mod0 ((x!0 Int) (x!1 Int)) Int
    (ite (and (= x!0 (- 1)) (= x!1 (- 1))) 0
      1))
)
```

PetarMax: Yup, well done z3, I thought /Int was floor. Will go through all /Int lemmas to assess.

ehildenb: At least need positivity on `B` and `C` it looks like, so could just change side condition to say `0 <Int B` and `0 <Int C`:

This gives back `unsat`:

```
(declare-const a Int)
(declare-const b Int)
(declare-const c Int)
(assert (not (= b 0)))
(assert (not (= c 0)))
; (assert (<= 0 a))
(assert (<= 0 b))
(assert (<= 0 c))
(assert (not (= (div (div (* a b) c) b) (div a c))))
(check-sat)
; (get-model)
```

As does this:

```
(declare-const a Int)
(declare-const b Int)
(declare-const c Int)
(assert (< 0 b))
(assert (< 0 c))
(assert (not (= (div (div (* a b) c) b) (div a c))))
(check-sat)
; (get-model)
```
```

**Comment by PetarMax** [comment]
Interestingly
```
rule (A *Int B) /Int C => B *Int (A /Int C)
      requires notBool C ==Int 0 andBool A modInt C ==Int 0
```
holds, since
```
(declare-const a Int)
(declare-const b Int)
(declare-const c Int)
(assert (not (= c 0)))
(assert (= 0 (mod a c)))
(assert (not (= (div (* a b) c) (* b (div a c)))))
(check-sat)
```
returns `unsat` both for `mod` and for `rem`.

**Comment by PetarMax** [comment]
I believe that all `/Int` are treated correctly now.

File: kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/int-simplification.k [hunk 2]
```diff
@@ -151,11 +155,15 @@ module INT-SIMPLIFICATION-COMMON
...
+    rule (B *Int A) /Int A => B requires notBool A ==Int 0 [simplification, preserves-definedness]
 
-    rule ((A *Int B) /Int C) /Int B => A /Int C requires B =/=Int 0 andBool C =/=Int 0 [simplification, preserves-definedness]
+    rule ((A *Int B) /Int C) /Int B => A /Int C requires 0 <Int B andBool 0 <Int C [simplification, preserves-definedness]
+    rule ((A *Int B) /Int C) /Int A => B /Int C requires 0 <Int A andBool 0 <Int C [simplification, preserves-definedness]
 
-    rule (A /Int 32) *Int 32 => A requires A modInt 32 ==Int 0 [simplification, preserves-definedness]
+    rule (A *Int B) /Int C => B *Int (A /Int C)
+      requires notBool C ==Int 0 andBool A modInt C ==Int 0
+      [simplification, concrete(A, C), preserves-definedness]
```
ehildenb, line 166: Just an observation, but maybe important for lemma efficiency? You'd know better than me at this point what makes the backend slow/fast. It looks to me that the first and last lemmas here are redundant:

```
    rule (A *Int B) /Int A => B requires notBool A ==Int 0 [simplification, preserves-definedness]
    rule (B *Int A) /Int A => B requires notBool A ==Int 0 [simplification, preserves-definedness]

    rule (A *Int B) /Int C => B *Int (A /Int C)
      requires notBool C ==Int 0 andBool A modInt C ==Int 0
      [simplification, concrete(A, C), preserves-definedness]
```

It seems that if we duplicated the 3rd rule (to handle the `(B *Int A) /Int C` case, then the first two lemmas would be fully redundant. They would basically reduce to `(A *Int B) /Int A => B *Int (A /Int A)` and `(B *Int A) /Int A => B *Int (A /Int A)`.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/lemmas.k
```diff
@@ -22,15 +26,16 @@ module LEMMAS [symbolic]
     rule 0 <=Int #ceil32(I)        => true requires 0 <=Int I [simplification, smt-lemma]
     rule 0 <=Int #ceil32(I) -Int I => true requires 0 <=Int I [simplification]
 
-    rule X *Int Y <Int pow256 => true requires Y <=Int maxUInt256 /Int X [simplification]
+    rule X *Int Y <Int pow256 => true requires Y <=Int maxUInt256 /Int X
+      requires 0 <=Int X orBool 0 <=Int Y [simplification]
```
ehildenb, line 30: Is this a typo? It looks like we have `requires ...` twice....

PetarMax: Yes, that's quite horrific, I didn't notice.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
```diff
@@ -0,0 +1,144 @@
...
+
+    // ###########################################################################
+    // Auxiliaries
+    // ###########################################################################
+
+    // Index of first bit that equals one
+    syntax Int ::=  #getFirstOneBit(Int) [function, total]
+
+    rule [gfo-succ]: #getFirstOneBit(X:Int) => log2Int ( X &Int ( ( maxUInt256 xorInt X ) +Int 1 ) ) requires          #rangeUInt(256, X) andBool X =/=Int 0
+    rule [gfo-fail]: #getFirstOneBit(X:Int) => -1                                                    requires notBool (#rangeUInt(256, X) andBool X =/=Int 0)
```
ehildenb, line 16: Huh, this is pretty clever non-looping algorithm. I guess we could also repeatedly divide by 2, but that could be slower.

I wonder though, should we mark this rule (or this fnuction) as `concrete`, so that it doesn't try to build this RHS expression for `X` when it's symbolic?

PetarMax: I did have that, but then it was not passed to the LLVM backend correctly and I was getting undefined evaluations. @jberthold advised to get rid of all attributes and then it worked. It is called only with concrete values in the associated simplifications below.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
```diff
@@ -0,0 +1,144 @@
...
+
+    // Index of first bit that equals zero
+    syntax Int ::= #getFirstZeroBit(Int) [function, total]
+
+    rule [gfz-succ]: #getFirstZeroBit(X:Int) => #getFirstOneBit ( maxUInt256 xorInt X ) requires         #rangeUInt(256, X)
+    rule [gfz-fail]: #getFirstZeroBit(X:Int) => -1                                      requires notBool #rangeUInt(256, X)
```
ehildenb, line 22:
```
#getFirstZeroBit(maxUInt256) =>
#getFirstOneBit(maxUInt256 xorInt maxUInt256) =>
#getFirstOneBit(0) =>
-1
```

Is this intended? I guess it makes sense, but just want to double-check with you the edge-case where we are in the "defined" case for `#getFirstZeroBit` but the "undefined" case for `#getFirstOneBit`.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/int-simplification.k
PetarMax: Yes yes, that's correct, let me take care of that.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
ehildenb: I wonder if we could have a test in `tests/specs/functional/slots-updates-spec.k` for cases like this, and we could add tests there as we discover more.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/int-simplification.k
PetarMax: Ah, the third one has `concrete(A, C)`, I can't rembmer if there was a specific reason for that...

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
PetarMax: Yes, this is intended.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/int-simplification.k
PetarMax: So, I think that the reason I put `concrete(A, C)` there is that I didn't want the backend to even think about symbolic `A modInt C ==Int 0`, especially given that it will now call Z3 to figure that out. The lemmas are redundant, but I think it's better for the backend to have them all.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
PetarMax: Sure, we could - what would you like to test, the behaviour of these auxiliary functions?

ehildenb: Yeah, just on a few specific hardcoded values basically. Even concrete ones are fine I think, make sure they'r ebehaving as intended in `KClaim`s.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
```diff
@@ -0,0 +1,144 @@
...
+    // Width of a mask, in bits and in bytes
+    syntax Int ::= #getMaskWidthBits(Int)  [function, total]
+    syntax Int ::= #getMaskWidthBytes(Int) [function, total]
+
+    rule [gmw-bits-succ-1]: #getMaskWidthBits(X:Int) => 256 -Int #getMaskShiftBits(X:Int)             requires  0 <=Int #getMaskShiftBits(X) andBool 0 ==Int X >>Int #getMaskShiftBits(X) [concrete(X), simplification]
+    rule [gmw-bits-succ-2]: #getMaskWidthBits(X:Int) => #getFirstOneBit(X >>Int #getMaskShiftBits(X)) requires  0 <=Int #getMaskShiftBits(X) andBool 0  <Int X >>Int #getMaskShiftBits(X) [concrete(X), simplification]
+    rule [gmw-bits-fail]:   #getMaskWidthBits(X:Int) => -1                                            requires -1 ==Int #getMaskShiftBits(X) [concrete(X), simplification]
+
+    rule [gmw-bytes-succ]: #getMaskWidthBytes(X:Int) => #getMaskWidthBits(X) /Int 8 requires           #getMaskWidthBits(X) modInt 8 ==Int 0   [concrete(X), simplification, preserves-definedness]
+    rule [gmw-bytes-fail]: #getMaskWidthBytes(X:Int) => -1                          requires notBool ( #getMaskWidthBits(X) modInt 8 ==Int 0 ) [concrete(X), simplification, preserves-definedness]
```
ehildenb, line 46: Could all the `-1` cases be replaced with `[owise]` instead? Maybe that makes it easier to reason about completeness...

ehildenb: Or maybe `[owise]` interacts poorly with `concrete`....

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
```diff
@@ -0,0 +1,144 @@
...
+    // VALUE and SLOT can also be concrete. In the former case, we need to extract the
+    // SHIFT appropriate, and in the latter case, the slot will appear on the LHS of the |Int.
+
+    // 1. Slot masking using &Int
+    rule [mask-b-and]:
+      MASK:Int &Int SLOT:Int =>
+        #asWord ( #buf ( 32, SLOT ) [ 32 -Int ( #getMaskShiftBytes(MASK) +Int #getMaskWidthBytes(MASK) ) := #buf ( #getMaskWidthBytes(MASK), 0 ) ] )
+        requires #rangeUInt(256, MASK) andBool #rangeUInt(256, SLOT)
+        andBool #isMask(MASK)
+        [simplification, concrete(MASK), preserves-definedness]
```
ehildenb, line 88: I guess `#rangeUInt(256, MASK)` is redundant with `#isMask(MASK)`? No problem leaving it in if there is some reason, even readability, but checking my understanding.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
PetarMax: Ah, hm, so these are simplifications now, but they probably needn't be, since it's again a function definition. If I remove all of the attributes, they could be `[owise]` in the function-rule sense, right?

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
PetarMax: Yes, it is redundant.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
```diff
@@ -0,0 +1,144 @@
...
+
+    // 2a. |Int and +Bytes, update to be done in left
+    rule [bor-update-to-left]:
+      A |Int #asWord ( B1 +Bytes B2 ) =>
+        #asWord ( #buf ( 32 -Int lengthBytes(B2), (A /Int (2 ^Int (8 *Int lengthBytes(B2)))) |Int #asWord ( #buf (32 -Int (lengthBytes(B1) +Int lengthBytes(B2)), 0) +Bytes B1 ) ) +Bytes B2 )
```
ehildenb, line 93:
```suggestion
        #asWord ( #buf ( 32 -Int lengthBytes(B2), (A /Int (2 ^Int (8 *Int lengthBytes(B2)))) |Int #asWord ( #buf (32 -Int (lengthBytes(B1) +Int lengthBytes(B2)), 0) +Bytes B1 ) ) +Bytes B2 )
```

Here, we have sub-expression:

```
#asWord(#buf(32 -Int (lengthBytes(B1) +Int lengthBytes(B2)), 0) +Bytes B1)
```

which I'm pretty sure is just the same as:

```
#asWord(B1)
```

PetarMax: It is the same, let me think for a moment why I put the other one instead.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
ehildenb: I think so, but I am not sure how `concrete` and `owise` interact, I think it may be poorly. So leaving it as is may be better.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
PetarMax: Yes, especially since there is this also:
```
    rule [asWord-concat-leading-zeros]:
      #asWord ( BZ +Bytes B ) => #asWord ( B )
      requires #asInteger ( BZ ) ==Int 0
      [simplification, concrete(BZ)]
```
and that will fire.

ehildenb:
It would be great ot have some tests of the masking procedure too as well in `tests/specs/functional/slot-updates-spec.k`.

Spceifically, any expressions of the form `VALUE *Int SHIFT |Int #asWord(B1 +Bytes ... +Bytes BN)`, especially ones you've run into "in the wild", and some hand-crafted ones exercising corner cases like writing `VALUE ==Int 0` or `VALUE ==Int MAX` for example.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
```diff
@@ -0,0 +1,144 @@
...
+        requires #rangeUInt(256, A) andBool A modInt (2 ^Int (8 *Int lengthBytes(B2))) ==Int 0 andBool lengthBytes(B1 +Bytes B2) <=Int 32
+        [simplification, preserves-definedness]
+
+    // 2b. |Int of +Bytes, update to be done in right
+    rule [bor-update-to-right]:
+      A |Int #asWord ( B1 +Bytes B2 ) =>
+        #asWord ( B1 +Bytes #buf ( lengthBytes(B2), A |Int #asWord ( B2 ) ) )
+        requires 0 <=Int A andBool A <Int 2 ^Int (8 *Int lengthBytes(B2))
+        [simplification, preserves-definedness]
```
ehildenb, line 102: We still may need restriction like `lengthBytes(B1 +Bytes B2) <=Int 32`, since `#asWord` puts a chop around the result. So say that `lengthBytes(B2) >Int 32`, and `A >Int pow256`, then the original expression `A |Int #asWord(B1 +Bytes B2)` would not be equal to the new expression, because in the new expression the upper bits of `A` would be truncated.

ehildenb: https://github.com/runtimeverification/evm-semantics/blob/b3198a80e462a13423cf64cf7d5a7f5e347415b4/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm-types.md?plain=1#L336

PetarMax: We can certainly ask for `lengthBytes(B2) <= 32`, and that will also make `A <Int pow256`. Then, the chopping would chop both LHS and RHS equally, am I right?

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/lemmas.k
PetarMax: Fixed.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
PetarMax: Trying with `[owise]` now.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
PetarMax: Done.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
PetarMax: Done.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
PetarMax: `lengthBytes(B2) <=Int 32` added.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
ehildenb: Well my concern specifically is that if a rule doesn't apply because of teh `concrete` attribute, perhaps then it will think "then Ishould apply the `owise` case!", and it shouldn't. It should only apply the owise case _if_ it can show that the first case _definitely doesn't_ apply.

PetarMax: Yes, so I've removed all `concrete` and `simplification`, now it's just `[owise]` and `[preserves-definedness]`

PetarMax: All of these auxiliaries are always called in the lemmas with concrete values, so nothing complex should pop up symbolically.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
```diff
@@ -0,0 +1,143 @@
...
+       andBool 0 <=Int X andBool X <Int 2 ^Int (8 *Int (32 -Int ( log2Int(SHIFT) /Int 8 )))
+       andBool 0 <=Int Y andBool Y <Int SHIFT
+       [simplification, concrete(SHIFT), comm, preserves-definedness]
+
+    // 3b. Buffer cropping
+    rule [buf-asWord-crop]:
+      #buf (W1:Int, #asWord(#buf(W2, X:Int) +Bytes B)) => #buf(W1 -Int lengthBytes(B), X) +Bytes B
+      requires 0 <=Int W1 andBool 0 <=Int W2 andBool lengthBytes(B) <=Int W1 andBool W1 <=Int W2 +Int lengthBytes(B)
+       andBool 0 <=Int X andBool X <Int 2 ^Int (8 *Int (W1 -Int lengthBytes(B)))
+      [simplification]
```
ehildenb, line 116: I think this one also suffers from moving something inside the `Word` field to outside the `Word` field. Specifically, `B` is chopped on the LHS of the rule, but free on the RHS.

Suppose:

- `lengthBytes(B) == 34`
- `W2 == 2`
- `W1 == 36`
- `X == 3`

Then, the LHS will basically fully chop the `X` off inside the `#asWord`, and will truncate part of `B` to be the rightmost 32 bytes, but the RHS will keep `X` intact at the front, and leave `B` unchopped.

I guess some restrictions you can add:

```
W2 +Int lengthBytes(B) <=Int 32
```

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
```diff
@@ -0,0 +1,143 @@
...
+
+    // 3c. Splitting the updated buffer into the updated value and the trailing zeros, explicit shift
+    rule [buf-split-l]:
+      #buf ( W, SHIFT *Int X ) => #buf( W -Int ( log2Int(SHIFT) /Int 8 ), X ) +Bytes #buf( log2Int(SHIFT) /Int 8, 0)
+      requires 0 <=Int W andBool W <=Int 32 andBool rangeUInt(256, SHIFT)
+       andBool SHIFT ==Int 2 ^Int log2Int(SHIFT)
+       andBool log2Int(SHIFT) modInt 8 ==Int 0
```
ehildenb, line 123: We see these conditions show up repeatedly, maybe ew could factor out a function:

```
syntax Bool ::= #isByteShift ( Int ) [function, total]
rule #isByteShift(X) => SHIFT ==Int 2 ^Int log2Int(SHIFT) andBool log2Int(SHIFT) modInt 8 ==Int 0
  requires 0 <Int SHIFT andBool SHIFT <Int pow256
rule #isByteShift(X) => false requires notBool (0 <Int SHIFT andBool SHIFT <Int pow256)
```

A thought.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
PetarMax: We do need a restriction, but I think the one we want is `W1 <=Int 32`, as the resulting buffer is of length `W1`.

With this in place, `X <Int 2 ^Int (8 *Int (W1 -Int lengthBytes(B)))` means that the chop has nothing to chop, is that right?

PetarMax: Adding `W1 <=Int 32` for now.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
PetarMax: Yes, done.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
ehildenb: I don't think so actually. `W1 <=Int 32` just means that the outer `#buf(...)` has a total length of `32`, but doesn't say anything about the length of the contents of `#asWord`, that can still be arbitrarily wide.

In fact, the above counterexample with just changing `W1 == 32` works again even with that restriction. `W1 == 32` just means that the second argument of `#buf(..., _)` needs to be within 32 bytes, but `#asWord(...)` ensures that it's the case, even if the contents are arbitrarily long, so `B` can still basically be unbounded in length and this lemma will apply, which will result in something of length than 32 bytes becoming something longer than 32 bytes.

When I'm looking to analyze these things, I basically look for values that start in(out) of the `Word` finite-field, and end out(in) the `Word` finite-field. In this case, because `B` starts inside an `#asWord`, it's effecitively constrained to being in the field on the LHS, but on the RHS it shows up outside the field.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
```diff
@@ -0,0 +1,149 @@
...
+      #buf ( W, SHIFT *Int X ) => #buf( W -Int ( log2Int(SHIFT) /Int 8 ), X ) +Bytes #buf( log2Int(SHIFT) /Int 8, 0)
+      requires 0 <=Int W andBool W <=Int 32 andBool #isByteShift(SHIFT)
+       andBool 0 <=Int X andBool X <Int 2 ^Int (8 *Int (W -Int ( log2Int(SHIFT) /Int 8)))
+       [simplification, concrete(W, SHIFT), preserves-definedness]
+
+    // 3d. Splitting the updated buffer into the updated value and the trailing zeros, implicit shift
+    rule [bor-split]:
+      X |Int Y => #asWord ( #buf ( 32 -Int #getFirstOneBit(X) /Int 8, X /Int ( 2 ^Int ( 8 *Int ( #getFirstOneBit(X) /Int 8 ) ) ) ) +Bytes
+                            #buf ( #getFirstOneBit(X) /Int 8, Y ) )
+      requires #rangeUInt(256, X) andBool 0 <=Int #getFirstOneBit(X)
```
ehildenb, line 137: I guess that `#rangeUInt(256, X)` and `0 <=Int #getFirstOneBit(X)` are somewhat redundant, given that if `notBool #rangeUInt(256, X)`, then #getFirstOneBit(X) => -1`.

So it could be either `#rangeUInt(256, X) andBool notBool X ==Int 0`, or it could be just `0 <=Int #getFirstOneBit(X)`. If you want to keep the original expression for readability that seems fine too.

ehildenb: Also, I think a potential counterexample to this one is:

- `X == 2`
- `Y == 1`
- `X |Int Y == 3`
- `#buf(32 -Int #getFirstOneBit(X) /Int 8, X /Int (2 ^Int (8 *Int (#getFirstOneBit(X) /Int 8)))) => #buf(32 -Int 0, X /Int (2 ^Int (8 *Int 0)))) => #buf(32, X /Int 2) => #buf(32, 1)`
- `#asWord(#buf(32, 1)) => 1`


So we have `3 == 1`.

I think restricting `#getFirstOneBit(X) modInt 8 ==Int 0` would suffice, which that condition alone would replace both the first two conditions.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
```diff
@@ -0,0 +1,149 @@
...
+       andBool 0 <=Int Y andBool Y <Int 2 ^Int ( 8 *Int ( #getFirstOneBit(X) /Int 8 ) )
+       [simplification, concrete(X), preserves-definedness]
+
+    // 3e. Clearing update leftovers that only have bits higher than the buffer accepts
+    rule [buf-bor-subsume]:
+      #buf ( N, X |Int Y ) => #buf ( N, X )
+      requires 0 <=Int N andBool N <=Int 32
+       andBool 0 <=Int X andBool X <Int 2 ^Int ( 8 *Int N )
+       andBool Y modInt 2 ^Int ( 8 *Int N ) ==Int 0
+       [simplification, concrete(X), preserves-definedness]
```
ehildenb, line 147: This should _never_ happen. It's an invariant of `#buf(W, X)` that it cannot be built ill-formed, we must always have `0 <=Int X andBool X <Int 2 ^Int (8 *Int W)`. If this is being violated, we need to track down where and fix it up. I suspect the prior rule may be to blame, it could build an ill-formed `#buf(1, 0)`.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
PetarMax: Hm, but in the counter-example, `lengthBytes(B) <=Int W1` doesn't hold if `W1 ==Int 32` and `lengthBytes(B) == 34`.

The length of the contents of `#asWord` can indeed be arbitrarily wide, but having `0 <=Int X andBool X <Int 2 ^Int (8 *Int (W1 -Int lengthBytes(B)))` together with `W1 <=Int 32` and `lengthBytes(B) <=Int W1` should ensure that there is no overflow.

The goal of the simplification is precisely to trim the leading zeros of the buffer when the overall length (`W2 +Int lengthBytes(B)`) is greater than 32.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
PetarMax: Yes, the counter-example holds, will think.

PetarMax:
```
    rule [bor-split]:
      X |Int Y => #asWord ( #buf ( 32 -Int #getFirstOneBit(X) /Int 8, X /Int ( 2 ^Int ( 8 *Int ( #getFirstOneBit(X) /Int 8 ) ) ) ) +Bytes
                            #buf ( #getFirstOneBit(X) /Int 8, Y ) )
      requires #rangeUInt(256, X) andBool 0 <=Int #getFirstOneBit(X)
       andBool 0 <=Int Y andBool Y <Int 2 ^Int ( 8 *Int ( #getFirstOneBit(X) /Int 8 ) )
```
I don't think that the counter-example works. The last clause
```
(H): Y <Int 2 ^Int ( 8 *Int ( #getFirstOneBit(X) /Int 8 ) )
```
should say that `Y` stops at most in the last byte before `X` begins. In the counter-example, we have that:
- `X ==Int 2`
- `#getFirstOneBit(X) ==Int 2`
- `#getFirstOneBit(X) /Int 8 ==Int 0`
- `2 ^Int ( 8 *Int ( #getFirstOneBit(X) /Int 8 ) ) == 2 ^Int 0 ==Int 1`
and so (H) becomes `Y <Int 1`, which is in contradiction with `Y ==Int 1` of the example.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
PetarMax: I know which engagement this was created for, so I'll re-run the proofs without this rule and see what happens.

PetarMax: `andBool Y modInt 2 ^Int ( 8 *Int N ) ==Int 0` is a very strange condition to request of a symbolic `Y` to hold.

PetarMax: I will remove this lemma and everything needs to pass without it.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
PetarMax: In any case, I have generalised it to:
```
    // 3b. Buffer cropping
    rule [buf-asWord-crop]:
      #buf (W:Int , #asWord(B:Bytes)) => #range(B, lengthBytes(B) -Int W, W)
      requires 0 <=Int W andBool W <=Int 32 andBool W <Int lengthBytes(B) 
       andBool #asWord ( #range(B, 0, lengthBytes(B) -Int W) ) ==Int 0
      [simplification, concrete(W), preserves-definedness]
```
and added another lemma that trims leading zeros off of a buffer if it's in `#asWord` and longer than 32.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
```diff
@@ -75,17 +75,27 @@ module BYTES-SIMPLIFICATION [symbolic]
...
 
+    rule [asWord-trim-leading-zeros-concat]:
+      #asWord ( BZ +Bytes B ) => #asWord ( B )
+      requires #asInteger ( BZ ) ==Int 0
+      [simplification, concrete(BZ)]
+
+    rule [asWord-trim-overflowing-zeros]:
+      #asWord ( B ) => #asWord ( #range(B, lengthBytes(B) -Int 32, 32) )
+      requires 32 <Int lengthBytes(B) andBool #asInteger(#range(B, 0, lengthBytes(B) -Int 32)) ==Int 0
+      [simplification]
```
ehildenb, line 97: Is the second condition here actually necessary? I think it doesn't matter if the higher bits are set, `#asWord(...)` will trim them away anyway, so we just need `32 <Int lengthBytes(B)` to prevent infinite loop, right?

PetarMax: For this particular lemma, yes, otherwise `lengthBytes(B) -Int 32 <Int 0`. I just wanted to have this lemma as well to contain the buffer size to max 32 whenever possible. There are others that will do the cropping in the `#buf ( X, #asWord ( ...` scenario, when we need to fit the number into a specified `X` bytes.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
PetarMax: Done now, to some extent.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
PetarMax: Sorry, misunderstood initially - no, it's not needed at all.

PetarMax: Removed now.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
```diff
@@ -0,0 +1,137 @@
...
+       andBool 0 <=Int X andBool X <Int 2 ^Int (8 *Int (32 -Int ( log2Int(SHIFT) /Int 8 )))
+       andBool 0 <=Int Y andBool Y <Int SHIFT
+       [simplification, concrete(SHIFT), comm, preserves-definedness]
+
+    // 3b. Buffer cropping
+    rule [buf-asWord-crop]:
+      #buf (W:Int , #asWord(B:Bytes)) => #range(B, lengthBytes(B) -Int W, W)
+      requires 0 <=Int W andBool W <=Int 32 andBool W <Int lengthBytes(B)
+       andBool #asWord ( #range(B, 0, lengthBytes(B) -Int W) ) ==Int 0
+      [simplification, concrete(W), preserves-definedness]
```
ehildenb, line 120: It seems to me that a rule like this we likely want to be lower priority than other rules, basically to let other more specific rules which may act on the actual structure of `B` to fire first, then let this one.

Just a thought.

Can be done with `simplification(60)`.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
```diff
@@ -0,0 +1,137 @@
...
+       andBool 0 <=Int X andBool X <Int 2 ^Int (8 *Int (W -Int ( log2Int(SHIFT) /Int 8)))
+       [simplification, concrete(W, SHIFT), preserves-definedness]
+
+    // 3d. Splitting the updated buffer into the updated value and the trailing zeros, implicit shift
+    rule [bor-split]:
+      X |Int Y => #asWord ( #buf ( 32 -Int #getFirstOneBit(X) /Int 8, X /Int ( 2 ^Int ( 8 *Int ( #getFirstOneBit(X) /Int 8 ) ) ) ) +Bytes
+                            #buf ( #getFirstOneBit(X) /Int 8, Y ) )
+      requires #rangeUInt(256, X) andBool 0 <=Int #getFirstOneBit(X)
+       andBool 0 <=Int Y andBool Y <Int 2 ^Int ( 8 *Int ( #getFirstOneBit(X) /Int 8 ) )
+       [simplification, concrete(X), preserves-definedness]
```
ehildenb, line 135: It seems that this generalizes 3a completely (`bor-update-with-shift`). Is there any reason to keep 3a then? Or, if 3a and this both apply, maybe we prefer the result we get with 3a, in which case we make this rule lower priority `simplification(60)`?

ehildenb: Although, this says `concrete(X)`, which is not more general.... so maybe both are needed.

tests/specs/mcd-structured/word-pack.k
```diff
@@ -0,0 +1,29 @@
...
+
+    syntax Int ::= #WordPackUInt48UInt48     (       Int , Int ) [function]
+                 | #WordPackAddrUInt48UInt48 ( Int , Int , Int ) [function]
+                 | #WordPackAddrUInt8        (       Int , Int ) [function]
+ // -----------------------------------------------------------------------------------------------------------------
+    rule #WordPackUInt48UInt48     (        UINT48_1 , UINT48_2 ) => #asWord (       #buf(20, 0) +Bytes #buf(6, UINT48_2) +Bytes #buf(6, UINT48_1) ) requires #rangeUInt(48, UINT48_1) andBool #rangeUInt(48, UINT48_2)
+    rule #WordPackAddrUInt48UInt48 ( ADDR , UINT48_1 , UINT48_2 ) => #asWord ( #buf(6, UINT48_2) +Bytes #buf(6, UINT48_1) +Bytes    #buf(20, ADDR) ) requires #rangeAddress(ADDR)      andBool #rangeUInt(48, UINT48_1) andBool #rangeUInt(48, UINT48_2)
+    rule #WordPackAddrUInt8        ( ADDR , UINT8               ) => #asWord (       #buf(11, 0) +Bytes #buf(1,    UINT8) +Bytes    #buf(20, ADDR) ) requires #rangeAddress(ADDR)      andBool #rangeUInt(8,     UINT8)
+
+endmodule
```
ehildenb, line 29: Wow, how beautiful!! :heart_eyes:

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
PetarMax: Ah, but simplifications are done bottom-up, so by the time we hit this one, `B` and `W` and `#asWord(B)` will have been simplified fully, and the only other simplifications in contention would be those that have the same LHS.

Can still make the change, if you think it's worth it?

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
PetarMax: This one is sort of a last resort, when `VALUE` is concrete and so `SHIFT *Int VALUE` is concrete and so we can't see the shift immediately. I think both are needed.

ehildenb (approved):
I put some comments about things that maybe look like they can be changed, but not sure if it's a net win.

I'm approving, the lemmas look sound to me. I would appreciate more tests, if you can dig up any more. One technique might be to remove each lemma one at a time, and see what breaks in downstream tests, and take those sub-expressions that show up and make sure they're in the test-suite. That way we know each lemma is exercised appropriately.

If you do make any more changes to the lemmas themselves, please get another round of review from me on the delta (I can just review the commits at that point, instead of reviewing the whole file).

tests/specs/mcd-structured/word-pack.k
PetarMax: Yeah, it just expands and works, it's amazing :)

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
ehildenb: Maybe then `simplification(60)` so that the other lemmas are prioritized over it if they also apply?

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
ehildenb: Up to you.

ehildenb:
If you don't have time for more tests that's fine too, but as we've seen in this review process, these things are fiddly and easy to get wrong, even just with typos. More tests is more better!

PetarMax:
Thank you - given time constraints, I will merge this as-is now, but I *will* come up with more tests when back. The real-world tests that are there were recorded as they came up and as the lemmas were created.

PR #2590, opened 2024-08-21
Include `which` in the nix dev shell

goodlyrottenapple:


jberthold (approved):
:broom: Works in the performance test script.

anvacaru (approved):

PR #2604, opened 2024-09-02
#asWord simplifications

PetarMax:
This PR stabilises and brings together all of the `#asWord`-related simplifications created during various engagements.


tests/specs/functional/lemmas-spec.k
```diff
@@ -532,33 +530,31 @@ module LEMMAS-SPEC
     claim [bit-n240-shl]: <k> runLemma ( notMaxUInt240 &Int ( X <<Int 240 ) ) => doneLemma( X <<Int 240 ) ... </k> requires #rangeUInt (  16 , X )
     claim [bit-n248-shl]: <k> runLemma ( notMaxUInt248 &Int ( X <<Int 248 ) ) => doneLemma( X <<Int 248 ) ... </k> requires #rangeUInt (   8 , X )
 
-    claim [bit-005-m-or-nm]: <k> runLemma ( maxUInt5   &Int ( X |Int ( notMaxUInt5   &Int T ) ) ) => doneLemma( X ) ... </k> requires 0 <=Int X andBool X <Int 2 ^Int 5 andBool 0 <=Int T
```
PetarMax, line 535: This and the following lemmas have been edited to include the `#rangeUInt(256, T)` side condition.

tests/specs/functional/lemmas-spec.k
```diff
@@ -532,33 +530,31 @@ module LEMMAS-SPEC
...
-    claim [bit-008-m-or-nm]: <k> runLemma ( maxUInt8   &Int ( X |Int ( notMaxUInt8   &Int T ) ) ) => doneLemma( X ) ... </k> requires #rangeUInt (   8 , X ) andBool 0 <=Int T
```
PetarMax, line 536: This and the following lemmas have been edited to include the `#rangeUInt(256, T)` side condition.

tests/specs/functional/lemmas-spec.k
```diff
@@ -504,7 +504,6 @@ module LEMMAS-SPEC
  // bitwise
  // -------
 
-    claim [bit-n005-0]: <k> runLemma ( notMaxUInt5   &Int X ) => doneLemma( 0 ) ... </k> requires #rangeUInt (   5 , X )
```
PetarMax, line 507: The four lemmas with `005` have been removed as they are not representative of the slot updates, from which all of these sets of lemmas stem, and which occur only for whole bytes only.

tests/specs/benchmarks/verification.k
```diff
@@ -17,8 +17,6 @@ module VERIFICATION-COMMON
 
     rule         #asWord(#ecrec(HASH, SIGV, SIGR, SIGS))             => 0    requires         #ecrecEmpty(HASH, SIGV, SIGR, SIGS) [simplification]
     rule 0  <Int #asWord(#ecrec(HASH, SIGV, SIGR, SIGS))             => true requires notBool #ecrecEmpty(HASH, SIGV, SIGR, SIGS) [simplification]
-    rule 0 <=Int #asWord(#ecrec(_   , _   , _   , _   ))             => true                                                      [simplification]
-    rule         #asWord(#ecrec(_   , _   , _   , _   )) <Int pow160 => true                                                      [simplification]
```
PetarMax, line 21: These are already included in the main lemmas.

tests/specs/functional/slot-updates-spec.k
```diff
@@ -25,36 +25,37 @@ module SLOT-UPDATES-SPEC
     claim [gfob-min]: <k> runLemma( #getFirstOneBit(0         ) ) => doneLemma( -1 ) ... </k>
     claim [gfob-max]: <k> runLemma( #getFirstOneBit(maxUInt256) ) => doneLemma(  0 ) ... </k>
 
-    claim [gfzb-min]: <k> runLemma( #getFirstZeroBit(0         ) ) => doneLemma(  0 ) ... </k>
```
PetarMax, line 28: This auxiliary function is not necessary as it is used in a single place.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/lemmas.k
```diff
@@ -157,10 +157,6 @@ module LEMMAS-HASKELL [symbolic]
   // Word Reasoning
   // ########################
 
-    rule #asWord(BUF) /Int D => #asWord(#range(BUF, 0, lengthBytes(BUF) -Int log256Int(D)))
-       requires 0 <Int D andBool D <Int pow256 andBool D ==Int 256 ^Int log256Int(D)
-        andBool log256Int(D) <=Int lengthBytes(BUF) [simplification, preserves-definedness]
-
```
PetarMax, line 163: Moved here: https://github.com/runtimeverification/evm-semantics/blob/596162ad26edbfad41b9373b71c8d38bdd0f5df0/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k#L94

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
```diff
@@ -301,15 +259,7 @@ module BYTES-SIMPLIFICATION [symbolic]
 
     rule [lengthBytes-leq-zero]: lengthBytes(B:Bytes) <=Int 0 => B ==K .Bytes [simplification]
 
-    // #asWord
-
-    rule #asWord(WS) >>Int M => #asWord(#range(WS, 0, lengthBytes(WS) -Int (M /Int 8) ))
```
PetarMax, line 306: Moved here:
https://github.com/runtimeverification/evm-semantics/blob/596162ad26edbfad41b9373b71c8d38bdd0f5df0/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k#L88-L89

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
```diff
@@ -244,24 +220,6 @@ module BYTES-SIMPLIFICATION [symbolic]
       #range(#padRightToWidth(_, BUF), 0, WIDTH) => BUF
       requires lengthBytes(BUF) ==Int WIDTH [simplification]
 
-    rule [range-eq-check]:
```
PetarMax, line 247: No longer required given 
https://github.com/runtimeverification/evm-semantics/blob/596162ad26edbfad41b9373b71c8d38bdd0f5df0/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k#L62

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
```diff
@@ -244,24 +220,6 @@ module BYTES-SIMPLIFICATION [symbolic]
...
-
-    rule [range-eq-check-ml]:
-      { BZ +Bytes #range ( B, S, W ) #Equals B } => #Top
-      requires 0 <=Int S andBool 0 <=Int W
-      andBool lengthBytes( B ) ==Int S +Int W
-      andBool #asInteger( BZ ) ==Int 0
-      andBool lengthBytes( BZ ) ==Int S
-      andBool #asWord ( B ) <Int 2 ^Int ( 8 *Int W )
-      [simplification, concrete(BZ, S, W), comm, preserves-definedness]
-
```
PetarMax, line 264: No longer required given 
https://github.com/runtimeverification/evm-semantics/blob/596162ad26edbfad41b9373b71c8d38bdd0f5df0/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k#L62

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
```diff
@@ -72,30 +72,6 @@ module BYTES-SIMPLIFICATION [symbolic]
...
-      [simplification, concrete(WB), preserves-definedness]
-
-    rule [buf-asWord-invert-rl-range]:
-      #asWord ( #range ( #buf ( WB:Int, X:Int ), S:Int, WR:Int ) ) => X
-      requires 0 <=Int WB andBool 0 <=Int X andBool 0 <=Int S andBool 0 <=Int WR
-       andBool S +Int WR ==Int WB
-       andBool X <Int 2 ^Int (8 *Int WR)
-      [simplification, concrete(WB, S, WR), preserves-definedness]
-
-    rule [asWord-trim-leading-zeros-concat]:
```
PetarMax, line 89: Moved here:
https://github.com/runtimeverification/evm-semantics/blob/596162ad26edbfad41b9373b71c8d38bdd0f5df0/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k#L62

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
```diff
@@ -72,30 +72,6 @@ module BYTES-SIMPLIFICATION [symbolic]
...
-      #asWord ( BZ +Bytes B ) => #asWord ( B )
-      requires #asInteger ( BZ ) ==Int 0
-      [simplification, concrete(BZ)]
-
-    rule [asWord-trim-overflowing-zeros]:
```
PetarMax, line 94: Moved here:
https://github.com/runtimeverification/evm-semantics/blob/596162ad26edbfad41b9373b71c8d38bdd0f5df0/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k#L68

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bitwise-simplification.k
```diff
@@ -41,57 +41,26 @@ module BITWISE-SIMPLIFICATION [symbolic]
...
-      [simplification, concrete(X, Z), comm, preserves-definedness]
-
-    // Generalization of:    maxUIntXXX &Int (Y |Int (notMaxUIntXXX &Int T)) =>    maxUIntXXX &Int Y
-    //                and notMaxUIntXXX &Int (Y |Int (   maxUIntXXX &Int T)) => notMaxUIntXXX &Int Y
-    rule X &Int (Y |Int (Z &Int T)) => X &Int Y
-      requires 0 <=Int X andBool 0 <=Int Y andBool 0 <=Int Z andBool 0 <=Int T
-       andBool (      X +Int 1 ==Int 2 ^Int log2Int (      X +Int 1 ) orBool // X is a maxUInt
-                 pow256 -Int X ==Int 2 ^Int log2Int ( pow256 -Int X ) )      // X is a notMaxUInt
-       andBool X +Int Z ==Int maxUInt256
-      [simplification, concrete(X, Z), comm]
```
PetarMax, line 94: Subsumed by slot update lemmas.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bitwise-simplification.k
```diff
@@ -41,57 +41,26 @@ module BITWISE-SIMPLIFICATION [symbolic]
...
     rule [bitwise-and-lt-zero]: (X &Int Y) <Int Z => false requires 0 <=Int X andBool 0 <=Int Y andBool Z <=Int 0 [concrete(Z), simplification]
 
-    rule [bitwise-and-lt]:       (X &Int Y) <Int Z      => true requires 0 <=Int X andBool 0 <=Int Y andBool (X <Int Z orBool Y <Int Z)          [simplification]
-    rule [bitwise-or-lt-pow256]: (X |Int Y) <Int pow256 => true requires 0 <=Int X andBool 0 <=Int Y andBool X <Int pow256 andBool Y <Int pow256 [simplification]
+    rule [bitwise-and-lt]: (X &Int Y) <Int Z => true requires 0 <=Int X andBool 0 <=Int Y andBool (X <Int Z orBool Y <Int Z) [simplification]
+
+    rule [bitwise-or-lt]:
+      (X |Int Y) <Int Z => true
+      requires 0 <=Int X andBool 0 <=Int Y andBool X <Int Z andBool Y <Int Z
+       andBool Z ==Int 2 ^Int ( log2Int( Z ) ) [simplification, concrete(Z)]
```
PetarMax, line 49: Generalised version of `[bitwise-or-lt-pow256]`.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
```diff
@@ -72,30 +72,6 @@ module BYTES-SIMPLIFICATION [symbolic]
...
       [simplification, concrete(W), preserves-definedness]
 
-    rule [buf-asWord-invert-rl]:
-      #asWord ( #buf ( WB:Int, X:Int ) ) => X
-      requires 0 <=Int WB andBool 0 <=Int X andBool
-               ( ( WB <=Int 32 andBool X <Int 2 ^Int (8 *Int WB) ) orBool
-                 ( WB  >Int 32 andBool X <Int pow256 ) )
-      [simplification, concrete(WB), preserves-definedness]
-
-    rule [buf-asWord-invert-rl-range]:
```
PetarMax, line 82: Moved here:
https://github.com/runtimeverification/evm-semantics/blob/596162ad26edbfad41b9373b71c8d38bdd0f5df0/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k#L74

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
```diff
@@ -72,30 +72,6 @@ module BYTES-SIMPLIFICATION [symbolic]
        andBool #asWord(B) <Int 2 ^Int (8 *Int W)
       [simplification, concrete(W), preserves-definedness]
 
-    rule [buf-asWord-invert-rl]:
```
PetarMax, line 75: Moved here:
https://github.com/runtimeverification/evm-semantics/blob/596162ad26edbfad41b9373b71c8d38bdd0f5df0/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k#L82-L83

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
```diff
@@ -57,31 +50,92 @@ module EVM-INT-SIMPLIFICATION-COMMON
...
-  rule [lt-asWord-range]:
-    X <Int #asWord ( #range ( B:Bytes, S, W ) ) => X <Int #asWord ( B )
+  // Unification: two `#asWord`s
+  rule [asWord-eq-asWord]:
+    #asWord ( B1 ) ==Int #asWord ( B2 ) => #buf ( 32 -Int lengthBytes(B1), 0 ) +Bytes B1 ==K #buf ( 32 -Int lengthBytes(B2), 0 ) +Bytes B2
+    requires lengthBytes(B1) <=Int 32 andBool lengthBytes(B2) <=Int 32
+    [simplification, preserves-definedness]
+
+  // Unification: two `#asWord`s
+  rule [asWord-eq-asWord-ml]:
```
PetarMax, line 60: Required for now, will go away once booster takes over.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
```diff
@@ -57,31 +50,92 @@ module EVM-INT-SIMPLIFICATION-COMMON
...
+                andBool (
+                  ( lengthBytes(B) <=Int 32 andBool #buf ( 32 -Int lengthBytes(B), 0 ) +Bytes B ==K #buf ( 32, X ) )
+                    orBool
+                  ( lengthBytes(B) >Int 32 andBool #range ( B, lengthBytes(B) -Int 32, 32 ) ==K #buf ( 32, X ) )
+                )
+    [simplification, concrete(X), comm, preserves-definedness]
+
+  // Unification: `#asWord` and concrete number
+  rule [asWord-eq-num-ml-l]: { #asWord ( B:Bytes ) #Equals X:Int } => { true #Equals #asWord ( B:Bytes ) ==K X:Int } [simplification, concrete(X), preserves-definedness]
+  rule [asWord-eq-num-ml-r]: { X:Int #Equals #asWord ( B:Bytes ) } => { true #Equals #asWord ( B:Bytes ) ==K X:Int } [simplification, concrete(X), preserves-definedness]
```
PetarMax, line 77: Required for now, will go away once booster takes over.

PetarMax:
TODO: add tests for the two new lemmas.

PetarMax:
The tests are now in and the PR is fully ready for review.

PetarMax:
I can also confirm that the two main engagement-related sets of proofs run marginally faster after these changes.

tests/specs/examples/storage-spec.md
```diff
@@ -26,7 +26,7 @@ requires "storage-bin-runtime.k"
 
 module VERIFICATION
     imports EDSL
-    imports LEMMAS
+    imports LEMMAS-WITHOUT-SLOT-UPDATES
```
PetarMax, line 29: This will be reverted once the `#buf` simplifications come in.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/lemmas.k
```diff
@@ -6,7 +6,7 @@ requires "evm-int-simplification.k"
 requires "int-simplification.k"
 requires "slot-updates.k"
 
-module LEMMAS [symbolic]
+module LEMMAS
```
PetarMax, line 9: Removing `[symbolic]` so that the LLVM backend gets access to the helper functions from the slot updates lemmas.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/slot-updates.k
```diff
@@ -1,6 +1,6 @@
 requires "word.md"
 
-module SLOT-UPDATES [symbolic]
+module SLOT-UPDATES
```
PetarMax, line 3: Removing `[symbolic]` so that the LLVM backend gets access to the helper functions.

PetarMax: @ehildenb @anvacaru Perhaps there is a better way of doing this - what I would really like is to keep these modules `[symbolic]`, but just remove the auxiliary functions and put them in some module with other auxiliary functions (or a new module) that would be included with the lemmas. Is there such a file already?

PetarMax: I think that could work, and they are about words, really. I've moved them, let's see.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
```diff
@@ -57,31 +50,91 @@ module EVM-INT-SIMPLIFICATION-COMMON
...
-    requires 0 <=Int S andBool 0 <=Int W
-     andBool 2 ^Int ( 8 *Int W ) <=Int X
-    [simplification, concrete(S, W, X), preserves-definedness]
-
-  rule [lt-asWord-range]:
-    X <Int #asWord ( #range ( B:Bytes, S, W ) ) => X <Int #asWord ( B )
+  // Unification: two `#asWord`s
+  rule [asWord-eq-asWord]:
+    #asWord ( B1 ) ==Int #asWord ( B2 ) => #buf ( 32 -Int lengthBytes(B1), 0 ) +Bytes B1 ==K #buf ( 32 -Int lengthBytes(B2), 0 ) +Bytes B2
+    requires lengthBytes(B1) <=Int 32 andBool lengthBytes(B2) <=Int 32
```
anvacaru, line 56: Should it also have `symbolic(B1, B2)` attribute to apply it only in symbolic cases?

PetarMax: No, there's no need. If either `B1` or `B2` were concrete, the corresponding `asWord` would be evaluated fully first.

ehildenb:
Can we have the code-motion and the type cleanups in a separate PR, that way I can review that quickly and have a smaller diff to review for actual new lemmas and tests? In general, we always request that code motion comes in its own PR for quicker/saner review process.

PetarMax:
I can move the typo corrections and the movement of the auxiliary slot-update lemmas, yes, if that's what you meant specifically. 

The rest I am not sure how to separate, given that the purpose of the PR is to consolidate `#asWord` reasoning, which involves movement of existing lemmas, changes to existing lemmas, introduction of new lemmas, and creation of appropriate tests.

ehildenb:
All the movement/consolidation should come in it's own PR, without semantic changes to anything. Semantic changes can come after. The reason being that it's easier to review a PR if I know it's all code movements, I just have to check "where did it come from where did it go?", not the actual contents of the lemmas. Then in a separate PR, which introduces or changes lemmas, I can inspect the actual content of the lemmas without. It reduces the overall amount of review I have to do.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/word.md
```diff
@@ -585,6 +585,81 @@ Range of types
...
+Slot updates are performed by the compiler with the help of masks, which are words of the form `1111-ONES-111100000-WIDTH-00000011111-SHIFT-11111`.
+The following auxiliary functions support the associated automated reasoning.
+
+-   `#getFirstOneBit` returns the 0-based index of the first non-zero bit of a word
+
+```k
+    syntax Int ::=  #getFirstOneBit(Int) [function, total]
+
+    rule [gfo-succ]: #getFirstOneBit(X:Int) => log2Int ( X &Int ( ( maxUInt256 xorInt X ) +Int 1 ) ) requires #rangeUInt(256, X) andBool X =/=Int 0 [preserves-definedness]
+    rule [gfo-fail]: #getFirstOneBit(_:Int) => -1                                                    [owise]
```
anvacaru, line 602: I think we can drop the spaces here
```suggestion
    rule [gfo-fail]: #getFirstOneBit(_:Int) => -1 [owise]
```

anvacaru (dismissed):

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bitwise-simplification.k
```diff
@@ -34,24 +34,26 @@ module BITWISE-SIMPLIFICATION [symbolic]
...
 
     // Generalization of: maxUIntXXX &Int Y => Y requires #rangeUInt(XXX, Y)
     rule [bitwise-and-maxUInt-identity]:
       X &Int Y => Y
-      requires 0 <=Int X
-       andBool X +Int 1 ==Int 2 ^Int log2Int(X +Int 1)
+      requires 0 <=Int X andBool X +Int 1 ==Int 2 ^Int log2Int(X +Int 1)
        andBool 0 <=Int Y andBool Y <Int X +Int 1
-       [concrete(X), simplification, comm, preserves-definedness]
+       [simplification(40), concrete(X), preserves-definedness]
```
ehildenb, line 49: Is it intentional to remove `comm` here?

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bitwise-simplification.k
```diff
@@ -34,24 +34,26 @@ module BITWISE-SIMPLIFICATION [symbolic]
...
 
-    // Generalization of: notMaxUIntXXX &Int Y => 0 requires #rangeUInt(XXX, Y)
+    // Generalization of: notMaxUIntXXX &Int Y => 0 requires #rangeUInt(256 -Int XXX, Y)
     rule [bitwise-and-notMaxUInt-zero]:
       X &Int Y => 0
-      requires #rangeUInt(256, X)
-       andBool pow256 -Int X ==Int 2 ^Int ( log2Int ( pow256 -Int X ) )
+      requires #rangeUInt(256, X) andBool pow256 -Int X ==Int 2 ^Int ( log2Int ( pow256 -Int X ) )
```
ehildenb, line 54: I wonder if factoring out `#isPower(X) => X ==Int 2 ^Int log2Int(X) requires 0 <Int X` would be helpful? That pattern is used a few places, not just in this PR.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
ehildenb, line 56: This seems a bit complicated, could it work like this?

```suggestion
  // Unification: two `#asWord`s
  rule [asWord-eq-asWord]:
    #asWord ( B1 ) ==Int #asWord ( B2 ) => B1 ==K B2
    requires lengthBytes(B1) <=Int 32 andBool lengthBytes(B1) ==Int lengthBytes(B2)
```

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
```diff
@@ -57,31 +50,91 @@ module EVM-INT-SIMPLIFICATION-COMMON
...
+  // Unification: `#asWord` and concrete number
+  rule [asWord-eq-num]:
+    #asWord ( B:Bytes ) ==Int X:Int =>
+      0 <=Int X andBool X <Int 2 ^Int (8 *Int minInt(32, lengthBytes(B)))
+                andBool (
+                  ( lengthBytes(B) <=Int 32 andBool #buf ( 32 -Int lengthBytes(B), 0 ) +Bytes B ==K #buf ( 32, X ) )
+                    orBool
+                  ( lengthBytes(B) >Int 32 andBool #range ( B, lengthBytes(B) -Int 32, 32 ) ==K #buf ( 32, X ) )
+                )
+    [simplification, concrete(X), comm, preserves-definedness]
```
ehildenb, line 73: Can we write this as multiple rules instead? The first clause can be a `requires` clause of both rules, and then each rule can make a choice based on `lengthBytes(B) <=Int 32` basically.

Are there cases where we cannot tell which of the cases in the `orBool` we fall into, but we can tell that _one_ of the cases applies?

ehildenb: I also wonder whether this lemma is needed, and if so do we have tests of it directly here? It's by far the most "expansive" in the sense that the RHS is much larger than the LHS (in term size) of the whole PR, so it doesn't necessarily feel like it's a simplificaiton.

ehildenb: Ah, I see `concrete(X)`, which means the RHS will be simpler by a decent amount. I still think the first condition can definitely by a side-condition, with a branch of rules:

```
rule #asWord(B) ==Int X => false requires notBool #rangeUInt(256, X)
rule #asWord(B) ==Int X => #buf(32 -Int lengthBytes(B), 0) +Bytes B ==K #buf(32, X) requires #rangeUInt(256, X) andBool lengthBytes(B) <=Int 32
rule #asWord(B) ==Int X => #range(B, lengthBytes(B) -Int 32, 32) ==K #buf(32, X) requires #rangeUInt(256, X) andBool 32 <Int lengthBytes(B)
```

I guess in general we should have pretty good knowledge of `B`, so it should be able to pick which rule applies here. Are there cases where we don't have that knowledge? Maybe when processing dynamically sized calldata? Do we have a test here of that case?

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
ehildenb: Can probably use `comm` attribute to only have one of these rules.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bitwise-simplification.k
PetarMax: Very probably, yes, and there are a few other helpers we could introduce. This can be the next PR.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bitwise-simplification.k
PetarMax: Yes, `&Int` forces concrete to the left, so since `X` is concrete, it can only be on the left.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
PetarMax: This is the generality that appears in the proofs, sometimes an equality question is asked on buffers of different lengths, but that does not automatically mean they hold a different value.

In this way, we have full generality of `#asWord` unification.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
PetarMax: Yes, we should know the length of B, so we should get one of the branches only. Happy to change.

PetarMax: I think that we should always be able to tell whether the length is smaller or greater than 32 in the context of asking for an `#asWord`, I don't see why the compiler why try to convert arbitrarily-sized dynamic data to a single word.

PetarMax: Ah, there's a problem. If we really can't choose a branch, then we would have to change both of these to each have three cases as well:
```
  rule [asWord-eq-num-ml-l]: { #asWord ( B:Bytes ) #Equals X:Int } => { true #Equals #asWord ( B:Bytes ) ==K X:Int } [simplification, concrete(X), preserves-definedness]
  rule [asWord-eq-num-ml-r]: { X:Int #Equals #asWord ( B:Bytes ) } => { true #Equals #asWord ( B:Bytes ) ==K X:Int } [simplification, concrete(X), preserves-definedness]
```
because otherwise the simplification would just go back and forth between the two.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bitwise-simplification.k
PetarMax: I should actually make sure that happens before by adding higher priority to those rules.

PetarMax: Done now.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
PetarMax: Re tests, yes, there are four, exercising all branches. They are starting from
https://github.com/runtimeverification/evm-semantics/blob/e73f8d93607156da28d22ea63b6e0f42c890726a/tests/specs/functional/lemmas-spec.k#L670

PetarMax: In summary for this one, I think that either it needs to stay as-is, knowing that the worst that can happen is that we get both branches on `lengthBytes(B)`, which, from my experience, is not likely, or we duplicate the reasoning for the ML-level lemmas.

ehildenb: Does using rule priorities make the simplificaitons non-looping?

PetarMax: Not in this case, no.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
```diff
@@ -57,31 +50,91 @@ module EVM-INT-SIMPLIFICATION-COMMON
...
+  rule [asWord-shr]:
+    #asWord( BA ) >>Int N => #asWord ( #range ( BA, 0, lengthBytes( BA ) -Int ( N /Int 8 ) ) )
+    requires 0 <=Int N andBool N modInt 8 ==Int 0
+    [simplification, concrete(N), preserves-definedness]
+
+  // Reduction: division in terms of `#range`
+  rule [asWord-div]:
+    #asWord( BA ) /Int N => #asWord ( #range ( BA, 0, lengthBytes( BA ) -Int ( log2Int( N ) /Int 8 ) ) )
+    requires 0 <Int N andBool N ==Int 2 ^Int log2Int ( N ) andBool log2Int( N ) modInt 8 ==Int 0
+    [simplification, preserves-definedness]
```
ehildenb, line 116: Potential counterexample (counterexamples, even potential ones, could be added as tests!):

- `BA == #buf(32, maxUInt256) +Bytes #buf(32, 1)`
- `N == pow248`


Reduction 1:

```
   #asWord(BA) /Int N
=> #asWord(#buf(32, maxUInt256) +Bytes #buf(32, 1)) /Int pow248
=> #asWord(#buf(32, 1)) /Int pow248 => 1 /Int pow248 => 0
```

Reduction 2:

```
   #asWord(BA) /Int N
=> #asWord(#range(BA, 0, lengthBytes(BA) -Int (log2Int(N) /Int 8))) /Int N
=> #asWord(#range(BA, 0, 64 -Int (248 /Int 8))) /Int pow248
=> #asWord(#range(BA, 0, 64 - 31)) /Int pow248
=> #asWord(#buf(32, maxUInt256) +Bytes #buf(1, 0)) /Int pow248
=> #asWord(#buf(31, maxUInt255) +Bytes #buf(1, 0)) /Int pow248
=> (maxUInt256 -Int maxUInt8) /Int pow248
=> 255
```

Basically, if `BA` is wider than a `Word`, then `#asWord` could chop it first then divide, or this routine will shift in bytes that are higher than `pow256` instead of dividing then chop.

PetarMax: Yes, great catch, will fix, we need to chop off the higher bytes before anything.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
```diff
@@ -57,31 +50,91 @@ module EVM-INT-SIMPLIFICATION-COMMON
...
+
+  // Comparison: `#asWord` is always in the range `[0, pow256)`
+  rule [asWord-lb]: 0 <=Int #asWord( _ )     => true [simplification, smt-lemma]
+  rule [asWord-ub]: #asWord( _ ) <Int pow256 => true [simplification, smt-lemma]
+
+  // Comparison: `#asWord(B)` is certainly less than something that does not fit into `lengthBytes(B)` bytes
+  rule [asWord-lt]:  #asWord ( B )  <Int X:Int => true  requires X >=Int   2 ^Int (8 *Int minInt(32, lengthBytes(B)))               [simplification, preserves-definedness]
+  // Comparison: `#asWord(B)` is certainly less than or equal to something that is not less than the max number fitting into `lengthBytes(B)` bytes
+  rule [asWord-le]:  #asWord ( B ) <=Int X:Int => true  requires X >=Int ( 2 ^Int (8 *Int minInt(32, lengthBytes(B))) -Int 1 )      [simplification, preserves-definedness]
```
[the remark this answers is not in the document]
PetarMax: Yes, agreed.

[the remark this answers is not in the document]
PetarMax: This was just so that it's a bit more human-friendly.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
```diff
@@ -57,31 +50,91 @@ module EVM-INT-SIMPLIFICATION-COMMON
...
+  rule [asWord-concat-lt]:
+    #asWord ( BA1 +Bytes BA2 ) <Int X:Int => #asWord ( BA1 ) <Int X /Int ( 2 ^Int ( 8 *Int lengthBytes ( BA2 ) ) )
+    requires #asWord ( BA2 ) ==Int X modInt ( 2 ^Int ( 8 *Int lengthBytes ( BA2 ) ) )
+    [simplification, concrete(BA2, X), preserves-definedness]
+
+  // Comparison: `#asWord` of `+Bytes` when lower bytes match, with `<=Int`
+  rule [asWord-concat-le]:
+    #asWord ( BA1 +Bytes BA2 ) <=Int X:Int => #asWord ( BA1 ) <=Int X /Int ( 2 ^Int ( 8 *Int lengthBytes ( BA2 ) ) )
+    requires #asWord ( BA2 ) ==Int X modInt ( 2 ^Int ( 8 *Int lengthBytes ( BA2 ) ) )
+    [simplification, concrete(BA2, X), preserves-definedness]
```
ehildenb, line 137: I think these both suffer similar counterexamples:

- `BA1 == #buf(1, 8)`
- `BA2 == #buf(64, maxUInt512)`
- `X == pow514 +Int maxUInt512`

Reduction 1:

```
   #asWord(BA1 +Bytes BA2) <Int X
=> maxUInt256 <Int pow514 +Int maxUInt512
=> true
```

Reduction 2:

```
   #asWord(BA1 +Bytes BA2) <Int X
=> #asWord(BA1) <Int X /Int (2 ^Int (8 *Int lengthBytes(BA2)))
=> 8 <Int (pow514 +Int maxUInt512) /Int pow512
=> 8 <Int pow2
=> 8 <Int 4
=> false
```

I think what's needed here are restrictions on the length of `BA1 +Bytes BA2`?

ehildenb: This can be added as a test.

PetarMax: I think that we need to put higher priority on the simplification that cuts a buffer inside asWord to be of max length 32.

PetarMax: It is of higher priority already. I will see what the tests do.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
PetarMax: `comm` doesn't work on ML-level rules, sadly.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
PetarMax: I have added two very general tests that show that the part of the buffer overflowing 32 bytes will be chopped off correctly before the simplification is applied.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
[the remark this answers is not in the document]
PetarMax: Done.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
PetarMax: I have added a general test that shows that that the part of the buffer overflowing 32 bytes will be chopped off correctly before the simplification is applied (analogous tests work for other lemmas mentioned further below). 

However, there is a problem with the new encoding of `/Int` that does not allow this claim to be run, it stalls in the implication request. This is the same issue that @jberthold noted in the process, the result of which was to change `/Int` to `divInt` in the appropriate test. I think that this needs to be analysed and fixed. Until then, this test remains commented.

ehildenb: Are we relying on the priority of the `simplification` rule then? That isn't sufficient, the priority attached to `simplificaiton` rule is a _suggestion_ for the backend, but not a _requirement_. If the backend _can't tell_ that a higher priority simplification rule may apply or not, it's free to try lower priority rules. If it _can't tell_ that a higher-priority equational rule may apply or not, then it must not try lower priority rules.

So we can't rely on the priority of `simplification` rules, they must be true standalone. Only can rely on `priority` of normal equational rules.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
ehildenb: Same as above, we cannot rely on priority of `simplification` rules, it's a suggestion for the backend, not a requirement.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
PetarMax: Understood, corrected now.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
PetarMax: Corrected.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
```diff
@@ -57,31 +50,94 @@ module EVM-INT-SIMPLIFICATION-COMMON
...
+  // Unification: `#asWord` and concrete number
+  rule [asWord-eq-num]:
+    #asWord ( B:Bytes ) ==Int X:Int =>
+      0 <=Int X andBool X <Int 2 ^Int (8 *Int minInt(32, lengthBytes(B)))
+                andBool (
+                  ( lengthBytes(B) <=Int 32 andBool #buf ( 32 -Int lengthBytes(B), 0 ) +Bytes B ==K #buf ( 32, X ) )
+                    orBool
+                  ( lengthBytes(B) >Int 32 andBool #range ( B, lengthBytes(B) -Int 32, 32 ) ==K #buf ( 32, X ) )
+                )
+    [simplification, concrete(X), comm, preserves-definedness]
```
ehildenb, line 73: I think we can remove the second case (`lengthBytes(B) >Int 32`) entirely, because the rule below `asWord-trim-overflowing` will apply in that case. So we'll end up with:

```
#asWord(#range(B, lengthBytes(B) -Int 32, 32)) ==Int X
```

Which will fall into the first case.

```suggestion
  // Unification: `#asWord` and concrete number
  rule [asWord-eq-num]:
    #asWord ( B:Bytes ) ==Int X:Int => 0 <=Int X andBool X <Int 2 ^Int (8 *Int minInt(32, lengthBytes(B)))
                               andBool #buf ( 32 -Int lengthBytes(B), 0 ) +Bytes B ==K #buf ( 32, X )
    requires lengthBytes(B) <=Int 32
    [simplification, concrete(X), comm, preserves-definedness]
```
```

**Review by ehildenb — APPROVED** [review]

File: kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k (hunk 29 above)
Comment (PetarMax) [reply kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k:73]: I think so, yes, but only if we can say with certainty whether the length of the buffer `B` is `<=Int` or `>Int` than 32, right? 

While I think that this should always be the case when `#asWord` is used, I don't think that we have a 100% guarantee that `#asWord` will not get a parameter whose length is sufficiently symbolic for us not to be able to tell that, in which case `asWord-trim-overflowing` won't fire and we we need to have both cases?

Comment (ehildenb) [reply kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k:73]: I guess, do we have an example of the case you're describing from client engagements? If so, definitely add a test for it. If not, I'm somewhat erring on the side of the simpler more auditable lemma.

Comment (PetarMax) [reply kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k:73]: I don't have an example from client engagements and I believe that the case I'm describing won't happen. What can happen is that the slot update mechanism adds leading zeros that take the buffer length beyond 32, but in those cases we would know the length precisely.

I'm happy to simplify the lemma.

Comment (PetarMax) [reply kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k:73]: Actually, can we do this, even simpler, we don't need either the minimum or the leading zeros?
```
  rule [asWord-eq-num]:
    #asWord ( B:Bytes ) ==Int X:Int =>
      0 <=Int X andBool X <Int 2 ^Int (8 *Int lengthBytes(B))
      andBool B ==K #buf ( lengthBytes(B), X )
    requires lengthBytes(B) <=Int 32
    [simplification, concrete(X), comm, preserves-definedness]
```

Comment (ehildenb) [reply kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k:73]: If we're simplifying, we can probably go even further, and have an extra `requires` clause `X <Int pow256`. We can have a separate case for otherwise:

```
rule #asWord(_) ==Int X => false requires X <Int 0 orBool pow256 <=Int X [simplification(40)]
```

Comment (ehildenb) [reply kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k:73]: That would allow us to instead have:

```suggestion
  // Unification: `#asWord` and concrete number
  rule [asWord-eq-num]:
    #asWord ( B:Bytes ) ==Int X:Int => #buf ( 32 -Int lengthBytes(B), 0 ) +Bytes B ==K #buf ( 32, X )
    requires lengthBytes(B) <=Int 32 andBool #rangeUInt(256, X)
    [simplification, concrete(X), comm, preserves-definedness]

  rule [asWord-neq-num]:
    #asWord(_) ==Int X => false requires notBool #rangeUInt(256, X) [simplification]
```

That actually is slightly more general since the second one is not restricting us to `concrete(X)`.

ehildenb: The first one is `concrete(X)` anyway, so I think that this shouldn't really be any less applicable.

PetarMax: and we will always end up only with `B ==K #buf ( lengthBytes(B), X )` really.

PetarMax: The second simplification you suggested, `[asWord-neq-num]`, if `X` is not concrete, I think would fire always for any `X` of arbitrary complexity, which means many calls to the SMT solver potentially about the range, and it almost always not applying.

PR #2608, opened 2024-09-05
Forcing sequential proof advancement in CI

PetarMax:
This is an attempt to stabilise the flakiness of tests in CI, specifically functional ones, by forcing the use of sequential proof advance rather than the parallel one with a single thread.


PetarMax:
The only way I have of assessing this is that the functional tests passed immediately here whereas I had to re-run them in almost every run of every other PR in the past several days.

.github/workflows/test-pr.yml
```diff
@@ -85,27 +85,27 @@ jobs:
         include:
           - name: 'Rules (booster)'
             test-suite: 'test-prove-rules'
-            test-args:
+            test-args: '--force-sequential'
```
tothtamas28, line 88: Shouldn't these arguments be baked into the Python tests so that locally we execute the same tests?

PetarMax: Yes, they should.

anvacaru (approved):

kevm-pyk/src/tests/integration/test_prove.py
```diff
@@ -340,7 +339,7 @@ def test_prove_rules(
         tmp_path,
         caplog,
         no_use_booster,
-        force_sequential,
+        True,
```
tothtamas28, line 342:
```suggestion
        force_sequential=True,
```

(Or whatever the name of that parameter is.)

tothtamas28 (dismissed):

kevm-pyk/src/tests/integration/test_prove.py
PetarMax: The other named parameters are left unnamed, should we name them all?

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -380,6 +380,8 @@ def create_kcfg_explore() -> KCFGExplore:
             ready = topological_sorter.get_ready()
             _LOGGER.info(f'Discharging proof obligations: {ready}')
             curr_claim_list = [all_claim_jobs_by_label[label] for label in ready]
+            for claim in curr_claim_list:
+                claim.up_to_date(digest_file)
```
PetarMax, line 384: Eliminating potential data races coming from line 385 below, since `up_to_date` both reads and writes to digest file.

tothtamas28: Nice catch!

tothtamas28: For `N` claims, this is `N` updates to the same file though, right?

If so, please open an issue to track this.

tothtamas28 (dismissed):

kevm-pyk/src/kevm_pyk/__main__.py
PetarMax: Yes, it is - what should the issue contain, though, I'm not sure?

tothtamas28: Instead of doing `N` file accesses like:
```python
for claim in claims:
    claim.up_to_date(file)
```

we should do a single file access:

```python3
up_to_date(clams, file)
```

PetarMax: We should fix this in this PR, I think.

Noting that there is a recursion happening in `_init_and_run_proof`, if this is of relevance.

kevm-pyk/src/kevm_pyk/__main__.py
```diff
@@ -190,7 +190,8 @@ def update_digest(self, digest_file: Path | None) -> None:
         if 'claims' not in digest_dict:
             digest_dict['claims'] = {}
         digest_dict['claims'][self.claim.label] = self.digest
-        digest_file.write_text(json.dumps(digest_dict, indent=4))
+        with SoftFileLock(f'{digest_file}.lock'):
```
tothtamas28, line 193: Reading the file should be also in the critical section to avoid a race condition.

PetarMax:
I've had five passes of the functional lemmas across three PRs, I think this is good to go.

tothtamas28 (approved):
:rocket:

PR #2616, opened 2024-09-10
Removing rangeBool reasoning

PetarMax:
The lemmas removed in this PR slow down execution of some real-world proofs considerably because the `rangeBool` constraints are SMT-checked for every expression of the form `notBool ( X ==Int 0 )` or `notBool ( X ==Int 1 )`. 

This will be resolved with the introduction of syntactic simplifications, because `#rangeBool` is always there syntactically.

### DIFF
--- a/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/lemmas.k
+++ b/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/lemmas.k
@@ -56,15 +56,6 @@ module LEMMAS-WITHOUT-SLOT-UPDATES [symbolic]
     rule #isPrecompiledAccount(#newAddr(_, _), _) => false [simplification]
 
   // ########################
-  // #rangeBool reasoning
-  // ########################
-
-    rule [rangeBool-not-zero-l]: notBool (X ==Int 0) => X ==Int 1 requires #rangeBool(X) [simplification]
-    rule [rangeBool-not-zero-r]: notBool (0 ==Int X) => X ==Int 1 requires #rangeBool(X) [simplification]
-    rule [rangeBool-not-one-l]:  notBool (X ==Int 1) => X ==Int 0 requires #rangeBool(X) [simplification]
-    rule [rangeBool-not-one-r]:  notBool (1 ==Int X) => X ==Int 0 requires #rangeBool(X) [simplification]
-
-  // ########################
   // bool2Word reasoning
   // ########################
 


PetarMax:
(Some of) the added lemmas are tested by some of the `dsvalue` tests in the `mcd` and `mcd-structured` test suites, via the compiler-introduced checks. 

I am not sure what a reasonable test that is not direct would look like.

tests/specs/examples/sum-to-n-foundry-spec.k
```diff
@@ -6,6 +6,8 @@ module VERIFICATION
     imports INFINITE-GAS
     imports EVM
 
+    rule [rangeBool-not-zero-l]: notBool (X ==Int 0) => X ==Int 1 requires #rangeBool(X) [simplification]
```
PetarMax, line 9: Added as a temporary fix for #2617.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/lemmas.k
```diff
@@ -99,6 +90,10 @@ module LEMMAS-WITHOUT-SLOT-UPDATES [symbolic]
     rule [b2w-bxor-one]:            1  xorInt bool2Word(Y) => bool2Word(notBool Y)                                             [simplification, comm]
     rule [b2w-bxor]:      bool2Word(X) xorInt bool2Word(Y) => bool2Word ( (X andBool notBool Y) orBool (notBool X andBool Y) ) [simplification]
 
+    // Relationship with `#rangeBool`
+    rule [b2w-rangeBool-eq-not-zero-l]: bool2Word (notBool (X ==Int 0)) => X requires #rangeBool(X) [simplification, comm]
+    rule [b2w-rangeBool-eq-not-zero-r]: bool2Word (notBool (0 ==Int X)) => X requires #rangeBool(X) [simplification, comm]
+
```
ehildenb, line 96: Do we also want the ones for `X ==Int 1` and `1 ==Int X`?

ehildenb (approved):

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/lemmas.k
PetarMax: I don't think so, at least not for now, because these ones (the one with `X ==Int 0` specifically, and then I added the commutative version) come from the compiler as a check that something is a bool.

PR #2625, opened 2024-09-18
Correction for `#asWord` lemmas

PetarMax:
This PR adds the necessary `requires` clauses to ML-level simplifications for `#asWord`, which prevent infinite looping observed in the Kontrol test suite.


ehildenb (approved):
Tests that are infinite looping but now are not would be great! I guess since something is failing on Kontrol, you ahve an example that can be merged here.

anvacaru (approved):

PetarMax:
I tried, but wasn't able to create one in KEVM. The test that does the looping in Kontrol is `TGovernance.getEscrowTokenTotalSupply`.

PR #2634, opened 2024-10-01
Memory update simplifications

PetarMax:
This PR brings in the memory update simplifications from various engagements, together with an associated test.

### DIFF
--- a/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
+++ b/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
@@ -227,26 +227,34 @@ module BYTES-SIMPLIFICATION [symbolic]
       requires S <Int 0
       [simplification]
 
-    rule [memUpdate-as-concat-inside]:
-      B1:Bytes [ S:Int := B2:Bytes ] =>
-        #range(B1, 0, S) +Bytes B2 +Bytes #range (B1 , S +Int lengthBytes(B2), lengthBytes(B1) -Int (S +Int lengthBytes(B2)))
-      requires 0 <=Int S andBool S +Int lengthBytes(B2) <Int lengthBytes(B1)
-      [simplification(60)]
+    rule [memUpdate-concat-in-left]: (B1 +Bytes B2) [ S := B ] => (B1 [ S := B ]) +Bytes B2
+      requires 0 <=Int S andBool S +Int lengthBytes(B) <=Int lengthBytes(B1)
+      [simplification(40)]
 
-    rule [memUpdate-as-concat-outside-1]:
-      B1:Bytes [ S:Int := B2:Bytes ] => #range(B1, 0, S) +Bytes B2
-      requires 0 <=Int S andBool lengthBytes(B1) <=Int S +Int lengthBytes(B2)
-      [simplification(60)]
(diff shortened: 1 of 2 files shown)


kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
```diff
@@ -227,27 +227,35 @@ module BYTES-SIMPLIFICATION [symbolic]
...
+    rule [memUpdate-as-concat-inside]:
+      B1:Bytes [ S:Int := B2:Bytes ] =>
+        #range(B1, 0, S) +Bytes B2 +Bytes #range(B1 , S +Int lengthBytes(B2), lengthBytes(B1) -Int (S +Int lengthBytes(B2)))
+      requires 0 <=Int S andBool S +Int lengthBytes(B2) <Int lengthBytes(B1)
+      [simplification(60)]
+
+    rule [memUpdate-as-concat-outside]:
+      B1:Bytes [ S:Int := B2:Bytes ] => #range(B1, 0, S) +Bytes B2
+      requires 0 <=Int S andBool lengthBytes(B1) <=Int S +Int lengthBytes(B2)
+      [simplification(60)]
```
ehildenb, line 257: I feel like these can be combined into one lemma (so no branching on `S +Int lengthBytes(B2) <Int lengthBytes(B1)` basically):

```
rule B1 [ S := B2 ] => #range(B1, 0, S) +Bytes B2 +Bytes #range(B1, S +Int lengthBytes(B2), lengthBytes(B1) -Int (S +Int lengthBytes(B2))) requires 0 <=Int S andBool S <=Int lengthBytes(B1) [simplification(60)]
```

I think that should handle both cases? The last `#range(...)` expression should go to empty if it can determine that the third argument is negative.

PetarMax: Yes, it will. I can't remember right now if there was a reason for splitting, that is, requiring that we know the branch before simplifying. I certainly prefer `+Bytes` over memory update, so let's try this and see what happens.

PetarMax: I think we might not need `S <=Int lengthBytes(B1)` either, because if it's greater `#range(B1, 0, S)` will pad with zeros (as intended) and the last `#range(...)` expression goes to `.Bytes`?

ehildenb: Yeah, like I said, my only concern in omitting that condition is that we could get a potentially huge 0-bytes expression there, which could slow things down.

PetarMax: This is happening right now with the lemmas as they are currently, and there are no issues, we're ok with reasoning about `+Bytes` and large constants.

PetarMax:
Yes, so, from the proofs failing and the associated claims, it seems that we might want to know in which branch of the two we are before applying the lemma.

PetarMax:
In any case, the two lemmas we are fiddling around with have not been introduced by this PR, they have been with us for a while and have worked well for a number of engagements.

ehildenb (approved):

PR #2641, opened 2024-10-09
Memory update lemma correction

PetarMax:
This PR fixes a typo in one of the recently upstreamed memory update lemmas, where `<=Int` should have been `<Int`.


PetarMax:
The tests is taken from a Kontrol test that was failing as a result of this change.

palinatolmach (approved):

PR #2666, opened 2024-12-17
Introducing the `optimize-kcfg` parameter

PetarMax:
This PR introduces the `optimize-kcfg` parameter to the proof options, which uses the recently-introduced infrastructure in `pyk` that optimizes the KCFG on-the-fly.

### DIFF
--- a/kevm-pyk/src/kevm_pyk/cli.py
+++ b/kevm-pyk/src/kevm_pyk/cli.py
@@ -381,6 +381,7 @@ class KProveOptions(Options):
     direct_subproof_rules: bool
     maintenance_rate: int
     assume_defined: bool
+    optimize_kcfg: bool
 
     @staticmethod
     def default() -> dict[str, Any]:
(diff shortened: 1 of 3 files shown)


kevm-pyk/src/kevm_pyk/cli.py
```diff
@@ -850,6 +852,13 @@ def kprove_args(self) -> ArgumentParser:
             action='store_true',
             help='Use the implication check of the Booster (experimental).',
         )
+        args.add_argument(
+            '--optimize-kcfg',
+            dest='optimize_kcfg',
+            default=None,
+            action='store_true',
+            help='Optimize the constructed KCFG on-the-fly.',
```
anvacaru, line 860: Can we make this a bit more detailed?
```suggestion
            help='Optimize the constructed KCFG on-the-fly by minimizing consecutive edges.',
```

anvacaru (approved):
Approved with a comment

kevm-pyk/src/kevm_pyk/cli.py
PetarMax: That's what it is now, but we don't know what else it will be in the future.

PR #2669, opened 2024-12-26
Removing SMT lemmas

PetarMax:
This PR removes several SMT lemmas from the main body of KEVM SMT lemmas and replaces them with appropriate simplifications.

The reason for the removal of the lemmas is that the functions in question:
- should never reach the SMT solver, in principle;
- overload Z3, making hard questions time out;
- are not required for either the test suite or engagement proofs to pass.


PetarMax:
Not required in the end.

PR #2672, opened 2025-01-06
Update `copytree` function to support Python 3.12

palinatolmach:
The `distutils.dir_util` module we've been using to access `copy_tree` function is not available in Python 3.12, which is now [being using](https://github.com/runtimeverification/evm-semantics/actions/runs/12611910880/job/35148076043#step:4:7) in the Code Quality Checks CI stage:
```
src/kevm_pyk/kdist/plugin.py:5: error: Cannot find implementation or library stub for module named "distutils.dir_util"  [import-not-found]
src/kevm_pyk/kdist/plugin.py:5: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
```

This PR imports this function from `shutil` instead, as discussed previously with @tothtamas28 and @Stevengre. It also makes the copied plugin folder and its files readable and writeable — otherwise, `nix build` complains about the missing permissions, because `shutil.copy*` functions handle permissions more conservatively than  `distutils.dir_util.copy_tree`.

### DIFF
--- a/kevm-pyk/src/kevm_pyk/kdist/plugin.py
+++ b/kevm-pyk/src/kevm_pyk/kdist/plugin.py
@@ -2,7 +2,6 @@ from __future__ import annotations
 
 import shutil
 import sys
-from distutils.dir_util import copy_tree
 from typing import TYPE_CHECKING
 
 from pyk.kbuild.utils import k_version
@@ -52,7 +51,7 @@ class KEVMTarget(Target):
 
 class PluginTarget(Target):
     def build(self, output_dir: Path, deps: dict[str, Any], args: dict[str, Any], verbose: bool) -> None:
-        copy_tree(str(config.PLUGIN_DIR), '.')
+        shutil.copytree(str(config.PLUGIN_DIR), '.')
         run_process_2(['make', '-j8'])
         shutil.copy('./build/krypto/lib/krypto.a', str(output_dir))
 
--- a/kevm-pyk/src/kevm_pyk/kdist/plugin.py
+++ b/kevm-pyk/src/kevm_pyk/kdist/plugin.py
@@ -51,7 +51,7 @@ class KEVMTarget(Target):
 
 class PluginTarget(Target):
     def build(self, output_dir: Path, deps: dict[str, Any], args: dict[str, Any], verbose: bool) -> None:
-        shutil.copytree(str(config.PLUGIN_DIR), '.')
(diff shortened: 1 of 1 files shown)


kevm-pyk/src/kevm_pyk/kdist/plugin.py
```diff
@@ -52,7 +52,9 @@ def context(self) -> dict[str, str]:
 
 class PluginTarget(Target):
     def build(self, output_dir: Path, deps: dict[str, Any], args: dict[str, Any], verbose: bool) -> None:
-        copy_tree(str(config.PLUGIN_DIR), '.')
+        shutil.copytree(str(config.PLUGIN_DIR), '.', dirs_exist_ok=True)
+        # Making the plugin directory readable and writable for nix
+        subprocess.run(['chmod', '-R', 'u+rw', '.'], check=True)
```
tothtamas28, line 57: I wonder why this is necessary for the Nix build. Are file permissions not preserved by `copytree` in this particular case?

tothtamas28 (approved):

kevm-pyk/src/kevm_pyk/kdist/plugin.py
palinatolmach: They are preserved! But in the Nix sandbox (I guess) during `nix build`, both `kevm-pyk/src/kevm_pyk/kproj/plugin` and the directory it's copied to are read-only (`dr-xr-xr-x`), even though when I run it locally it's writeable: 
```
❯ ls -ld kevm-pyk/src/kevm_pyk/kproj/plugin
drwxr-xr-x@ 15 palina  staff  480 Jan  6 12:05 kevm-pyk/src/kevm_pyk/kproj/plugin
```
Without this `chmod`, the Nix CI jobs are failing due to the permissions error, while other jobs are passing: https://github.com/runtimeverification/evm-semantics/actions/runs/12629197158/job/35186671303.

tothtamas28: > both kevm-pyk/src/kevm_pyk/kproj/plugin and the directory it's copied to are read-only

What puzzles me though, why does it work with `distutils` then? (It's not a blocker, I'd just like to understand the root cause.)

palinatolmach: Thank you @F-WRunTime! I think here we are copying files within the same file system, and it doesn't always involve a Docker container, but I'll keep that in mind!

PR #2679, opened 2025-01-14
Filter GST discarded keys in the Python harness

anvacaru:
Changes:
- Migrating the filtering of `discardKeys` from driver.md to `gst_to_kore`.
- Dropping the `quicksort` of the GST input


kevm-pyk/src/kevm_pyk/gst_to_kore.py
```diff
@@ -28,6 +28,44 @@
 SORT_MODE: Final = SortApp('SortMode')
 SORT_ETHEREUM_SIMULATION: Final = SortApp('SortEthereumSimulation')
 
+GST_DISCARD_KEYS: Final = [
```
tothtamas28, line 31: Consider making this a `set`.

tothtamas28 (approved):

kevm-pyk/src/kevm_pyk/gst_to_kore.py
anvacaru: Is there a reason to use `set` instead of `frozenset`?

tothtamas28: No, `frozenset` is probably even better.

Also, if these constants are not supposed to be public, then mark them with a `_`.

ehildenb (approved):
Love it!

PR #2680, opened 2025-01-16
Add option to automatically generate failing.llvm

anvacaru:
- Changes to the `test_conformance.py` harness:
    1. add a flag `--save-failing` to run and save failing tests to failing.llvm
    2. When the flag is active, all the tests of a GST file will be executed, otherwise, it will raise at the first failing test.
-  Regenerating failing.llvm

Before using the new flag, delete the contents of `failing.llvm`.

invocation example:
```sh
# make sure the --maxfail flag is not passed or it is set to a large number.
poetry run pytest src/tests/integration --verbose --durations=0 --dist=worksteal --numprocesses=8 --save-failing -k test_conformance.py
```

PR #2681, opened 2025-01-16
Update Ethereum-tests to its latest commit.

Robertorosmaninho:
This PR is one of the steps through fully supporting an Ethereum execution using the `Cancun` schedule. We need to test all the existing conformance tests. We updated the `failing.llvm` using the new `--save-failing` introduced by #2680.


tests/failing.llvm
anvacaru, line 1: I just realized that the order of the test files is arbitrary, depending on multiple factors like which test does Pytest pick up first to run or which test finishes first. Could you sort this file and stage the changes again? This way, we can track changes more quickly.

Robertorosmaninho: Sure!

Robertorosmaninho: Done!

anvacaru (approved):

dwightguth:
I see a bunch of tests vanished from the list of failing tests. I would expect that if we were implementing a missing feature or fixing a bug, but it feels odd for that to happen when we're updating to a new version of the tests. Have you been able to manually verify that the tests removed from the failing list in fact no longer exist?

Robertorosmaninho:
@dwightguth, The latest commits at `ethereum-tests` were deleting some tests, but the majority of the deletions on this file are related to using tests only for `Cancun`. Tests for other schedules were moved to Legacy tests. See: https://github.com/ethereum/tests/pull/1380

dwightguth (approved):

dwightguth:
Tests are probably just flaky; run them again.

PR #2682, opened 2025-01-16
Updates to the driver.md GST handler

anvacaru:
This PR makes a few minor updates to the `driver.md` and `gst_to_kore.py` test handlers.

1. `expectException` is an optional key for a block in a General State Test used to highlight an invalid block. This field represents the name of the exception which is expected to be thrown when importing this block to the  chain.
Up until now this key was in the `discardedKeys` set, but GSTs would still fail final assertions. Changes here are added to check if the status code is an `ExceptionalStatusCode` when `expectException` is present, discarding the rest of the checks for the current block.
Other vm implementations follow the same approach: 
[py-evm](https://github.com/ethereum/py-evm/blob/f2219d6df83ede8f9574eab48cbe4bc33197066f/tests/json-fixtures/blockchain/test_blockchain.py#L1406) 
[rust-evm](https://github.com/rust-ethereum/evm/blob/bb2532e7d746fa228a82a61583088cdc9affe9bb/jsontests/src/run.rs#L222)

2. Implementation of [ EIP-3706](https://eips.ethereum.org/EIPS/eip-3607) which requires the sender of a transaction to be an EOA (externally owned account, without any code deployed). 
For this EIP I added a new rule for `loadTx` to throw `EVMC_FAILURE` when the origin of the transaction has deployed code.
I couldn't find this EIP linked under any hardfork update, however it is present in the [Yellowpaper](https://ethereum.github.io/yellowpaper/paper.pdf). (under 6. Transaction Execution).

3. Initial checks for `balance underflow`, `depth exceeded`, and `nonce exceeded` were also omitted from the `loadTx` rules, which caused a number of tests to fail.

4. Changed `filter_discard_keys` to remove discarded keys from the GST file/fixture recursively.


Robertorosmaninho (approved):
LGTM, but I would wait for a second approval.
As we had some conflicts on `failing.llvm` I should ask if this is expected to really remove so many tests of it?

anvacaru:
Thank you! The conflicts were expected because the previous PR changed `failing.llvm`.

Most of the tests removed from `failing.llvm` come from `stEIP1559/intrinsic.json` and failed because the `expectException` was not handled.

PR #2687, opened 2025-01-17
Sorting `failing.llvm` using  unix `sort` with `en_US.UTF-8` locale

Robertorosmaninho:
The idea of this PR is to set a common, simple and easy-to-reproduce approach to sort `failing.llvm` when modifying it.


anvacaru:
Looks good. Can you post here the command used? Or perhaps write a new `make` command to regenerate the `failing.llvm` file and then sort the files.

Something like this (not tested)
```make
# Conformance Tests

test-conformance: poetry
	$(MAKE) -C kevm-pyk/ test-integration PYTEST_ARGS+="-k test_conformance.py"

conformance-failing-list: poetry
	echo "" > tests/failing.llvm
	$(MAKE) -C kevm-pyk/ test-integration PYTEST_ARGS+="-k test_conformance.py --save-failing --max-failing=10000"
	sort -o tests/failing.llvm tests/failing.llvm
```

anvacaru (approved):

PR #2692, opened 2025-01-22
Update loadTx rules in driver.md

anvacaru:
1. For tests such as `NoSrcAccount1559.json,*`, where the sender of the transaction is not part of the account set provided, the execution would get stuck on `loadTx(ACCT)`. To address this I initially tried adding a rule such as
```k
    rule <k> loadTx(ACCTFROM:Int) => startTx ... </k>
         <statusCode> _ => EVMC_OUT_OF_GAS </statusCode>
         <txPending> ListItem(_TXID:Int) REST => REST </txPending>
         <accounts> ACCTS </accounts>
      requires notBool `AccountCellMap:in_keys`(<acctID> ACCTFROM </acctID>, ACCTS:AccountCellMap)
```
However, I was not able to run the test, having the interpreter crash with 
```bash
src/tests/integration/test_conformance.py:53: in _test
    res = interpret({test_name: test}, schedule, mode, chainid, usegas, check=False)
src/kevm_pyk/interpreter.py:24: in interpret
    kore = KoreParser(proc_res.stdout).pattern()
../../../../.cache/pypoetry/virtualenvs/kevm-pyk-jkAxdTav-py3.11/lib/python3.11/site-packages/pyk/kore/parser.py:202: in pattern
    name = self._match(TokenType.ID)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <pyk.kore.parser.KoreParser object at 0x72589e645d90>
token_type = <TokenType.ID: 11>

    def _match(self, token_type: TokenType) -> str:
        if self._la.type != token_type:
>           raise ValueError(f'Expected {token_type.name}, found: {self._la.type.name}')
E           ValueError: Expected ID, found: EOF
```

So instead I ended up adding an `[owise]` rule.

2). Tests such as `test_contract_creating_tx[fork_Cancun-blockchain_test-over_limit_ones]` were failing for Cancun because of the BeaconRoots contract storage check. The storage of this account is modified at the start of a block with the timestamp value.  Because the transactions in the test are invalid ( and so are the blocks) the storage of the BeaconRoots contract is expected to be empty at the end of the tests. I updated the rule to skip the transaction instead of starting it.

### DIFF
--- a/kevm-pyk/src/kevm_pyk/gst_to_kore.py
+++ b/kevm-pyk/src/kevm_pyk/gst_to_kore.py
@@ -37,6 +37,7 @@ _GST_DISCARD_KEYS: Final = frozenset(
         'transactionSequence',
         'chainname',
         'lastblockhash',
+        'hasBigInt',
     ]
 )
 _GST_LOAD_KEYS: Final = frozenset(
--- a/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/driver.md
+++ b/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/driver.md
@@ -83,9 +83,10 @@ To do so, we'll extend sort `JSON` with some EVM specific syntax, and provide a 
 
     syntax EthereumCommand ::= loadTx ( Account ) [symbol(loadTx)]
  // --------------------------------------------------------------
-    rule <k> loadTx(_) => #end EVMC_OUT_OF_GAS ... </k>
+    rule <k> loadTx(_) => startTx ... </k>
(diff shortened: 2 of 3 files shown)


Robertorosmaninho (approved):
LGTM.
Although, I'm a bit unsure about the `[owise]` idea. Ideally, we would have the very specific case. However, as you said we could quickly catch any false positive case, I'll approve it.

PR #2693, opened 2025-01-23
Implementing `BLOBHASH` Opcode

Robertorosmaninho:
This PR implements the `BLOBHASH` Opcode as part of the implementation of [EIP-4844](https://eips.ethereum.org/EIPS/eip-4844#opcode-to-get-versioned-hashes).
To implement this new operation, the type of `txVersionedHashes` had to be changed to `List`.


kevm-pyk/src/kevm_pyk/kproj/evm-semantics/state-utils.md
```diff
@@ -479,6 +487,7 @@ The `"rlp"` key loads the block information.
          => #if ( notBool Ghasbasefee << SCHED >> )
                 orBool TXTYPE ==K Legacy
                 orBool TXTYPE ==K AccessList
+                orBool TXTYPE ==K Blob
```
Robertorosmaninho, line 490: I'm just unsure about this part. Do you have any thoughts, @anvacaru? Please see this [Gas Accounting](https://eips.ethereum.org/EIPS/eip-4844#gas-accounting) section of EIP-4844 for context.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
```diff
@@ -916,11 +916,19 @@ Some operators don't calculate anything, they just push the stack around a bit.
...
 
 ```k
-    syntax UnStackOp ::= "MLOAD"
- // ----------------------------
+    syntax UnStackOp ::= "MLOAD" | "BLOBHASH"
+ // -----------------------------------------
     rule <k> MLOAD INDEX => #asWord(#range(LM, INDEX, 32)) ~> #push ... </k>
          <localMem> LM </localMem>
 
+   rule <k> BLOBHASH INDEX => 0 ~> #push ... </k>
```
dwightguth, line 924: These rules seem fine but the placement of them, between MLOAD and MSTORE, seems off. I would place it next to BLOCKHASH, which it is much more similar to.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/state-utils.md
dwightguth: This seems wrong. Blob transactions are newer than DynamicFee transactions and should have the fields of all prior types, including the DynamicFee fields.

dwightguth (changes requested):

dwightguth:
There is one more problem. You are matching directly on the txVersionedHashes field, but this field exists inside a cell Map and you are not specifying which transaction you are accessing the hashes of, which means that if the block contains multiple transactions, you might conceivably return the hashes associated with a past or future transaction within the block. You probably want to either the field from the transaction into the `<evm>` cell when `loadTx` is executed. This will involve a change to the configuration and thus to the spec files in the test suite, but I don't see anything that is currently in the EVM cell that could be reliably used to tell the semantics which transaction you are interested in accessing.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
```diff
@@ -917,7 +919,7 @@ These operations are getters/setters of the local execution memory.
 
 ```k
     syntax UnStackOp ::= "MLOAD"
- // ----------------------------
+ // -----------------------------------------
```
dwightguth, line 922: This line in the diff looks like it might be a stray change you didn't intend.

dwightguth (commented):
Code looks good. I wonder why some of the blobhash related tests still fail. You may want to take a look, but obviously the first priority is getting the proofs passing. As I said before, please timebox that to a couple days at most and let me know if it seems like it will take longer than that.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/asm.md
```diff
@@ -105,6 +105,7 @@ Operator `#revOps` can be used to reverse a program.
     rule #asmOpCode( GASLIMIT       ) =>  69
     rule #asmOpCode( CHAINID        ) =>  70
     rule #asmOpCode( SELFBALANCE    ) =>  71
+    rule #asmOpCode( BLOBHASH       ) =>  73
```
anvacaru, line 108: Hmm, not related to this PR  but it seems like we're missing a rule for `BASEFEE` here.

```suggestion
    rule #asmOpCode( SELFBALANCE    ) =>  71
    rule #asmOpCode( BASEFEE        ) =>  72
    rule #asmOpCode( BLOBHASH       ) =>  73
```

Robertorosmaninho: Indeed, it seems we're not implementing [EIP-3198](https://eips.ethereum.org/EIPS/eip-3198)

anvacaru: Can you double check this please? I can see the opcode being handled [here](https://github.com/runtimeverification/evm-semantics/blob/master/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md?plain=1#L1083) , so I think it was just ommited in `#asmOpCode` rules. If it's unimplemented, then yes, we can have a new PR, otherwise I think we can just add it here.

Robertorosmaninho: Yeah, the feature seems to be the case. Thanks for pointing that out! I'll add it here! Good catch!

Robertorosmaninho: Done!

dwightguth (approved):

tests/failing.llvm
```diff
@@ -2,19 +2,35 @@ BlockchainTests/GeneralStateTests/Cancun/stEIP4844-blobtransactions/blobhashList
 BlockchainTests/GeneralStateTests/Cancun/stEIP4844-blobtransactions/blobhashListBounds4.json,*
 BlockchainTests/GeneralStateTests/Cancun/stEIP4844-blobtransactions/blobhashListBounds5.json,*
 BlockchainTests/GeneralStateTests/Cancun/stEIP4844-blobtransactions/blobhashListBounds6.json,*
+BlockchainTests/GeneralStateTests/Cancun/stEIP4844-blobtransactions/blobhashListBounds7.json,*
```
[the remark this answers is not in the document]
Robertorosmaninho: From what I saw, it's mostly because these tests were supposed to fail. However, they were failing for the wrong reasons and now, they're "passing" when a failure is expected. 

The reasons vary a lot, from wrong balance (lack of implementation of gas accounting) to the [block validation](https://eips.ethereum.org/EIPS/eip-4844#execution-layer-validation).

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/word.md
```diff
@@ -557,6 +558,7 @@ Range of types
     rule #rangeNonce    (   X          ) => #range ( 0               <= X < maxUInt64        )
     rule #rangeSmall    (   X          ) => #range ( 0               <= X < 10               )
     rule #rangeBlockNum (   X          ) => #range ( 0               <= X <= maxBlockNum     )
+    rule #rangeNegUInt64(   X          ) => #range ( pow56           <= X < pow64            )
```
dwightguth, line 561: if this is intended to tell you whether a UInt64 when cast to an Int64 is negative, it is implemented wrong. The sign of an integer is a bit, not a whole byte.

Robertorosmaninho: Not necessarily, the idea is actually to see if a `word` of UInt64 is negative. This was inspired by this `sgn` function: https://github.com/runtimeverification/evm-semantics/blob/bfd9e29ce15203b4db5f823ad362d0ab6a8b30ca/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm-types.md?plain=1#L45-L50

PR #2698, opened 2025-02-01
Refactoring the custom_step harness

anvacaru:
This PR brings changes to the custom_step logic mechanism in the `KEVMSemantics` module to improve clarity and scalability:
 1. as pointed out, there's not much sense in caching the substitution between matching the pattern and applying the `execute_custom_step` function, as these two occur one after the other. So I'm removing `_cached_subst`, and now the `Subst` is passed directly as an argument to the `execute_custom_step` function (closing https://github.com/runtimeverification/kontrol/issues/939).
 2. adding `_custom_steps` as an immutable list of `Tuple[abstract_pattern, pattern_custom_step_function]`, described below
 3. Instead of checking patterns using `if` statements ( [see instance in subclass here](https://github.com/runtimeverification/kontrol/blob/master/src/kontrol/foundry.py#L285)), I'm changing `custom_step` to iterate over `_custom_step_definitions`, trying to match the `K_CELL` against each pattern and executing the corresponding function if a match is found. This approach removes the overloaded instances of `custom_step` and `can_take_custom_step` from subclasses ([Example of updated usage here](https://github.com/runtimeverification/kontrol/pull/956/files#diff-59cc0840d8befef59d646ac05de0d54e60f10330329cfb480e8b586751a70230R114-R119)).


kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
```diff
@@ -1438,7 +1438,8 @@ The various `CALL*` (and other inter-contract control flow) operations will be d
            ...
          </account>
 
-    rule <k> #call ACCTFROM ACCTTO ACCTCODE VALUE APPVALUE ARGS STATIC
+    rule [call.false]:
+         <k> #call ACCTFROM ACCTTO ACCTCODE VALUE APPVALUE ARGS STATIC
```
ehildenb, line 1442: Should we label the corresponding other one `call.true` then?

kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -53,12 +53,14 @@
...
+        tuple[KSequence, Callable[[Subst, CTerm, CTermSymbolic], KCFGExtendResult | None]], ...
+    ]
 
     def __init__(self, auto_abstract_gas: bool = False, allow_symbolic_program: bool = False) -> None:
         self.auto_abstract_gas = auto_abstract_gas
         self.allow_symbolic_program = allow_symbolic_program
-        self._cached_subst = None
+        self._custom_step_definitions = ((self._load_pattern, self._exec_load_custom_step),)
 
     @staticmethod
```
ehildenb, line 65: Does Kontrol override this `custom_step_definitions` itself with some more patterns? If so, maybe it should be an argument to the constructor `__init__` that can be `| None`, and if it's `None` we set it to this one specific pattern?

anvacaru: Yes, Kontrol will add more patterns on top of those defined here. In the `KontrolSemantics.__init__` I'm calling `super().__init__` to instantiate KEVMSemantics, then I'm overriding `custom_step_definitions` as 
```py
    def __init__(
        self, auto_abstract_gas: bool = False, allow_symbolic_program: bool = False, provider_url: str | None = None
    ) -> None:
        super().__init__(auto_abstract_gas=auto_abstract_gas, allow_symbolic_program=allow_symbolic_program)
        ...
        self._custom_step_definitions = self._custom_step_definitions + (
            (self._rename_pattern, self._exec_rename_custom_step),
            (self._forget_branch_pattern, self._exec_forget_custom_step),
            (self._call_fork_pattern, self._exec_fork_call_custom_step),
            (self._sload_fork_pattern, self._exec_fork_sload_custom_step),
        )
```
([link](https://github.com/runtimeverification/kontrol/pull/956/files#diff-59cc0840d8befef59d646ac05de0d54e60f10330329cfb480e8b586751a70230R114))

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
anvacaru: Yes, I don't think the `call.true` will be used in the short future, but I added it for consistency together with other labels.

kevm-pyk/src/kevm_pyk/kevm.py
anvacaru: I updated the `__init__` function to take the optional argument. I like this approach better.

kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -53,12 +53,26 @@
 class KEVMSemantics(DefaultSemantics):
     auto_abstract_gas: bool
     allow_symbolic_program: bool
-    _cached_subst: Subst | None
+    _custom_step_definitions: tuple[
+        tuple[KSequence, Callable[[Subst, CTerm, CTermSymbolic], KCFGExtendResult | None]], ...
```
tothtamas28, line 57: Consider extracting a class with an attribute (`KSequence`) and a method (`Callable[....]`).

anvacaru: Hmm, should I also encapsulate the execution inside of the class? For example, adding a method like 
```py
    def try_execute(self, cterm: CTerm, cterm_symbolic: CTermSymbolic) -> KCFGExtendResult | None:
        subst = self.pattern.match(cterm.cell('K_CELL'))
        if subst is not None:
            return self.exec_fn(subst, cterm, cterm_symbolic)
        return None
```

tothtamas28: Sounds reasonable to me.

anvacaru: I implemented the class. Not sure if I should instantiate the `CustomStep` elements in the `__init__` itself or somewhere else.

kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -47,18 +47,51 @@
...
+
+
 # KEVM class
 
 
 class KEVMSemantics(DefaultSemantics):
     auto_abstract_gas: bool
     allow_symbolic_program: bool
-    _cached_subst: Subst | None
+    _custom_steps: tuple[CustomStep]
```
tothtamas28, line 79:
```suggestion
    _custom_steps: tuple[CustomStep, ...]
```

kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -47,18 +47,51 @@
 
 _LOGGER: Final = logging.getLogger(__name__)
 
+
+class CustomStep:
+    """Encapsulates a custom step definition consisting of an abstract pattern and its execution function."""
+
+    pattern: KSequence
+    exec_fn: Callable[[Subst, CTerm, CTermSymbolic], KCFGExtendResult | None]
```
tothtamas28, line 55: You can turn this into a method.

tothtamas28: I guess probably not, but you can consider extracting another class for it.

kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -47,18 +47,51 @@
 
 _LOGGER: Final = logging.getLogger(__name__)
 
+
+class CustomStep:
```
tothtamas28, line 51: If `__init__` only sets the attributes then this can become a `NamedTuple`:

```suggestion
class CustomStep(NamedTuple):
```

kevm-pyk/src/kevm_pyk/kevm.py
```diff
@@ -76,13 +70,13 @@ def try_execute(self, cterm: CTerm, cterm_symbolic: CTermSymbolic) -> KCFGExtend
...
     allow_symbolic_program: bool
-    _custom_steps: tuple[CustomStep]
+    _custom_steps: tuple[CustomStep, ...]
 
     def __init__(
         self,
         auto_abstract_gas: bool = False,
         allow_symbolic_program: bool = False,
-        custom_step_definitions: tuple[CustomStep] = None,
+        custom_step_definitions: tuple[CustomStep] | None = None,
```
tothtamas28, line 79:
```suggestion
        custom_step_definitions: tuple[CustomStep, ...] = (),
```

tothtamas28 (approved):

kevm-pyk/src/kevm_pyk/kevm.py
anvacaru: Added a new type alias in https://github.com/runtimeverification/evm-semantics/pull/2698/commits/012224537ccaaa948f06b58132bb193cbbe0a3ac.

PR #2702, opened 2025-02-06
Upstream symbolic `bytes` lookup, `List` membership lemmas

palinatolmach:
Closes https://github.com/runtimeverification/evm-semantics/issues/2476.

This PR adds lemmas for lookups in partially symbolic `bytes` arrays; specifically, in runtime bytecode of contracts that contain `immutable` variables that are initialized in a constructor with a symbolic value. It also adds `List` membership lemmas that facilitate checking if a performed call is allowed in Kontrol when the `allowCalls` cheatcode is used.

Once this PR is merged, the relevant lemmas should be removed from `KONTROL-AUX-LEMMAS` and https://github.com/runtimeverification/kontrol/blob/master/src/tests/integration/test-data/symbolic-bytes-lemmas.k.


lucasmt (approved):
I'm approving the PR, just have one comment above that I think needs to be addressed before it's merged

palinatolmach:
I opened a PR in Kontrol to test if upstreamed lemmas don't negatively affect the tests and are applied if I take out the same lemmas from Kontrol aux files. The tests are passing: https://github.com/runtimeverification/kontrol/pull/1059 (there's a minor expected output diff in some `end-to-end` tests that don't seem relevant). I updated the bytes lookup lemma according to @lucasmt's suggestion too. @anvacaru I'd merge it tomorrow if it looks good to you too.

anvacaru (approved):
LGTM!

PR #2712, opened 2025-03-04
Implementing Gas Accounting from EIP-4844

Robertorosmaninho:
This PR introduces `#calcBlobFee` and `#totalBlobGas` helper functions to calculate the correct amount of `wei` to decrease from the balance of the sender following the EIP-4844 specification: https://eips.ethereum.org/EIPS/eip-4844#gas-accounting


anvacaru (approved):

Robertorosmaninho:
@Stevengre Can you help me to understand why a test is failing in the new test-suite you introduced recently?

Stevengre:
I'm sorry for this confusing, but you might need to run it a second time.

Stevengre:
It looks like a weird SMT time out error that I need to investegate. You can skip it in my test in integration/test-summarize.py.

Robertorosmaninho:
> It looks like a weird SMT time out error that I need to investegate. You can skip it in my test in integration/test-summarize.py.

It fails in 2 different tests in the first and the second time. Should I skip both?

Stevengre:
Yes, you can. But please leave me some useful message when you skip them, so that I can investigate after it.

PR #2723, opened 2025-03-14
Use the upstream `llvm_interpret` function

Stevengre:
- Simplified the `interpret` function by removing error handling for subprocess calls and directly utilizing `llvm_interpret`.
- Updated the `interpret` function to improve clarity and efficiency by integrating the filtering and conversion of GST data into a single line.
- Enhanced documentation to clarify the purpose of the `interpret` function.

### DIFF
--- a/kevm-pyk/src/kevm_pyk/interpreter.py
+++ b/kevm-pyk/src/kevm_pyk/interpreter.py
@@ -1,10 +1,11 @@
 from __future__ import annotations
 
+from subprocess import CalledProcessError
 from typing import TYPE_CHECKING
 
 from pyk.kdist import kdist
 from pyk.kore.parser import KoreParser
-from pyk.utils import run_process_2
+from pyk.ktool.krun import llvm_interpret_raw
 
 from .gst_to_kore import filter_gst_keys, gst_to_kore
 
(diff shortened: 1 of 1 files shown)


kevm-pyk/src/kevm_pyk/interpreter.py
```diff
@@ -21,6 +21,9 @@ def interpret(gst_data: Any, schedule: str, mode: str, chainid: int, usegas: boo
     if check:
         proc_res.check_returncode()
 
+    if len(proc_res.stderr) > 0:
+        raise Exception(proc_res.stderr)
```
tothtamas28, line 25: This seems a bit convoluted. If `check` is `True`, then this code is I think not reachable. Otherwise, you do not want an exception to be raised.

tothtamas28: The call to `run_process_2` in `_interpret` should probably be replaced by `llvm_interpret` from `pyk.ktool.krun`.

Stevengre: > This seems a bit convoluted. If check is True, then this code I think is not reachable. Otherwise, you do not want an exception to be raised.

If `check` is `True`, we still need this. Because the `returncode` is empty, and `proc_res.check_returncode()` doesn't work for this case.

kevm-pyk/src/kevm_pyk/interpreter.py
```diff
@@ -16,18 +17,18 @@
 
 
 def interpret(gst_data: Any, schedule: str, mode: str, chainid: int, usegas: bool, *, check: bool = True) -> Pattern:
-    proc_res = _interpret(gst_data, schedule, mode, chainid, usegas)
+    try:
```
tothtamas28, line 20: At this point, this is basically `llvm_interpret`. This is good, since we can just call that instead of `_interpret`. Please check if `_interpret` is called elsewhere, and if not, let's eliminate it in favor of `llvm_interpret`.

Stevengre: Sure. `_interpret` is only used in `interpreter.py`. I think that I can remove it.

kevm-pyk/src/kevm_pyk/interpreter.py
```diff
@@ -3,31 +3,17 @@
...
-
-def _interpret(gst_data: Any, schedule: str, mode: str, chainid: int, usegas: bool) -> CompletedProcess:
-    gst_data_filtered = filter_gst_keys(gst_data)
-    interpreter = kdist.get('evm-semantics.llvm') / 'interpreter'
-    init_kore = gst_to_kore(gst_data_filtered, schedule, mode, chainid, usegas)
-    proc_res = run_process_2([str(interpreter), '/dev/stdin', '-1', '/dev/stdout'], input=init_kore.text, check=False)
-    return proc_res
+    """Interpret the given GST data using the LLVM backend."""
+    init_kore = gst_to_kore(filter_gst_keys(gst_data), schedule, mode, chainid, usegas)
+    return llvm_interpret(kdist.get('evm-semantics.llvm-summary'), init_kore, check=check)
```
tothtamas28, line 19:
```suggestion
    return llvm_interpret(kdist.get('evm-semantics.llvm'), init_kore, check=check)
```

tothtamas28 (approved):

PR #2730, opened 2025-03-25
Fixing loose ends in driver.md testing harness

anvacaru:
Fixing two bugs that were causing some tests to fail.
1. The available gas cell would not reset between successive transactions, causing the `gasUsed`(`txGasLimit - gasAvailable`) value to mismatch and even become negative in some cases.
2. Transient storage values were not being reverted at the end of a transaction.


kevm-pyk/src/kevm_pyk/kproj/evm-semantics/driver.md
```diff
@@ -73,6 +73,7 @@ To do so, we'll extend sort `JSON` with some EVM specific syntax, and provide a
     rule <k> startTx => loadTx( #sender( #getTxData(TXID), TW, TR, TS, B ) ) ... </k>
          <txPending> ListItem(TXID:Int) ... </txPending>
          <chainID> B </chainID>
+         <gas> _ => 0 </gas>
```
anvacaru, line 76: Initially, I tried using `clearTX` after `finalizeTx(_)`; however, this increased the number of failed tests.

Robertorosmaninho (approved):
The modification makes sense to me and as it doesn't add any regression it ought to be fine to assume this was the missing feature to fix the behavior of self-destruction in KEVM.

palinatolmach (approved):

PR #2732, opened 2025-04-02
Optimize summary rules

Stevengre:


kevm-pyk/src/kevm_pyk/summarizer.py
```diff
@@ -289,6 +289,35 @@ def get_todo_list() -> list[str]:
     return todo_list
 
 
+def stack_added(opcode_id: str) -> int:
```
tothtamas28, line 292:
```suggestion
def stack_added(opcode: str) -> int:
```

(Similarly for the rest.)

tothtamas28 (commented):
LGTM. I'll let @JuanCoRo approve the PR though, as he has a better understanding of the implications of this change on the proving process.

kevm-pyk/src/kevm_pyk/summarizer.py
Stevengre: I'm working on the better rule id and will fix it latter!

Stevengre: Done!

JuanCoRo: Hey @Stevengre! Should the `-SUMMARY-USEGAS` labeling be in place already? For what I see in the last commit it's still using the old one (e.g., `rule [ADD-SUMMARY-0]:`)
Is this meant for a follow up PR? Thank you!

Stevengre: Sure, I'll provide the better names in the next PR. Just after merging this one.

Stevengre: For this one, I make the require clause more feasible for some opcode and reduce the rules to make it easier to generate better names.

Stevengre: There are still some problems for some opcodes, like `balance`. But I think that I can resolve it when you need.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/summaries/pushzero-summary.k
```diff
@@ -63,14 +35,12 @@ module PUSHZERO-SUMMARY
            ...
          </kevm>
       requires ( USEGAS_CELL:Bool
-       andBool ( ( notBool #sizeWordStack ( WS:WordStack , 0 ) <Int 0 )
-       andBool ( ( notBool 1023 <Int #sizeWordStack ( WS:WordStack , 0 ) )
+       andBool ( #sizeWordStack ( WS:WordStack , 0 ) <=Int 1023
```
JuanCoRo, line 38: These conditions here are not exactly equivalent, right? I think we would still need something like
```
0 <=Int #sizeWordStack ( WS:WordStack , 0 )
```
Am I missing something?

Stevengre: This is generated automatically, since I don't provide any not underflow condition for the opcodes. This follows the same principle as the not overflow before opcode execution that you might need to add this assumption in the theorems.

JuanCoRo: Do you know why there's an overflow requirement but not an underflow? I'm just trying to understand where this condition get s dropped.

Stevengre: Wait a minute... Let me check the semantics...

Stevengre: Reason for this:

The following new-introduced code provides `not overflow` precondition for the opcode summaries.

https://github.com/runtimeverification/evm-semantics/blob/e7498c3478754bd25c9f15450eb47d6d64cd2778/kevm-pyk/src/kevm_pyk/summarizer.py#L624

As a result, the following rule applies without a split according to the conditions.

https://github.com/runtimeverification/evm-semantics/blob/e7498c3478754bd25c9f15450eb47d6d64cd2778/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md?plain=1#L347-L356

----

However, without this new precondition, there will be a split for 

```
notBool ( #stackUnderflow(WS, OP) orBool #stackOverflow(WS, OP) ) 
andBool notBool ( STATIC andBool #changesState(OP, WS) ) 
```

And it seems like the haskell backend doesn't reduce `notBool ( #stackUnderflow(WS, OP) orBool #stackOverflow(WS, OP) ) ` to its simplest conjunctive normal form (CNF) -- `notBool #stackUnderflow(WS, OP) andBool notBool #stackOverflow(WS, OP)`.

Therefore, when `notBool ( #stackUnderflow(WS, OP) orBool #stackOverflow(WS, OP) )` is false, since `#stackUnderflow(WS, OP)` is always false for pushzero, then this constraint is reduced to `#stackOverflow(WS, OP)` and evaluate to `1023 <Int #sizeWordStack ( _WS:WordStack , 0 )`.

When `notBool ( #stackUnderflow(WS, OP) orBool #stackOverflow(WS, OP) )` is true, ???

---

Sorry, forget it... It looks strange to me. It might also relate to the existing simplifcation rules like:

https://github.com/runtimeverification/evm-semantics/blob/e7498c3478754bd25c9f15450eb47d6d64cd2778/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/lemmas.k#L49

https://github.com/runtimeverification/evm-semantics/blob/e7498c3478754bd25c9f15450eb47d6d64cd2778/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/int-simplification.k#L218-L219

And it might also relate to the way that the backend tackle the side conditions and simplication rules.

---

I'm sorry, Juan. I'm actually not quite sure why such a strange condition was generated previously. This is indeed an interesting issue. I think @ehildenb  and @jberthold  might have more thoughts on what could be causing this result.

jberthold: > These conditions here are not exactly equivalent, right? I think we would still need something like
> 
> ```
> 0 <=Int #sizeWordStack ( WS:WordStack , 0 )
> ```
> 
> Am I missing something?

> Reason for this:
> ...
> https://github.com/runtimeverification/evm-semantics/blob/e7498c3478754bd25c9f15450eb47d6d64cd2778/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/lemmas.k#L49

I think the simplification that @Stevengre points out makes the `0 <= ...` condition redundant (specialise `N == 0` in the lemma).


And just to clarify:

> And it seems like the haskell backend doesn't reduce `notBool ( #stackUnderflow(WS, OP) orBool #stackOverflow(WS, OP) ) ` to its simplest conjunctive normal form (CNF) -- `notBool #stackUnderflow(WS, OP) andBool notBool #stackOverflow(WS, OP)`.

Correct for the `booster` backend, we are trying to not transform predicates to CNF (the only transformation we do is to turn ML predicates into 1st-order ones). The legacy backend might have logic to transform to CNF but this is actually not a desired thing - we want all simplifications to be readable K code rather than backend magic.

In this case, we could maybe just write the CNF into the rule in the first place?

JuanCoRo: Thank you @Stevengre and @jberthold!!
It makes sense to me now, so I'll approve the PR!

Stevengre: Thank you @jberthold! It makes a lot of sence! But I think that we can also make this CNF during the frontend process. I'll make an issue to record this.

PR #2737, opened 2025-04-07
Update opcode summaries to standardize labels and improve clarity

Stevengre:
- Changed labels for various opcode summaries to follow a consistent naming convention, replacing instances of `-1` with `-NOGAS` or `-USEGAS` where applicable.
- This update enhances readability and maintains uniformity across the summary definitions, ensuring better understanding and maintenance of the codebase.


kevm-pyk/src/kevm_pyk/summarizer.py
```diff
@@ -407,6 +407,63 @@ def _transform_dash_aux(inner: KInner) -> KInner:
...
+    le0 = False
+    le_length_bytes = False
+
+    def _transform_rule_id_nogas(_inner: KInner) -> KInner:
+        nonlocal flag
+        if (
+            isinstance(_inner, KApply)
+            and _inner.label.name == 'notBool_'
+            and _inner.args[0] == KVariable('USEGAS_CELL', 'Bool')
+        ):
```
tothtamas28, line 422: Consider something like

```suggestion
        match _inner:
            case KApply(KLabel('notBool_'), (KVariable('USEGAS_CELL', 'Bool'), *_)):
```

(And similarly for other such statements.)

kevm-pyk/src/kevm_pyk/summarizer.py
```diff
@@ -407,6 +407,63 @@ def _transform_dash_aux(inner: KInner) -> KInner:
     return top_down(_transform_dash_aux, inner)
 
 
+def _transform_rule_id(proof_id: str, requires: KInner) -> str:
+    flag = 'USEGAS'
+    noberlin = False
+    le0 = False
+    le_length_bytes = False
```
tothtamas28, line 414: It looks like these variables are each set from exactly one `_transform_*` function, so consider consolidating this logic a bit:

```python3
flag = _compute_flag(requires)
```

where

```python3
def _compute_flag(inner: KInner) -> bool:
    flag = 'USEGAS'
    
    def _transform_rule_id_nogas(inner: KInner) -> KInner:
        nonlocal flag
        ...

    top_down(_transform_rule_id_nogas, inner)
    return flag
```

kevm-pyk/src/kevm_pyk/summarizer.py
Stevengre: you mean just using one aux function and `match-case` to achieve the current goal?

Stevengre: This seems to conflict with the following consolidation requirement.

Stevengre:
```python
def _compute_flag(_inner: KInner) -> KInner:
    nonlocal ...
    match _inner:
        case: ...
            nogas = True
        case: ...
            noberlin = True
...
```

What about this?

tothtamas28: > This seems to conflict with the following consolidation requirement.

They are independent from one another:
* one is about what's inside a `_transform_*` function: change the `if` to a `match`
* the other how these `_transform` functions are called: nest them within a `_calculate_*` function that wraps the `nonlocal`.

Stevengre: I'm trying to use `match-case`, but it doesn't work as `if`. The following is what I tried.

```
        if _inner == KApply('_<Int_', [KApply('lengthBytes(_)_BYTES-HOOKED_Int_Bytes', [KVariable('OUTPUT_CELL', 'Bytes')]), KApply('_+Int_', [KVariable('W1', 'Int'), KVariable('W2', 'Int')])]):
            invalid = True
        # match _inner:
        #     case KApply(
        #         '_<Int_', (KApply('lengthBytes(_)_BYTES-HOOKED_Int_Bytes', (KVariable('OUTPUT_CELL', 'Bytes'), *_)))
        #     ):
        #         invalid = True
```

Stevengre: I updated the implementation. But I cannot use `match-case` to replace the current `if` statement to achieve the same result.

tothtamas28 (approved):

PR #2742, opened 2025-04-16
Update to conformance test suite

anvacaru:
Adding support for the new execution-spec-tests conformance suite of the EVM.
Creating a new dir under `tests/execution-spec-tests` with:
1. a new `get-execution-spec-tests.sh` script inspired by the one provided in the documentation ([source](https://github.com/ethereum/execution-spec-tests/blob/84d28dbc27fb36e040c8779d822be99f80b16547/docs/consuming_tests/index.md)) that downloads a given version of the JSON fixtures from the official GH release page. 
2. a new `kevm-pyk/src/tests/integration/test_execution_spec_tests.py` pytest file that executes four categories of tests: blockchain, blockchain-engine, state and transactions. Duplicated logic from the old kevm-`pyk/src/tests/integration/test_conformance.py` has been refactored under `kevm-pyk/src/tests/utils.py`
3. new `execution-spec-tests/failing.llvm` file.
4. I added the `test-fixtures` and `fixtures-failing-list` Makefile rules that download the JSON fixtures. They are used on CI.


tests/execution-spec-tests/get_execution_spec_tests.sh
```diff
@@ -0,0 +1,32 @@
...
+# Download and extract
+echo "Downloading fixture artifacts $ARTIFACT version $VERSION..."
+curl -LO $DOWNLOAD_URL
+
+echo "Extracting artifacts to $TARGET_DIR..."
+tar -xzf "$ARTIFACT" --strip-components=1 -C "$TARGET_DIR"
+
+# Cleanup
+echo "Deleting tar file..."
+rm $ARTIFACT
```
ehildenb, line 32: This seems like something we can have directly int he python testing harness instead of a separate script, and use the pytest fixtures instead of the makefile to manage downloading and extracting all the tests ahead of time.

tests/execution-spec-tests/slow.llvm
ehildenb, line 1: It looks like this is empty, maybe we can just omit it and not use it for now?

ehildenb:
So now the `test-integration` will run both the old tests in the old format, and the new ones? And I guess the plan is that eventually we'll pass all the new ones and phase out the old ones?

tests/execution-spec-tests/get_execution_spec_tests.sh
anvacaru: Yes, that was my initial approach. However, because each list of tests is computed when the module is loaded ([as, for example, here](https://github.com/runtimeverification/evm-semantics/blob/660cf3dff7bc8dc542d19ee58b18ce254b588798/kevm-pyk/src/tests/integration/test_execution_spec_tests.py#L29)), I couldn't have a fixture to download the tests before that.

tests/execution-spec-tests/slow.llvm
anvacaru: Agreed.

tests/execution-spec-tests/get_execution_spec_tests.sh
ehildenb: Ah OK. Thanks for teh explanation.

tests/execution-spec-tests/slow.llvm
anvacaru: I removed it.

tests/execution-spec-tests/get_execution_spec_tests.sh
```diff
@@ -0,0 +1,32 @@
...
+# - fixtures_develop.tar.gz: Generated up to and including the latest dev fork.
+# As of March 2024, dev is Prague, deployed is Cancun.
+
+ARTIFACT="fixtures_develop.tar.gz"
+TARGET_DIR="fixtures"
+
+OWNER="ethereum"
+REPO="execution-spec-tests"
+VERSION="v4.2.0" #"latest"
+DOWNLOAD_URL="https://github.com/$OWNER/$REPO/releases/download/$VERSION/$ARTIFACT"
```
ehildenb, line 18: Maybe we should source the `VERSION` from its own file directly, that way it's easy to update it in place automatically if we want later. But this is fine for now.

tests/execution-spec-tests/get_execution_spec_tests.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+set -e
```
ehildenb, line 2:
```suggestion
#!/bin/bash
set -euxo pipefail
```

This set of options makes it so that it will fail the whole script in any part fails.

Makefile
```diff
@@ -51,6 +51,29 @@ conformance-failing-list: poetry
...
 	fi
 
+download-json-fixtures:
+	@if [ ! -d "tests/execution-spec-tests/fixtures" ]; then \
+		echo "Fixtures directory not found. Downloading..."; \
+		cd tests/execution-spec-tests; \
+		sh get_execution_spec_tests.sh; \
+	else \
+		echo "Fixtures directory found. Skipping download."; \
+	fi
```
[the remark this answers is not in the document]
anvacaru: Downloading the file should take well under a minute. I agree with your solution, but instead of `sh`, I used `bash` to invoke the script as `sh` would complain about `-o pipefail`.

tests/execution-spec-tests/get_execution_spec_tests.sh
anvacaru: Done! I added a new `VERSION` file and a resolution order:
1. `FIXTURE_VERSION` (env override)
2. `VERSION` file
3. ` latest` string literal if the previous two are not defined or the file does not exist.

tests/execution-spec-tests/get_execution_spec_tests.sh
anvacaru: Additionally, I removed any `echo`s in favor of the `x` trace option.

Makefile
```diff
@@ -51,6 +51,25 @@ conformance-failing-list: poetry
 		sed -i '1{/^[[:space:]]*$$/d;}' tests/failing.llvm ;\
 	fi
 
+download-json-fixtures:
+	rm -rf tests/execution-spec-tests/fixtures
+	cd tests/execution-spec-tests ;\
+	bash get_execution_spec_tests.sh
```
ehildenb, line 57:
```suggestion
	cd tests/execution-spec-tests && bash get_execution_spec_tests.sh
```

Using `&&` ensures that if the first command fails (if the directory isn't there), then the whole line will fail.

ehildenb (approved):
LGTM!

PR #2747, opened 2025-04-23
Add rules for new Prague schedule

anvacaru:
- defining new `Prague` schedule and setting it as main schedule for testing
- bringing in some changes from https://github.com/runtimeverification/evm-semantics/pull/2746:
  - reverting a change in the `interpret` function to automatically identify a schedule as it is not implemented correctly yet.
  - Fixing typos in Makefile for `fixtures-failing-list`
  - Add missing rules to load env values for `Randao` and `excessBlobGas`


kevm-pyk/src/kevm_pyk/cli.py
```diff
@@ -480,7 +480,7 @@ class EVMChainOptions(Options):
     @staticmethod
     def default() -> dict[str, Any]:
         return {
-            'schedule': 'CANCUN',
+            'schedule': 'PRAGUE',
```
ehildenb, line 483: Is this changing the default to `PRAGUE`? Should we wait for us to support it fully first? Or do we already support a lot of the functionality?

ehildenb (approved):

kevm-pyk/src/kevm_pyk/cli.py
anvacaru: Ah, yes! Good catch. I'll revert this change and set it as the default later.

palinatolmach (approved):

PR #2759, opened 2025-05-08
EIP 2537 - Precompile for BLS12-381 curve operations

anvacaru:
- Implementing [EIP-2537](https://eips.ethereum.org/EIPS/eip-2537). 
- Imported from https://github.com/Pi-Squared-Inc/evm-semantics/pull/42 and adjused for our KEVM configuration.
- BLS hooks have been added in https://github.com/runtimeverification/blockchain-k-plugin/pull/213.

#### Overview
- Adds seven new BLS12 precompiled contracts (addresses 11-17) for the Prague fork
- Extends precompiled contract upper bound from 10 to 17

#### New Precompiled Contracts
- `BLS12_G1ADD`: Point addition in G1 (curve over base field)
- `BLS12_G1MSM`: Multi-scalar-multiplication in G1
- `BLS12_G2ADD`: Point addition in G2 (curve over quadratic extension)
- `BLS12_G2MSM`: Multi-scalar-multiplication in G2
- `BLS12_PAIRING_CHECK`: Pairing operations between G1 and G2 points
- `BLS12_MAP_FP_TO_G1`: Maps base field element to G1 point
- `BLS12_MAP_FP2_TO_G2`: Maps extension field element to G2 point

- Adds utility functions for BLS12 curve operations
- Implements validation logic for coordinates, scalars, and curve points
- Defines error handling for invalid inputs

#### Gas costs
- G1 operations: add (375), mul (12000)
- G2 operations: add (600), mul (22500)
- Mapping: Fp→G1 (5500), Fp2→G2 (23800)
- Pairing: base (37700) + per pair (32600)
- MSM operations include discount tables based on batch size


tests/execution-spec-tests/failing.llvm
```diff
@@ -1639,57 +1639,33 @@ blockchain_tests/cancun/eip6780_selfdestruct/dynamic_create2_selfdestruct_collis
...
+blockchain_tests/prague/eip2537_bls_12_381_precompiles/bls12_g2add/invalid.json,tests/prague/eip2537_bls_12_381_precompiles/test_bls12_g2add.py::test_invalid[fork_Prague-blockchain_test_from_state_test--invalid_point_b_5-]
+blockchain_tests/prague/eip2537_bls_12_381_precompiles/bls12_g2msm/invalid.json,tests/prague/eip2537_bls_12_381_precompiles/test_bls12_g2msm.py::test_invalid[fork_Prague-blockchain_test_from_state_test--bls_g2_truncated_input-]
+blockchain_tests/prague/eip2537_bls_12_381_precompiles/bls12_g2mul/invalid.json,tests/prague/eip2537_bls_12_381_precompiles/test_bls12_g2mul.py::test_invalid[fork_Prague-blockchain_test_from_state_test--bls_g1_add_input_invalid_length-]
+blockchain_tests/prague/eip2537_bls_12_381_precompiles/bls12_g2mul/invalid.json,tests/prague/eip2537_bls_12_381_precompiles/test_bls12_g2mul.py::test_invalid[fork_Prague-blockchain_test_from_state_test--bls_g2_truncated_input-]
+blockchain_tests/prague/eip2537_bls_12_381_precompiles/bls12_g2mul/invalid.json,tests/prague/eip2537_bls_12_381_precompiles/test_bls12_g2mul.py::test_invalid[fork_Prague-blockchain_test_from_state_test--bls_g2mul_short_input-]
```
Robertorosmaninho, line 1667: I'm curious to understand why these tests are still failing. Is there some block validation we are missing here?

anvacaru: I'm not sure right now. Execution fails on matching the `<gasUsed>` cell, so it might be because [EIP-7623](https://eips.ethereum.org/EIPS/eip-7623) is not implemented in this branch. I tried merging https://github.com/runtimeverification/evm-semantics/pull/2756 locally, but these proofs still fail.

Robertorosmaninho: From my experience, this mismatch on call gas usually means that we're executing something we shouldn't or not executing something we should. Could this be the case here?

anvacaru: Thanks for your suggestion.

Indeed, the remaining tests in `invalid.json` you see above fail because eip-7623 is not implemented in this branch. ([see this draft PR with both eips](https://github.com/runtimeverification/evm-semantics/pull/2761/files#diff-67864c9d56a0a12e77fe295edac0122a4e913d5edbaf101b3236055d4de40db5L1618-L1668)). The remaining `valid.json` tests are failing because I tried using `#asWord` instead of `Bytes2Int`. I guess that the use of `chop` is messing up the calculation. These are also addressed in the PR linked earlier. I will push a commit to revert the use of `#asWord`.

The remaining `precompile_before_fork.json` fails because we do not account for the block number when deciding which schedule is active.

Robertorosmaninho (approved):
LGTM!

tests/execution-spec-tests/failing.llvm
Robertorosmaninho: Perfect! Thanks for the clarification!

PR #2834, opened 2025-12-18
eip-7825: transaction gas limit cap

anvacaru:
Adding a check in `#isValidTransaction` rules to check that the transaction gas limit is below the cap in schedules where [EIP-7825](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-7825.md) is active.


palinatolmach (approved):

tests/execution-spec-tests/failing.llvm
```diff
@@ -614,6 +614,7 @@ blockchain_tests/osaka/eip7951_p256verify_precompiles/test_valid.json,*
 blockchain_tests/osaka/eip7951_p256verify_precompiles/test_wycheproof_extra.json,*
 blockchain_tests/osaka/eip7951_p256verify_precompiles/test_wycheproof_invalid.json,*
 blockchain_tests/osaka/eip7951_p256verify_precompiles/test_wycheproof_valid.json,*
+blockchain_tests/osaka/eip7825_transaction_gas_limit_cap/test_transaction_gas_limit_cap_at_transition.json,*
```
palinatolmach, line 617: This can be addressed in a follow up (or removed altogether once the tests pass), but `test_transaction_gas_limit_cap_at_transition` is already excluded in line 426 (where all 7825 fixtures are excluded, so this functionality is not being tested, I believe). This entry also could be grouped with other eip7825 entries - Claude is telling me that `make fixtures-failing-list` would fix that.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/state-utils.md
```diff
@@ -612,6 +612,7 @@ The `"rlp"` key loads the block information.
       andBool TX_MAX_PRIORITY_FEE <=Int TX_MAX_FEE
       andBool BAL >=Int TX_GAS_LIMIT *Int TX_MAX_FEE +Int VALUE
       andBool TX_GAS_LIMIT <=Int BLOCK_GAS_LIMIT
+      andBool (notBool Ghastxgaslimit << SCHED >> orBool TX_GAS_LIMIT <=Int Gmaxtxgaslimit < SCHED >)
```
palinatolmach, line 615: This could be extracted into a helper looking something like this, but I think as is it's fine too:
```
    syntax Bool ::= #hasValidTxGasLimit ( Int , Schedule ) [symbol(#hasValidTxGasLimit), function]
 // ------------------------------------------------------------------------------------------------
    rule #hasValidTxGasLimit(TX_GAS_LIMIT, SCHED) => notBool Ghastxgaslimit << SCHED >> orBool TX_GAS_LIMIT <=Int Gmaxtxgaslimit < SCHED >
```

palinatolmach (commented):
One more thing: the description of `#isValidTransaction(TXID, SENDER)` in `state-utils.md` could be updated to include the new transaction gas limit checking.

PR #2835, opened 2025-12-19
eip-7918: blob base fee bounded by execution cost

anvacaru:
This PR implements [EIP-7918](https://eips.ethereum.org/EIPS/eip-7918), which introduces a reserve price mechanism for the blob base fee to ensure proper fee market functioning and fair pricing for blob verification compute costs.

- Modified Cexcessblob to accept a 4th parameter (`BASE_FEE`) for reserve price calculation, representing the base fee of the parent block. Rules have been updated to follow this logic:
```python
def calc_excess_blob_gas(parent: Header) -> int:
    target_blob_gas = GAS_PER_BLOB * blobSchedule.target
    if parent.excess_blob_gas + parent.blob_gas_used < target_blob_gas:
        return 0

    if BLOB_BASE_COST * parent.base_fee_per_gas > GAS_PER_BLOB * get_base_fee_per_blob_gas(parent):
        return parent.excess_blob_gas + parent.blob_gas_used * (blobSchedule.max - blobSchedule.target) // blobSchedule.max
    else:
        return parent.excess_blob_gas + parent.blob_gas_used - target_blob_gas
```
- A new `previousBaseFee` cell has been added to track the base fee of the parent block.


tests/execution-spec-tests/failing.llvm
```diff
@@ -584,22 +417,6 @@ blockchain_tests/osaka/eip7883_modexp_gas_increase/test_vectors_from_legacy_test
 blockchain_tests/osaka/eip7883_modexp_gas_increase/test_vectors_from_legacy_tests.json,tests/osaka/eip7883_modexp_gas_increase/test_modexp_thresholds.py::test_vectors_from_legacy_tests[fork_Osaka-blockchain_test_from_state_test-legacy-case-9]
 blockchain_tests/osaka/eip7918_blob_reserve_price/test_blob_base_fee_with_bpo_transition.json,*
```
palinatolmach, line 418: These two EIP-7918 transition tests remain skipped while the other EIP-7918 tests pass and are removed from this file. Is this because BPO/blob-schedule transitions are not supported yet? Let's merge as-is, but it'd be good to track this as a follow-up.

palinatolmach (approved):

tests/execution-spec-tests/failing.llvm
palinatolmach: Opened an issue to track this: https://github.com/runtimeverification/evm-semantics/issues/2879.

PR #2846, opened 2026-05-14
Fix `<codeAddr>` tracking for delegated calls

tothtamas28:


tothtamas28:
@anvacaru, @RaoulSchaffranek, this PR fixes `<codeAddr>` tracking for delegated calls, but breaks EIP-7702 delegation in the context of precompiles (see newly added failing tests: 0b91714e151d5a053816f98f95a3726cb41accee). Let's discuss how to properly decouple these two concerns.

anvacaru:
@tothtamas28 One possible solution could be to add an additional argument in `#mkCall`, named `ACCTEXEC` below. `ACCTCODE` would be the one used in the `#precompiled?` rule, whereas `ACCTEXEC` is set in the `<codeAddr>`. 

```
rule <k> #mkCall ACCTFROM ACCTTO ACCTCODE ACCTEXEC BYTES APPVALUE ARGS STATIC
     => … ~> #precompiled?(ACCTCODE, SCHED) ~> #execute
    …
    </k>
    …
    <codeAddr> _ => ACCTEXEC </codeAddr> 
```

Then, for the `call.delegatedAuthority` rule, you would pass both `ACCTCODE` and `DELEGATED_ACCOUNT`.
```
#callWithCode ACCTFROM ACCTTO ACCTCODE #getAccountCode(DELEGATED_ACCOUNT) VALUE APPVALUE ARGS STATIC
``` 
But, this means that for the remaining `call.*` rules, we have to duplicate the `ACCTCODE` argument when calling `#callWithCode`.

PR #2847, opened 2026-05-21
Improve performance of #computeValidJumpDests

ehildenb:
The existing implementation of `#computeValidJumpDests` was optimized for the LLVM backend a long time ago, and then mangled in various PRs since, and so doesn't really make sense. This optimizes the performance by a factor of 2X, as measured by the test which is added (and watching the time specifically of the `simplify` requests).

Claude did all the work here.

Here is the performance table Claude produced:

```
  ┌────────────────────────────────────────┬──────────┬───────────┐                                                                                                                                                                            
  │                 Phase                  │ Baseline │ Optimized │                                                                                                                                                                            
  ├────────────────────────────────────────┼──────────┼───────────┤                                                                                                                                                                            
  │ Spec parsing/digest check              │ ~1.6s    │ ~1.6s     │                                                                                                                                                                            
  ├────────────────────────────────────────┼──────────┼───────────┤                                                                                                                                                                            
  │ kore-rpc-booster server startup        │ ~7.1s    │ ~7.2s     │                                                                                                                                                                            
  ├────────────────────────────────────────┼──────────┼───────────┤                                                                                                                                                                            
  │ simplify #001 (definedness constraint) │ 21.9s    │ 10.7s     │                                                                                                                                                                            
  ├────────────────────────────────────────┼──────────┼───────────┤                                                                                                                                                                            
  │ simplify #002                          │ 1.3s     │ 1.3s      │                                                                                                                                                                            
  ├────────────────────────────────────────┼──────────┼───────────┤                                                                                                                                                                            
  │ simplify #003 (initial/target nodes)   │ 23.0s    │ 11.6s     │                                                                                                                                                                            
  ├────────────────────────────────────────┼──────────┼───────────┤                                                                                                                                                                            
  │ simplify #004                          │ 1.2s     │ 1.4s      │                                                                                                                                                                            
  ├────────────────────────────────────────┼──────────┼───────────┤                                                                                                                                                                            
  │ add-module                             │ 1.5s     │ 1.5s      │                                                                                                                                                                            
  ├────────────────────────────────────────┼──────────┼───────────┤                                                                                                                                                                            
  │ execute #001                           │ 6.3s     │ 5.6s      │                                                                                                                                                                            
  ├────────────────────────────────────────┼──────────┼───────────┤                                                                                                                                                                            
  │ implies (subsumption check)            │ 3.2s     │ 3.2s      │                                                                                                                                                                            
  ├────────────────────────────────────────┼──────────┼───────────┤                                                                                                                                                                            
  │ kevm "Proof timing"                    │ 11.16s   │ 10.35s    │                                                                                                                                                                            
  ├────────────────────────────────────────┼──────────┼───────────┤                                                                                                                                                                            
  │ Total wall time                        │ 67s      │ 44s       │                                                                                                                                                                            
  └────────────────────────────────────────┴──────────┴───────────┘  
```

I also asked Claude to take the generic things it learned about th erepo and produce a `CLAUDE.md`, and did one round of edits on that with Claude. So hopefully other times using Claude to do simple fixes will go quickly too.

### DIFF
--- a/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
+++ b/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
@@ -1747,18 +1747,13 @@ The various `CALL*` (and other inter-contract control flow) operations will be d
     rule <k> #accessAccounts ADDRSET:Set => .K ... </k>
          <accessedAccounts> TOUCHED_ACCOUNTS => TOUCHED_ACCOUNTS |Set ADDRSET </accessedAccounts>
 
-    syntax Bytes ::= #computeValidJumpDests(Bytes)             [symbol(computeValidJumpDests),    function, memo, total]
-                   | #computeValidJumpDests(Bytes, Int, Bytes) [symbol(computeValidJumpDestsAux), function             ]
- // --------------------------------------------------------------------------------------------------------------------
-    rule #computeValidJumpDests(PGM) => #computeValidJumpDests(PGM, 0, padRightBytes(.Bytes, lengthBytes(PGM), 0))
-
-    syntax Bytes ::= #computeValidJumpDestsWithinBound(Bytes, Int, Bytes) [symbol(computeValidJumpDestsWithinBound), function]
- // --------------------------------------------------------------------------------------------------------------------------
-    rule #computeValidJumpDests(PGM, I, RESULT) => RESULT requires I >=Int lengthBytes(PGM)
-    rule #computeValidJumpDests(PGM, I, RESULT) => #computeValidJumpDestsWithinBound(PGM, I, RESULT) requires I <Int lengthBytes(PGM)
-
-    rule #computeValidJumpDestsWithinBound(PGM, I, RESULT) => #computeValidJumpDests(PGM, I +Int 1, RESULT[I <- 1]) requires PGM [ I ] ==Int 91
-    rule #computeValidJumpDestsWithinBound(PGM, I, RESULT) => #computeValidJumpDests(PGM, I +Int #widthOpCode(PGM [ I ]), RESULT) requires notBool PGM [ I ] ==Int 91
+    syntax Bytes ::= #computeValidJumpDests(Bytes)                  [symbol(computeValidJumpDests),    function, memo, total]
+                   | #computeValidJumpDests(Bytes, Int, Bytes, Int) [symbol(computeValidJumpDestsAux), function             ]
+ // -------------------------------------------------------------------------------------------------------------------------
+    rule #computeValidJumpDests(PGM)                 => #computeValidJumpDests(PGM, 0, padRightBytes(.Bytes, lengthBytes(PGM), 0), lengthBytes(PGM))
+    rule #computeValidJumpDests(  _, I, RESULT, LEN) => RESULT                                                                           requires I >=Int LEN
+    rule #computeValidJumpDests(PGM, I, RESULT, LEN) => #computeValidJumpDests(PGM, I +Int 1,                        RESULT[I <- 1], LEN) requires I <Int LEN andBool PGM [ I ] ==Int 91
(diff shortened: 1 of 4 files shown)


ehildenb:
My initial prompt:

 > I want to optimize computeValidJumpDetsts. Right now, it's broken into two mutually recursive functions, which is not necessary. It just comes from historical usage. Also, we are recomputing lengthBytes every iteration of checking a jumpdest, rather than once and storing it. I think we can compute it once for the top-level #computeValidJumpDests rule, and store it as an argument to the tail-recursive variant for comparison as we go. We can also avoid the two mutually recursive functions too. We should make a test of this, which exercises the haskell backend on a large byte-array with some symbolic values (so it doesn't just defer it to the LLVM backend). To do that, you can make a runLemma/doneLemma style claim with #computeValidJumpDests(10k zeros +Bytes 91 +Bytes 10k zeros +Bytes #buf(32, INT_VAR)) as the LHS, and check that it takes a while to execute this claim. If so, then adjust the definition of computevalidjumpdests and check that it goes down in execution time. If the initial execution time is longer than 2 minutes, shorten the 10k byte arrays. Can you make a comprehensive investigation and plan?

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
```diff
@@ -1747,18 +1747,13 @@ The various `CALL*` (and other inter-contract control flow) operations will be d
...
+    syntax Bytes ::= #computeValidJumpDests(Bytes)                  [symbol(computeValidJumpDests),    function, memo, total]
+                   | #computeValidJumpDests(Bytes, Int, Bytes, Int) [symbol(computeValidJumpDestsAux), function             ]
+ // -------------------------------------------------------------------------------------------------------------------------
+    rule #computeValidJumpDests(PGM)                 => #computeValidJumpDests(PGM, 0, padRightBytes(.Bytes, lengthBytes(PGM), 0), lengthBytes(PGM))
+    rule #computeValidJumpDests(  _, I, RESULT, LEN) => RESULT                                                                           requires I >=Int LEN
+    rule #computeValidJumpDests(PGM, I, RESULT, LEN) => #computeValidJumpDests(PGM, I +Int 1,                        RESULT[I <- 1], LEN) requires I <Int LEN andBool PGM [ I ] ==Int 91
+    rule #computeValidJumpDests(PGM, I, RESULT, LEN) => #computeValidJumpDests(PGM, I +Int #widthOpCode(PGM [ I ]), RESULT,          LEN) requires I <Int LEN andBool notBool PGM [ I ] ==Int 91
```
jberthold, line 1756: Nice, just what I thought...
While at it, maybe it would be worth trying the Booster's evaluator performance in comparison?
Currently none of these rules is applied because the `...Aux` function is partial.
We cannot really add a `total` attribute to the `...Aux` function... (it assumes `I < lengthBytes(PGM)`) but all rules could be marked `preserves-definedness` and booster would start applying them.
(and [`widthOpCode` can be marked as a total function](https://github.com/runtimeverification/evm-semantics/blame/abedccff6393846f8243ef41b362809c9ceb8b0d/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md#L1815-L1818) BTW).

jberthold (approved):
LGTM

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
ehildenb: I tried iterating on the other changes related to marking this as preserving definedness or marking functions as total (such as `widthOpCode`), but they didn't result in immediate wins because other things get in the way.

It did lead me to launching into a large-scale experiment of tryign to get Claude to stop us from falling back to Kore, and that is proving successful. So there will be more follow-up PRs to this one.

PR #2848, opened 2026-05-25
lemmas: add bool2Word simplification for symbolic Bool expressions

Stevengre:
Add two `[simplification]` lemmas in `lemmas/lemmas.k` so that
`bool2Word(B)` reduces when the path condition implies `B` but does not
syntactically equate it. The function rules in `evm-types.md` are
unchanged — concrete inputs still take the original zero-overhead path.

```
rule [b2w-true]:  bool2Word(B) => 1 requires         B [simplification]
rule [b2w-false]: bool2Word(B) => 0 requires notBool B [simplification]
```

In Kore, the existing function rules pattern-match on `true`/`false` in
the argument position, which only runs the simplifier (no SMT against
the path condition). The new simplification lemmas bind `B`
unconditionally and discharge the side condition with the full path
condition, so a goal like `bool2Word(X >=Int 0) => 1` under path
condition `X >Int 0` now reduces.

### Tests

`tests/specs/functional/bool2Word-symbolic-spec.k` pins this with 6
claims:

| # | argument | path condition | without lemmas | with lemmas |
|---|---|---|---|---|
| 1 | `B` | `B` | pass | pass |
| 2 | `B` | `notBool B` | pass | pass |
| 3 | `X >Int Y` | `X >Int Y` | pass | pass |
| 4 | `X >Int Y` | `notBool (X >Int Y)` | pass | pass |
| 5 | `X >=Int 0` | `X >Int 0` | **stuck** | pass |
| 6 | `X <Int 0` | `X >Int 0` | **stuck** | pass |

## History

The original commits on this branch modified the `bool2Word` function
rules in `evm-types.md` directly. Per @ehildenb's review feedback,
switched to the `[simplification]` lemma approach — equivalent on this
spec, less invasive, opt-in via `imports LEMMAS`.

## Test plan

- [ ] `make test-prove-functional` (in particular `lemmas-spec.k`
  regressions — many existing `bool2Word` lemmas live next to the new
  ones)
- [ ] `make test-prove-rules` / `test-prove-optimizations`
- [ ] Conformance smoke

Draft pending full suite run.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm-types.md
+++ b/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm-types.md
@@ -29,8 +29,8 @@ Primitives provide the basic conversion from K's sorts `Int` and `Bool` to EVM's
 ```k
     syntax Int ::= bool2Word ( Bool ) [symbol(bool2Word), function, total, injective, smtlib(bool2Word)]
  // ----------------------------------------------------------------------------------------------------
-    rule bool2Word( true  ) => 1
-    rule bool2Word( false ) => 0
+    rule bool2Word( B:Bool ) => 1 requires         B ==Bool true
+    rule bool2Word( B:Bool ) => 0 requires notBool B ==Bool true
 
     syntax Bool ::= word2Bool ( Int ) [symbol(word2Bool), function, total]
  // ----------------------------------------------------------------------
--- a/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm-types.md
+++ b/kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm-types.md
@@ -29,8 +29,8 @@ Primitives provide the basic conversion from K's sorts `Int` and `Bool` to EVM's
 ```k
     syntax Int ::= bool2Word ( Bool ) [symbol(bool2Word), function, total, injective, smtlib(bool2Word)]
  // ----------------------------------------------------------------------------------------------------
-    rule bool2Word( B:Bool ) => 1 requires         B ==Bool true
-    rule bool2Word( B:Bool ) => 0 requires notBool B ==Bool true
+    rule bool2Word( true  ) => 1
+    rule bool2Word( false ) => 0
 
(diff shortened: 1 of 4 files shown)


kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm-types.md
```diff
@@ -29,8 +29,8 @@ Primitives provide the basic conversion from K's sorts `Int` and `Bool` to EVM's
 ```k
     syntax Int ::= bool2Word ( Bool ) [symbol(bool2Word), function, total, injective, smtlib(bool2Word)]
  // ----------------------------------------------------------------------------------------------------
-    rule bool2Word( true  ) => 1
-    rule bool2Word( false ) => 0
+    rule bool2Word( B:Bool ) => 1 requires         B ==Bool true
+    rule bool2Word( B:Bool ) => 0 requires notBool B ==Bool true
```
ehildenb:
```suggestion
    rule bool2Word( B:Bool ) => 1 requires         B
    rule bool2Word( B:Bool ) => 0 requires notBool B
```

Does it work like this instead?

ehildenb: And you may consider, instead, just writing _additional_ simplification lemmas in `lemmas.k`, and see if that works:

```
rule bool2Word(B) => 1 requires         B [simplification]
rule bool2Word(B) => 0 requires notBool B [simplification]
```

ehildenb: That way we get hte better booster performance, and we keep LLVM backend fast path.

Stevengre: Switched to this approach in 6757536 — simplification lemmas in lemmas/lemmas.k, evm-types.md untouched. Verified all 6 test claims still pass on the new path (legacy form: 2 stuck; either approach: 6 trivial). Timed both approaches at ~2.5s on the micro-spec — indistinguishable within noise.

ehildenb (approved):

PR #2849, opened 2026-05-25
lemmas: booster-only gap fixes — widthOpCode, int/bytes simplifications, preserves-definedness

ehildenb:
The booster backend (booster-only-simplify mode) fails to discharge several side conditions and lemmas that kore handles correctly, because booster's equation engine has stricter matching requirements.  This PR closes the gaps found while running the functional test-suite in booster-only mode. Note that I collected the information needed (or Claude did, more precisely) using functionality that's not merged yet (https://github.com/runtimeverification/k/pull/4919), but that the improvements here are independent of that.

**Changes:**
- Updates `CLAUDE.md` with some basic information about how to debug stuck proofs, mostly focused on getting logs out of the haskell backend for analysis.
- `evm.md`: add `total` to `#widthOpCode` declaration; add `preserves-definedness` to all four `#computeValidJumpDests` rules so booster's definedness checker passes
- `lemmas.k`: add `smtlib(widthOpCode)` annotation + `smt-lemma` bounds for `#widthOpCode` (0 ≤ x ≤ 33); add `widthOpCode-concat-{left,right}` rules for booster's single-step equation engine; add `Bool` tautology rules; add `KSet` membership/dedup rules with comments explaining which are permanently indeterminate in booster (backend limitation, not fixable from K source)
- `int-simplification.k`: add `minInt`/`maxInt` comparison rules, commutativity normalisation (`==Int` concrete-on-left), arithmetic and `Bool` simplifications
- `evm-int-simplification.k`: add `bit-mask-asWord-m` and `bit-notmask-asWord-nm`
- `bytes-simplification.k`: add `preserves-definedness` to `lookup-as-asWord`; add concrete `b""` concat identity rules
- `compute-valid-jump-dests-spec.k`: add a fast `N=5` functional claim alongside the existing `N=1000` performance claim for quicker CI feedback
- `set-dedup-spec.k`: new functional spec exercising the `KSet` membership and dedup rules under kore. This one is still not passing with `--booster-only-simplify`.

**Validation:**
- Booster-only functional suite: 8/10 specs pass after these fixes (avg ~2× wall-clock speedup with `--booster-only-simplify` flag vs kore baseline). The two remaining failures (`lemmas-spec` with many claims, `merkle-spec`) are engine-level limitations unrelated to these lemmas.
- compute-valid-jump-dests specifically: 1.23× wall-clock speedup in booster-only mode on the N=1000 symbolic claim; with `--use-booster-dev` (pure booster, no kore proxy) the same proof drops from **~200 s to 0.68s**. The kore fallback is the bottleneck, not the semantics.
- int-simplifications-spec improves 3.48× in booster-only mode after the `minInt`/`maxInt` and commutativity normalisation rules.
- `widthOpCode` smt-lemma bounds let Z3 discharge #widthOpCode-containing side conditions without concrete evaluation, which was previously causing fallbacks.
- set-dedup-spec is intentionally kore-only: it documents the booster matchSets() limitation (booster cannot match non-empty KSet patterns — Match.hs:673-676) so the permanently-dead rules in lemmas.k have a test-level explanation.


kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
```diff
@@ -267,8 +271,8 @@ module BYTES-SIMPLIFICATION [symbolic]
     rule [lengthBytes-geq-zero]:   0 <=Int lengthBytes ( _ )                     => true                                     [simplification, smt-lemma]
     rule [lengthBytes-geq-nonPos]: X <=Int lengthBytes ( _ )                     => true requires X <=Int 0                  [simplification, concrete(X)]
     rule [lengthBytes-concat]:     lengthBytes(BUF1 +Bytes BUF2)                 => lengthBytes(BUF1) +Int lengthBytes(BUF2) [simplification]
-    rule [lengthBytes-buf]:        lengthBytes(#buf(S, _))                       => S requires 0 <=Int S                     [simplification]
-    rule [lengthBytes-range]:      lengthBytes(#range(_, S:Int, W:Int))          => maxInt(0, W) requires 0 <=Int S          [simplification]
+    rule [lengthBytes-buf]:        lengthBytes(#buf(S, _))                       => S requires 0 <=Int S                     [simplification, preserves-definedness]
+    rule [lengthBytes-range]:      lengthBytes(#range(_, S:Int, W:Int))          => maxInt(0, W) requires 0 <=Int S          [simplification, preserves-definedness]
```
ehildenb, line 275: I'm not sure if these rules actually do preserve definedness.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
```diff
@@ -101,6 +101,25 @@ module EVM-INT-SIMPLIFICATION
...
+  // General reduction: #asWord of a range extracted from a buf.
+  // #buf(N, X) is the big-endian N-byte encoding of X.  Taking W bytes
+  // starting at byte S gives (X >> (8*(N-S-W))) & (2^(8*W)-1).
+  // Activated with concrete N, S, W so Booster can evaluate the arithmetic.
+  rule [asWord-range-buf]:
+    #asWord ( #range ( #buf ( N:Int, X:Int ), S:Int, W:Int ) ) =>
+      X /Int ( 2 ^Int ( 8 *Int ( N -Int S -Int W ) ) ) modInt ( 2 ^Int ( 8 *Int W ) )
+    requires 0 <=Int S andBool 0 <Int W andBool S +Int W <=Int N
+     andBool 0 <=Int X andBool X <Int 2 ^Int ( 8 *Int N )
+    [simplification, concrete(N, S, W), preserves-definedness]
```
ehildenb, line 121: This lemma seems unnecessarily complicated, we should track down which test triggered adding it, and see if we can find a smaller/simpler version that works.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/int-simplification.k
```diff
@@ -232,6 +242,35 @@ module INT-SIMPLIFICATION-COMMON
...
+    // the non-linear shift rules above no longer match; these linear rules catch the result.
+    rule [minint-lt-maxint-shift-a-factored]:
+        E:Int +Int minInt(A:Int, _B:Int) <Int F:Int +Int maxInt(C:Int, _D:Int) => true
+        requires A <Int C andBool E <=Int F
+        [simplification(40), preserves-definedness]
+    rule [minint-leq-maxint-shift-b-factored]:
+        minInt(_A:Int, B:Int) +Int E:Int <=Int maxInt(_C:Int, D:Int) +Int F:Int => true
+        requires B <=Int D andBool E <=Int F
+        [simplification(40), preserves-definedness]
+
```
ehildenb, line 273: Some of these seem like they can be simplified with some other reasoning steps in the way. For eaxmple some of the rules could be combined with a disjunction in the requires clause, and others can first do arithmetic simplification on the expression then rely on the other rules here to fire afterwards.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/lemmas.k
```diff
@@ -36,12 +36,29 @@ module LEMMAS-WITHOUT-SLOT-UPDATES [symbolic]
...
+    rule X:Int in (REST:Set |Set SetItem(Y:Int)) => X in REST requires notBool ( X ==Int Y ) [simplification]
+
+    // Deduplication of repeated concrete Int elements in a symbolic |Set expression.
+    // Required in normal mode (where kore's simplify step fires it) but INDETERMINATE in booster.
+    // Booster cannot match SetItem(X) in the pattern: matchSets() is unimplemented for non-empty
+    // KSets (booster/Match.hs:673-676). Cannot be fixed from K source; requires backend change.
+    // Context: #touchAccounts in evm.md adds addresses via |Set; when the same address is added
+    // twice to a symbolic touchedAccounts, this rule is needed to normalize back to a 2-element form.
+    rule S:Set |Set SetItem(X:Int) => S requires X in S [simplification]
+
```
ehildenb, line 61: @jberthold how much work would it be to support these types of simplification rules in the booster backend directly? We don't need to support sets in general, but just these expressions.

The attached `set-dedup-spec.k` file should capture claims that need to pass with the new rules.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/lemmas.k
```diff
@@ -255,4 +302,9 @@ module LEMMAS-HASKELL [symbolic]
     rule [notBool-or]:  notBool ( A  orBool B ) => ( notBool A ) andBool ( notBool B ) [simplification(60)]
     rule [notBool-and]: notBool ( A andBool B ) => ( notBool A )  orBool ( notBool B ) [simplification(60)]
 
+    rule [bool-or-not-self]:      B:Bool  orBool notBool B:Bool => true  [simplification]
+    rule [bool-not-or-self]:  notBool B:Bool  orBool B:Bool     => true  [simplification]
+    rule [bool-and-not-self]:     B:Bool andBool notBool B:Bool => false [simplification]
+    rule [bool-not-and-self]: notBool B:Bool andBool B:Bool     => false [simplification]
```
ehildenb, line 306: Formatting needs fixed.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
```diff
@@ -1750,10 +1750,10 @@ The various `CALL*` (and other inter-contract control flow) operations will be d
...
                    | #computeValidJumpDests(Bytes, Int, Bytes, Int) [symbol(computeValidJumpDestsAux), function             ]
  // -------------------------------------------------------------------------------------------------------------------------
-    rule #computeValidJumpDests(PGM)                 => #computeValidJumpDests(PGM, 0, padRightBytes(.Bytes, lengthBytes(PGM), 0), lengthBytes(PGM))
-    rule #computeValidJumpDests(  _, I, RESULT, LEN) => RESULT                                                                           requires I >=Int LEN
-    rule #computeValidJumpDests(PGM, I, RESULT, LEN) => #computeValidJumpDests(PGM, I +Int 1,                        RESULT[I <- 1], LEN) requires I <Int LEN andBool PGM [ I ] ==Int 91
-    rule #computeValidJumpDests(PGM, I, RESULT, LEN) => #computeValidJumpDests(PGM, I +Int #widthOpCode(PGM [ I ]), RESULT,          LEN) requires I <Int LEN andBool notBool PGM [ I ] ==Int 91
+    rule #computeValidJumpDests(PGM)                 => #computeValidJumpDests(PGM, 0, padRightBytes(.Bytes, lengthBytes(PGM), 0), lengthBytes(PGM))                                                                             [preserves-definedness]
```
ehildenb, line 1753: This function is already marked as total.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
```diff
@@ -1750,10 +1750,10 @@ The various `CALL*` (and other inter-contract control flow) operations will be d
...
+    rule #computeValidJumpDests(  _, I, RESULT, LEN) => RESULT                                                                           requires I >=Int LEN                                                                   [preserves-definedness]
+    rule #computeValidJumpDests(PGM, I, RESULT, LEN) => #computeValidJumpDests(PGM, I +Int 1,                        RESULT[I <- 1], LEN) requires I <Int LEN andBool PGM [ I ] ==Int 91                                      [preserves-definedness]
+    rule #computeValidJumpDests(PGM, I, RESULT, LEN) => #computeValidJumpDests(PGM, I +Int #widthOpCode(PGM [ I ]), RESULT,          LEN) requires I <Int LEN andBool notBool PGM [ I ] ==Int 91                              [preserves-definedness]
```
ehildenb, line 1756: Is it true that these are preserving definedness?

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
```diff
@@ -1812,7 +1812,7 @@ System Transaction Configuration
 ```
 
 ```k
-    syntax Int ::= #widthOpCode(Int) [symbol(#widthOpCode), function]
+    syntax Int ::= #widthOpCode(Int) [symbol(#widthOpCode), function, total, smtlib(widthOpCode)]
  // -----------------------------------------------------------------
```
ehildenb, line 1816: formatting

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/lemmas.k
jberthold: Related code is here: https://github.com/runtimeverification/haskell-backend/blob/00479fbdb86b70fc602aefca9fabab5afc8bad5f/booster/library/Booster/Pattern/Match.hs#L659-L676
At the moment, only empty sets are matched. It looks like we could consider matching singleton sets like the ones above, but there is the danger of discarding a match prematurely. For the above rules, it should be OK because there is always an explicit `|Set` operation separating the sets and they are singletons where it matters... :thinking:

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
jberthold: I think this rule should still be marked `preserves-definedness` ... the RHS `#computeValidJumpDests(_, _, _, _)` is not total but should be defined for these arguments.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/evm.md
jberthold: I guess not, but it is pragmatically the only choice to avoid having `lengthBytes(PGM)` again here.
This is only OK if the length of `PGM` and `RESULT` is `> I`  but at the call sites they are the same length `LEN`.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/bytes-simplification.k
jberthold: `#buf` and `#range` are total so I don't see a problem with these.

ehildenb:
I'm going to re-open this is chunks, where I can more concretely quantify how each chunk helps.

PR #2545, opened 2024-07-31
EIP-6780: SELFDESTRUCT only in same transaction

anvacaru:
- adding a new `<createdAccounts>` cell in the `<substate>` to keep track of the accounts created in the current transaction.
-  modified the python test harness to support a handful of tests that run on `chainId` 0. 
- implemented eip-6780 to remove the current account in `SELFDESTRUCT` only if it was created in the same tx ([ EIP-6780](https://eips.ethereum.org/EIPS/eip-6780) )
- enabling ~3000 tests from `failing.llvm`.
- updating `.k` proof files with the new cell.

PR #2854, opened 2026-06-01
Haskell-backend comparison tooling, proof-running docs, and booster-dev test cleanup

ehildenb:
Adds documentation and tooling for comparing two haskell-backend builds on the proof suite — both proof performance and booster-dev (pure-booster, no kore fallback) pass rate — while holding K fixed at `deps/k_release`. Also lands the test-list and repo-hygiene fixes surfaced while using it.

**Changes:**
- `docs/kup-override.md`: install K with a `kup --override` of the haskell-backend input — local checkout or a remote bare git ref (branch/tag/commit) — to test a backend build without editing the project flake.
- `scripts/compare-junit-runs.py` + CLAUDE.md "Comparing haskell-backend builds": tooling and a documented procedure to compare two backends' proof runs — per-test and total wall-clock timing, plus the booster-dev fail↔pass diff (JUnit in, aligned-markdown + JSON out).
- `kevm-pyk/pyproject.toml`: add `pytest-timeout` dev dependency (the documented booster-dev run uses `--timeout` to bound stuck specs).
- `tests/failing-symbolic.haskell-booster-dev`: unskip 4 specs that prove under pure booster-dev on stock K v7.1.329 (2 functional, 2 kontrol).
- `.gitignore` + untrack: ignore generated `<spec>-spec.json` claim-parse caches; remove `tests/specs/mcd-structured/vat-spec.json`, accidentally committed in #2545.

**Validation:**
- `scripts/compare-junit-runs.py` passes the project's flake8/mypy/black config and was exercised end-to-end on full baseline-vs-override suite runs.
- The 4 unskipped specs were confirmed passing in a full `--use-booster-dev` run on stock K v7.1.329 (skip-list neutralised so every spec actually ran).
- The new gitignore rule matches all 154 generated caches; `git ls-files 'tests/specs/**/*.json'` returns 0 after untracking.
- All cross-references in CLAUDE.md and `docs/kup-override.md` resolve — no dangling doc/tooling references. No semantics or harness logic changed: additive docs/tooling, one dev-dependency, and two data-file edits.


docs/kup-override.md
```diff
@@ -0,0 +1,151 @@
+# Testing a Local Haskell Backend with kup
+
+`kup` can install K (or any managed package) with one of its Nix flake inputs
```
anvacaru: Do we want to commit to a line-wrap format for the entire project? If so, it should be stated in the `AGENTS.md`/`CLAUDE.md`.

anvacaru (approved):

docs/kup-override.md
ehildenb: I made it consistent with the one-sentence-per-line style guide, and also prefixed the documenattion with the date.

PR #2859, opened 2026-06-04
Simplification lemmas to improve Haskell booster coverage

ehildenb:
This PR adds a curated first tier of simplification lemmas, plus the supporting attribute annotations, so that common arithmetic, Bool, bytes, and `#asWord` goals discharge directly in the Haskell booster instead of falling back to kore-rpc. Each new rule is paired with a focused coverage claim in the functional specs, and the change set includes the SMT support needed to keep the MCD/DSS proofs passing once the new lemmas simplify their keccak side-conditions away.
  
**Changes:**
- Add simplification lemmas across `int-simplification.k`, `evm-int-simplification.k`, `bytes-simplification.k`, `lemmas.k`, and `buf.md`: `==Int` commutativity normalization (concrete argument on the left), arithmetic and Bool tautologies, `A <=Int A +Int B` short-circuit when `0 <=Int B`, concrete bytes-concat identities, `#asWord` comparison-to-false cases, `minInt`/`maxInt` base comparisons, `eq-false-lt`, `#widthOpCode` bounds, and a concrete `#powByteLen` comparison rule.
- Mark `#widthOpCode` `[total, smtlib]` and `#asAccount` `[total]`; add `[preserves-definedness]` to the rules that preserve definedness.
- `tests/specs/mcd/verification.k`: mark the keccak disequality rule `[smt-lemma]` so Z3 can discharge `keccak(_) =/=Int <small constant>` independently of the path condition. This is required once the new lemmas simplify that constraint away, and fixes the `heal`/`fold`/`frob` DSS proofs and the `flopper-tick`/`end-pack`/`flopper-dent` rules proofs.
- Add coverage claims for the new rules in `lemmas-no-smt-spec.k`, `int-simplifications-spec.k`, and `abi-spec.k`.

**Validation:**
- Full prove suite (functional, rules, optimizations, DSS) on K v7.1.330.
- Per-spec timing comparison against master: the lemma set is performance-neutral on the large majority of specs; the heaviest MCD/DSS proofs see roughly +20–25% from the cumulative per-step cost of the larger rule set, and the keccak `[smt-lemma]` itself adds ~1%.
- Confirmed that all slowdown disappers when coupled with this Haskell backend PR: https://github.com/runtimeverification/haskell-backend/pull/4153
- Confirmed that this PR reduces fallbacks to Kore dramatically on the KEVM test-suite (and on Kontrol).


kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
```diff
@@ -123,6 +131,18 @@ module EVM-INT-SIMPLIFICATION
...
+  // Loop-safe with asWord-le-false: requires clauses have #asWord on the RIGHT, so
+  // neither rule fires on the other's requires.
+  rule [asWord-lt-false]: #asWord ( B:Bytes )  <Int X:Int => false requires X <=Int #asWord ( B ) [simplification]
+  rule [asWord-le-false]: #asWord ( B:Bytes ) <=Int X:Int => false requires X  <Int #asWord ( B ) [simplification]
+  // Comparison: `#asWord(B) == X` is false when the path condition says X =/= #asWord(B).
+  // The requires X =/=Int #asWord(B) expands to notBool(X ==Int #asWord(B)) via the K builtin;
+  // since ==Int is comm, this matches the path condition fact notBool(#asWord(B) ==Int X).
+  // No loop: our rule matches #asWord on the LEFT; after builtin expansion the recursive
+  // term X ==Int #asWord(B) has X on the LEFT, so this rule does not fire on it.
+  rule [asWord-eq-false]: #asWord ( B:Bytes ) ==Int X:Int => false requires X =/=Int #asWord ( B ) [simplification]
```
ehildenb, line 144: These lemmas seem like they could be made more generic, just with `Int` variables on each side of the binary comparson operator instead. The last one may result in a loop because `=/=Int` may simplify down to `==Int`.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/lemmas.k
```diff
@@ -245,6 +245,30 @@ module LEMMAS-HASKELL [symbolic]
...
+    // steps so the full concrete-prefix lookup can be reduced in one go.
+    rule [widthOpCode-concat-left]:
+      #widthOpCode ( ( B1:Bytes +Bytes _B2:Bytes ) [ I:Int ] ) => #widthOpCode ( B1 [ I ] )
+      requires 0 <=Int I andBool I <Int lengthBytes ( B1 )
+      [simplification, preserves-definedness]
+
+    rule [widthOpCode-concat-right]:
+      #widthOpCode ( ( B1:Bytes +Bytes B2:Bytes ) [ I:Int ] ) => #widthOpCode ( B2 [ I -Int lengthBytes ( B1 ) ] )
+      requires lengthBytes ( B1 ) <=Int I
+      [simplification, preserves-definedness]
```
ehildenb: These seem like they should already have the relevant `Bytes` simplification lemmas applying, or could just be written as bytes-simplification lemmas.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/buf.md
```diff
@@ -37,10 +37,17 @@ module BUF
...
 
     rule 0    <Int #powByteLen(SIZE) => true requires 0 <=Int SIZE [simplification, preserves-definedness]
     rule SIZE <Int #powByteLen(SIZE) => true requires 0 <=Int SIZE [simplification, preserves-definedness]
+    // Concrete comparison: CONST < #powByteLen(N) evaluates when both arguments are concrete.
+    // Needed because 2^Int(8*N) is rewritten to #powByteLen(N) even for concrete N (via the
+    // symbolic(N) rule firing before path-condition lookup makes N concrete), but #powByteLen
+    // has no-evaluators so the result can't be computed directly.
+    rule [powByteLen-lt-concrete]: CONST <Int #powByteLen(N) => true
+        requires 0 <=Int N andBool CONST <Int 2 ^Int (8 *Int N)
+        [simplification, concrete(CONST, N), preserves-definedness]
```
ehildenb, line 50: Dose this generalize the previous lemma? I guess not since the previous one applies for all `SIZE`, not just `concrete` `SIZE`. Also, I think this lemma could be non-concrete in `N` and it should be OK, if `N` is concrete we simplify down to basic linear arithmetic inequalities, and so that shouldn't cause any solver issues.

ehildenb: Right, it doesn't generalize the diagonal SIZE <Int #powByteLen(SIZE) (same var both sides). Relaxed to concrete(CONST) — once N is concrete the requires is a linear inequality; symbolic N just doesn't fire. 6cd5f0deb

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/lemmas.k
ehildenb: Agreed — bytes-concat-lookup-{left,right} already exist, so these duplicated them as a #widthOpCode-descent workaround. Removed; better fixed in the backend if descent is still an issue. 761f1f30d

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
ehildenb: The naive generalization loops the same way =/= does: generic lt/le rewrite each other's <=/< requires. The #asWord LHS-vs-requires asymmetry is what keeps these loop-safe, so I kept them specific. Happy to add a loop-safe (e.g. concrete-guarded) generic variant if you'd prefer.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
```diff
@@ -101,6 +101,14 @@ module EVM-INT-SIMPLIFICATION
...
     [simplification, concrete(WB), preserves-definedness]
 
+  // Specialisation for 32-byte buf with rangeUInt(256) side-condition: avoids
+  // the minInt(2^256, pow256) evaluation that Booster cannot match against the
+  // path condition's 2^Int 256 tree form.
+  rule [asWord-buf-inversion-rangeUInt256]:
+    #asWord ( #buf ( 32, X:Int ) ) => X
+    requires #rangeUInt ( 256, X )
+    [simplification]
+
```
ehildenb, line 111: Doesn't this end up being a less general rule than the `asWord-buf-inversion` above?

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
ehildenb: Correction to my earlier reply: the `concrete(CONST)`-only relaxation (6cd5f0deb) does cause a solver/rewrite loop, so I reverted it in 293a1a469 — the rule is back to `concrete(CONST, N)`. With `N` symbolic the rule *does* fire (the match is just `CONST <Int #powByteLen(N)`), and evaluating its `requires CONST <Int 2 ^Int (8 *Int N)` re-introduces `2 ^Int (8 *Int N)`, which another simplification rewrites straight back to `#powByteLen(N)` → non-termination. Found by bisecting a real regression: `benchmarks/bytes00-spec` went from ~64s to non-terminating, first failing exactly at 6cd5f0deb. So both arguments stay concrete; supporting symbolic `N` would need a formulation that doesn't round-trip through `2^(8*N)`.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/buf.md
ehildenb: Confirmed empirically that the loose form loops, so keeping these `#asWord`-specific. The termination argument is the LHS-vs-`requires` asymmetry: these match `#asWord(B)` on the left of the comparison while their `requires` puts `#asWord(B)` on the right, so neither `-lt-false`/`-le-false` fires on the other's side condition. A fully generic `A <Int B => false requires B <=Int A` (plus its `<=` twin) rewrites inside each other's `requires` indefinitely, and `==Int`/`=/=Int` collapse the same way — that's the loop you flagged. I'd rather not add a generic concrete-guarded variant here: it adds loop surface without buying coverage we need today. Leaving as-is.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/lemmas/evm-int-simplification.k
ehildenb: Yes, deliberately — it's a Booster-coverage specialization of `asWord-buf-inversion` for the 32-byte case, not a replacement. The general rule's side condition is `X <Int minInt(2 ^Int (8 *Int WB), pow256)`; at `WB = 32` that's `minInt(2 ^Int 256, pow256)`, which Booster can't match/evaluate against the path condition (whose bound is carried as a `2 ^Int 256` tree), so the general rule silently fails to fire there. The specialization restates the bound as `#rangeUInt(256, X)`, which matches the path condition directly, so the inversion actually fires in the common case. Semantically redundant with the general rule, but not operationally in Booster. Can be dropped once the backend matches `minInt(2^256, pow256)` against `2^Int 256`.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/buf.md
ehildenb: Correction to my earlier reply: the `concrete(CONST)`-only relaxation (6cd5f0deb) does cause a solver/rewrite loop, so I reverted it in 293a1a469 — the rule is back to `concrete(CONST, N)`. With `N` symbolic the rule *does* fire (the match is just `CONST <Int #powByteLen(N)`), and evaluating its `requires CONST <Int 2 ^Int (8 *Int N)` re-introduces `2 ^Int (8 *Int N)`, which another simplification rewrites straight back to `#powByteLen(N)` → non-termination. Found by bisecting a real regression: `benchmarks/bytes00-spec` went from ~64s to non-terminating, first failing exactly at 6cd5f0deb. So both arguments stay concrete; supporting symbolic `N` would need a formulation that doesn't round-trip through `2^(8*N)`.

ehildenb:
## (i) Performance changes

Caveat: no clean perf benchmark was run. Numbers below are **recover-mode wall-clock under concurrent np (≈2× RPC, no caching, cross-run contention)** — directionally fine, not authoritative. The dedicated isolated normal-mode methodology (CLAUDE.md "Comparing haskell-backend builds") is the right tool for real perf and is a recommended follow-up.

| scope | master | tier1 | delta |
|---|---|---|---|
| functional suite wall-clock | 1583 s | 1601 s | +1% |
| rules suite wall-clock | 3255 s | 4019 s | +23%* |
| rules, sum over 124 both-passed specs | 11949 s | 12319 s | **+3% (noise)** |
| dss (vat-spec, capped) | 10800 s (timeout) | 10800 s (timeout) | — |

*The +23% rules wall-clock is **not** a slowdown: tier1 *completes 14 more proofs* that master fails/crashes out of quickly. The fair per-spec view (both-passed specs) is **+3%, i.e. flat within contention noise**. Per-spec movers are mixed and small (largest: `mcd/flopper-dent-guy-same` +105 s, `ecrecoverloop02-sigs-valid` −56 s) — consistent with noise, not a systematic regression.

**Takeaway:** no evidence of a performance regression; a proper isolated perf run is needed to claim any speedup.

ehildenb:
## (ii) Fallback changes (the headline — clear reduction)

| suite | metric | master | tier1 | delta |
|---|---|---|---|---|
| functional | total handoffs (488 common claims) | 21 | **11** | −48% |
| functional | flavour | implies 21 | implies 11 | implies-only |
| rules | total handoffs (124 both-passed specs) | 73 | **58** | −21% |
| rules | execute / implies | 30 / 43 | **16** / 42 | execute ~halved |
| rules | all 264 common claims | 105 | 81 | −23% |
| dss (vat-spec, partial) | total handoffs (36 common claims) | 120 | **33** | −72% |
| dss | execute / implies | 109 / 11 | **21** / 12 | execute ~−80% |

**The dominant, consistent effect is a large reduction in `execute`-flavour fallback** (the booster aborting a symbolic-execution rewrite and handing the step to Kore): ~½ on rules, ~⅕ on DSS. `implies`-flavour fallback is flat on rules/dss and is the entire (smaller) functional effect. Reductions concentrate where tier1 added lemmas: the `ecrecover`/`ecrecoverloop` benchmarks, MCD `flapper`/`flopper`/`end`/`pot` proofs, and the DSS vat proofs (helped further by tier1's keccak-disequality `[smt-lemma]` fix).

Even stronger than the counts — **recover-mode provability** (does the proof close booster-first at all):

| suite | master pass/fail | tier1 pass/fail |
|---|---|---|
| rules | 124 / 18 | **138 / 4** |

**14 rules specs that master can only close *with* Kore now close booster-side on tier1; 0 regressions.** (The 4 fail-on-both are recover-mode limitations — see anomalies.)

ehildenb:
## (iii) Booster-dev changes (skip-list)

Verified by running the standalone `booster-dev` binary on both branches over the full 137-entry `failing-symbolic.haskell-booster-dev` list (neutralised, then restored). **Methodological note: recover-mode "0 handoffs" is *not* a proxy for booster-dev provability** — of 82 specs that looked booster-only in recover-mode, only 5 actually pass `booster-dev`. The verification was essential.

**Removed from the skip-list on this branch (5; all pass `booster-dev` on tier1):**

| spec | master booster-dev | tier1 booster-dev | attribution |
|---|---|---|---|
| `mcd-structured/dsvalue-peek-pass-rough-spec.k` | fail | **pass** | **tier1-specific** |
| `functional/bitwise-asword-spec.k` | pass | pass | stale |
| `functional/bitwise-simplifications-spec.k` | pass | pass | stale |
| `mcd/dsvalue-read-pass-summarize-spec.k` | pass | pass | stale |
| `mcd-structured/dsvalue-read-pass-summarize-spec.k` | pass | pass | stale |

1 removal is tier1's doing; 4 are stale entries the list had drifted on (already passing on master). Skip-list: 137 → 132 lines.

ehildenb:
## (iv) Anomalies / non-passing proofs

| # | anomaly | detail |
|---|---|---|
| 1 | **booster-dev regression** | `functional/bitwise-mask-shift-spec.k` passes `booster-dev` on master, **fails on tier1** (stable recheck, 1 *failing* node — not a timeout). Still passes in normal & recover mode; skip-listed so no CI break. **Kept on the list.** A tier1 lemma edit weakened the standalone-booster path for this claim — worth the author's look. |
| 2 | recover-mode fail-on-both | 4 specs fail under recover-mode on *both* branches — `examples/sum-to-n-foundry`, `kontrol/test-{allowchangestest-testfailallowchangestostorage, countertest-testincrement, owneruponlytest-testincrementasowner}-0`. These are recover-mode limitations (stuck/indeterminate), independent of the lemmas; all pass in normal mode. |
| 3 | minor handoff regression | `examples/erc721-spec.md` rises 1→2 implies-handoffs on tier1 (single spec, implies-flavour). |
| 4 | implies-count artifact (rules) | Over *all* common claims, implies-handoffs appear to rise 45→53 on tier1 — an artifact of tier1 completing 14 extra proofs that master never reached. On the both-passed set, implies is flat (43→42). |
| 5 | DSS incomplete | Both branches hit the 3 h cap on the 2-spec dss test; the −72% is over the identical first tranche (`mcd/vat-spec.k`, 36 claims), **not** an end-to-end completion. recover-mode is too slow for full DSS; a normal-mode `count-aborts.py` pass is the right end-to-end instrument. |

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/buf.md
```diff
@@ -37,10 +37,17 @@ module BUF
     syntax Int ::= #powByteLen ( Int ) [symbol(#powByteLen), function, no-evaluators]
  // ---------------------------------------------------------------------------------
  // rule #powByteLen(SIZE) => 2 ^Int (8 *Int SIZE)
-    rule 2 ^Int (8 *Int SIZE) => #powByteLen(SIZE) [symbolic(SIZE), simplification]
+    rule 2 ^Int (8 *Int SIZE) => #powByteLen(SIZE) [symbolic(SIZE), simplification, preserves-definedness]
```
ehildenb, line 40: Is this preserving definedness? I would think it's only the case if `0 <=Int SIZE`, otherwise I don't think `^Int` is defined.

ehildenb: I guess since `#powByteLen` doesn't have a definition, we can just say it preserves definedness. But I'm also wondering if `#powByteLen` is even necessary as a construct.

ehildenb:
Overall, these lemmas alone are a mild performance hit (~6%). But coupled with https://github.com/runtimeverification/haskell-backend/pull/4153 with the optional local simplification recursion set to 20, become performance neutral overall (and performance positive in many simplification-heavy cases). These lemmas alone eliminate _many_ fallbacks to Kore though, so this gets us closer ot our goal of eliminating the Kore backend altogether.

kevm-pyk/src/kevm_pyk/kproj/evm-semantics/buf.md
ehildenb: Yes, by definition it's only defined when `SIZE >= 0` anyway.

anvacaru (approved):

PR #2867, opened 2026-06-18
kevm-pyk: thread step_timeout through run_prover

Stevengre:
Adds `step_timeout` to `run_prover`, forwarded to `APRProver` (which enforces the per-step budget: interrupt + halve `execute_depth` + retry, see runtimeverification/k#4930).

`step_timeout` is only honored on the sequential `advance_proof` path — `parallel_advance_proof` ignores it. So when `step_timeout` is set, `run_prover` runs sequentially regardless of `force_sequential`, warning if `max_frontier_parallel > 1`. Default `None` is unchanged.

Enables Kontrol's `--step-timeout`. Making the parallel path honor `step_timeout` is a pyk follow-up.


kevm-pyk/src/kevm_pyk/utils.py
```diff
@@ -134,13 +135,25 @@ def create_prover() -> APRProver:
...
+            run_sequential = force_sequential
+            if step_timeout is not None and not force_sequential:
+                if max_frontier_parallel > 1:
+                    _LOGGER.warning(
+                        f'Proof {proof.id}: step_timeout is only enforced when running sequentially; '
+                        f'ignoring max_frontier_parallel={max_frontier_parallel} and running sequentially.'
+                    )
+                run_sequential = True
+
+            if run_sequential:
```
ehildenb: Why do we introduce a new variable `run_sequential`? Is that necessary? Can we just use `force_sequential`, and set it to `True` if needed?

Stevengre: Thank you! Dropped the extra variable and just reassign `force_sequential`.

ehildenb (approved):

PR #2868, opened 2026-06-18
tests/failing-symbolic.haskell-booster-dev: drop specs now passing under booster-dev

ehildenb:
Re-validated the booster-dev skip-list against stock K 7.1.337 by running the suite under --use-booster-dev with the list neutralised. Three specs now close under pure booster-dev (no Kore fallback): functional/bitwise-mask-shift-spec.k, mcd/cat-exhaustiveness-spec.k, mcd-structured/cat-exhaustiveness-spec.k. Also drops the stale functional/slot-updates.k entry, which matched no spec file.


palinatolmach (approved):
nice!

PR #2876, opened 2026-07-27
 EIP-8037 + EIP-7778 + EIP-8038: state gas reservoir and refund-free block gas accounting

anvacaru:
Implements [EIP-8037](https://eips.ethereum.org/EIPS/eip-8037): state-creation costs move to a per-transaction state-gas reservoir; execution gas and state gas are tracked separately, and `gasUsed` is their max. The new `#finalizeTx` rules also implement [EIP-7778](https://eips.ethereum.org/EIPS/eip-7778): block gas is accumulated from pre-refund totals while sender/miner settlement still applies the refund. Aligned against the EELS reference implementation.

This PR also implements [EIP-8038](https://eips.ethereum.org/EIPS/eip-8038).The two are kept together because their specifications cross-reference each other: EIP-8038 declares `Requires: EIP-8037`, defers state-creation costs to it, and its SSTORE cost table spans both the execution-gas and state-gas dimensions, so neither can be validated against the fixtures alone.

> ~**Blocked by #2834 (EIP-7825).** This branch does not kompile on its own: `#txStateGasReservoir` is defined as the transaction gas above the EIP-7825 cap (`maxInt(0, GLIMIT -Int Gmaxtxgaslimit < SCHED >)`), and the per-transaction block-gas-remaining admission check bounds the limit by that same cap, so `Gmaxtxgaslimit`/`Ghastxgaslimit` must already be declared. #2834 has to merge first; CI here will fail until it does. Verified locally that `plugin` and `llvm` kompile clean once #2834 is merged in.~

**Changes:**
  - Three new `<callState>` cells (`stateGasReservoir`, `stateGasSpilled`, `newAccountCharged`) with frame push/pop, CALL/CREATE/SELFDESTRUCT charge sites, and refill on create-tx failure, plus block-level `<gasUsedRegular>`/`<gasUsedState>`.
  - Amsterdam SSTORE regular-gas restructure (cold XOR warm access + first-change write cost, simplified refunds). This is specified by EIP-8038, but lands in an EIP-8037-labelled commit because it predates the two being combined.
  - EIP-8038 constants: `COLD_ACCOUNT_ACCESS` 3000, `STORAGE_WRITE` 10000, `ACCOUNT_WRITE` 9000, `CALL_VALUE` 11300, `CREATE_ACCESS` 12000, `REFUND_STORAGE_CLEAR` 11616, access lists 2900/2000, and the second-database-read surcharge on `EXTCODESIZE`/`EXTCODECOPY`. `COLD_STORAGE_ACCESS` stays at 2100; `BALANCE`/`EXTCODEHASH` get no surcharge.
  - State-creation constant increases; per-tx block-gas-remaining admission preserved on pre-state-gas forks.
  - All `tests/specs/` closed-form `<callState>` configs gain the three cells: 213 configuration blocks across 157 spec files in `benchmarks`, `erc20`, `kontrol`, `mcd` and `mcd-structured`.


palinatolmach (approved):
lgtm overall, with two follow-ups once Glamsterdam EIPs are finalized:
1. implementing the successful-child merge repayment according to the updated EIP-8037,
2. reverting EIP-7702 delegations in case of a pre-execution OOG revent, once EIP-2780 is implemented in KEVM.

PR #2878, opened 2026-08-09
Enable CI, disable CD

palinatolmach:
This PR re-enables CI on this repository and removes all secret usage from it, which makes it safe to run again. It also drops the secret-based CD entirely rather than converting it: the release workflow is removed here and comes back, secret-less, in a follow-up PR.

It performs the following updates:
- Removes auto version bumps and release-cutting on `master`, along with `release.yml` itself. There is no release automation left in this repository. **This will be updated in a follow-up PR.**
- Removes Nix/Cachix cache and `gh-pages` docs deploys, Slack notifications, the devops dispatch, and the `check-cachix-pin.sh` helper that verified the Cachix push.
- Removes the dependency-bump automation that synced `deps/*` into `kevm-pyk/pyproject.toml`, `kevm-pyk/uv.lock` and `flake.{nix,lock}`; the `_update-deps/*` cron is not running, so that sync is manual for now.
- Bumps all pulled-in actions to the latest (checkout v7.0.1, setup-python v7.0.0, setup-uv v9.0.0) and pins every one of them to its commit SHA.
- Sets `permissions: contents: read` on `test-pr.yml` and `persist-credentials: false` on its checkouts, so the CI token is minimal regardless of the repository default, and no credential is left behind in the workspace.
- Passes the `with-docker` action's inputs through `env:` instead of interpolating them into the shell block, which removes the template-injection exposure in the action that runs PR-authored code on the self-hosted runners.
- Makes the Docker teardown best-effort and switches the deprecated `--time` flag to `--timeout`.


.github/workflows/release.yml
```diff
@@ -7,166 +7,36 @@ concurrency:
   group: ${{ github.workflow }}
   cancel-in-progress: true
```
yiyi-wang-rv: `cancel-in-progress: true` is riskier than it was. The old workflow's first job only *drafted* a release; this one bumps, commits, pushes, and publishes in a single script.

If a second push to `release` lands (or someone cancels) in the window between `git push origin HEAD:release` on line 38 and `gh release create` on line 39, the branch keeps a `Set Version: vX` commit with no tag and no release, and the next run bumps to `vX+1` — so `vX` is silently skipped and never published, with nothing in the logs pointing at why.

Suggest `cancel-in-progress: false` here. Releases are cheap to queue and expensive to half-apply.

.github/workflows/release.yml
```diff
@@ -7,166 +7,36 @@ concurrency:
...
-    needs: draft-release
-    steps:
-      - name: 'Check out code'
-        uses: actions/checkout@v4
+      - name: 'Check out release branch'
+        uses: actions/checkout@v7
         with:
-          ref: ${{ github.event.push.head.sha }}
+          ref: release
           fetch-depth: 0
```
yiyi-wang-rv: The `with:` block can just be dropped:

```suggestion
      - name: 'Check out release branch'
        uses: actions/checkout@v7
```

- `ref: release` is not just redundant, it is slightly worse than the default. For a `push` event, the default checkout gives you `refs/heads/release` at exactly `github.sha` — the commit that triggered this run. `ref: release` re-resolves the branch tip at job start, so under a race it can bump and release a *different* commit than the one that fired the workflow.
- `fetch-depth: 0` existed only for the `git merge-base origin/master origin/release` logic that lived in `master-push.yml`. Nothing left in this job reads history, so a full-history clone is pure cost.

.github/workflows/release.yml
```diff
@@ -7,166 +7,36 @@ concurrency:
...
+      - name: 'Bump version, commit, push, and create release'
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         run: |
-          set -x
-          VERSION=v$(cat package/version)
-          gh release delete ${VERSION}                  \
+          set -euxo pipefail
+          ./package/version.sh bump
+          ./package/version.sh sub
```
yiyi-wang-rv: `version.sh sub` rewrites `version = "..."` in `kevm-pyk/pyproject.toml`, but `kevm-pyk/uv.lock` also records the workspace package version (`name = "kevm-pyk"` / `version = "1.0.678"` at HEAD) and nothing re-locks. So every release commit ships a lockfile that disagrees with its own manifest.

This doesn't break anything today — `uv run`/`uv sync` re-lock silently — but any `--locked`/`--frozen` invocation against the release branch would fail, and it makes the tagged tree subtly inconsistent. The gap predates this PR (`master-push.yml` had it too), but this is now the only place it happens and the fix is one line: `uv --project kevm-pyk lock` after `sub`, or extend `version_sub()` in `package/version.sh` so the two stay together.

.github/workflows/release.yml
```diff
@@ -7,166 +7,36 @@ concurrency:
...
-      - name: 'Post success to channel'
-        if: success()
-        uses: slackapi/slack-github-action@v1.24.0
-        with:
-          channel-id: "#kevm-notifications"
-          slack-message: "Created KEVM release: https://github.com/${{ github.repository }}/actions/runs/${{ github.run_id }}"
-        env:
-            SLACK_BOT_TOKEN: ${{ secrets.SLACK_BOT_TOKEN }}
+            --title "${VERSION}"                       \
+            --target "$(git rev-parse HEAD)"
```
yiyi-wang-rv: Two things:

**1. The release is now published unverified.** The old flow created it as a `--draft`, built with Nix on x86_64 *and* ARM64, and only then ran `gh release edit --draft=false`; a failed build deleted the draft. Now a public `v<version>` tag and release are cut from whatever is sitting on `release`, with no build of that tree at all. `test-pr.yml` covers PRs into `master`, but not the merge commit that lands on `release`, so a bad merge resolution publishes cleanly. The self-hosted runners are still available, so consider either keeping `--draft` and letting a human promote, or gating this step on a build job via `needs:`. If publishing unverified is a deliberate trade (releases are now just source tags, since the Nix binaries are published off-GitHub), it's worth saying so in the header comment.

**2. `--repo runtimeverification/evm-semantics` is redundant.** `gh` resolves the repo from `GITHUB_REPOSITORY`/the checkout. Hardcoding it means a fork that pushes `release` tries to create a release in *this* repo and fails confusingly instead of doing the harmless local thing. Dropping the flag also drops a line.

Nit: with no `--notes`/`--generate-notes` the release body is empty. Not a regression — the old workflow did the same — but `--generate-notes` gives a real changelog for free now that nothing else publishes release information.

.github/workflows/test-pr.yml
```diff
@@ -75,7 +75,7 @@ jobs:
     timeout-minutes: 45
     steps:
       - name: 'Check out code'
-        uses: actions/checkout@v4
+        uses: actions/checkout@v7
```
yiyi-wang-rv: Worth confirming with devops before merging, because this is the change most likely to break the run.

`actions/checkout@v7` declares `using: node24`; v4 was `node20`, and there is no intermediate step — v5 and v6 are already node24. Same for `setup-python@v7` and `setup-uv@v9`. The `ubuntu-latest` jobs are fine, but this checkout and the ones in `test-prove` (line 141) and `nix` (line 172) run on self-hosted runners: `[self-hosted, linux, normal]`, `[self-hosted, linux, fast]`, and the `normal`/`ARM64` labels. Those hosts need an actions-runner release that ships Node 24 or every conformance, proof, and Nix job fails at the checkout step before doing any work.

The ARM64 runner is the one I'd check first. If any of them can't be updated yet, `@v4` is the only node20 option — the bump would have to wait rather than land at v5/v6.

.github/workflows/test-pr.yml
```diff
@@ -13,17 +13,17 @@ jobs:
...
+        uses: actions/setup-python@v7
         with:
           python-version: '3.10'
       - name: 'Get uv release'
         id: uv_release
         run: |
           echo uv_version=$(cat deps/uv_release) >> "${GITHUB_OUTPUT}"
       - name: 'Install uv'
-        uses: astral-sh/setup-uv@v6
+        uses: astral-sh/setup-uv@v9.0.0
```
yiyi-wang-rv: The rationale in the commit message checks out: `astral-sh/setup-uv` publishes no floating major tag for v8/v9 (`git/matching-refs/tags/v9` returns only `refs/tags/v9.0.0`), unlike `actions/checkout` and `actions/setup-python`, which both do publish `v7`. So the exact pin is right here and the asymmetry with the other two actions is justified.

One suggestion, since this PR is explicitly about CI's security posture: `v9.0.0` is still a *mutable* tag, and `setup-uv` is now the only non-`actions/*` action left in the repo. GitHub's hardening guidance is to pin third-party actions by full commit SHA:

```yaml
uses: astral-sh/setup-uv@<sha>  # v9.0.0
```

Cheap to do here — there's no floating tag to give up, so a SHA pin costs nothing in maintenance beyond what the exact-version pin already costs.

.github/workflows/master-push.yml
```diff
@@ -1,43 +0,0 @@
...
-          if git diff --exit-code ${old_master} ${new_master} -- package/version; then
-            git merge --no-edit origin/master
-            ./package/version.sh bump
-          else
-            git merge --no-edit --strategy-option=theirs origin/master
-          fi
-          ./package/version.sh sub
-          if git add --update && git commit --no-edit --allow-empty --message "Set Version: $(cat package/version)"; then
-            git push origin release
-            git tag "release-$(cat package/version)" origin/master
```
yiyi-wang-rv, line 41: Deleting this also drops the `release-<version>` tags on `master` — they exist up through `release-1.0.921` today. The new `release.yml` only creates `v<version>`, pointing at the bump commit on the `release` branch.

The consequence is that there is no longer any pointer from a published release back to the `master` commit it was cut from, which is the thing you want when bisecting "which master commit is in v1.0.9xx". Probably fine if nothing downstream consumes those tags, but it's currently a side effect of the deletion rather than a stated decision — worth a line in the PR description either way. If it does matter, `release.yml` can create the same tag with the built-in token.

.github/workflows/update-version.yml
```diff
@@ -1,84 +0,0 @@
...
-
-  update-versions:
-    name: 'Update Dependency Versions'
-    runs-on: ubuntu-latest
-    steps:
-      - name: 'Check out code'
-        uses: actions/checkout@v4
-        with:
-          submodules: true
-          token: ${{ secrets.JENKINS_GITHUB_PAT }}
```
yiyi-wang-rv, line 26: This is the one deletion I'd push back on.

The `_update-deps/*` branches are still being pushed by the external `rv-jenkins` cron — #2864, #2865, and #2866 all landed from `_update-deps/runtimeverification/k` in the last few days. Without this job, those branches carry *only* the changed `deps/*` file, and none of the downstream sync happens: the `kframework==` pin in `kevm-pyk/pyproject.toml`, `kevm-pyk/uv.lock`, `deps/uv_release`, `deps/blockchain-k-plugin_release`, and `flake.{nix,lock}`. So every dependency-update PR from here on will fail CI until a human reproduces all of that by hand — which is a real recurring cost, not a one-off.

And the secrets here were replaceable rather than load-bearing:
- `JENKINS_GITHUB_PAT` was only used for `checkout` + the final `git push` → the built-in `GITHUB_TOKEN` plus `permissions: contents: write` covers that.
- `CACHIX_PUBLIC_TOKEN` only fed a build cache → dropping the `cachix-action` step makes the job slower, not broken.

The one genuine catch, which may well be *why* deletion was chosen: a push made with `GITHUB_TOKEN` does not trigger `pull_request` workflows, so the auto-synced commit wouldn't re-run `test-pr.yml` on its own the way the PAT-driven push did. That's a real limitation of the conversion and worth stating explicitly.

Either way — converted or deleted — the PR description should say what replaces this, so dep bumps don't quietly become an undocumented manual chore.

yiyi-wang-rv (commented):
Reviewed the whole diff. The core goal is achieved and verified: `grep -rn "secrets\." .github/` now returns exactly one hit, the auto-issued `secrets.GITHUB_TOKEN` in `release.yml`. I also checked the two subtle things this design leans on, and both hold:

- **No self-trigger loop.** `release.yml` pushes to `release`, which is its own trigger branch — but pushes made with the built-in `GITHUB_TOKEN` do not create new workflow runs, so this terminates. Worth a one-line comment in the file, since it is load-bearing and invisible.
- **`git push` still works after the checkout bump.** `actions/checkout` v6 moved credentials to a separate file, but `persist-credentials` still defaults to `true` in v7, so the pushes in the release job authenticate.

I verified the action versions too: `actions/checkout@v7` and `actions/setup-python@v7` both exist and publish floating major tags; `astral-sh/setup-uv` genuinely has no `v8`/`v9` floating tag (`git/matching-refs/tags/v9` returns only `v9.0.0`), so the exact pin in commit 2 is correct and necessary. `gh release create` without `--notes` does not error in a non-TTY (`create.go` only prompts when `IO.CanPrompt()`), so the release step will not fail on that.

Findings inline. The two I would resolve before merging:

1. **node24 on the self-hosted runners** — `checkout@v7`/`setup-python@v7`/`setup-uv@v9` are all `using: node24`; v4 was the last node20 checkout. The self-hosted `normal`/`fast`/`ARM64` runners need a recent enough actions-runner or every conformance/proof/nix job dies at the checkout step. This is the most likely thing to actually break.
2. **Deleting `update-version.yml` breaks the dependency-bump pipeline**, which is still live (#2864/#2865/#2866 all came from `rv-jenkins` pushing `_update-deps/runtimeverification/k` in the last few days). Its secrets were replaceable rather than load-bearing — details inline.

Smaller: `cancel-in-progress: true` on a workflow that now pushes commits and cuts releases can half-apply and silently skip a version; the release is published with no build verification, where the old flow gated promotion on a successful Nix build; and `test-pr.yml` has no `permissions:` block.

One nit on the PR description: it lists a DockerHub image publish among the removals, but I don't see one anywhere in the diff — the only Docker usage is `.github/workflows/Dockerfile` via the local `with-docker` action, which is untouched CI infrastructure.

.github/workflows/test-pr.yml
yiyi-wang-rv, line 1: This workflow has no `permissions:` block, so its `GITHUB_TOKEN` inherits the repo/org default — which can still be read-write depending on how the org is configured. Given the goal of this PR, and given that `release.yml` now sets its scope explicitly, adding

```yaml
permissions:
  contents: read
```

at the top of the file would make the CI token minimal regardless of the repo default, and make the release workflow's `contents: write` a visible exception rather than the same thing everything else silently gets. Nothing in `test-pr.yml` writes to the repo, so this should be a no-op functionally.

(Also worth knowing, not necessarily acting on in this PR: these jobs run PR-authored code on self-hosted runners. That's pre-existing, but it is the reason the token scope here is worth being explicit about.)

.github/workflows/test-pr.yml
anvacaru: @yiyi-wang-rv, it seems like the checkout code (`actions/checkout@v7`, node24) succeeded on every self-hosted runner in both runs, so I think this is safe.

.github/workflows/release.yml
anvacaru: Applied this change in https://github.com/runtimeverification/evm-semantics/pull/2878/commits/a279462a9feab25a03c377ac9e4dc6d8cda4e508.

.github/workflows/release.yml
anvacaru: Applied in https://github.com/runtimeverification/evm-semantics/pull/2878/commits/42767ea44423979df417bcd5e06caa38b550beeb.

.github/workflows/release.yml
anvacaru: Done in https://github.com/runtimeverification/evm-semantics/pull/2878/commits/95fdf0f1270625d9395b3b9c2899b74651fc6ebe.

Re-locking now happens in `version_sub()`, so the manifest and the lockfile can't drift apart regardless of who runs the script. I've used `test-pr.yml` as an example to set up `uv` in the release job. 

However, the new step tripped a **zizmor** `cache-poisoning` finding: `setup-uv` enables caching by default on hosted runners, which would give the one workflow that publishes an input written outside its own trust boundary. `enable-cache: false` should clear it and costs nothing, since the job locks exactly once.

.github/workflows/test-pr.yml
anvacaru: Replaced all tags to SHA hashes in https://github.com/runtimeverification/evm-semantics/pull/2878/commits/940a37c82cec03628a3a2fece79fa4eddc089591.

.github/workflows/release.yml
anvacaru: Done in https://github.com/runtimeverification/evm-semantics/pull/2878/commits/785673fccd9091d30cd81ddd42d1b4adc77dc7aa, added the `draft` and `--generate-notes` flag and dropped the `--repo` one.

.github/workflows/test-pr.yml
anvacaru: Addressed in https://github.com/runtimeverification/evm-semantics/pull/2878/commits/d825b52e137cdeacd9ae18206cb40b1573bbfdf8.

.github/workflows/master-push.yml
anvacaru: Afaik, kontrol pins `v<version>` in both `flake.nix` and `pyproject.toml`, and I don't think that we use the  `release-1.0` tags anywhere. Also, the bisect pointer should still be recoverable without them, because the bump commit's parent is always the `master` to `release` merge.

yiyi-wang-rv (commented):
Re-audited all 8 new commits. This is in good shape — everything I raised is addressed, and I verified each fix rather than taking the reply at face value. One new finding, non-blocking, inline on `package/version.sh`.

## Verified fixes

**SHA pins** — resolved all three against the tag objects; every one matches its comment:

| pin | tag object | claimed |
|---|---|---|
| `actions/checkout@3d3c42e5…` | `3d3c42e5aac5ba805825da76410c181273ba90b1` | v7.0.1 ✅ |
| `actions/setup-python@5fda3b95…` | `5fda3b95a4ea91299a34e894583c3862153e4b97` | v7.0.0 ✅ |
| `astral-sh/setup-uv@c771a70e…` | `c771a70e6277c0a99b617c7a806ffedaca235ff9` | v9.0.0 ✅ |

**node24 on self-hosted runners** — agreed, and confirmed independently: run [31393430707](https://github.com/runtimeverification/evm-semantics/actions/runs/31393430707) is green on all 12 jobs, covering `[self-hosted, linux, normal]`, `[self-hosted, linux, fast]`, and both `normal` and `ARM64`. That was the finding I was least able to settle from the diff, so having it exercised end-to-end closes it properly.

**Dropping the `with:` block** — checked the two things this now depends on. `persist-credentials` still defaults to `true` in checkout v7 (v6 moved credentials to a separate file but kept the default), so `git push origin HEAD:release` still authenticates; and for a `push` event the default ref is `refs/heads/release` at exactly `github.sha`, which is strictly better than the old `ref: release`. 

**`enable-cache: false`** — valid and does what the comment says: `getEnableCache()` in `setup-uv/src/utils/inputs.ts` is `enableCacheInput === "true"`, so anything other than `"true"`/`"auto"` disables it. Good instinct not to restore a cache into the one workflow holding `contents: write`.

**`version.sh` cleanup** — behaviour-preserving, and worth saying why so nobody re-adds the guard: the old code read `version` and `current_version` from the *same file*, so `major == major && minor == minor` was always true and the patch was always incremented. The deleted branch was unreachable, not a major/minor code path. The new comment about a hand-set line starting at `X.Y.1` is accurate.

**`--draft` / `--generate-notes` / `--repo` dropped, `cancel-in-progress: false`, `contents: read`** — all good, and the header comment now explains the one gap CI doesn't cover, which is the right thing to write down.

**`release-<version>` tags** — your reply checks out: kontrol's `flake.nix` pins `kevm.url = "github:runtimeverification/evm-semantics/v1.0.921"` alongside `deps/kevm_release`, so it consumes `v<version>`, not `release-*`. No consumer to break.

**Docker teardown** — I was initially wary of `|| true` on a teardown step, since it can hide a real leak on a long-lived self-hosted runner. Pulling the log from the pre-fix run (d825b52e1) shows the change is well-motivated: the registry mirror returned EOF, `Set up Docker` failed, and teardown then emitted `No such container: kevm-ci-concrete-<sha>` plus a second `##[error] Process completed with exit code 1` — exactly the duplicate-failure noise the comment describes. `--time` is also genuinely deprecated now (`Flag --time has been deprecated, use --timeout instead` on Docker 28), and the self-hosted daemons accept `--timeout`: the teardown steps in the green run *succeeded* rather than being silently swallowed by `|| true`, which is the distinction I wanted to confirm.

## `update-version.yml`

Agreed, and your reasoning is better than mine — if the cron is down, nothing regresses today, and this PR shouldn't be blocked on rebuilding an automation whose trigger doesn't fire. One ask: file a tracking issue rather than leaving it in the PR thread. The failure mode is delayed rather than absent — when the cron comes back it will start opening `_update-deps/*` PRs that carry only the `deps/*` file and fail CI, and at that point the context for *why* the workflow went away will be hard to recover.

## Operational note on `--draft`

Not a code issue, just worth writing into the release runbook: a draft release does not create the git tag until it is published, and the Slack notifications are gone, so nothing announces that a draft is waiting. Until someone clicks publish, downstream pins like kontrol's `github:runtimeverification/evm-semantics/v<version>` will 404 on the new version. That is the intended trade for gating publication on a human, but the human now has to know to show up.

## Verdict

Approvable from my side. The `uv lock` point below is a follow-up, not a blocker.

For the kontrol port, the three checks that actually earned their keep here: resolve every pinned SHA against its tag before merging, exercise the self-hosted runners once for the node24 bump rather than reasoning about runner versions, and check whether kontrol has its own `deps/`-driven update-version flow that would be deleted along the same lines — the same delayed-failure question applies there.

package/version.sh
```diff
@@ -29,6 +24,8 @@ version_sub() {
     local version
     version="$(cat $version_file)"
     sed --in-place 's/^version = ".*"$/version = "'${version}'"/' kevm-pyk/pyproject.toml
+    # uv.lock records the workspace package version too; re-lock so it cannot drift from the manifest.
+    uv --project kevm-pyk lock
```
yiyi-wang-rv: New finding, and one I own — it's a consequence of my earlier comment, and having measured it I think my original framing was stronger than the facts supported.

`uv lock` is a full re-resolution, not a targeted edit of the version field. Reproduced with uv **0.9.14**, exactly what `deps/uv_release` pins and what the new `setup-uv` step installs in `release.yml`:

```console
$ sed -i 's/1.0.678/1.0.679/' kevm-pyk/pyproject.toml
$ uv --project kevm-pyk lock
Resolved 85 packages in 7.74s
Updated kevm-pyk v1.0.678 -> v1.0.679

$ git diff kevm-pyk/uv.lock
 [[package]]
 name = "exceptiongroup"
 dependencies = [
-    { name = "typing-extensions", marker = "python_full_version < '3.13'" },
+    { name = "typing-extensions", marker = "python_full_version < '3.11'" },
 ]
 [[package]]
 name = "kevm-pyk"
-version = "1.0.678"
+version = "1.0.679"
```

So the first release after this merges will produce **two** changes, not one, inside a commit whose message is only `Set Version: v1.0.679` — and that commit is what gets tagged and released.

The part that makes this awkward is that you cannot gate it in CI. `uv --project kevm-pyk lock --check` **passes on HEAD today** (exit 0): `--check` verifies that the recorded lock still satisfies the manifests, not that it is byte-identical to a fresh resolution. So this drift is invisible on `master` and first appears at release time. I also tried `uv --project kevm-pyk version 1.0.679` hoping for a narrower write — same marker rewrite, plus it syncs the venv as a side effect. There is no "update only my own version field" operation.

Which cuts both ways on my original comment: nothing in this repo uses `--locked`/`--frozen`, and `uv lock --check` tolerates a stale workspace version, so the pre-PR drift was harmless in practice. Re-locking trades that for "release commits can absorb unreviewed resolution changes" — a yanked wheel or a marker-inference change would land in the tagged tree without review or CI.

**Suggested follow-up, non-blocking:** run `uv --project kevm-pyk lock` once against `master` — here or in a separate PR — and commit the resulting marker change. It then gets reviewed and CI-tested like any other change, and the first release-time `uv lock` touches only the version line. Keep `version_sub()` exactly as it is; the re-lock belongs there, it just shouldn't be the thing that discovers accumulated drift.

yiyi-wang-rv (approved):

.github/actions/with-docker/action.yml
```diff
@@ -83,3 +83,11 @@ runs:
...
       docker cp . ${CONTAINER_NAME}:/home/${USER}/workspace
       docker exec ${CONTAINER_NAME} chown -R ${USER}:${GROUP} /home/${USER}
+    env:
+      INPUTS_CONTAINER_NAME: ${{ inputs.container-name }}
+      INPUTS_SUBDIR: ${{ inputs.subdir }}
+      INPUTS_OS: ${{ inputs.os }}
+      INPUTS_DISTRO: ${{ inputs.distro }}
+      INPUTS_DOCKERFILE: ${{ inputs.dockerfile }}
+      INPUTS_LLVM: ${{ inputs.llvm }}
+      INPUTS_TAG_NAME: ${{ inputs.tag-name }}
```
tothtamas28: Consider moving this above `shell`.

.github/workflows/test-pr.yml
```diff
@@ -7,23 +7,30 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
   cancel-in-progress: true
 
+# No job here writes to the repository, so the token stays read-only regardless of
+# the repository default.
+permissions:
+  contents: read
```
tothtamas28, line 13: Nice.

tothtamas28 (approved):

.github/actions/with-docker/action.yml
anvacaru: Addressed in https://github.com/runtimeverification/evm-semantics/pull/2878/commits/d3bc3cb96d4b9094469b63bc70a96fd046cc1135.

.github/workflows/lint-workflows.yml
```diff
@@ -0,0 +1,56 @@
...
+          echo "uv_version=$(cat deps/uv_release)" >> "${GITHUB_OUTPUT}"
+      - name: 'Install uv'
+        uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9  # v9.0.0
+        with:
+          version: ${{ steps.uv_release.outputs.uv_version }}
+          enable-cache: false
+      - name: 'Run zizmor'
+        env:
+          ZIZMOR_VERSION: '1.29.0'
+        run: uvx "zizmor@${ZIZMOR_VERSION}" --offline .github/
```
tothtamas28: Consider: https://docs.zizmor.sh/integrations/#via-zizmorcorezizmor-action

anvacaru: Addressed in https://github.com/runtimeverification/evm-semantics/pull/2878/commits/6e75b7c90e27fb4b6a18e56f5be5ab7382cfbef9.

SARIF uploads are enabled so findings annotate the diff, by `security-events: write`. However, this might cause failures in PRs that come from forks, where GitHub forces `GITHUB_TOKEN` to be read-only regardless of the permissions block.

.github/workflows/lint-workflows.yml
```diff
@@ -0,0 +1,53 @@
...
+      - name: 'Check out code'
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+        with:
+          persist-credentials: false
+      - name: 'Run zizmor'
+        uses: zizmorcore/zizmor-action@3dc1ecc9bcb9e94e9b2c709687979e1298497054  # v0.6.2
+        with:
+          version: '1.29.0'
+          online-audits: false
+          inputs: .github/
```
tothtamas28:
```suggestion
```

Default `.` is fine.

tothtamas28 (approved):

package/version.sh
anvacaru: I've reverted this change in `package/version.sh` and removed the release flow for now. Will address it in a follow-up PR.

